PR #286, opened 2015-01-08
Updated SAML docs to include feature flags that are very useful

technovangelist:


content/guides/saml.md
```diff
@@ -0,0 +1,108 @@
...
+
+After enabling the feature (and waiting for caches to clear) the customer will need to get a new version of the SP Metadata, which will have a different, org-specific AssertionConsumerService endpoint to send assertions to.
+
+### SAML Strict Mode
+For an org that has the saml_strict_mode feature flag enabled, users can not log in using a username and password on the login page - they must log in via their org's SAML IdP.
+
+Normally users who are invited to an org using SAML will not have a password, so they can't use the login page anyway. However, there are possibilities where a user at a SAML org can have a password
+
+* The user was created before the Org enabled SAML
+* Support added a temporary password and the user reset it In these cases (or in any case where a user has a password in the postgres database) a user can log in either with SAML (via the IdP) or with the login page using their Datadog username and password.
```
dcoleman17: This might benefit from a bit of a re-write.  The original doc was written with internal, support users in mind.  For instance, I don't think we should mention postgres database in a public document.

technovangelist: oops, i had cleaned it up a bit, but missed this one. any other issues?

dcoleman17: I'd suggest collapsing the first bullet into the paragraph and eliminating the second bullet:

Normally users who are invited to an org using SAML will not have a password, so they can’t use the login page anyway. However, there are possibilities where a user at a SAML org can have a password, for instance if the user was created before the org enabled SAML.

For the IdP Initiated Login section, the following might be more clear (though I am not sure 'Customer' and 'Org' are appropriate words to use in documentation):

When Datadog is configured for SP Initiated Login (which is the default), upon loading an org's "Single Sign On Login URL" the browser is redirected to the customer IdP with a SAML AuthnRequest. On the customer IdP the user types in credentials. If the credentials are correct the IdP redirects back to Datadog with a SAML Assertion.  Before validating the SAML Assertion, Datadog checks to see if the SAML Assertion was issued in response to a SAML AuthnRequest issued by Datadog.

Some IdPs have the ability to send an assertion directly to Datadog without first getting an AuthnRequest, which is referred to as IdP Initiated Login.

If Datadog is configured for SP Initiated Login (which is the default) and receives an assertion from an IdP trying to initiate and  IdP Initiated Login session, Datadog will not be able to match the Assertion with an AuthnRequest (since no AuthnRequest was made).  This will result in an error page with a message saying that SAML Response is missing “InResponseTo” attribute.

Datadog support can configure IdP Initiated Login for an org.  After IdP Initiated Login is enabled, the customer the customer will need to get a new version of the SP Metadata, which will have a different, org-specific AssertionConsumerService endpoint to send assertions to.

technovangelist: I just got rid of that feature. It seems like a rare circumstance that is better to just let support answer when it comes up.

PR #288, opened 2015-01-09
AWS JSON example.

lewg:
It needed commas.


technovangelist:
Hi @lewg - Thanks for helping out with this. It looks like one of the support engineer made the change around the same time that you made the pr. Have you noticed any other issues in the docs? If so, create another issue or PR and it won't take us as long to get it updated. 

Thanks again.

PR #308, opened 2015-02-26
Basic metrics-collection info for etcd

johnaxel:


johnaxel:
Since we're going to announce our etcd integration soon, @kky and I thought it would be good to have some basic docs to point to. Does this look OK to you @technovangelist @degemer?

degemer:
@johnaxel looks good to me.
I don't know if it's relevant, but you could add that etcd metrics are tagged with `etcd_state:leader` or `etcd_state:follower`, depending on the node status.

alq666:
Excellent point. Tags do matter.

johnaxel:
Thanks, Quentin! Just pushed a small change to add the tagging info.

PR #312, opened 2015-02-27
Running rake now checks for broken links

technovangelist:
Every now and then there are issues brought up for broken links. Now, when you run rake, it compiles, refreshes the browser, AND checks for broken links...which resulted in finding about 5 broken links no one had reported yet

Signed-off-by: Technovangelist <m@technovangelist.com>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,4 +15,5 @@ group :development do
   gem 'guard-rack'
   gem 'rack-contrib'
   gem 'rack-livereload'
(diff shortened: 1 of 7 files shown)


miketheman:
@technovangelist you may want to put that into the jenkins job too.

technovangelist:
nanoc check returns an error if there are broken links (along with a list of the broken links). Can we set jenkins to not deploy when there is an error here?

Issue #209, opened 2014-08-19
Route documentation by browser language


alq666:
@jhotta is translating our documentation to Japanese and has a branch on a fork to serve documentation out of `/ja/`.

If you are browsing with a Japanese browser, you should be served pages in Japanese without having to specify a specific url.

technovangelist:
This is now in place and works wonderfully

PR #318, opened 2015-02-28
Auto redirect based on japanese

technovangelist:
If a user has their browser pref set to japanese and they are not on a ja page they will be redirected to the same url with ja at the beginning. If they are not using japanese and they are on a ja page, they will be redirected to the same url minus the ja.

If you are not using ja and want to see japanese version, then add ?overridelang to the url. The logic to redirect will be ignored. Same goes if using ja and want to see english version.

Signed-off-by: Technovangelist <m@technovangelist.com>

### DIFF
--- a/content/static/datadog-docs.js
+++ b/content/static/datadog-docs.js
@@ -46,4 +46,15 @@
 
     // Export to global scope.
     window.DD_docs = DD_docs;
+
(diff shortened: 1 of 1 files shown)


technovangelist:
This resolves #209

PR #327, opened 2015-03-06
Added "change" option for time_aggr

rabdill:
Metric alerts can be based on the delta of a metric, but it was left out of the options listed.


technovangelist:
Hi @rabdill - Thanks so much for catching this. We are going to merge it in right now. Turns out there is yet another option as well: pct_change. I will update that separately. 

thanks again, and if you find any other issues please let us know.

rabdill:
Not a problem at all! Glad I could chip in a little.

PR #333, opened 2015-03-16
Celene/update monitoring guide

celenechang:
- added template var section under monitor notifications
- small update to FAQ


content/guides/monitoring.md
```diff
@@ -336,7 +336,50 @@ right people get notified so the problem can be resolved as soon as possible.
...
+    monitor event display a different message depending on whether the event is a
+    trigger, resolve, or no data notification. These variables use simple if-else
+    logic with the following syntax:
+
+        {{#var}} This text will show {{/var}} only if the variable `var` is true.
+        {{^var}} This text will show {{/var}} only if `var` is NOT true.
+
+    The conditional variables available are is_alert, is_recovery, and is_no_data.
+    These can also be seen in the "Use message template variables" help box in
+    Step 3 of the monitor editor.
```
conorbranagan: I think it would be beneficial to add a real example here. Just something simple but that shows a real use of the conditional variables

content/guides/monitoring.md
```diff
@@ -336,7 +336,50 @@ right people get notified so the problem can be resolved as soon as possible.
...
+
+    ![template var](/static/images/monitor/templatevar2.png)
+
+    The tag template variables available depend on the tag group selected in Step 1
+    of the monitor editor. The possible options will automatically populate at the
+    bottom of the "Use message template variables" help box in Step 3 of the editor.
+    These variables can also be used in the monitor titles (names), but note that
+    the variables are only populated in the text of Datadog child events (not the
+    parent, which displays an aggregation summary).
+
```
conorbranagan: Similarly it might be useful to show an example message here, to see how it translates. I'd even go a step further with your example and show how you can use it in follow-up steps, e.g.:
```
1. ssh to {{host.ip}}
2. Run command `fix_disk_space`
3. ...
```

I think we might have samples in our account...? Maybe even showcase using more than one, although that's not definitely necessary.

content/guides/monitoring.md
```diff
@@ -347,11 +390,13 @@ right people get notified so the problem can be resolved as soon as possible.
 
 - *Can you alert on a function?*
 
-  Not currently but this feature is in the works.
+  Yes, with a feature called 'query alerts'. Selecting the 'Source' tab of a
```
conorbranagan: I'd be in favor of not mentioning 'query alerts' at all as the user doesn't really need to care about that name. So just say "Yes, selecting the 'Source' tab..."

conorbranagan:
Some small comments but looks good overall, thanks :+1:

celenechang:
cool thanks @conorbranagan !

PR #334, opened 2015-03-17
replace deprecated DogApi library with datadogpy

brettlangdon:
according to https://github.com/DataDog/dogapi it is deprecated and people should be using https://github.com/DataDog/datadogpy instead, so I replaced the dogapi reference on the libraries page with the new one.

however, looking back at it, probably should have kept dogapi and just listed as deprecated and added datadogpy?


remh:
:+1: Thanks @brettlangdon !

PR #345, opened 2015-03-24
samantha pointed out that event and alert were mixed up

technovangelist:


conorbranagan:
This is actually inaccurate. We *do* aggregate on event_type/aggregation_key/hostname. Is there a case where this wasn't happening?

PR #355, opened 2015-03-30
Azure integration doc

2bethere:
This the V1 doc. Doesn't include extension deployment instructions as that's still under development.


technovangelist:
Hey Will, before merging this in can you make one change? I see you manually added the list of integrations. This means that the list appears twice. Can you remove your manual list?

2bethere:
Reverted to Matt's version of auto list generation.

technovangelist:
Before doing a pull request, can you do a rake clean, then rake. That will catch any errors in the build. It currently will not build due to the changes in the index file.

2bethere:
My bad, should have built again after reverting. Now there is no error.

technovangelist:
i merged it and then added another pr to add kind: integration and integration title: Microsoft Azure so that the doc is auto added to the generated list of integrations

PR #385, opened 2015-04-28
Adding multi bool in response for monitor actions

Viq111:
PR for the actual code is  #7802


Viq111:
@conorbranagan Also, how do I deploy this ? :question:

conorbranagan:
:+1:  Will send you deploy info

PR #399, opened 2015-05-07
deleted old alerting doc

technovangelist:


LotharSee:
Is this going to redirect the old URL to the new one?
We should (instead of returning a 4xx or 5xx) for SEO purpose.

PR #445, opened 2015-06-30
[api] wrong argument metric_type -> type

degemer:
Our python lib was using `metric_type` instead of `type`, but it's the second one which is correct. Our Ruby and Shell scripts are already correct, and our python script didn't use it.


yannmh:
Let's ship it :ship: !

degemer:
@technovangelist Our python library has been updated, it's a one-line fix, I deployed it on staging, looked fine, going to deploy it on prod.

PR #457, opened 2015-07-21
Update api-tags-get-host.sh

sysadmin-x:
Since the original example provided the use of jq, I've included it as a part of an 'if' statement.  If jq is not found in the system path, then use "cut" to correctly parse host_name from the JSON response.


sysadmin-x:
@technovangelist 

Matt,

No problem at all. I've created a new patch and submitted a pull request with a simplified solution. 

Thanks and Regards,
Aaron

PR #472, opened 2015-07-30
Removing host_id mentions in the api docs

MartinDatadog:
following PR #465 and according to trello https://trello.com/c/ylgRWFZb/701-docs-api-mentions-host-id-which-is-not-exposed-to-customers

- `host_id` has been replaced with `host_name` for ruby and shell code snippets
- `host_id` has been removed from the signatures.


MartinDatadog:
@technovangelist Mind reviewing this and share your opinion when you get a chance?

irabinovitch:
@technovangelist *ping*

PR #487, opened 2015-08-20
Martin/faq replace dogapi refs

MartinDatadog:


technovangelist:
i don't understand this. you seem to have a pr with a boatload of commits from lots of people....and these are already in master.  that doesn't feel right.

PR #496, opened 2015-09-06
Add extra whitespace in kafka docs

macobo:
It seems the markdown engine for the docs is a bit different for github and datadog, the suggestion to edit `conf.d/kafka_consumer.yaml` gets put in the code block


technovangelist:
Yes, they are both two different supersets of markdown.

PR #497, opened 2015-09-10
Updates for August pricing changes

irabinovitch:
Updates based on the new pricing released in August.


alq666:
Will not merge as we want to push people to annual plans, and the monthly price per host with an annual commit is still $15.

irabinovitch:
OK.  Do we want to clarify the yearly vs monthly rate?

PR #498, opened 2015-09-10
[outliers] Update outlier docs algs section

githomin:


content/guides/outliers.md
```diff
@@ -80,75 +80,51 @@ There are two different outlier detection algorithms you can use on your data: D
...
-The threshold is calculated as follows. We create a new median time series by taking the median of the values from the existing time series at every time point. Then we calculate the distance between each host and the median series. The threshold is the median of those distances.
-
-In pseudocode:
-
-~~~ python
-    median_series = pointwise_median(series_list)
-    dists = [dist(series,median_series) for series in series_list]
-    threshold = median(dists)
-~~~
+<img src="/static/images/outliers/outliers-dbscan-2d.png" style="width:100%; border:1px solid #777777"/>
```
StephenKappel: Not sure if it's different where this is going to be published, but GitHub doesn't like this path to the image. I think `/content` would need to be prepended to the path for this to work.

content/guides/outliers.md
```diff
@@ -80,75 +80,51 @@ There are two different outlier detection algorithms you can use on your data: D
...
 
 #### Parameters
 
-DBSCAN accepts the following parameters:
+We use a simplified form of DBSCAN to detect outliers on time series. (We consider each host to be a point in d-dimensions, where d is the number of values in the time series.) Any point can agglomerate, and any point that is not in the largest cluster will be considered an outlier.
```
StephenKappel: I think you can probably drop the parenthesis around the second sentence. It fits right it with the flow; it doesn't seem like a side note.

content/guides/outliers.md
```diff
@@ -80,75 +80,51 @@ There are two different outlier detection algorithms you can use on your data: D
...
 
-* **alpha** *(default: 3.0)*: Sets the constant by which the threshold is multiplied.
+We set the initial distance threshold as follows. We create a new median time series by taking the median of the values from the existing time series at every time point. Then we calculate the distance between each host and the median series. The threshold is the median of those distances.
 
-For example, to use DBSCAN with an alpha of 3 on system load for a pool of Cassandra workers, we'd use:
+The only parameter we take is: alpha. This sets the constant by which the initial threshold is multiplied to yield DBSCAN’s distance parameter 𝜀. Here is DBSCAN with an alpha of 3.0 in action on a pool of Cassandra workers:
```
StephenKappel: I'm being a complete nitpick, but... Might alpha be offset with code formatting and the first two sentenced combined? I just think it might read a little smoother that way.

> The only parameter we take is `alpha`, the constant by which ... Here is DBSCAN with an `alpha` of 3.0 ...

StephenKappel: Or, better yet, instead of using code formatting, **alpha** should be bold to match the formatting for **scale** and **pct** below.

content/guides/outliers.md
```diff
@@ -80,75 +80,51 @@ There are two different outlier detection algorithms you can use on your data: D
...
 
-The [Median Absolute Deviation](https://en.wikipedia.org/wiki/Median_absolute_deviation) (MAD) algorithm is a robust method of calculating variability among data. It is designed to be usable for both normal and non-normal distributions.
+<img src="/static/images/outliers/outliers-mad-az.png" style="width:100%; border:1px solid #777777"/>
 
-We use MAD to calculate an *outlier_factor* which we will compare to points from a series. If the percentage of points in the series that are above the *outlier_factor* is greater than some given threshold, then that series will be considered an outlier.
+The scale parameter should be tuned depending on expected variability of the data. For example if data is generally within a small range of values then this should be small. On the other hand if points can vary greatly then you want a higher scale so these variabilities do not trigger a false positive.
```
StephenKappel: Missing commas after "For example," "range of values," "On the other hand," "vary greatly."

content/guides/outliers.md
```diff
@@ -80,75 +80,51 @@ There are two different outlier detection algorithms you can use on your data: D
...
 
-<img src="/static/images/outliers/outliers-dbscan-banding.png" style="width:100%; border:1px solid #777777"/>
+For a given set of data D = {d<sub>1</sub>, ..., d<sub>n</sub>}, the deviations are the difference between each d<sub>i</sub> and median(D). The MAD is then the median of the absolute values of all the deviations. For example if D = {1, 2, 3, 4, 5, 6, 100}, then the median is 4, the deviations are {-3, -2, -1, 0, 1, 2, 96}, and the MAD is 2. (Note that the standard deviation by contrast is 33.8.)
 
-Meanwhile a higher setting of alpha only identifies the host handling significantly less traffic as an outlier.
+#### Parameters
 
-<img src="/static/images/outliers/outliers-dbscan-no-banding.png" style="width:100%; border:1px solid #777777"/>
+In our case, the data set is the set of all points in every time series. We take the MAD of all the points then multiply it by 1.4826 and our first parameter, **scale**. The first constant normalizes MAD so that it is comparable to the standard deviation of the normal distribution. The scale parameter then specifies how many “deviations” a point has to be away from the median for it to be considered an outlier.
```
StephenKappel: > The first constant normalizes...

Can we drop the word "first"? It's the only constant, right?

content/guides/outliers.md
```diff
@@ -80,75 +80,51 @@ There are two different outlier detection algorithms you can use on your data: D
...
-    outlier_factor = (scale * mad) / 0.6745
-~~~
+In the following image, we see a group of hosts flushing their buffers together while one host is flushing its buffer slightly later. DBSCAN picks this up as an outlier whereas MAD does not. This is a case where we would prefer to use MAD, as we don’t care about when the buffers get flushed. The synchronicity of the group is just an artifact of the hosts being restarted at the same time. On the other hand, if instead of flushed buffers, the metrics below represented a scheduled job that actually should be synchronized across hosts, DBSCAN would be the right choice.
 
-#### Parameters
+<img src="/static/images/outliers/outliers-flushing.png" style="width:100%; border:1px solid #777777"/>
 
-MAD accepts the following parameters:
+### Setting up alerts
+When setting up an outlier alert, an important parameter is the size of the time window. If the window size is too large, by the time an outlier is detected, the bad behavior might have been going on for longer than one would like. If the window size is too short, the alerts will not be as resilient to spikes.
```
StephenKappel: > ...the alerts will not be resilient to spikes.

Would "normal volatility" be better than "spikes." One might ask what the usefulness of outlier detection is if it ignores spikes.

StephenKappel:
@githomin - I've made some nitpicky comments above, but overall it looks good.

PR #505, opened 2015-09-23
[dogstatsd] Mention that the message field in SC should be last

olivielpeau:
So that the docs are consistent with the way we parse service check
packets in the agent's dogstatsd server.

See also https://github.com/DataDog/dd-agent/issues/1927


tmichelet:
:+1: let's merge this.

olivielpeau:
Thanks! Merging.

PR #525, opened 2015-10-16
switch to using sass, autoprefixer, and update rules

technovangelist:


technovangelist:
going to update to a newer bootstrap, but getting on sass is going to make that a bit easier. plus, lets rely on autoprefixer to get the prefixing right. And i had a stupid rule in there before to always rebuild the api doc even when it wasn't needed.

PR #537, opened 2015-10-23
[mongodb] Adding metrics to list

JohnLZeller:
MongoDB integration doc was missing some metrics


JohnLZeller:
I'm thinking it could also be useful to add a link to this blog post from mongodb: http://blog.mongolab.com/2014/01/how-big-is-your-mongodb/

What are your thoughts?

technovangelist:
sounds great. go for it

PR #539, opened 2015-10-23
[jmx] Add docs on JBoss/WildFly and Tomcat

olivielpeau:
NB: Should be merged when v5.6.0 of the agent is released.


yannmh:
@technovangelist can we reset the `staging` branch please ? It looks like it's conflicting with `master` branch state.

yannmh:
Thanks for putting all this together @olivielpeau. 
I find the instructions very clear :green_heart:

PR #554, opened 2015-11-05
[faq] Add note about trailing underscores

JohnLZeller:
An issue came up that exposed that we both trim trailing spaces from submitted tags, and that the documentation suggests otherwise. This adds a note to make sure it's documented.

(internal) https://trello.com/c/yXQz0bwl/632-trailing-underscores-in-tags


JohnLZeller:
@conorbranagan spare a r? :)

conorbranagan:
:+1:

PR #555, opened 2015-11-12
speed up compile time

technovangelist:


technovangelist:
Lots of optimizations here. Nanoc was cycling through every item several times when building each item.  this dropped the erb filter processing time from 12+ seconds to as little as 1 second. Plus there is a 'cache_buster' added that adds the hash of the file to the end of the file name of the 2 main css files, but running on 760 files. filtering this to only work on 2 files dropped the runtime of that one thing from 30+sec to 0.02 seconds. A few other optimizations brings compile down from 60+ seconds to somewhere between 2 and 20 seconds.

PR #562, opened 2015-11-14
filled out tagging to reference integrations

technovangelist:


technovangelist:
Added info on what integrations automatically add what tags.

PR #566, opened 2015-11-17
Specifying a datadog.conf restriction in the tagging guide, following…

Martiflex:
… conversation in ticket 38159


technovangelist:
I am going to deny this one but you pointed out a valid issue. The problem is not that datadog.conf is a yaml file that doesn't accept a format, but rather that datadog.conf is not a yaml file but a ini file (historical reasons). I will fix it in a separate pr. Thanks

PR #581, opened 2015-12-01
Removing antique parameter dogstatsd_interval parameter, now absent f…

Martiflex:
…rom the datadog-agent code. See discussion here: https://github.com/DataDog/dd-agent/issues/805


wilcollins:
Any reason for not having this configurable other than the lack of implementation? I'll be happy to PR if it's not a lost cause based on any current disposition on the matter.

PR #598, opened 2015-12-22
Updated NGINX and DynamoDB doc

JayJayM:
Added default dashboard screenshot on NGINX and links to our series of publication.


irabinovitch:
Minor nit: we should probably do each doc as a separate PR/commit in the future. Otherwise LGTM. Merging.

PR #604, opened 2016-01-04
Drop python-dogstatsd from libraries page as it's deprecated

conorbranagan:
You can perform all functionality in the `datadogpy` library now. The README also shows the library as deprecated when you visit the repo.


irabinovitch:
LGTM.

PR #619, opened 2016-01-22
Update nokogiri in response to security update

miketheman:
Again, non-critical, as we only use the library during site generation,
but should be updated to prevent security warnings for insecure
dependencies.

https://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2015-7499

Fixed versions: 1.6.7.2
Solution: Upgrade to latest version.

Source:
https://groups.google.com/forum/#!topic/ruby-security-ann/Dy7YiKb_pMM


miketheman:
@technovangelist merge at will.

PR #620, opened 2016-01-25
Thresholds is no longer service check only

ojongerius:


ojongerius:
Doco has been updated.

PR #649, opened 2016-02-29
Add nginx LUA integration to integration list

irabinovitch:
* alphabetize library / integration listings
* Add nginx LUA integration to integration list


technovangelist:
alphabetization? how did we not think of that one before.....

PR #656, opened 2016-03-07
update config directory

phoet:
this should also be changed, but i'm not sure if that is true for all `BasicAgentUsage-nix`

> Configuration
> The configuration file for the Agent is located at ~/.datadog-agent/datadog.conf
>
> Configuration files for integrations are located in ~/.datadog-agent/conf.d/

### DIFF
--- a/content/guides/basic_agent_usage/osx.md
+++ b/content/guides/basic_agent_usage/osx.md
@@ -3,7 +3,7 @@ title: Basic Agent Usage for OS X
 kind: documentation
 servicename: /usr/local/bin/datadog-agent
 serviceinfoname: /usr/local/bin/datadog-agent info
-configdirectory: ~/.datadog-agent/
+configdirectory: /opt/datadog-agent/etc/
 logdirectory: /var/log/datadog/
 supervisorlog: /var/log/datadog/supervisord.log
 os: osx


technovangelist:
the install creates a symlink to those directories in ~/.datadog-agent. so the current docs are accurate. Thanks for submitting the PR though

phoet:
@technovangelist ah, looks like an old installation had it installed with different access rights. i removed the `~/.datadog-agent/` directory and re-installed the agent. now it looks fine:

```
lrwxr-xr-x   1 peter  staff    31  7 Mär 16:51 checks.d -> /opt/datadog-agent/etc/checks.d
lrwxr-xr-x   1 peter  staff    29  7 Mär 16:51 conf.d -> /opt/datadog-agent/etc/conf.d
lrwxr-xr-x   1 peter  staff    35  7 Mär 16:51 datadog.conf -> /opt/datadog-agent/etc/datadog.conf
```

another note, in the [osx setup guides](https://app.datadoghq.com/account/settings#agent/mac) the `/opt/datadog-agent/etc/conf.d` is used as well.

technovangelist:
ok, but maybe this points to something we should look for in the install. Thanks for pointing it out. 

I will look into the setup guides

PR #682, opened 2016-03-22
[integration] bugsnag

irabinovitch:
Add doc for bugsnag integration.

### DIFF
--- a/content/integrations/bugsnag.md
+++ b/content/integrations/bugsnag.md
@@ -1,0 +1,42 @@
+---
+title: Datadog-Bugsnag Integration
+integration_title: Bugsnag
+kind: integration
+doclevel: basic
+---
+
+### Overview
+
+Bugsnag provides software teams with an automated crash detection platform for their web and mobile applications. Bugsnag automatically captures and alerts you of errors as they happen. Integrating Datadog with Bugsnag will send error notifications to your Datadog event stream.
+
+With this integration:
+
+- Get a summary of the error in your Datadog event stream
+- Get notified when a project has a spike in error rates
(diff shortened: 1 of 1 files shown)


content/integrations/bugsnag.md
```diff
@@ -0,0 +1,42 @@
...
+
+1. In [Bugsnag](https://bugsnag.com/) go to _Settings_ for the project you would like to configure to send notifications to Datadog
+1. Select _Team Notifications_ and then _Datadog_
+1. Customize the notifications you'll see in Datadog by selecting error notification triggers.
+1. Apply custom filters to your notification triggers to see errors from specific release stages and severities.
+1. Enter your Datadog API key.
+1. Select _Test Notification_ to test the configuration. A test error from Bugsnag will appear in Datadog's event stream.
+1. _Save_ your settings.
+1. Add more streams from the same project to see error events based on a different set of notification criteria.
+
```
technovangelist: is there any way to test that it works?

content/integrations/bugsnag.md
```diff
@@ -0,0 +1,42 @@
...
+
+### Metrics
+
+This integration does not include metrics at this time.
```
technovangelist: i have been not including the section if there is nothing to put there, but i think the inclusion and stating that it does not include metrics is much better going forward

content/integrations/bugsnag.md
irabinovitch: "1. Select _Test Notification_ to test the configuration. A test error from Bugsnag will appear in Datadog's event stream."

PR #696, opened 2016-04-05
Updated Native Database setups steps to include reference to the inte…

avaughan89:
…gration tile steps

### DIFF
--- a/content/integrations/awsrds.md
+++ b/content/integrations/awsrds.md
@@ -90,7 +90,7 @@ There are 3 options for monitoring RDS instances. You can choose to use standard
 
 * **Native Database Integration**
 
-  1.  Configure an agent and connect to your RDS instance by editing the appropriate yaml file in your conf.d directory.
+  1.  Configure an agent and connect to your RDS instance by editing the appropriate yaml file in your conf.d directory. Please complete the setup steps for the database found in their [integration tile](https://app.datadoghq.com/account/settings)  
       a.  If you are using MySQL, MariaDB, or Aurora, then edit mysql.yaml:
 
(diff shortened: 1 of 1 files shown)


irabinovitch:
@avaughan89  since most of these have docs here in documentation maybe just link them directly from here?

avaughan89:
@irabinovitch Updated to include links to the 3 docs.

technovangelist:
Im not sure i see the value of this addition. What is the flow supposed to be? It looks like you are saying, before you update the yaml file as it is shown here, go to this other page, update the yaml, then come back here and update the yaml again. So what is the value of going to the configuration on the other page??

avaughan89:
@technovangelist A customer referenced that the doc did not make it clear you needed to complete the database steps before editing the yaml file.  This change is not *necessary* but it addresses their concerns and clears up the confusion.  My original commit followed that a little closer.

technovangelist:
if the change is not needed, why have them do it

irabinovitch:
Theres definitely room for improving this doc.  I'm not sure if repeating the docs for each of the native integrations here makes sense. I like Andrew's approach of mentioning that native metrics are available, and linking to the appropriate pages.

technovangelist:
the problem was that it was a change to the docs that wasn't tested and wasn't known to make any difference. and then it was telling people to follow some instructions on a different page, but not directing them to 'which' of the page of instructions, since there were some that you should not do, even if it did make a difference....

I think this should be collapsed a bit. people choose which of the 3 integrations they want and do it. making people click around to other pages gets confusing. When i do that I can't remember where i was.

PR #703, opened 2016-04-15
correction to results for screenboard api calls

irabinovitch:
results from the screenboard.get_all api are nested under a top level screenboards array


technovangelist:
looks good

PR #704, opened 2016-04-15
first commit for catchpoint

technovangelist:


technovangelist:
oops, just noticed there is another pr for this

PR #719, opened 2016-04-29
added initconfig item to set min collect interval

technovangelist:


content/guides/agent_checks.md
```diff
@@ -182,6 +183,7 @@ instances:
       password: 5678
 EOF
 %>
+`min_collection_interval` can be added to the init_config section to define how often the check should be run. If it is greater than the interval time for the agent collector, a line will be added to the log  stating that collection for this script was skipped.
```
irabinovitch: Should we mention what the default is, or what the lowest setting it can be is?

technovangelist: Yes, that sounds good

technovangelist:
Added clarification around this.

PR #728, opened 2016-05-05
Document custom log parser return value for non-matching lines

kindlyseth:


technovangelist:
Thanks for the suggestion. The current doc says what to return if it found a value which is what is shown in the example. If there is no match, then it doesn't explicitly return anything which is the same as returning 'None'. Do you think it is necessary to call that out? I am not sure I see the need to do so.

kindlyseth:
Reading carefully through the original document (without my suggested change) again, I see no place where it is even implicitly mentioned that the parsing function is also a matching function and need not produce a value for each line. I realize that, with your knowledge of monitoring systems, this behavior is obviously the only sensible thing, but for someone like myself, who is familiar with parsers but new to monitoring, it is not at all obvious.

Assuming it's accepted that the above needs to be called out in some way, there is the question of how. While it's obvious to even slightly experienced programmers of languages with an implicit return value (ex: JavaScript returning undefined; Python returning None; etc.) that using that feature is the obvious way to indicate no match, not all Datadog users are even slightly experienced Python programmers. (In my case, it's been years since I used Python more than in passing, and I'd forgotten that detail.)

Calling `return None` out explicitly serves both purposes, and is therefore my recommendation, although I can see the case for a larger rewrite making the matching part of the parsing function's role more explicit.

technovangelist:
excellent, thanks for explaining that. I can see your point and am merging it in. Thanks so much for your help.

PR #729, opened 2016-05-09
updated hdfs docs

technovangelist:
Updated to reflect tile

### DIFF
--- a/content/integrations/hdfs.md
+++ b/content/integrations/hdfs.md
@@ -3,38 +3,72 @@ title: Datadog-Hadoop HDFS Integration
 integration_title: Hadoop HDFS
 kind: integration
 git_integration_title: hdfs
+newhlevel: true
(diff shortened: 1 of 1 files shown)


content/integrations/hdfs.md
```diff
@@ -3,38 +3,72 @@ title: Datadog-Hadoop HDFS Integration
...
+      -------------
+          - instance #0 [OK]
+          - Collected 8 metrics & 0 events
+
+
+
+
+# Metrics
 
 The metrics available are collected using df from Spotify's Snakebite. hdfs.in_use is calculated by dividing used by capacity.
```
irabinovitch: We should link to df and/or snakebite

irabinovitch: Also need to confirm this is true after the we redid the integration in 5.7

PR #732, opened 2016-05-11
'exclude' is a child element of the hashmap

thedebugger:
- conf is an array of hashmap, and 'exclude' and 'include' are hashmap
  elements


technovangelist:
Looks like you caught a hard to spot typo. Thanks so much. Merging now

PR #733, opened 2016-05-11
[Azure] Add overview section

tmichelet:
to summarize what the integration does


technovangelist:
looks good, thanks. At some point soon we should update the format to reflect the standard integration format, but this is a good interim step. Thanks.

PR #734, opened 2016-05-11
Fixing the number of jvm metrics collected out of the box.

Martiflex:
There are 11 metrics and not 13, see https://github.com/DataDog/jmxfetch/tree/master/src/main/resources and the metric list provided on the same page.


technovangelist:
so 13ish??? good catch.

PR #758, opened 2016-05-25
fix spacing in yaml blocks

jescholl:
The yaml was inconsistent throughout the documentation and the example
that ships with the agent.  Previously the examples here would result in
validation errors.


technovangelist:
doh! i looked at a more recent PR first that addressed the same issue. You noticed it first though. The issue is resolved though. 

thanks

Issue #756, opened 2016-05-21
Events documentation


ash2k:
Please document that `date_happened` should be specified in seconds for clarity. I was sending milliseconds by mistake and detailed docs would have saved me some time :)

Also descriptions of event fields are inconsistent on those two pages:

http://docs.datadoghq.com/api/#events
http://docs.datadoghq.com/guides/dogstatsd/

PR #760, opened 2016-06-01
clarify measure is in seconds and not ms

irabinovitch:
#756 suggested we clarify that we require timestamps be in seconds and not milliseconds.


irabinovitch:
@technovangelist @jyee Minor PR. Can I get a quick 👍  ?

technovangelist:
looks good. is it worth saying mentioning why this comment is relevant? Some time commands output the time in milliseconds and it can be hard to troubleshoot until you get the right value. Otherwise, looks good to go.

PR #764, opened 2016-06-02
Updated Agent Check doc to mention restarting Agent for changes to ta…

avaughan89:
…ke effect


technovangelist:
Not sure 'please' should be used in docs. Could you rephrase it to something like Restart the agent to ensure.... etc...

avaughan89:
@technovangelist Updated

technovangelist:
thanks

PR #765, opened 2016-06-02
fixes example yaml errors on JMX integration docs

matthewmain:


technovangelist:
Im all for clarifying, but there isn't a need for those indents is there? does the agent error on that? I believe its valid YAML.

matthewmain:
We received a complaint about this, and we’re seeing that in a parser the present yaml is producing null values. For example:

conf:
- include:
  domain: org.apache.cassandra.db

is yielding:

{
  "conf": [
    {
      "domain": "org.apache.cassandra.db", 
      "include": null
    }
  ]
}
Am I overlooking something?

technovangelist:
ahh, yes, you are right. I misread what that was doing

matthewmain:
cool, thanks!

PR #766, opened 2016-06-02
Corrected link to local image for windows service event

avaughan89:


technovangelist:
Perfect. Thanks for fixing this

PR #767, opened 2016-06-03
More yaml spacing fixes

jescholl:
Most of the YAML issues got fixed in #765, but these two got missed.


irabinovitch:
@jescholl Thanks for helping catch these. We will look at adding some validation in our documentation build process to catch future spacing issues.

PR #768, opened 2016-06-03
Section names in API docs should be visible links

jyee:
- Added hyperlinks to headers
- Moved IDs from header tags to divs and styled to provide vertical offset (appear below nav menu)
- Added some JS to preserve language choice via URL
- Added some JS to try to maintain scroll to visible section when switching languages (doesn't appear to work in IE10 or lower).

Trello card: https://trello.com/c/6opUrbfd/915-documentation-section-names-in-api-docs-should-visible-links


irabinovitch:
LGTM! Thanks.

PR #769, opened 2016-06-07
remove documentation about deprecated window and notify functionality

irabinovitch:
remove documentation about deprecated window and notify functionality


technovangelist:
looks good

PR #802, opened 2016-07-07
Remove trailing slash from base_url

jyee:
Per Docs at http://nanoc.ws/doc/reference/helpers/
  base_url should not have a trailing slash. The slash
  was causing the sitemap to have errant URLs with
  double slashes.


technovangelist:
good catch

PR #803, opened 2016-07-07
standardize links to example conf and check files for local integrations

technovangelist:
Many of the local integrations used different text to include links to the example config and check files. And many others didn't include any links and it's been a request from a number of support folks for a while. This update standardizes the text and adds it to every integration that runs at the agent. If we want to change the text, it's easy to do for the entire set of integrations at once


jyee:
LGTM!

PR #805, opened 2016-07-08
[AWS] update the role based authentication part now that it is the new default

deterralba:
Follows https://github.com/DataDog/dogweb/pull/13362


tmichelet:
👍

PR #810, opened 2016-07-11
[api] Add monitor tags filter for get_all

Viq111:


technovangelist:
looks good. good catch on including the ja docs as well.

Viq111:
@technovangelist Actually this wasn't added to the code yet :/ Waiting on rewording from @celenechang

PR #812, opened 2016-07-11
added integration doc for powerdns recursor

technovangelist:
also cleaned up some text in activemq doc...

### DIFF
--- a/content/integrations/powerdns.md
+++ b/content/integrations/powerdns.md
@@ -1,0 +1,50 @@
+---
+title: Datadog-PowerDNS Integration
+integration_title: PowerDNS Recursor
+kind: integration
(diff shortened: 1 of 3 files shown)


content/integrations/powerdns.md
```diff
@@ -0,0 +1,50 @@
...
+doclevel: basic
+---
+
+# Overview
+
+Connect your PowerDNS Recursor to Datadog to:
+
+* Visualize its performance
+* Understand query latency
+* Get alerted when something fails or when you're under attack.
```
jyee: One tiny formatting issue: the first two points have no period, but the last one does. I think for consistency, you should remove the period.

PR #814, opened 2016-07-18
jyee/ruby integration docs

jyee:
Ruby code samples added for both dogstatsd-ruby and dogapi-rb.
Added a couple links in the Python integration overview to match.


technovangelist:
Looks good

PR #817, opened 2016-07-19
different spark screenshot

technovangelist:


technovangelist:
oops, looks like i forgot to add the image

jyee:
:) I was wondering where that image was going. It's definitely nicer than the giant dashboard with very flat graphs.
Looks good to merge.

PR #818, opened 2016-07-20
Mattw/integrationformatchecker

technovangelist:
This new check will verify that each integration is using the new format. So it has an Overview and either an Installation and/or a Configuration section. The Overview has an image. And its using the new hlevel. 

The check only runs on the local dev box and not on the server


Checks
```diff
@@ -6,3 +7,63 @@ check :no_api_keys do
...
+        # New hlevel
+        #############
+
+        # Does the document have the right headings?
+        headings = doc.xpath("//div[@id='pagecontent']/h1").collect {|node| node.text.strip}
+        if headings.include? 'Overview'
+          if (headings & ['Configuration', 'Installation']).empty?
+            missingfiles += "\n  - Missing Config or Installation section"
+          else
+            overviewimage = doc.xpath("//img[preceding::h1[contains(text(), 'Overview')] and following::h1[(contains(text(), 'Configuration')) or (contains(text(), 'Installation'))]]/@src")
```
jyee: @technovangelist I was still getting the "missing overview image" on my [Route 53 doc](https://github.com/DataDog/documentation/pull/839/files#diff-01762e5e6b207681b1e38cce59ae38b0R13) and I think there are 2 issues:

1. Images are getting wrapped in P tags and it seems that the "preceding" selector is nearly the same as "preceding-sibling" and [doesn't include ancestors](https://www.w3.org/TR/xpath/#axes)... so when an img is on it's own, it has no preceding elements.
2. contains() also didn't seem to be working for me... and I'm not sure why. But unless someone is changing the id's, they should just be "overview", "configuration" and "installation".

In any case, this xpath seemed to work better for me:

```
overviewimage = doc.xpath("//p[preceding::h1[@id='overview'] and following::h1[(@id='configuration') or (@id='installation')]]/img/@src")
```

PR #823, opened 2016-07-20
if no navigation specified, all articles get a toc

technovangelist:
If you have a sidebar and nav defined in the meta, then you get a table of contents in the sidebar. If you don't then you get nothing. This PR changes that. If you have nothing, you get an indented table of contents that shows each h1, h2, h3,etc. Specify navigation in the meta and the auto toc is overridden, though they all have the header Table of Contents to keep things consistent.

### DIFF
--- a/lib/helpers_.rb
+++ b/lib/helpers_.rb
@@ -1,6 +1,7 @@
 include Nanoc::Helpers::XMLSitemap
 include Nanoc::Helpers::Rendering
 include Nanoc::Helpers::LinkTo
+require 'nokogiri'
 
 # general functions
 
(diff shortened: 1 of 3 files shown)


lib/helpers_.rb
```diff
@@ -1,6 +1,7 @@
 include Nanoc::Helpers::XMLSitemap
 include Nanoc::Helpers::Rendering
 include Nanoc::Helpers::LinkTo
+require 'nokogiri'
```
jyee: We'll need to add nokogiri to the gemfile so that it gets loaded on bundle install.

technovangelist: Its already part of it. nanoc needs it

jyee:
LGTM

PR #825, opened 2016-07-20
all links get little link icon

technovangelist:
You know how when you look at a readme on github, there is a little link icon to the left of each heading that appears on hover? well this adds that to our docs for everything except the top page title and the api page. 
example


jyee:
This is a nice touch for usability.

PR #830, opened 2016-07-21
FAQ markdown cleanup

jyee:
- removed html.
- renamed link references to [section-#] (e.g. `[agent-1]`) and placed them at the bottom of each section.
- converted triple backticks to singles (multi backticks are only necessary if you intend to have a literal backtick. see [here](https://daringfireball.net/projects/markdown/syntax#code)).
- tried to wrap lines as best as possible for easier reading.

### DIFF
--- a/content/faq.md
+++ b/content/faq.md
@@ -35,20 +35,22 @@ sidebar:
 -->
 
 ### Agent
-{: #agent}
 
 <!-- this is not currently true
-<h4 id="dogstatsd-flush-interval">Is the DogStatsD flush interval configurable?</h4>
-<p>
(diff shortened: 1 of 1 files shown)


technovangelist:
lgtm

technovangelist:
actually, can i make one request? try removing the nav section in the front matter. it will autogenerate a toc on the left

jyee:
Yeah, I gave that a try, but the TOC got gigantic because of all the h4s for every question. I was wondering if it'd be worthwhile to create a new frontmatter setting for autonav minimum, so e.g. you could set it to 3 and it'd only autonav for h1, h2 and h3.

PR #833, opened 2016-07-22
Update sqlserver.md

avaughan89:


technovangelist:
thanks for that. could have sworn I checked every integration to make sure those links worked.

Issue #56, opened 2013-01-07
How do I capture custom Windows performance counters in Datadog?


conorbranagan:
This should have its own guide I think, stepping through adding your performance counters to a .NET application, get the WMI class of the performance counters and handling common issues (like [counters not being published to WMI](http://blogs.msdn.com/b/dougste/archive/2009/02/11/net-performance-counters-missing-when-using-wmi-to-access-them.aspx))

We still need to figure out the step from performance counter => WMI class.

mhinze:
I found this, which is a good start: https://help.datadoghq.com/hc/en-us/articles/205016075-How-to-retrieve-WMI-metrics

PR #834, opened 2016-07-22
update wmi doc to be useful

technovangelist:
Resolves issue #56 amongst other things.


content/integrations/wmi.md
```diff
@@ -1,15 +1,134 @@
...
+
+    The metrics definitions include three components: class property in WMI, metric name as it appears in Datadog, and the metric type. 
+
+4.  Restart the agent.
+
+## Configuration Options
+
+Each WMI query has 2 required options, `class` and `metrics` and six optional options, `host`, `namespace`, `filters`, `tag_by`, `constant_tags` and `tag_queries`.
+
+`host` is the optional target of the WMI query, `localhost` is assumed by default if you set this option, make sure that Remote Management is enabled on the target host [see here](https://technet.microsoft.com/en-us/library/Hh921475.aspx) for more information.
```
jyee: I think this is missing punctuation (it's a bit of a run-on). How about:

`host` is the optional target of the WMI query and `localhost` is assumed by default if you set this option. Make sure that Remote Management is enabled on the target host. [See here](https://technet.microsoft.com/en-us/library/Hh921475.aspx) for more information.

content/integrations/wmi.md
```diff
@@ -1,15 +1,134 @@
...
+
+`class` is the name of the WMI class, for example `Win32_OperatingSystem` or `Win32_PerfFormattedData_PerfProc_Process`. You can find many of the standard class names on the [MSDN docs](http://msdn.microsoft.com/en-us/library/windows/desktop/aa394084.aspx).
+
+The `Win32_FormattedData_*` classes provide many useful performance counters by default.
+
+`namespace` is the optionnal WMI namespace to connect to (default to `cimv2`). 
+
+`provider` is the optional WMI provider (default to `32` on Datadog Agent 32-bit or `64`). It is used to request WMI data from the non-default provider. Available options are: `32` or `64`.
+For more information: [https://msdn.microsoft.com/en-us/library/aa393067.aspx](https://msdn.microsoft.com/en-us/library/aa393067.aspx)
```
jyee: I'm not sure if we have a style guide, but I usually prefer not to have plain urls. Maybe, this could be:

Reference [the MSDN docs](https://msdn.microsoft.com/en-us/library/aa393067.aspx) for more information.

content/integrations/wmi.md
```diff
@@ -1,15 +1,134 @@
...
+
+4.  Restart the agent.
+
+## Configuration Options
+
+Each WMI query has 2 required options, `class` and `metrics` and six optional options, `host`, `namespace`, `filters`, `tag_by`, `constant_tags` and `tag_queries`.
+
+`host` is the optional target of the WMI query, `localhost` is assumed by default if you set this option, make sure that Remote Management is enabled on the target host [see here](https://technet.microsoft.com/en-us/library/Hh921475.aspx) for more information.
+
+`class` is the name of the WMI class, for example `Win32_OperatingSystem` or `Win32_PerfFormattedData_PerfProc_Process`. You can find many of the standard class names on the [MSDN docs](http://msdn.microsoft.com/en-us/library/windows/desktop/aa394084.aspx).
```
jyee: Should we arrange these option explanations in the order listed in line 71? So class and metrics would be above host?

PR #836, opened 2016-07-22
[api] add event delete api docs for shell

samanthadrago:
Still waiting on PRs to get merged for Python and Ruby libraries, so just added docs for shell for now and placeholders for Py and Rb.


technovangelist:
looks good

PR #843, opened 2016-07-24
update hadoop yarn + image

technovangelist:


irabinovitch:
We seem to be inconsistent about what we're using for images.  In some docs we are show a dashboard in others just a single graph.  Given we have a fairly nice hadoop dashboard I'd lean towards using that here. What do you think?

CC: @technovangelist  @jyee

technovangelist:
I have tended to using whatever is in the integration tile. when there isn't an image then i make one up. on the rds one, the images were going to be boring like this. but an entire dashboard is a bit cluttered. so I took just the minimum, a few graphs that have some color and the RDS logo. I think its better than the alternatives. doing that here as well makes sense.

PR #859, opened 2016-07-29
Update the docs; the /intake url now needs an API key to work. Checki…

Martiflex:
…ng the machine time against the aap.agent.datadoghq.com url is simpler. See https://trello.com/c/DrHIGkIs/2129-machine-clock-time-check-with-intake-doesn-t-work-anymore


technovangelist:
nice catch. interesting that the intake url was used before. pr looks good to me.

PR #868, opened 2016-08-02
update varnish for 5.8.x release

irabinovitch:
update varnish to include instructions from 5.8 around around group membership, as well as to conform with our new docs templates.


technovangelist:
looks good to me

PR #871, opened 2016-08-02
added more info about the basic algorithm

githomin:


StephenKappel:
Seems fine to me 👍 

Can we also update adaptive's description to not use "more" as if it is being compared to a single other algorithm?

githomin:
good catch!

PR #880, opened 2016-08-05
Stephen/anomalies kwargs

StephenKappel:
Add documentation for keyword arguments in `anomalies()` function.


technovangelist:
That looks great.

StephenKappel:
At standup, we decided to leave this undocumented for now. We generally want the users to stick with default behavior, but will have this in our back pocket just in case.

PR #902, opened 2016-08-08
update chef for formatting and add install

technovangelist:


content/integrations/chef.md
```diff
@@ -2,10 +2,13 @@
...
 kind: integration
-doclevel: basic
 git_integration_title: chef
+newhlevel: true
+updated_for_agent: 5.8.5
 ---
-### Overview
+# Overview
+
+![Chef Event](/static/images/chefevent.png)
```
irabinovitch: we have an amazing chef default dashboard, can we use that instead?

PR #929, opened 2016-08-23
Adjusted series argument

jyee:
Series argument is required for console and optional for python. It
does not apply for Ruby.


technovangelist:
lgtm

PR #932, opened 2016-08-24
Translated the latest Anomaly and Outlier guides into Japanese.

toolyee:


technovangelist:
this is crashing for me. did you verify it compiles on your side?

technovangelist:
This looks much better. Thanks for updating and getting this going.

PR #934, opened 2016-08-24
add link to screenboardupdate

technovangelist:
The trello card refers to POST vs PUT, but looks like we fixed that already. It also mentioned that the link to the PUT was broken. so this fixes that issue

### DIFF
--- a/content/api/index.html
+++ b/content/api/index.html
@@ -1276,7 +1276,7 @@ kind: documentation
       </div>
     </div>
 
-    <h4 id="screenboards-post">Update a Screenboard</h4>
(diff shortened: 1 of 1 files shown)


stephenlechner:
Still looking like [the example code snippet](https://github.com/DataDog/documentation/blob/abad52b2927729d7cbff2b0045f46c52f676102f/content/api/index.html#L1299) is using "POST" instead of "PUT" and "POST" still doesn't work for screenboard updates. Anywhere we can update that example code snippet? Happy to do it myself if someone can point me where I'd be able to find it.

PR #935, opened 2016-08-25
add new metric list endpoint

technovangelist:


content/api/index.html
```diff
@@ -161,6 +161,30 @@
         </p>
       </div>
     </div>
+
+    <div class="int-anchor" id="metrics-list"></div><a href="#metrics-list"><h4>Get list of active metrics</h4></a>
+    <div class="row">
+      <%= left_side_div %>
+        <p>
+          Get the list of actively reporting metrics from a given time until now. This endpoint is not available in the Python and Ruby libraries.
```
jyee: @technovangelist: Should we just add classes to this so that it isn't shown when on Python and Ruby?

jyee:
Assuming that we don't want to hide this section for Python and Ruby, this looks good to merge.

PR #937, opened 2016-08-25
Initial commit of testing doc

jyee:
Documentation of the AgentCheckTest class that should help with people writing agent checks or new-style integrations

https://trello.com/c/F9Lm8Kst/1010-write-documentation-on-testing-checks

### DIFF
--- a/content/guides/testing.md
+++ b/content/guides/testing.md
@@ -1,0 +1,234 @@
+---
+title: Testing Integrations
+kind: guide
+listorder: 6
+autotocdepth: 2
+---
+<style>
+.func-p{margin-left:20px;}
+.func-code{margin-left:40px;}
+</style>
+
+### Overview
+
+Automated unit tests help ensure that your integration is working as designed and guard against regression.
+
+Though we don't require tests for each metric collected, we strongly encourage you to provide as much coverage as possible.
+
+### Adding Tests
+
+Tests can be easily added by extending the `AgentCheckTest` class. Additionally, you may import Attributes from Nose to manage requirements.
+
+~~~
+from nose.plugins.attrib import attr
+from tests.checks.common import AgentCheckTest
+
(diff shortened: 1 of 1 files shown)


content/guides/testing.md
```diff
@@ -0,0 +1,234 @@
...
+autotocdepth: 2
+---
+<style>
+.func-p{margin-left:20px;}
+.func-code{margin-left:40px;}
+</style>
+
+### Overview
+
+Automated unit tests help ensure that your integration is working as designed and guard against regression.
```
irabinovitch: *guards
*regressions

content/guides/testing.md
```diff
@@ -0,0 +1,234 @@
...
+
+Though we don't require tests for each metric collected, we strongly encourage you to provide as much coverage as possible.
```
irabinovitch: Do we we want to set some minimum bar for inclusion? 

CC: @remh , @gmmeyer , @truthbk

irabinovitch:
Great start! Should we plan a seperate doc on how to write good tests? Or walking someone through the anatomy of a simple test?

content/guides/testing.md
gmmeyer: I'm not sure, the minimum bar should probably be testing all of the important functionality, but that doesn't necessarily mean testing every metric.

irabinovitch: "testing the important things" is probably hard to measure as well.

irabinovitch: @gmmeyer @truthbk @remh Can we come to consensus on what the minimum requirement for merging a check is?

content/guides/testing.md
jyee: This should be good right? I think the only difference is guard vs. guards - and it should be guard, since the subject is plural tests. That is "tests guard against regression"

PR #939, opened 2016-08-25
Add AWS Elasticsearch list and describe permissions

jyee:
https://trello.com/c/110jk43Y/1078-aws-elasticsearch


technovangelist:
lgtm

irabinovitch:
l'm not sure we need these. Seems like we get these these metrics cloudwatch only. I dont believe we call list/describe?

tmichelet:
Up to date permissions can be find at https://gist.github.com/tmichelet/fc51d6c1db4b2b6b6e6cbe76746a9488

We don't need those today. However, we'll add a `ES` integration at some point, so in the future we will

PR #940, opened 2016-08-26
jp integration missing the beta filter

technovangelist:


jyee:
lgtm.
Although, maybe remove the () and add spaces around the != to match the rest of the code style?

PR #962, opened 2016-09-13
removing warning that no one can explain.

githomin:
also removed line about incorporating as_count and rollup into the query editor as that has happened.


mstepniowski:
lgtm 👍

PR #964, opened 2016-09-14
Better anomalies docs

StephenKappel:
https://trello.com/c/HNRN1PED/436-add-faq-about-using-rollups-to-the-documentation
https://trello.com/c/X1OK1v37/419-revision-pass-on-ad-guide


content/guides/anomalies.md
```diff
@@ -45,10 +45,10 @@ Start by navigating to the [New Monitor](https://app.datadoghq.com/monitors#/cre
...
+  <li>This number is equivalent to the <code>bounds</code> parameter used in the <code>anomalies</code> function in dashboards; it controls the width of the gray band. We recommend using a value of 2 or 3.</li>
   <li>If you only care about unusually high or unusually low values, you can choose to only alert on values above or below the bounds.</li>
-  <li>As with other alerts, smaller time windows lead to faster alerting but can yield more false positives.</li>
-  <li>You can change the anomaly detection algorithm used here.</li>
+  <li>We recommend using a window size of at least 15 minutes. (A 30 minute window works well in most cases.) With small windows (<= 10 minutes), metrics often appear noisy, making it difficult to visualize the difference between anomalies and noise. Note that setting this window size to X minutes doesn't require an anomaly to last X minutes before an alert is triggered. You can tune the threshold to control how long an anomaly must last to trigger an alert. For example, with this window size set to 30 minutes, you can can alerted when an anomaly lasts for just five minutes by setting the threshold to 5/30 = 17%.</li>
```
githomin: This is too much information here. I'd say after the 30 min suggestion, say go to FAQ for more.

content/guides/anomalies.md
```diff
@@ -57,13 +57,13 @@ Continue with steps (3) and (4) as you would for any other monitor.
 
 We currently offer four different anomaly detection algorithms.
 
-* Basic: This uses a simple lagging rolling quantile computation to determine the range of expected values. It uses very little data, and adjusts quickly to changing conditions. However, it has no knowledge of seasonal behavior or longer trends.
+* _Basic_: Use this algorithm for metrics that have no repeating seasonal pattern. _Basic_ uses a simple lagging rolling quantile computation to determine the range of expected values. It uses very little data, and adjusts quickly to changing conditions but has no knowledge of seasonal behavior or longer trends.
```
githomin: ", but it"

githomin: remove comma before and

content/guides/anomalies.md
```diff
@@ -90,3 +90,19 @@ The gray band here makes sense; it is wide enough to capture the noise in the ti
 Again, the band seems to be reasonably sized, because the non-anomalous data from 8:50 - 9:00 and from 9:10 - 9:20 is inside the band. A band any narrower would start to highlight normal data as anomalous. Notice the band in this graph is ~8x wider than the one in the previous graph. The anomalous period from 9:00 - 9:10 looks a little different from the rest of the series, but it is not extreme enough to fall outside of the band.
 
 In general, if an anomaly disappears when you zoom in, this doesn't mean that it's not an anomaly. It means that, while the individual points in the zoomed-in view are not anomalous in isolation, the fact that many slightly unusual points occur together is anomalous.
+
+### How can I alert on drops in a metric that routinely has points at/near zero?
```
githomin: Maybe this should be: "How do I narrow the bounds for shorter timespans?"

StephenKappel: I'm not a huge fan or my original version, but not sure if this is any clearer.

githomin:
Add the word "seasonal" to the phrase "taking into account (seasonal) day-of-week and time-of-day patterns" in the first graph (re: corey's comment) https://trello.com/c/X1OK1v37

StephenKappel:
Re: Cory's comments -- I'm not sure if adding the word seasonal actually answers his question any more than the sentence already does. Maybe we should add a FAQ (or two) stating that we use weekly seasonality with up to 6 weeks of historical data?

githomin:
Agreed, but it'd be good to include that (or something like it) in case someone wonders what "seasonal" means. As for the amount of data, maybe we should say that the last 3 algs use "a couple months" of data? I'd rather not specify seasonality=weekly, lookbacks=6 in the docs

content/guides/anomalies.md
```diff
@@ -73,6 +73,14 @@ No. Anomaly detection is designed to assist with visualizing and monitoring metr
 
 Also, anomaly detection requires historical data to make good predictions. If you have only been collecting a metric for a few hours or a few days, anomaly detection probably won't be very useful.
 
+### For the seasonal algorithms (_robust_, _agile_, and _adaptive_), how much history is analyzed? What types of seasons are included?
+
+All of the seasonal algorithms may use up to a couple of months of historical data when calculating a metric's expected normal range of behavior. By using a significant amount of past data, the algorithms are able to avoid giving too much weight to abnormal behavior that might have occurred in the recent past.
+
+_Robust_ and _agile_ assume that the metric has weekly seasonality.
```
githomin: can we take out these lines? we'll probably use auto-seasonality detection soon

githomin: and below too

content/guides/anomalies.md
```diff
@@ -91,9 +99,11 @@ Again, the band seems to be reasonably sized, because the non-anomalous data fro
...
 
-### How can I alert on drops in a metric that routinely has points at/near zero?
+### Is it possible to capture anomalies that occur within the bounds?
+
+If the reason anomalies are occurring within the bounds is that the volatility of a metric leads to wide bounds that mask true anomalies (as described in the FAQ above), you may be able apply functions to the series to reduce its volatility, leading to narrower bounds and better anomaly detection.
 
-Many important metrics represent the success of some user-driven action. For example, `successful.logins` or `checkout.completed`, etc. It can be useful to monitor for anomalous drops in one of those metrics, as this may be an indication that something is preventing successful completion of these events and that the user experience is suffering.
+For example, many important metrics (e.g `successful.logins`, `checkouts.completed`, etc.) represent the success of some user-driven action. It can be useful to monitor for anomalous drops in one of those metrics, as this may be an indication that something is preventing successful completion of these events and that the user experience is suffering.
```
githomin: e.g.,

content/guides/anomalies.md
```diff
@@ -90,3 +98,21 @@ The gray band here makes sense; it is wide enough to capture the noise in the ti
...
+
+It's common that these metrics have points that are at or near zero, especially when viewing the metric over a short window of time. Unfortunately, this results in the bounds of the anomaly detection forecast include zero, making it impossible to detect anomalous drops in the metric. An example is shown below.
+
+<img src="/static/images/anomalies/raw_profile_updates.png" style="width:500px; border:1px solid #777777"/>
+
+How can we work around this problem? One approach is to add a `rollup()` to force the use of a larger interval. `rollup()` takes as an argument the number of seconds that should be aggregated into a single point on the graph. For example, applying `rollup(120)` will lead to a series with one point every two minutes. With larger intervals, zeros become rare and can correctly be categorized as anomalies. Here's the same series as above but with a 2-minute rollup applied.
```
githomin: Can you make this more specific? Like "when taking sums of count metrics..."  and then `rollup(sum, 120)`.

content/guides/anomalies.md
StephenKappel: shall we just remove this FAQ all together?

githomin: i guess that makes it a bit short. maybe add the graph about data to the algorithms section? Like "We currently offer four different anomaly detection algorithms. All of the seasonal... "

content/guides/anomalies.md
StephenKappel: I intentionally left the time aggregator out, because I thought it would only make it more confusing. While sum is the correct aggregator for counts, it will not be the right choice for other situations. By only advising them to override the interval, I'm hoping to avoid folks overriding the time aggregator incorrectly.

githomin: It's confusing because you're using sum in your example though. What if you say, "One approach that works well for sums of count metrics is to add..."

StephenKappel: Oh. Maybe I should fix my example to remove the sum. I forgot I put that in the title. The chart will look exactly the same with or without including sum in the function call.

githomin:
👍

PR #966, opened 2016-09-15
Mattw/jenkins59

technovangelist:


irabinovitch:
@technovangelist  we probably still want to have docs here about the jenkins plugin, but im not sure if that goes in another page.  

Having a deprecation note about the agent-check makes sense. We should add one in the YAML config in dd-agent, and in the tile as well.

technovangelist:
keeping docs on the plugin that is maintained in a different repo increases chances that it's going to get out of sync. I guess the same could be said about other integrations, but this seems even more likely here.

irabinovitch:
We develop the plugin, so having things like metrics metrics metadata and other similar niceties of our docs site keeps it on par with our other integrations.

technovangelist:
cool, will update

technovangelist:
ok, integration updated

PR #968, opened 2016-09-16
added warning about multi-alerts

githomin:


content/guides/anomalies.md
```diff
@@ -71,10 +71,18 @@ All of the seasonal algorithms may use up to a couple of months of historical da
...
+No. Anomaly detection is designed to assist with visualizing and monitoring metrics that have predictable patterns. For example, `my_site.page_views{*}` might be driven by user traffic and thus vary predictably by time of day and day of week. If your metric does not have any sort of repeated/predictable pattern, then a simple chart overlay or threshold alert might be better than anomaly detection.
 
 Also, anomaly detection requires historical data to make good predictions. If you have only been collecting a metric for a few hours or a few days, anomaly detection probably won't be very useful.
 
+Take care when creating multi-alerts. A metric such as `service.requests_served{*}` could be a good candidate for anomaly detection, but `service.requests_served{*} by {host}`is probably not. If your hosts are load-balanced, then an [outlier monitor](https://docs.datadoghq.com/guides/outliers/) will be better for detecting hosts that are behaving abnormally. If your service scales up, each new host won’t be monitored at all until there is a minimum amount of history for anomaly detection to kick in, and even then the predictions are not likely to be good.
```
StephenKappel: `even then the predictions are not likely to be good` <-- makes it sounds like the algorithm is bad. 

`even then alerts might be noisy due to instability in the number of requests handled by a single host`?

content/guides/anomalies.md
```diff
@@ -71,10 +71,18 @@ All of the seasonal algorithms may use up to a couple of months of historical da
...
+
+### How should I set the window size and alert threshold?
+
+Smaller window sizes will lead to faster alerts, however, with very small windows (<= 10 minutes), metrics often appear noisy, making it difficult to visualize the difference between anomalies and noise.
+
+Note that setting this window size to X minutes doesn't require an anomaly to last X minutes before an alert is triggered. You can tune the threshold to control how long an anomaly must last to trigger an alert. For example, with this window size set to 30 minutes, you can can alerted when an anomaly lasts for just five minutes by setting the threshold to 5/30 = 17%. That said, we have found that anomaly alerts are most reliable when the window size is between 15 minutes and an hour and the threshold is on the higher side (> 70%).
```
StephenKappel: i'd prefer `this window size` --> `the window size` in a couple of spots

StephenKappel: I'm not sure I agree with the last sentence. I think a 25% threshold on a 30 minute window would be a good alert.

githomin: At least from what I've seen in #anomalies-test the < 50 % ones have all been very noisy.

githomin: i'll move that 70 down to 40?

StephenKappel: yeah; 40 or 50 would be better 👍

PR #972, opened 2016-09-19
Mattw/core extras integrations install

technovangelist:


jyee:
LGTM, I think we might want to add a link to the integration sdk guide (though not entirely necessary). Also, I should set the list order on the sdk guide so that it follows directly after this on the list of guides

technovangelist:
sounds good, @jyee . the doc didn't exist before, but I have added a link to it now.

PR #976, opened 2016-09-20
Remove mentions to the kubernetes docker tag

hkaj:
The tag will soon be deprecated and its replacement is already available. Let's prepare a smooth transition.


irabinovitch:
@hkaj  should this wait for 5.9?

hkaj:
It's ready and not tied to the agent release as it's docker related so I don't think we need to. However I have to open other PRs related to this change (in dogweb notably) so let's wait for that.

PR #978, opened 2016-09-22
updated images and edited addl features in saml

technovangelist:


irabinovitch:
@technovangelist @jyee We have a number of KB articles about how to work with individual SAML providers. We should link to these.

technovangelist:
added links to the specific providers

PR #980, opened 2016-09-22
first commit to move to using local repos vs gh

technovangelist:
this is the first try at getting metrics to be pulled from a local clone. 

Not sure how to get staging to clone dogweb, extras, and core. There is a new rake task: `rake pull` that will pull all those repos to the parent of documentation, but not sure if roy has access to clone dogweb. Also, haven't tried to see what happens if external tries to build this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ tested.code
 .sass-cache
 github_metrics
 fileformaterrors
+localmetricsstore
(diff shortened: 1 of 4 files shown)


technovangelist:
i think i am going to switch to doing it the old way on roy since that works, but if running local in dev then use the new process.

irabinovitch:
I don't think it makes sense to support two methods of this. Just more code to maintain.

technovangelist:
closing since i need to think about this some

PR #985, opened 2016-09-22
fixed br tag error

technovangelist:
every time you build locally, you get 2 errors about an unclosed br tag. this fixes that.

### DIFF
--- a/content/integrations/ntp.md
+++ b/content/integrations/ntp.md
@@ -57,7 +57,7 @@ To validate your installation and configuration, restart the agent and execute t
 
 # Metrics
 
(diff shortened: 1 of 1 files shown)


irabinovitch:
Shouldn't this just  come from the metadata.csv in dogweb?

technovangelist:
it would if it existed there

irabinovitch:
If it's missing would it make more sense to add it there, than manually here in the docs?

technovangelist:
agree, will post ticket to appropriate team. I think there are a couple other integrations that have no folders in dogweb so will go through all of them to post those to that team too too

PR #991, opened 2016-09-27
Azure integration add new services

DanielLanger:
One liner to add new azure services we now support


technovangelist:
For a one liner change, why are you including a lot of jira files?

PR #999, opened 2016-09-29
Added name & monitor tags filters to api

lypnol:


content/api/index.html
```diff
@@ -790,10 +790,16 @@
...
+          <%= argument 'name', "A string to filter monitors by name",
+          :default => "None" %>
           <%= argument 'tags', "A comma separated list indicating
           what tags, if any, should be used to filter the list of monitors
           by scope",
           :default => "None" %>
+          <%= argument 'monitor_tags', "A comma separated list indicatinig
+          what custom tags/services, if any, should be used to filter the list
+          of monitors",
+          :default => "None" %>
```
celenechang: How about: "A comma-separated list indicating what service and/or custom tags, if any, should be used to filter the list of monitors"

celenechang: of course, this will only apply after the 'application' -> 'service' migration..

PR #1012, opened 2016-10-13
add note that heapster must be running in standalone mode

irabinovitch:
add note that heapster must be running in standalone mode


jyee:
Is there a better link to stand alone docs? The link to the repo doesn't seem as helpful as it ought to be.

Maybe this documentation? https://github.com/vishh/heapster-1/blob/master/docs/standalone.md
it's a fork of the main heapster repo, but it looks like the current master for heapster no longer includes those docs.

jyee:
Or I guess we could link to https://github.com/kubernetes/heapster/blob/heapster-old/docs/standalone.md (old branch)

technovangelist:
hmm, 2 months. Any updates needed or are we good to go @jyee @irabinovitch

irabinovitch:
I haven't found a better repo / docs to link to.

PR #1020, opened 2016-10-20
fix histogram tags example

sseg:
Examples in the guides were not sending any data.


technovangelist:
Oh you wanted content in that histogram??? doh!

PR #1025, opened 2016-10-27
Add better messaging when a metadata CSV cannot be read

jyee:
Add better messaging when a metadata CSV cannot be read to help identify bad files from dogweb.


jyee:
I should have probably broken this up into two different branches.

The first commit just adds better error messaging if the CSV parser fails to parse the metadata csv. 

The second commit allows you to set an environment variable for a dogweb branch (or better, pass it on the command line to ruby, e.g. `rake clean && rake branch="jyee/my_dogweb_branch"`). This should let us better test when things break on that end.

irabinovitch:
@technovangelist any concerns?

technovangelist:
cool, works well. sorry for the delay on my part

PR #1047, opened 2016-11-19
Fix Screenboard Update with PUT (not POST)

jyee:
https://trello.com/c/DsrqN2hr/995-screenboard-update-endpoint-uses-put-currently-a-doc-say-post


stephenlechner:
@jyee Thank you so much for this! very sorry i missed your note earlier in the trello card, and thanks again for finding the time to fix this on your end. Really appreciate it.

PR #1058, opened 2016-12-05
Update No Data timeframe default for monitor API

Ricky-Thomas:
Currently if a monitor is created without the no_data_timeframe option, it defaults to 1x the alert timeframe, rather than 2x


technovangelist:
Looks good, thanks.

PR #1065, opened 2016-12-16
Clarified group_states, added usage to example code

jyee:
https://trello.com/c/vpaR4DLe/1208-update-get-all-monitors-function-arguments-definition-in-ruby

Because group_states isn't clear (and a bit hard to describe), I updated the argument description (both get_monitor and get_all_monitors) and added it to the example in the get_monitor function.


irabinovitch:
@technovangelist  mind doing a quick review so we can merge?

PR #1068, opened 2016-12-22
[outliers] add documentation for outliers scaled algos

jeerim:
Add descriptions for the outliers scaled algos, scaledMAD and scaledDBSCAN.


githomin:
I'd use "ScaledDBSCAN" instead of "Scaled DBSCAN" throughout. Also, instead of outliers being "spurious", I would put it as: if scale-free outliers don't interest you, or something like that.

content/guides/outliers.md
```diff
@@ -37,11 +37,11 @@ To set up an outlier detection graph for your data you add a metric to the graph
 
 To do so, create a new timeseries graph on your dashboard with your chosen metric. Your screen should look like:
 
-![](/static/images/outliers/outliers-dash-choose-metrics-newer.png)
+![](/static/images/outliers/outliers-dash-choose-metrics-updated.png)
 
-Now, click on the + icon (Add functions and modifiers) on the right side of the second metrics line. In the "Modify your query" box, choose the "outliers" function:
+Now, click on the + icon (Add functions and modifiers) on the right side of the metrics line. In the popup box, choose "Algorithms" then one of the four outlier algorithms:
```
githomin: Now, click on the + icon on the right side of the metrics line. Choose "Algorithms" from the function categories, then one of the four outlier algorithms:

content/guides/outliers.md
```diff
@@ -77,9 +77,9 @@ To ensure that your alert is properly calibrated, you can set the time window at
 {: #algorithms}
 
 
-There are two different outlier detection algorithms you can use on your data: DBSCAN and Median Absolute Deviation (MAD). We recommend starting with the default algorithm, DBSCAN. If you have trouble detecting the right outliers, you can adjust the parameters to DBSCAN or try the alternate algorithm, MAD. Explanation of each algorithm and its parameters follows.
+There are two different types of outlier detection algorithms you can use on your data: DBSCAN/Scaled DBSCAN and MAD (Median Absolute Deviation)/Scaled MAD. We recommend starting with the default algorithm, DBSCAN. If you have trouble detecting the right outliers, you can adjust the parameters to DBSCAN or try the alternate algorithm, MAD. If your metrics have a larger scale and you are seeing outliers within a closely clustered group, try the scaled algorithms, Scaled DBSCAN or Scaled MAD. Explanation of each algorithm and its parameters follows.
```
githomin: ScaledDBSCAN, ScaledMAD

githomin: Don't write out MAD here.

githomin: "If you have metrics on a larger scale that look to be closely clustered where the DBSCAN/MAD algorithms are marking some as outliers, and this is undesired behavior, try the scaled algorithms."

content/guides/outliers.md
```diff
@@ -91,15 +91,25 @@ Traditionally, DBSCAN takes: 1) a parameter 𝜀 that specifies a distance thres
...
+
+The ScaledDBSCAN algorithm scales the distance threshold according to the relative magnitudes of the median series and the hosts’ distances to the median series. In most situations, it will behave the same as regular DBSCAN does. However, when the median series is large compared to the distances to the median series, the distance threshold becomes proportional to the size of the median series. As a result, assessing whether two time series are close depends on the scale of the median series.
+
+Here is a comparison of DBSCAN and ScaledDBSCAN with tolerances of 3 on field data size in a group of Elasticsearch nodes:
+
+![](/static/images/outliers/outliers-scaled-dbscan-es.png)
+
+### Median Absolute Deviation (MAD)/ScaledMAD
```
githomin: Just "MAD" here.

content/guides/outliers.md
```diff
@@ -91,15 +91,25 @@ Traditionally, DBSCAN takes: 1) a parameter 𝜀 that specifies a distance thres
...
 The only parameter we take is `tolerance`, the constant by which the initial threshold is multiplied to yield DBSCAN’s distance parameter 𝜀. Here is DBSCAN with a tolerance of 3.0 in action on a pool of Cassandra workers:
 
 ![](/static/images/outliers/outliers-dbscan-cassandra.png)
 
 You should set the tolerance parameter depending on how similarly you expect your group of hosts to behave—larger values allow for more tolerance in how much a host can deviate from its peers.
 
-### Median Absolute Deviation (MAD)
+#### ScaledDBSCAN
+
+The distance threshold of the DBSCAN algorithm is independent of the overall scale of the metrics. This can lead to seemingly spurious outliers when the overall scale of the metrics is large compared to the distances between hosts and the median time series. Consider a group of constant time series with values {1000, 1001, 1002, 1005, 1015}. The median series will be a constant series at 1002. DBSCAN with a tolerance of 3.0 will identify the series at 1015 to be an outlier, even though it may be almost indistinguishable from the other series visually on the graph.
```
githomin: strike this sentence: This can lead to seemingly spurious outliers when the overall scale of the metrics is large compared to the distances between hosts and the median time series.

githomin: visually on the graph with the y-axis fixed at 0.

content/guides/outliers.md
```diff
@@ -115,14 +125,26 @@ Now to mark a time series as an outlier, we use the second parameter, `pct`. If
...
+Like for DBSCAN, the MAD algorithm is designed to be independent of the overall magnitude of the metrics. If D = {1000, 1001, 1002, 1005, 1010}, the median is 1002, and the MAD is 2. Even though the point at 1010 seems close to the median in terms of their relative scales, it is still an outlier point for a tolerance of 3.
 
-So which algorithm should you use? For most outliers, both algorithms will perform well at the default settings. However, there are subtle cases where one algorithm is more appropriate than the other.
+The ScaledMAD algorithm, like ScaledDBSCAN, considers the relative scales of the MAD and the median. In most cases, it will behave the same as the MAD algorithm does. However, when the MAD of the data set becomes small compared to the median, the measure of deviation becomes proportional to the median. Therefore, determining whether a point is an outlier depends on the overall scale of the metrics.
+
+Here is an example of MAD and ScaledMAD algorithms for comparing the usable memory in Cassandra hosts. Both have tolerances of 3 and pct of 20:
+
+![](/static/images/outliers/outliers-scaled-mad-cassandra.png)
+
+### DBSCAN/ScaledDBSCAN vs. MAD/ScaledMAD
```
githomin: DBSCAN vs. MAD / Scaled vs. Not Scaled

irabinovitch:
Do we have a particular date we want to launch these? Can you ping someone on #documentation or #baklava for a quick review of large docs updates?

githomin:
Sure, we didn't think this qualified as a large update, but the more reviews the better. We don't have a particular date in mind; just whenever the docs and such are finished.

content/guides/outliers.md
jeerim: "If you have metrics on a larger scale that look to be closely clustered but DBSCAN/MAD algorithms are identifying some as outliers, try the scaled algorithms." (I think when this is undesired behavior is implicit)

content/guides/outliers.md
jeerim: Since all our graphs have a fixed y-axis starting from 0, this seems redundant.

content/guides/outliers.md
githomin: add a "the" in front of DBSCAN but 👍

content/guides/outliers.md
githomin: It's the default but it can be changed. I think it's worth noting since it's the reason the series look close.

jeerim: "...visually on the graph, especially when the y-axis is fixed to start at 0"

jeerim: "...visually on the graph when the origin of the y-axis is at 0"

githomin:
line 22:  "We’ve added a new query function..." should be changed since it's not new anymore.

githomin:
line 36: "by adding *the* outliers"

githomin:
line 60:  stick to 2nd person; instead of "we'd", "you can", and "the" instead of "our"

githomin:
line 68: strike redundant lines: "By default we have chosen DBSCAN with a tolerance value of 3 because this works for many cases. More information about the outlier functions and their parameters is available below."

githomin:
(sorry for all the edits to parts you didn't write; they're probably all my fault)

githomin:
👍

jeerim:
Changing line 68 to "Then select an algorithm and parameter values to use for outlier detection."

content/guides/outliers.md
```diff
@@ -19,7 +19,7 @@ Outlier Detection is an algorithmic feature that allows you to detect when some
 
 ## How to Use Outlier Detection on Your Data
 
-We’ve added a new query function called `outliers` to our query language. This function will return the usual results but outlier series will be marked.
+We’ve added a query function called `outliers` to our query language. This function will return the usual results but outlier series will be marked.
```
technovangelist: i don't think we should have language like 'we have added...'. This could be in the docs for a long time and there are no dates on the docs, so when someone reads it in 2 years, did we just do it last week? Just describe the function.

content/guides/outliers.md
```diff
@@ -33,15 +33,15 @@ For example, here is a graph of gunicorn requests by host with outlier detection
 
 You can see that one of the series is an outlier: it is handling significantly lower traffic than the others for the time window in question.
 
-To set up an outlier detection graph for your data you add a metric to the graph showing all series in the groups. You apply the outlier detection algorithm by adding `outliers` function on your data. After applying the function, outlier series will be colored with a bold, warm palette, while all other series will be colored with a lightweight, greyscale color palette.
+To set up an outlier detection graph for your data you add a metric to the graph showing all series in the groups. You apply the outlier detection algorithm by adding the `outliers` function on your data. After applying the function, outlier series will be colored with a bold, warm palette, while all other series will be colored with a lightweight, greyscale color palette.
 
 To do so, create a new timeseries graph on your dashboard with your chosen metric. Your screen should look like:
```
technovangelist: there are two sentences that talk about setting up an outlier detection graph...or at least I assume there is. The second says 'to do so' so I have to guess what its about. can you clarify this? If i got it right, then which of the two steps do I follow?

content/guides/outliers.md
```diff
@@ -33,15 +33,15 @@ For example, here is a graph of gunicorn requests by host with outlier detection
...
 
-![](/static/images/outliers/outliers-dash-choose-metrics-newer.png)
+![](/static/images/outliers/outliers-dash-choose-metrics-updated.png)
 
-Now, click on the + icon (Add functions and modifiers) on the right side of the second metrics line. In the "Modify your query" box, choose the "outliers" function:
+Now, click on the + icon on the right side of the metrics line. Choose "Algorithms" from the function categories, then one of the four outlier algorithms:
```
technovangelist: don't assume that images will be shown. mention them then let the image support it.

content/guides/outliers.md
```diff
@@ -91,15 +91,25 @@ Traditionally, DBSCAN takes: 1) a parameter 𝜀 that specifies a distance thres
 
 We use a simplified form of DBSCAN to detect outliers on time series. We consider each host to be a point in d-dimensions, where d is the number of elements in the time series. Any point can agglomerate, and any point that is not in the largest cluster will be considered an outlier.
 
-We set the initial distance threshold as follows. We create a new median time series by taking the median of the values from the existing time series at every time point. Then we calculate the (Euclidean) distance between each host and the median series. The threshold is the median of those distances, multiplied by a normalizing constant
+We set the initial distance threshold as follows. We create a new median time series by taking the median of the values from the existing time series at every time point. Then we calculate the (Euclidean) distance between each host and the median series. The threshold is the median of those distances, multiplied by a normalizing constant.
```
technovangelist: why is euclidean in parens? do you need them?

content/guides/outliers.md
jeerim: Yes, this is a bit awkward. I've changed the wording a bit - hopefully this reads better.

content/guides/outliers.md
jeerim: Good point. Changing to  'The `outliers` query function, when applied to your query, will return the usual results but with outlier series marked.'

content/guides/outliers.md
jeerim: Yes, this is a bit awkward. The second sentence is an elaboration of the first. I've changed the wording a bit - hopefully this reads better.

content/guides/outliers.md
jeerim: Removed 'Your screen should look like:'

content/guides/outliers.md
jeerim: Yes, I think this is necessary, since there are many definitions of distance and we're using the Euclidean one.

technovangelist: so whats the benefit of the parens?

jeerim: Ah, I misunderstood. I don't think it really matters whether it's in parenthesis or not. I can remove them.

PR #1073, opened 2016-12-29
[API] Fix Time Variables so they make sense

ziquanmiao:
Setting EndTime to Curr_time + 100 doesn't make sense, the data within the time frame doesn't exist yet

### DIFF
--- a/code_snippets/api-events-stream.rb
+++ b/code_snippets/api-events-stream.rb
@@ -6,7 +6,7 @@ app_key='87ce4a24b5553d2e482ea8a8500e71b8ad4554ff'
 
 dog = Dogapi::Client.new(api_key, app_key)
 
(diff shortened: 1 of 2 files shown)


code_snippets/api-events-stream.rb
```diff
@@ -6,7 +6,7 @@
 
 dog = Dogapi::Client.new(api_key, app_key)
 
-start_time = Time.now.to_i
-end_time = Time.now.to_i + 100
+end_time = Time.now.to_i
```
jyee: Looks like an extra space at the end here, but once you delete that, it's good to merge.

ziquanmiao:
Sounds good. I removed the whitespace character right after `+end_time = Time.now.to_i`

PR #1084, opened 2017-01-13
Documented metric alias creation via json for graph

PerplexedSphex:


content/graphingjson.md
```diff
@@ -190,6 +190,33 @@ or if you're looking to display all errors:
...
+      {
+        "q": "avg:system.load.1{*}",
+        "type": "line",
+        "conditional_formats": [],
+        "aggregator": "avg",
+        "alias": "Your metric"
+      }
+    ]
+
+Note that you can only define one alias per JSON expression.
```
irabinovitch: But we can still define multiple per graph?

PerplexedSphex: Correct: https://cl.ly/3q242z0M2J2P

PR #1096, opened 2017-01-18
Added filter

technovangelist:
Update common filters in config file to get links added to phrases in the docs

So the config yaml file has a phrase "info command" and a link that points to the agent guide. Now every mention of info command in the docs will get replaced with a [info command](http://docs.datadoghq.com/guides/basic_agent_usage/ubuntu/). You can add as many other phrase/link combos as you like. info command was the reason for adding this.

### DIFF
--- a/Rules
+++ b/Rules
@@ -158,8 +158,8 @@ compile '*' do
 
     unless id.match('/api/')
       filter :header_links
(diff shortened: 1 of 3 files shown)


jyee:
LGTM

PR #1097, opened 2017-01-18
If an integration is marked beta, message appears at top

technovangelist:
The message is labeled beta and describes what beta means.


jyee:
LGTM!

PR #1101, opened 2017-01-18
Auto populates ja directory with files that have no translation

technovangelist:
function looks for all english pages, all japanese pages, and creates a list of pages that don't exist in japanese that do in english. Then it adds an item to the @items array that is the same as the english version except that the language metadata is 'ja' and it has a info block that says this is a translation.


technovangelist:
thanks @jhotta for adding the translation

PR #1108, opened 2017-01-19
Change 'monitoring' to 'monitors'

jyee:
Updates the "Guide to Monitoring" to "Guide to Monitors", including redirects for english and japanese guides.

Trello: https://trello.com/c/TriP8mkY/1286-update-guide-to-monitoring-title


technovangelist:
Why?

technovangelist:
since this is just about a name change, this looks good. but lets think about taking this down to just a single page rather than 2.

PR #1114, opened 2017-01-20
added `put` statements to log auto generated pages (original source list)

jhotta:


lib/generate_ja_pages.rb
```diff
@@ -7,7 +7,9 @@ def generate_ja_pages
 
   missingpages = english_pages-japanese_pages
 
-
+  File.open("autogen_files.log", "w") do |f|
+    f.puts(missingpages)
```
technovangelist: can you update this so that it only does it for us and not when jenkins runs?

technovangelist:
is this pr still necessary with pr #1115?

irabinovitch:
@jhotta Did you see Matt's question?

lib/generate_ja_pages.rb
jhotta: Apology!
I will work on it.

PR #1132, opened 2017-02-06
Jyee/trace agent api

jyee:
Tracing API docs.
https://trello.com/c/BhaGvHxk/1277-add-apm-apis-to-api-docs


awang:
Looks great to me.  At some point we need to have better documentation around service vs resource vs name terminology, that one is tracked here: https://trello.com/c/yJP5BKyG/1007-better-span-name-docs

irabinovitch:
@jyee CAn we push this to staging so we can review there?

PR #1138, opened 2017-02-08
checked and updated translation on libraries  page

jhotta:


irabinovitch:
translation page was fairly heavily refactored, we'll need to revisit translating.

PR #1147, opened 2017-02-10
fix 'dog' binary path for splunk integration

alvin-huang:
This documentation hasn't changed in two years so this is probably a result of that. 

The example script refers to `dog` directly while the current binary absolute path of `/opt/datadog-agent/bin/dog` should be called instead. This is probably because at some points `dog` was inside `/usr/local/bin/dog`, part of $PATH, or was symlinked but is no longer there (sorry I haven't been using Datadog that long to know about past releases).


technovangelist:
Thanks for the pr and for doing both en and ja. I know it worked last time I checked less than six months ago, but being more specific is probably a good thing. I verified with the devs that dog is always going to be at /opt/datadog-agent/bin going forward from v 5.7.

Thanks, merging.

PR #1154, opened 2017-02-13
updated headers and linkes

jhotta:


technovangelist:
Looks good, didn't realize there were some prs that were not translation

PR #1161, opened 2017-02-14
fixing a bug that stoping compiling html

jhotta:
RuntimeError: Could not parse YAML for content/libraries.md: (<unknown>): mapping values are not allowed in this context at line 13 column 11


CharlyF:
👍

PR #1175, opened 2017-02-21
Fix monitor downtime update bash example (API)

joemeszaros:
# Problem

The bash example, which updates a monitor downtime contains a bad comment. It creates a downtime which is then updated, but the comment line is referring to delete:

`# Create a downtime to delete`

# Solution

Update the comment in the example to 

`# Create a downtime to update`


technovangelist:
Wow, that is an awesomely commented pr. Thanks. Looks good to merge

PR #1179, opened 2017-02-23
Initial commit of docker dev and related docs

jyee:
Fixed & ready for review.
This adds a DockerfileDev and instructions to build the dev environment in Docker.

Note that the Docker image uses Nginx to serve the docs site. There appears to be a disconnect between guard, rack and webrick so it wasn't able to serve content to anything other than localhost (i.e. couldnt serve outside the container). This isn't perfect, but it should get contributors up and running quickly.

Also note that by running nginx and rake, both dump to stdout, so you get a mix of both logs - that might be a bit messy.


README.md
```diff
@@ -34,6 +37,28 @@ export github_personal_token=[paste access token here]
...
 You should then run `source ~/.bash_profile` to reload the settings.
 
+## Setup using Docker
+
+An alternative to setting up the development environment as explained above is to use the Docker container development environment.
+
+```
+# From within the documentation folder:
+# Build the docker image. This can take 15-20 minutes.
+docker build -f DockerfileDev -t dd-docs:latest .
```
irabinovitch: should we just put this on quay?

jyee:
Docker image pushed to DockerHub. Also simplified the documentation and created a `rake compileslow` task (similar to the `rake slow` task, but doesnt run guard)

irabinovitch:
@jyee I think we should go ahead and merge this as is, and add a follow up task on setting up Guard or equivalent.

irabinovitch:
Sent over a PR that makes the CMD/ENTRYPOINT guard. This watches and rebuilds properly for docker environments.

README.md
```diff
@@ -62,6 +64,27 @@ export github_personal_token=[paste access token here]
 
 You should then run `source ~/.bash_profile` to reload the settings.
 
+## Setup using Docker
```
kmshultz: Should this be a subheading in the general 'Setup' section above? Also we should clearly explain at the beginning of 'Setup' that there are two setup options: non-Docker (i.e. the legacy way) and Docker. If readers follow the legacy Setup instructions and _then_ see the Docker Setup - and want to do that instead - they will have done the legacy setup for no reason.

jyee:
@kmshultz One last look for the updated README (docker focused) and ADVANCED (where local setup has moved)?

PR #1204, opened 2017-03-07
Jhotta/adding jp message to insert example links method

jhotta:


irabinovitch:
this is handled elsewhere in the migration to hugo.

Issue #1207, opened 2017-03-09
Stops compilation, if I add "<%= get_metrics_from_git() %>" to a translated md file. 


jhotta:
Adding `<%= get_metrics_from_git() %>` into integration markdown files under `content/ja/integrations` will brake HTML compilation.

```
17:35:19 - INFO - running compile
bundle exec nanoc compile
Loading site data… done
Compiling site…

Captain! We’ve been hit!

Message:

Encoding::CompatibilityError: incompatible character encodings: ASCII-8BIT and UTF-8

Compilation stack:

  - [item]   /ja/integrations/awslambda/ (rep default)

Stack trace:

  0. item /ja/integrations/awslambda/ (rep default):4:in `concat'
  1. item /ja/integrations/awslambda/ (rep default):4:in `get_binding'
  2. /Users/papago/.rbenv/versions/2.3.3/lib/ruby/2.3.0/erb.rb:864:in `eval'
  3. /Users/papago/.rbenv/versions/2.3.3/lib/ruby/2.3.0/erb.rb:864:in `result'
  4. /Users/papago/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/nanoc-3.8.0/lib/nanoc/filters/erb.rb:33:in `run'
  5. /Users/papago/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/nanoc-3.8.0/lib/nanoc/base/compilation/filter.rb:117:in `setup_and_run'
  6. /Users/papago/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/nanoc-3.8.0/lib/nanoc/base/result_data/item_rep.rb:343:in `filter'
  7. /Users/papago/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/nanoc-3.8.0/lib/nanoc/base/compilation/item_rep_proxy.rb:45:in `filter'
  8. /Users/papago/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/nanoc-3.8.0/lib/nanoc/base/compilation/rule_context.rb:59:in `filter'
  9. /Users/papago/DD-support/en_documentation/Rules:148:in `block in parse'
  ... 41 more lines omitted. See full crash log for details.

If you believe this is a bug in nanoc, please do report it at
-> https://github.com/nanoc/nanoc/issues/new <-

A detailed crash log has been written to ./crash.log.
[...]

PR #1208, opened 2017-03-09
bug fix: In some case, formatmetrics() returens non UTF-8 codec.

jhotta:
And HTML compilation stops.

Issue#: 1207
Stops compilation, if I add "<%= get_metrics_from_git() %>" to a translated md file. #1207


technovangelist:
that seems like a good fix.

jhotta:
Thank you for reviewing.

PR #1215, opened 2017-03-17
Remove a broken link on the libraries page

kmshultz:


content/libraries.md
```diff
@@ -207,10 +207,6 @@ Some great folks have written their own libraries to help interact with Datadog.
 
   * This [library][51] will allow you to generate process information from StatsD, given pid files. It was created by [GitterHQ][52].
 
-#### Reverse Color Scheme
-
-  * To get a darker background than default, you can use [this CSS library][55] to flip the colors.
-
```
irabinovitch: does line 276 need to be remove as well? also might want to check with support and PM on if its ok to mention the dark mode beta.

kmshultz:
Ahh, so that's what this was about? I didn't even know what this referred to.

Thanks - I'll check with them.

PR #1222, opened 2017-03-22
Use s3 command from jenkins' staging build

kmshultz:
Already pushed to `staging` - it works.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -45,7 +45,7 @@ namespace :deploy do
   desc 'Deploy to staging S3 bucket; Should be used by `rake release:staging`'
   task :staging do
     sh("cp staging-robots.txt output/robots.txt")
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -45,7 +45,7 @@ namespace :deploy do
   desc 'Deploy to staging S3 bucket; Should be used by `rake release:staging`'
   task :staging do
     sh("cp staging-robots.txt output/robots.txt")
-    sh("cd output && aws s3 sync --delete --size-only . s3://datadog-docs-staging --acl public-read")
+    sh("cd output && aws s3 sync --size-only . s3://datadog-docs-staging --acl public-read --cache-control=\"max-age=600\"")
```
irabinovitch: Do we want --delete in place so it removes files we delete from git?

kmshultz: @irabinovitch added back.

PR #1223, opened 2017-03-22
Bring prod's s3 command in line with staging's

kmshultz:
Just to make this symmetrical with #1222.


kmshultz:
Will edit deploy-doc-prod job in jenkins to the following after this is merged: 

```
#!/bin/bash
source ~/.rvm/scripts/rvm
bundle install --local --path=/cache/ --without development
echo "uploading..."
bundle exec rake release:prod
echo "complete..."
```

PR #1224, opened 2017-03-27
Add option Further Reading section to Integration doc guidance

kmshultz:
Some of our docs reference Docs guides and Datadog blog posts related to the integration at hand, but these references are peppered throughout different sections of the READMEs in no orderly way.

What about adding an optional Further Reading section?


jyee:
I like this!

technovangelist:
Great idea.

PR #1229, opened 2017-03-30
Update the tag section location on the kubernetes integration doc

oceanef:
Discussing with Haissam, it looks more appropriate to put the tags section at the instance level rather that at the init_config level


oceanef:
@irabinovitch integration-core looks already settled this way : https://github.com/DataDog/integrations-core/blob/master/kubernetes/conf.yaml.example#L112-L115

PR #1245, opened 2017-04-12
New guide for Composite Monitors

kmshultz:
Initially, this is based on 1) an internal Google doc from @sboak and 2) this KB article: https://help.datadoghq.com/hc/en-us/articles/207529246-Creating-Composite-Monitors

I may have made some assumptions that are not true. @isaacdd Please review and let me know if:

1. Anything is glaringly wrong
2. We should document more caveats/gotchas
3. We should document any features/behavior in more detail


kmshultz:
Items from the KB article and Google doc that this PR does not encapsulate (either because I didn't understand or wasn't sure of meaning):

---
Note: The following options will not apply to composite monitors:
  notify_no_data
  no_data_timeframe"


---
In the example below, both monitors are grouped by host, but they do not share any hosts in common. These two monitors are filtered to different AWS availability-zones and will never overlap, **but it is possible that temporary circumstances would yield a similar result** [what does the last bit mean?].

---
There are some special considerations for calculating NO-DATA state for composite monitors:
At least one of the monitors used in the composite needs to be set to notify in NO-DATA state for a composite no-data state to be computed
Alerting on NO-DATA state will follow the "Trigger when" rules you specify for the composite monitor
Calculations of other alert states (e.g. CRITICAL) take precedence over NO-DATA state, and alert first.

isaacdd:
I, unfortunately, not the most familiar with composite monitors. I think @dd-caleb @athap @celenechang, @jesusmaldonado or @sboak know more to review these docs.

kmshultz:
Whoops, apologies @isaacdd. If the others could weigh in that'd be great!

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,161 @@
...
+!(a && b)
+a || b && !c
+(a || b) && (c || d)
+~~~
+
+Outside of a composite monitor's New Monitor and Edit forms (e.g. on its Status page), its individual monitors are known by their numeric IDs:
+
+![composite-status](/static/images/composite_monitors/composite-status.png)
+
+In the API, a composite monitor's trigger condition is called its query. While a non-composite monitor's query can encapsulate many things — a metric, tags, an `avg` function, a group-by, etc — a composite monitor's query is simply its trigger condition, defined in terms of its constituent monitors.
```
kmshultz: sure, will re-word.

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Choose individual monitors
+
+You can choose up to 10 individual monitors to use in the new composite monitor. The individual monitors cannot themselves be composite monitors. 
+
+After you choose your first monitor, the UI will indicate its alert type and current status:
+
+![create-composite-2](/static/images/composite_monitors/create-composite-2.png)
+
+Recall that each Datadog monitor has one of two alert types: simple, or multi-alert. You may create composite monitors from individual monitors of different alert types (e.g. one simple alert monitor and one multi-alert monitor), but if you choose more than one multi-alert monitor, then *all* multi-alert monitors in the composite monitor must use the same grouping. In the screenshot below, monitor 'a' is grouped by `availability-zone`, but monitor 'b' is grouped by `host`:
```
kmshultz: Well, grouping only seems relevant to multi-alert monitors, unless I'm mistaken? At least in the context of composite monitors. I'll think of a way to re-word it.

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,161 @@
...
+
+### How to alert?
+
+Consider a composite monitor that uses three individual monitors — A, B, and C — and a trigger condition `A && B && C`. How many simultaneous alerts might you potentially receive from the composite monitor? That depends on its individual monitors' alert types. 
+
+Let's look at three examples with varying individual monitors. We'll consider how the composite monitor behaves in each case.
+
+#### All simple alert monitors
+
+The composite monitor will also have a simple alert type, i.e. it will only send up to one alert per evaluation cycle. The composite monitor triggers when the queries for A, B, and C are all true at the same time.
```
kmshultz: I thought about that too, but I wasn't to avoid just restating the example title. I thought the lead-in sentence explained what the examples mean:

"Let's look at three examples with varying individual monitors. We'll consider how the composite monitor behaves in each case."

If it doesn't explain it well enough, maybe I'll re-word the lead-in?

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,161 @@
...
+#### One multi-alert monitor (monitor A)
+
+The composite monitor will have a multi-alert type. If monitor A has 4 reporting sources — hosts web01 through web04 — you may receive up to 4 alerts each time Datadog evaluates the composite monitor. In other words: for a given evaluation cycle, Datadog has 4 cases to consider. For each case, monitor A's status may vary across its sources, but the statuses of monitors B and C — which are simple alert types — are unchanging. Here's an example cycle:
+
+| monitor A | monitor B | monitor C | composite status |
+| --------- | --------- | --------- | ---------------- |
+| T (web01) | T         | T         | T - alert!       |
+| F (web02) | T         | T         | F                |
+| T (web03) | T         | T         | T - alert!       |
+| F (web04) | T         | T         | F                |
```
kmshultz: Sure, there is room for true and false. I think T/F or True/False is best here since the trigger condition uses boolean operators.

Also to say 'triggered' or 'not triggered' on the individual monitors is misleading, because they in themselves trigger nothing.

The example in the final section re-works one of these examples to show actual statuses.

It might be worth saying 'triggered!' instead of 'alert!' for the triggered cases.

content/guides/composite_monitors.md
kmshultz: We need eng feedback on this one. From the KB article we've written and the google doc draft, this is how I understand what's legal.

> Individual monitors must share at least one common reporting source

I don't think that's true. If two individual monitors are simple alert types, they don't need to share any reporting sources in common (far as I understand it).

> If more than one multi-alert monitor is used, all individual monitors must be multi-alert.

This is also not true.

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,161 @@
...
+
+#### All simple alert monitors
+
+The composite monitor will also have a simple alert type, i.e. it will only send up to one alert per evaluation cycle. The composite monitor triggers when the queries for A, B, and C are all true at the same time.
+
+#### One multi-alert monitor (monitor A)
+
+The composite monitor will have a multi-alert type. If monitor A has 4 reporting sources — hosts web01 through web04 — you may receive up to 4 alerts each time Datadog evaluates the composite monitor. In other words: for a given evaluation cycle, Datadog has 4 cases to consider. For each case, monitor A's status may vary across its sources, but the statuses of monitors B and C — which are simple alert types — are unchanging. Here's an example cycle:
+
+| monitor A | monitor B | monitor C | composite status |
```
jesusmaldonado: And also restate the composite query (A && B && C) for each of these tables?

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,161 @@
...
+| --------- | --------- | --------- | ---------------- |
+| T (web01) | T         | T         | T - alert!       |
+| F (web02) | T         | T         | F                |
+| T (web03) | T         | T         | T - alert!       |
```
jesusmaldonado: `Alert`
and 
`No alert`

or 
`Triggered` and `Not Triggered`
(because some composite monitors may Trigger a Warn State, not just Alert)

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,161 @@
...
+In this cycle, you would receive two alerts.
+
+#### Many multi-alert monitors (monitors A and B)
+
+The composite monitor will have a multi-alert type, but the number of alerts per cycle will be, at most, the number of common sources between monitors A and B. If web01 through web05 are reporting for monitor A, and web04 through web09 are reporting for monitor B, the composite monitor _only_ considers the common sources: web04 and web05. You can only receive up to two alerts in a cycle. Here's an example cycle:
+
+| monitor A | monitor B | monitor C | composite status |
+|-----------|-----------|-----------|------------------|
+| T (web04) | T (web04) | T         | T - alert!       |
+| F (web05) | T (web05) | T         | F                |
```
jesusmaldonado: it might be useful to add a third example 
```
| T (web01) | T (web05) | T         | Not triggered               |
```

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,161 @@
...
+
+### When to alert?
+
+The previous section considered monitor statuses in binary terms: true, or false. But a monitor can have statuses other than simply `Alert` and `OK`. Possible statuses include (in order of increasing severity): `OK`, `Skipped`, `Ignored`, `No Data`, `Unknown`, `Warn`, and `Alert`. Above, what true and false actually refer to is whether or not a given status should trigger an alert (true) or not (false). In other words: the status' **alert-worthiness**. 
+
+Composite monitors consider statuses from `No Data` up to `Alert` to be alert-worthy. When a composite monitor evaluates as alert-worthy, it inherits the most severe status among its individual monitors. When a composite monitor does not evaluate as alert-worthy, it inherits the _least_ severe status.
+
+Let's look again at the previous example that used only one multi-alert monitor. This table includes status detail, but still indicates the alert-worthiness of each status with T or F:
```
jesusmaldonado: Reiterate that it's `A && B && C` where A is multi, B and C are simple.

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,167 @@
...
+
+In Datadog, go to the New Monitor page and select 'Composite Monitor' from the list of monitor types:
+
+![choose-composite-type](/static/images/composite_monitors/select-monitor-type.png)
+
+### Choose individual monitors
+
+You can choose up to 10 individual monitors to use in the new composite monitor. When choosing monitors, the following rules apply:
+
+* No individual monitor may itself be a composite monitor
```
kmshultz: @jyee I put the chosen-monitor requirements in a list -- good idea, I think it's clearer.

Does it make more sense than before. Still need Eng to review it for correctness.

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,161 @@
...
+  4 OK
+
+The monitor was last triggered at Fri Mar 24 2017 15:56:21 EDT (28 secs ago)
+~~~
+
+After setting any other miscellaneous options, click 'Save'.
+
+## How composite monitors work
+
+In this section, we look at a few examples to show **how** and **when** Datadog will alert on composite monitors.
```
jesusmaldonado: State that T is a possible state of `No Data`, `Unknown`, `Warn` or `Alert` and F is a possible state of `OK` `Skipped` `Ignored`

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,161 @@
...
+
+In this cycle, you would receive one alert.
+
+Remember: all multi-alert monitors used in a composite monitor _must_ use the same group-by. And if they do, and there are multiple groups in the group-by, the groups must appear in the same order. For example, a monitor grouped by `{host,environment}` and another grouped by `{environment,host}` cannot be used together in a composite monitor.
+
+### When to alert?
+
+The previous section considered monitor statuses in binary terms: true, or false. But a monitor can have statuses other than simply `Alert` and `OK`. Possible statuses include (in order of increasing severity): `OK`, `Skipped`, `Ignored`, `No Data`, `Unknown`, `Warn`, and `Alert`. Above, what true and false actually refer to is whether or not a given status should trigger an alert (true) or not (false). In other words: the status' **alert-worthiness**. 
+
+Composite monitors consider statuses from `No Data` up to `Alert` to be alert-worthy. When a composite monitor evaluates as alert-worthy, it inherits the most severe status among its individual monitors. When a composite monitor does not evaluate as alert-worthy, it inherits the _least_ severe status.
```
jesusmaldonado: It might also be worth saying explicitly that this means that an alert may happen even though they do not all have the same status.

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,167 @@
...
+
+#### All simple alert monitors
+
+The composite monitor will also have a simple alert type, i.e. it will only send up to one alert per evaluation cycle. The composite monitor triggers when the queries for A, B, and C are all true at the same time.
+
+#### One multi-alert monitor (monitor A)
+
+The composite monitor will have a multi-alert type. If monitor A has 4 reporting sources — hosts web01 through web04 — you may receive up to 4 alerts each time Datadog evaluates the composite monitor. In other words: for a given evaluation cycle, Datadog has 4 cases to consider. For each case, monitor A's status may vary across its sources, but the statuses of monitors B and C — which are simple alert types — are unchanging. Here's an example cycle:
+
+| monitor A    | monitor B| monitor C | composite status |
```
jesusmaldonado: And also restate the composite query (A && B && C) for each of these tables?

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,167 @@
...
+![choose-composite-type](/static/images/composite_monitors/select-monitor-type.png)
+
+### Choose individual monitors
+
+You can choose up to 10 individual monitors to use in the new composite monitor. When choosing monitors, the following rules apply:
+
+* No individual monitor may itself be a composite monitor
+* Individual monitors may be all simple alert monitors, all multi-alert monitors, or a combination of simple and multi-alert
+* If more than one multi-alert monitor is chosen, then **all** multi-alert monitors in the composite monitor must:
+  1. Use the same grouping (e.g. a monitor grouped by `{host}` cannot be used with one grouped by `{environment}`) AND the same group ordering (e.g. a monitor grouped by `{host,device}` cannot be used with one grouped by `{device,host}`)
```
jesusmaldonado: this is not exactly true
e.g. a multi-alert with a average_by: host result for host:i-3892738 and another multi-alert with a average_by:device_host and a result for device_host:i-3892738 --> these monitor results will undergo whatever operation is in the query and be evaluated

so even though the tag (host:i-3892738 !== device_host:i-3892738) does not equal each other the result will be evaluated because the group VALUE i-3892738 == i-3892738 is the same

of course availability-zone and host will likely never lead to the same reporting source, so it LOOKS like the same group_by keys must be necessary but it is only an indirect consequence of those group-bys never being able to produce the same reporting sources.

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,167 @@
...
+  2. Have at least one group element in common (e.g. a monitor that applies only to `environment:dev` cannot be used with one that applies only to `environment:prod`)
```
jesusmaldonado: we should say have one reporting source in common

content/guides/composite_monitors.md
jesusmaldonado: to be clear, the first portion of this bullet 1 is not exactly true. the group ordering portion of bullet 1 is correct

content/guides/composite_monitors.md
kmshultz: I also was tripped up and first (mistakenly) understood it as @jyee did when I first read the google doc draft and the KB article. I had hoped my re-wording in here avoided confusing the reader, but apparently it did not.

I've put the chosen-monitor requirements in a list as @jyee recommended - I hope it's much clearer now. Eventually, we'll want to run the final draft of this PR by someone who's totally fresh to composite monitors to make sure they aren't confused on first read.

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,161 @@
...
+
+After you choose your first monitor, the UI will indicate its alert type and current status:
+
+![create-composite-2](/static/images/composite_monitors/create-composite-2.png)
+
+Recall that each Datadog monitor has one of two alert types: simple, or multi-alert. You may create composite monitors from individual monitors of different alert types (e.g. one simple alert monitor and one multi-alert monitor), but if you choose more than one multi-alert monitor, then *all* multi-alert monitors in the composite monitor must use the same grouping. In the screenshot below, monitor 'a' is grouped by `availability-zone`, but monitor 'b' is grouped by `host`:
+
+![create-composite-4](/static/images/composite_monitors/create-composite-4.png)
+
+You cannot create such a composite monitor.
```
kmshultz: @jesusmaldonado hmm, interesting. Perhaps we don't want to allow that, then? or is it impossible since group-bys and reporting sources can change in the future?

content/guides/composite_monitors.md
kmshultz: @jesusmaldonado yes, it's stated below in the 'When' section.

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,161 @@
...
+The monitor was last triggered at Fri Mar 24 2017 15:56:21 EDT (28 secs ago)
+~~~
+
+After setting any other miscellaneous options, click 'Save'.
+
+## How composite monitors work
+
+In this section, we look at a few examples to show **how** and **when** Datadog will alert on composite monitors.
+
+### How to alert?
```
kmshultz: or maybe most accurate: How many alerts per cycle? (or per evaluation)

jesusmaldonado: per evaluation

content/guides/composite_monitors.md
jesusmaldonado: yeah but the user hasn't read that when they get up here...

content/guides/composite_monitors.md
kmshultz: That's only relevant to the first section, though. And it's so simple and obvious to state - there's no need for a T/F table - perhaps it doesn't need its own section.

content/guides/composite_monitors.md
kmshultz: The table  points that out, but after showing the table, perhaps it's worth drawing attention to one row in the table where that's the case.

content/guides/composite_monitors.md
kmshultz: good call, will do.

content/guides/composite_monitors.md
kmshultz: Ok, this is crucial for us to explain clearly. it sounds like the only thing that matters is the **reporting source**, not what the group-bys actually are?

This would explain why `{host,device}` and `{device,host}` cannot work together, because host,device will always be like `i-123,/dev/sda` and device/host like `/dev/sda,i-123`, i.e. they'll never collide.

> e.g. a multi-alert with a average_by: host result for host:i-3892738 and another multi-alert with a average_by:device_host and a result for device_host:i-3892738 --> these monitor results will undergo whatever operation is in the query and be evaluated

I am still not understanding this.

content/guides/composite_monitors.md
kmshultz: understood.

jesusmaldonado:
**Clarifications**
```
Note: The following options will not apply to composite monitors:
notify_no_data
no_data_timeframe"
```
No longer true...

notify_no_data: If ANY of the constituent monitors are notify on no data state, the composite monitor will by default take into account NO DATA in its evaluation cycle. **note** this can be overriden by the toggle so that the composite will alert on no data even though constituent monitors don't. You either do or don't count NO DATA for ALL monitors, there's no way to change it for each monitor.
no_data_timeframe: this doesn't matter. Either the evaluation of the individual monitor resulted in NO DATA or it didn't.
```
In the example below, both monitors are grouped by host, but they do not share any hosts in common. These two monitors are filtered to different AWS availability-zones and will never overlap, but it is possible that temporary circumstances would yield a similar result [what does the last bit mean?].
```
This is what i was talking about availability-zone-1:ASIA may somehow come from the same reporting source as group-by availability-zone-1 (i.e. `availability-zone-1:ASIA`) and would produce the same result. If for some weird reason your `env` is named the same scheme as your `host` group-bys this can also happen i.e. `env:i-123` and `host:-i123` would lead to a computation and potential trigger.
```  
There are some special considerations for calculating NO-DATA state for composite monitors:
At least one of the monitors used in the composite needs to be set to notify in NO-DATA state for a composite no-data state to be computed
```
This can be overriden. It's all or nothing.
```
Alerting on NO-DATA state will follow the "Trigger when" rules you specify for the composite monitor
Calculations of other alert states (e.g. CRITICAL) take precedence over NO-DATA state, and alert first.
```
This just means the set of alert severity could be either
`NO DATA, WARN, UNKNOWN, ALERT`
or
`WARN, UNKNOWN, ALERT`

content/guides/composite_monitors.md
kmshultz: It sounds like, technically, the second part of bullet 1 - the ordering thing - is in theory also not true; only true in practice?

if you had some tags which could have overlapping values, e.g. if you had some host named `prod` (and who would, but follow me..) and you had two monitors with group-bys of `{environment,host}` and `{host,environment}`, then you could have reporting sources of say `prod,prod` reporting to both monitors, and a composite monitor using these two monitors *would* be able to calculate a result. Correct?

content/guides/composite_monitors.md
kmshultz: @jyee "an aggregation function like `avg`" 👍 👎 ?

content/guides/composite_monitors.md
kmshultz: Perhaps it's best to put the alert-worthiness section first and the how-many-alerts section second? I think I ordered them how they are because thinking in terms of T/F is simpler, but since alert-worthiness is more real, maybe we put it first.

content/guides/composite_monitors.md
jesusmaldonado: yeah that's true actually. i guess we need to alter the language to `may not produce a result` and then talk about how they must produce the same reporting sources as well

jesusmaldonado: you understand it.
ex.
`A && B`
A: `host:i-123` (Warn)
B `device_host:i-123` (Warn)
--> Composite will trigger as WARN

content/guides/composite_monitors.md
jesusmaldonado: No data is also only included if it's toggled yes/no on the composite page.

content/guides/composite_monitors.md
jesusmaldonado: Yeah. I think it's fine to be more specific. Composite users are generally more advanced, you have to be because you need to have created other monitors to make a composite query, so they should be used to seeing the alerting. And then we could keep the frequency section but use triggerable states (No data, alert, warn, unknown) instead of T or F and therefore show more precise examples as well?

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,167 @@
...
+In the API, a composite monitor's trigger condition is called its query. While a non-composite monitor's query can encapsulate many things — a metric, tags, an `avg` function, a group-by, etc — a composite monitor's query is simply its trigger condition, defined in terms of its constituent monitors.
+
+For two non-composite monitors with the following queries:
+
+~~~
+"avg(last_1m):avg:system.mem.free{role:database} < 2147483648" # monitor ID: 1234
+"avg(last_1m):avg:system.cpu.system{role:database} > 50" # monitor ID: 5678
+~~~
+
+a composite monitor's query is simply `"1234 && 5678"`, `"!1234 || 5678"`, etc.
```
jesusmaldonado: maybe here talk about the all or nothing role of the no data toggle?

content/guides/composite_monitors.md
kmshultz: As I understand it, this is why we must always *allow* users to choose multi-alerts with different group-bys, even if it's not likely to produce a result. Even if the multi-alerts have no reporting sources in common at the time the composite monitor is created, they may at a future date.

But the UI warns them, at least, if they're creating a composite that is likely to never be triggered.

content/guides/composite_monitors.md
kmshultz: I will work that in some place or another.

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,161 @@
...
+
+![create-composite-2](/static/images/composite_monitors/create-composite-2.png)
+
+Recall that each Datadog monitor has one of two alert types: simple, or multi-alert. You may create composite monitors from individual monitors of different alert types (e.g. one simple alert monitor and one multi-alert monitor), but if you choose more than one multi-alert monitor, then *all* multi-alert monitors in the composite monitor must use the same grouping. In the screenshot below, monitor 'a' is grouped by `availability-zone`, but monitor 'b' is grouped by `host`:
+
+![create-composite-4](/static/images/composite_monitors/create-composite-4.png)
+
+You cannot create such a composite monitor.
+
+Furthermore, even if all chosen multi-alert monitors use the same grouping, they must *also* have at least one reporting source (i.e. host) in common. In the screenshot below, both monitors are grouped by `host`, yet the UI indicates that the two monitors are incompatible:
```
kmshultz: >  they never have the same value

i.e. they don't have any values in common *right now* (hence the warning), but they may in the future, yes?

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,167 @@
...
+
+If you choose a multi-alert monitor, the UI will show its group-by clause (e.g. `host`) and how many unique sources (e.g. how many hosts) are currently reporting. When combining many multi-alert monitors, you should almost always choose monitors that have the same group-by. If you don't, the UI will warn you that such a composite monitor may never trigger:
+
+![create-composite-4](/static/images/composite_monitors/create-composite-4.png)
+
+Even if you do choose multi-alert monitors with the same group-by, the UI may still warn you about the selection. In the following screenshot, both monitors are grouped by `host`:
+
+![create-composite-5](/static/images/composite_monitors/create-composite-5.png)
+
+Since there's still a 'Group Matching Error' despite matching group-bys, we can assume that these monitors currently have no reporting sources in common. If there are no common reporting sources, Datadog cannot compute a status for the composite monitor. For better understanding, [read below](#multi-alert-monitors-and-common-reporting-sources).
```
jesusmaldonado: ✌🏻

kmshultz:
@jesusmaldonado please review this again when you have a moment. I have added further explanations for common reporting sources andfor `notify_no_data`. Hopefully they're correct!

I first considered add these two notes in a 'Caveats' type section at the bottom of the document, but  for now I've included them in 'How composite monitors work'. They may work best as some kind of note/footnote; with no style delineation (e.g. yellow 'Note' background), I feel they disrupt the reader's flow with too much detail on something many readers won't care about (until they need to care about it).

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,175 @@
...
+
+---
+
+Composite monitors let you combine many individual monitors into one so that you can define more specific alert conditions. You can choose up to 10 existing monitors - monitor A and monitor B, say - and then set a trigger condition using boolean operators (e.g. “A && B”). The composite monitor will trigger when its individual monitors' statuses simultaneously have values that cause the composite's trigger condition to be true.
+
+With regard to configuration, a composite monitor is independent of its constituent monitors. You can modify the notification policy of a composite monitor without affecting the policies of its constituent monitors, and vice versa. Furthermore, deleting a composite monitor will not delete its constituent monitors; a composite monitor does not own other monitors, it only uses them. Many composite monitors may reference the same individual monitor.
+
+_Note: this guide refers variously to 'individual monitors', 'constituent monitors', and 'non-composite monitors'. They all mean the same thing: the monitors that a composite monitor uses to calculate its status._
```
kmshultz: As it's written now, this feels necessary to say. I'd like to avoid using three different terms to refer to the same thing, but looking across all usages, I can find appropriate (IMO) usages of each. I'm open to scrapping one, though, if others think three is too many.

adimusset:
I have a few remarks: 
- screenshots 3, 4 and 5 are outdated, the UI has changed a bit
- it may be important to add somewhere how the NOT function works : 
if a is in any alert-worthy state (including no data), !a = OK
if b is in any alert-unworthy state !b = Alert
- for the no data paragraph, the most important thing to say is that there is a switch, the default behavior is a bit confusing

kmshultz:
@adimusset thank you, I'll get some newer screenshots. However, why does the UI still show Composites as beta? ('Composite (Beta)')

> for the no data paragraph, the most important thing to say is that there is a switch, the default behavior is a bit confusing

Is what is written on `No Data` inaccurate, or just confusing?

adimusset:
Yeah we should remove the beta sign

About the no data paragraph, I think this is accurate but it feels too complicated, the default value is just computed by the front-end to pre-populate the switch. You directly see the result and you can change it if you disagree.

kmshultz:
@adimusset is the 'Beta' something we can remove in short order? I'd like to get screenshots after that's removed. It looks kinda cheesy for our official docs to have 'Beta' in the screenshots.

content/guides/composite_monitors.md
jesusmaldonado: I think constituent monitors is most accurate but if you're going to use three I think having this note is good. Subjective call

jesusmaldonado:
@kmshultz you could just mention both ways to be very clear: "When creating a composite monitor through the UI, override the notify_no_data condition by setting the toggle `screenshot here of toggle maybe`. If being created through the API, set the `notify_no_data` field manually.

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,181 @@
...
+
+If no constituent monitors have `notify_no_data` enabled but you want to receive alerts on `No Data` for the composite monitor, enable `notify_no_data` for the composite monitor only.
+
+A composite monitor cannot be configured to consider `No Data` alert-worthy for some of its constituent monitors but not for others; it's all or none.
+
+#### How composite monitors select common reporting sources
+
+As [explained above](#many-multi-alert-monitors), composite monitors that use many multi-alert monitors only consider the individual monitors' _common reporting sources_. In the example, the common sources were `host:web04` and `host:web05`, but there's a subtle caveat: in identifying common reporting sources, composite monitors only look at tag _values_ (i.e. `web04`), not tag names (i.e. `host`). This technically makes it possible for composite monitors to trigger on multi-alert monitors that group by different tags.
```
jesusmaldonado: This is probably the best explanation I've seen 🐸

content/guides/composite_monitors.md
```diff
@@ -0,0 +1,181 @@
...
+
+For example, if the example above included a multi-alert monitor 'D' grouped by `environment`, and that monitor had a single reporting source, `environment:web04`, then the composite monitor would consider `web04` the single common reporting source between A, B, and D, and would compute its trigger condition.
+
+Often, two monitors grouped by different tags tend to have reporting sources whose values never overlap, e.g. `web04` and `web05` for monitor A, `dev` and `prod` for monitor D. But if and when they do overlap, a composite monitor that uses two such monitors becomes capable of triggering an alert.
```
jesusmaldonado: yes 🤓

sboak:
This is very thorough and a good overview. It's somewhat unrelated to this work, but we need to update the image for composite monitors in step 1 of monitor creation - it's currently just reusing the metric/threshold image. When that's ready these docs will need updating, but otherwise I think this looks really good.

kmshultz:
@adimusset ok, reading this after the weekend, I agree the no-data bit is overexplained and unnecessary. I have removed it. 👍 

I just pushed a few other changes, as well. Soon, I'll push some updated screenshots sans 'Beta' and this should be ready for final review.

kmshultz:
@sboak New screenshots up and some final text changes. Ready for your final review.

PR #1246, opened 2017-04-12
Remove old 'officially integrating with dd' guide

kmshultz:
This older guide (it's not really a 'guide') seems unnecessary now that there's http://docs.datadoghq.com/guides/integration_sdk. If you concur with removing it, I'll delete the file.

In the docs redesign, we may want to reintroduce this invitation somewhere, i.e. on the main 'Integrations' page, linking to http://docs.datadoghq.com/guides/integration_sdk.


irabinovitch:
Agreed its not a guide but it seems to have slightly different content than the SDK page.  Could we find a new home for it and and links to the SDK page?

jeremy-lq:
If we keep this guide, linking to and briefly explaining integrations-extras may also make sense. The "Submitting Your integration" bit on the SDK page should also be updated based on recent discussions.

technovangelist:
whats the next step on this one?

kmshultz:
I think we can ignore it until the Docs redesign, which is when I imagine we'll do a lot of refactoring of content.

technovangelist:
so should we kill the pr?

kmshultz:
Sure.

PR #1248, opened 2017-04-13
update trace port to 8126 in code snippets

irabinovitch:
The tracing api port was switched from 7777 to 8126 after the beta.


jyee:
LGTM!

PR #1250, opened 2017-04-17
Moore/aws lambda

themsquared:
Just a quick fix on the lambda code page and a CSS update for <code> blocks. This may impact other pages with multiple lines of code though so it'd be smart to review before merging.

### DIFF
--- a/content/integrations/awslambda.md
+++ b/content/integrations/awslambda.md
@@ -26,7 +26,8 @@ To send custom metrics to Datadog, you must print a log line from your Lambda, u
 Please ensure the `unix_epoch_timestamp` is in seconds (not milliseconds).
 
 For example, here is sample snippet for printing a valid custom metric, from your Lambda function (in Python):
-<code>unix_epoch_timestamp = int(time.time())
+~~~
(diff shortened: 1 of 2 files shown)


kmshultz:
@themsquared Thanks submitting this.

Actually, instead of `<code>`,  let's use backticks for in-line code and ~~~ for multi-line code blocks, e.g.:

~~~
unix_epoch_timestamp = int(time.time())
value = 42
...
~~~

(see https://kramdown.gettalong.org/quickref.html#code-blocks)

This way, we should get multi-line code - with breaks - and we can avoid messing with CSS.

Thanks again!

themsquared:
Works for me.  Just want to make the multi-lines readable 👍

kmshultz:
@themsquared Yep, thanks for spotting it! If you don't mind updating it as I described, I'll merge. Otherwise we can close this and I'll fix it in a different PR - it's up to you.

themsquared:
Updated. Please review and merge if good to go.

kmshultz:
@themsquared Looks good - thanks again!

Issue #1249, opened 2017-04-15
Document API downtime.monitor_id and monitor.options.evaluation_delay


bshelton229:
Please document the ability to create/update a downtime object attached to a monitor via sending a `monitor_id` field. The API seems to support sending a `monitor_id` on POST and PUT. Please add that to the api documentation.

The API also seems to support sending an `evaluation_delay` option to the monitor endpoint. Please add this to the api documentation as well.

We'd like for these two features to eventually make it into the terraform datadog provider and it would be great to have this officially documented.

This PR enables these properties in go-datadog-api
https://github.com/zorkian/go-datadog-api/pull/106

Thanks!

PR #1253, opened 2017-04-18
Document API downtime.monitor_id and monitor.options.evaluation_delay

bshelton229:
This is an attempt to add documentation for #1249.

I'm not totally sure `evaluation_delay` is only for metric monitors, but it seems to be the case.

Thanks!

### DIFF
--- a/content/api/index.html
+++ b/content/api/index.html
@@ -729,11 +729,11 @@ kind: documentation
               <p>Example: <code>{'critical': 90, 'warning': 80}</code></p>
               </li>
               <li><code>evaluation_delay</code> Time (in seconds) to delay
-              evaluation, as a non-negative integer.  For example, if the
-              value is set to 300 (5min) and the time is 7:00, the monitor
-              will evaluate data at 6:55. This is useful for AWS CloudWatch
-              and other backfilled metrics to ensure the monitor will
-              always have data during evaluation.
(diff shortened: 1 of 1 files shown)


content/api/index.html
```diff
@@ -728,6 +728,13 @@
               specified using the thresholds option.
               <p>Example: <code>{'critical': 90, 'warning': 80}</code></p>
               </li>
+              <li><code>evaluation_delay</code> Time (in seconds) to delay
+              evaluation, as a non-negative integer.  For example, if the
+              value is set to 300 (5min) and the time is 7:00, the monitor
+              will evaluate data at 6:55. This is useful for AWS CloudWatch
```
dd-caleb: perhaps instead of 

> value is set to 300 (5min) and the time is 7:00

something like:

> value is set to 300 (5min), the timeframe is set to `last_5m` and the time is 7:00, the monitor will evaluate data from 6:50 to 6:55.

kmshultz:
@bshelton229 thank you for adding this. You are correct, this option is only for metric monitors currently.

bshelton229:
@dd-caleb I think that's a lot clearer. Updated. Thanks.

kmshultz:
Thanks for this @bshelton229! By the way, I am doing the other doc updates in #1255.

PR #1254, opened 2017-04-18
Add Docs for new Azure Storage Integration

DanielLanger:
Currently in Beta, so the docs have a beta tag.

Reason for publishing is folks are asking :)


content/integrations/azure_storage.md
```diff
@@ -0,0 +1,37 @@
...
+integration_title: Microsoft Azure Storage
+kind: integration
+git_integration_title: azure_storage
+newhlevel: true
+beta: true
+---
+
+# Overview
+
+This feature is currently in Beta. To enable it for your account, please reach out to <support@datadoghq.com>
```
kmshultz: I would finish the Overview with this rather than lead in.

kmshultz: Actually, the `beta: true` tag causes a red-background BETA warning to be displayed below the title and above the Overview. However it does not say 'Please reach out at support@datadog to enable it', so we could keep that part here. I would still add it at the end, though.

DanielLanger:
@kmshultz done

irabinovitch:
:+1:

PR #1260, opened 2017-04-20
Adding some highlight and notes for the service discovery

aerostitch:
Adding some highlight and notes for the service discovery using Kubernetes annotations.
I must say I've bumped into some issues with my current config to use the service discovery via kubernetes annotations in my environment.
I must say it was mostly because I was reading the doc too quickly and didn't realized that the agent wouldn't check the deployments metadata, just the pods metadata.
So I highlighted a few elements that, I hope, will help other people get their configuration ready more quickly.
And added another example based on my current config.

### DIFF
--- a/content/guides/servicediscovery.md
+++ b/content/guides/servicediscovery.md
@@ -171,7 +171,7 @@ For example, if a container has this label configured as `com.datadoghq.sd.check
 
 ## Configuration templates with Kubernetes annotations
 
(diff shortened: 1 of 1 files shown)


kmshultz:
@aerostitch Thanks for pointing this out, and sorry for the delayed response! I'm documenting this caveat in #1357, so I'll close out this PR.

PR #1262, opened 2017-04-24
Added Datadog 101 and cleaned up css around videos

technovangelist:


Gemfile.lock
```diff
@@ -143,7 +138,6 @@ DEPENDENCIES
   rouge
   sassc
   therubyracer
-  video_info
```
irabinovitch: These changes look unrelated?

Gemfile
```diff
@@ -9,7 +9,6 @@ gem 'builder'
 gem 'dogapi', '~> 1.22.0'
 gem 'rouge'
 gem 'oj'
-gem 'video_info'
```
irabinovitch: Intentional?

irabinovitch:
Can we push this up to staging first?

Gemfile.lock
technovangelist: video info was a plugin that made it easier to get the still image for the video. api changed and no longer needed  a long hash that was different from video id. so video_info no longer needed. removed it.

Gemfile
technovangelist: yes

PR #1281, opened 2017-05-07
Override helper function for generating yaml check for the Java Integ…

nmuesch:
…ration page


technovangelist:
i looked at overriding the function that creates the set of links but its such a specialized case that it makes sense to do it here instead

PR #1292, opened 2017-05-10
Remove the older of two AWS OpsWorks pages.

kmshultz:


kmshultz:
I suspect the confusion was that the original page was 'opsworks.md' and someone noticed that there was no 'awsopsworks.md' and decided to write one.

Issue #1265, opened 2017-04-26
There is no "Manage > Connections" in SumoLogic


PabloCastellano:
http://docs.datadoghq.com/integrations/sumologic/

The second step in the **SumoLogic to Datadog** paragraph is:

- From the main menu, choose Manage -> Connections.

Please make a note saying that you need to have administrator permissions to view the Connections button, otherwise it is confusing.

PR #1293, opened 2017-05-10
SumoLogic configuration requires SumoLogic Administrator rights

kmshultz:
A user noticed this in #1265.

### DIFF
--- a/content/integrations/sumologic.md
+++ b/content/integrations/sumologic.md
@@ -13,10 +13,9 @@ Enable the Sumo Logic integration to:
 
 ## Installation
 
-Datadog integration with SumoLogic is two-part: send alerts and events from Datadog to Sumologic, or have SumoLogic send data to Datadog. You can enable one or both.
+Datadog integration with SumoLogic is two-part: send alerts and events from Datadog to Sumologic, or have SumoLogic send log data to Datadog. You can enable one or both.
 
 ### Datadog to SumoLogic
-
 
 1. Login to SumoLogic as a user with Administrator rights.
 2. From the main menu, choose **Manage** -> **Collection**.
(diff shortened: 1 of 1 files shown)


content/integrations/sumologic.md
```diff
@@ -13,11 +13,12 @@ Enable the Sumo Logic integration to:
 
 ## Installation
 
-There are two parts to the SumoLogic integration, based on the direction of information between the products. You can send alerts and events from Datadog to Sumologic and you can have SumoLogic send data to Datadog.
+Datadog integration with SumoLogic is two-part: send alerts and events from Datadog to Sumologic, or have SumoLogic send data to Datadog. You can enable one or both.
```
technovangelist: this feels a bit more awkward than the original.  'there are two parts' vs 'xxx is two-part'

kmshultz: How about:

"Datadog integrates with Sumo Logic in two ways: you can forward Sumo Logic log data to your Datadog event stream, and/or you can use Sumo Logic as a notification channel from Datadog alerts and events. In other words, either service can inform the other."

technovangelist: i would use that to describe the rds integration (with three instead of two) since there are 3 different ways that you may choose and you probably arent going to use all of them

this has two parts or components and you are probably going to use both. Maybe just eliminate it: The datadog integration with sumologic can send...blah blah

kmshultz: *Are* users likely to use both? How do we know?

In any case, I don't think it's up to us to tell users they should pick one or use all of the above, but are you saying 'either' makes it sound like they have to make a choice? How about '*each* service can inform the other', that sounds less mutual-exclusiony.

technovangelist: that would be better i think. but most will use both, kinda like in pagerduty. Although im not sure what was wrong with the original.

kmshultz: The 'and/or' also implies one or both, but just 'and' may be best.

I don't think it's that different from the original, just updated it while I'm here to fix the other thing.

PR #1294, opened 2017-05-10
Aliasing has not been implemented yet; remove it from docs for now

kmshultz:


technovangelist:
actually, isn't it better to roll back the original pr on this?

technovangelist:
they do the same thing though

PR #1301, opened 2017-05-12
first try at getting twitter cards for videos

technovangelist:
Verified that it does nothing bad on staging. was hoping to be able to test with twitter's validation tool but staging has robots.txt that disallows it.


layouts/_head.html
```diff
@@ -30,4 +30,16 @@
 
     </script>
     <script src="https://use.fontawesome.com/700fac0342.js"></script>
+    <% if @item.identifier.match("/videos/") && @item[:kind]== 'video' %>
+        <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
+        <meta name="twitter:card" content="player" />
+        <meta name="twitter:site" content="@datadoghq" />
+        <meta name="twitter:title" content="Sample Player Card" />
```
jyee: This probably needs to be a token to fill in the title, right?

technovangelist: doh

PR #1305, opened 2017-05-15
clarify to use port 17123 if using another statsd server

technovangelist:


content/integrations/statsd.md
```diff
@@ -12,7 +12,9 @@ Enable the Datadog StatsD integration to:
   * Monitor uptime and availability of a StatsD server
   * Collecting metrics from a StatsD server's admin interface
 
-**Please note the Datadog Agent includes [dogstatsd](http://docs.datadoghq.com/guides/dogstatsd/) which serves as a StatsD forwarder. This integration is intended for monitoring external StatsD servers, and is not needed to send metrics to Datadog using the StatsD protocol.**
+**Please note the Datadog Agent includes [dogstatsd](http://docs.datadoghq.com/guides/dogstatsd/) which serves as a StatsD forwarder. If you want to send metrics from a statsd server, just point it to the forwarder listening on port 17123.**
```
irabinovitch: @technovangelist I think this should read "from an application using statsd, just point it to the forwarder listening on port 8125".

kmshultz: Well, I've just read this like three times and I'm still not sure what the current content or proposed edit are saying, or if what Ilan is saying is correct, so, take that as you will. :)

What is this about forwarding? Can't we just tell readers that this check monitors StatsD itself, and link them to the DogStatsD page if they want to instrument metrics?

"Monitor the availability of your StatsD servers and track how many metrics they're processing.

This Agent check does not help you instrument your application metrics; it only monitors the status of StatsD itself. See [our DogStatsD guide] to learn how to instrument application metrics."

^ Though this doesn't cover the forwarding thing. I'm not sure myself how that's supposed to work - is it documented somewhere?

content/integrations/statsd.md
```diff
@@ -12,7 +12,9 @@ Enable the Datadog StatsD integration to:
...
+
+**This integration is intended for monitoring external StatsD servers, and is not needed to send metrics to Datadog using the StatsD protocol.**
```
kmshultz: I don't think we want to say 'external'. The check could be running locally to a StatsD daemon.

content/integrations/statsd.md
kmshultz: Also, s/statsd/StatsD

kmshultz: Re: Ilan's comment, looking at the Agent config file, I think Matt to advise forwarding StatsD-collected metrics to the Agent Forwarder (port 17123), not to DogStatsD.

But if this is right thing to recommend, it's confusing to first state that DogStatsD serves as a StatsD forwarder, and then to advise forwarding to the Agent Forwarder process.

In any case, we shouldn't provide instruction on this here. We should document it elsewhere (one of the Agent pages) if it isn't already and link to it from here.

technovangelist: the problem is that folks install the integration thinking that they need to to get metrics from a 3rd party statsd agent thingy. this is only for monitoring that statsd thing

content/integrations/statsd.md
technovangelist: this is for monitoring a statsd server that is not us

kmshultz: Define 'us'? I understand it's not for monitoring DogStatsD, it's for monitoring a vanilla StatsD. The question is, are people more often using it to monitor StatsD daemons running on their own infrastructure, or running elsewhere?

content/integrations/statsd.md
kmshultz: > the problem is that folks install the integration thinking that they need to to get metrics from a 3rd party statsd agent thingy.

I don't follow...

Whatever the confusion is, linking them to the DogStatsD page doesn't help, as that page doesn't describe how to forward StatsD metrics to Datadog. Even if it did describe it, the sentence is still confusing:

> Please note the Datadog Agent includes DogStatsD, which serves as a StatsD forwarder.

Readers who don't follow the link may think: "Great, yes, that's what I want to do! Sounds like I'm all set." The sentence makes it sound like StatsD forwarding is working out of the box as long as dd-agent is installed.

kmshultz: Also the second bullet point:

> Collecting metrics from a StatsD server's admin interface

*does* make it sound like this will connect to StatsD and pull in all the application metrics it is collecting.

Updated proposal:

> Monitor the availability of your StatsD servers and track how busily they're processing application metrics.
>
> This Agent check does NOT help you forward application metrics from your StatsD servers to Datadog. See [whatever documentation] to learn how to do that.

irabinovitch: I agree that the original text being and the text being proposed is unclear at the moment. It's better than what we had but it will still confuse people.

kmshultz: What about my proposed text? (in the comment above yours) How can it be better?

content/integrations/statsd.md
```diff
@@ -7,12 +7,9 @@ git_integration_title: statsd
 
 # Overview
 
-Enable the Datadog StatsD integration to:
+This Agent check monitors the availability and uptime of non-Datadog StatsD servers. It also tracks the rate StatsD servers are receiving metrics.
```
kmshultz: Clean and nice! Maybe end with "receiving application metrics", though maybe "application" qualifier is unnecessary.

PR #1307, opened 2017-05-15
Update Monitor Guide with new Manage Monitors page/query syntax.

kmshultz:
WIP, need to add a few images still.

The query syntax is mentioned in the 'Write a Query' section. I've elected to not copy all the detail from Elasticsearch's page - I'd rather let readers go there, and then read on our page how our syntax is unlike ES. @sboak @adimusset if you disagree with this approach, let me know, and I can add more comprehensive detail.


content/guides/monitors.md
```diff
@@ -226,10 +226,68 @@ To schedule downtime, click the "Schedule Downtime" button in the upper right.
...
+* The Muted checkbox appears under the Status field, but Muted is actually its own field; checking it adds `muted:true` to your query, not `status:muted`.
+* The Metric/Check field is always called `metric` in the query, e.g. selecting the check `http.can_connect` adds `metric:http.can_connect` to your query.
+
+For fields that have an arbitrary (i.e. large) number of values across all monitors—Service tag, Scope, Metric/Check, Notification—use the field-specific search bars to find the value you're looking for.
+
+When you need to run a more complex search than the checkboxes allow, use the search bar to edit your query or write a new one.
+
+#### Write a query
+
+The most common reason to write a query is to search for specific text across all monitor names (i.e. titles) and message bodies. A simple search of `postgresql` will return all monitors with `postgresql` anywhere in the name or message body. To search on title or message body, but not both, qualify the search term with the field name, e.g. `name:postgresql`.
```
adimusset: we have not decided yet if we call the monitor title `name` or `title`.

content/guides/monitors.md
```diff
@@ -226,10 +226,68 @@ To schedule downtime, click the "Schedule Downtime" button in the upper right.
...
+
+Otherwise, you can use boolean operators (AND, OR, and NOT) and parentheses to write complex queries using any monitor fields. The search syntax is very similar to that of [Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/query-dsl-query-string-query.html#query-string-syntax), so it's easiest to describe how it is *not* like Elasticsearch syntax:
+
+* Regular expressions are not supported
+* Single-character wildcard (`?`) is not supported, but the general wildcard (`*`) is
+* Proximity searches are not supported, but the [fuzzy](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/query-dsl-query-string-query.html#_fuzziness) operator is
+* Ranges are not supported
+* Boosting is not supported
+* The set of [reserved characters](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/query-dsl-query-string-query.html#_reserved_characters) for monitor search is more limited. The characters are: `-`, `(`, `)`, `"`, `~`, `*`, `:`
```
adimusset: the special characters are `(`, `)`, `"`,  `*`, `:` in our syntax - the only way to escape them is to use quotes. (So you can't escape `"`). Important to note that all other special characters (eg `-`) need to be used with quotes (`.` is an exception as it is very common, maybe `-` should be an exception too)

adimusset: ` ` is also a special character

adimusset:
I left two comments, looks good ! Just to be clear the current list of facets available is 
```
"type" / "status"  / "creator" / "scope" / "env" / "service" / 
"team" / "notification" / "metric" / "muted" / "monitor_type_id" /
"title" / "message" / "id"
```
`monitor_type_id` is an internal concept
About the redirection to ES docs, it may be confusing because we do not support parentheses between facets. For example `(metric:cpu OR type:apm) notification:slack` does not work, but `metric:cpu OR NOT type:apm` does work without problem.

content/guides/monitors.md
adimusset: we went with `title`

kmshultz:
@adimusset I corrected the list of special characters and the `title` vs `name` thing. Since this is live on prod now, I'll go ahead and merge this.

I think I'm still fuzzy on what's allowed with parens and also how you've described (far as I read it) two different sets of special characters.

Some examples for each would be instructive (though you already provided one about the parens).

I'll go ahead and merge this, but I'm guessing we'll add more to it on Monday, including perhaps a few screenshots.

kmshultz:
Oops, I screwed something up when pushing the branch. Abandoning this in favor of #1338.

content/guides/monitors.md
kmshultz: @adimusset the column in the UI is still labeled 'NAME'. I suppose we want to change this to 'TITLE'?

Issue #1242, opened 2017-04-07
Missing information about escaping line break in dogstatsd event's text


hush-hush:
This documentation: http://docs.datadoghq.com/guides/dogstatsd/#fields-1

State that `text — Event text. Supports line breaks.`. Actually line break are used to separate multiple metrics in one payload to DogStatsD.

People must escape their line break in event's text to work ("\n" -> "\\\\n"). We should explain that in the documentation.

Thanks.

PR #1308, opened 2017-05-15
Add caveat that line breaks in DogStatsD fields must escape the slash.

kmshultz:
@hush-hush brought this up in #1242.

### DIFF
--- a/content/guides/dogstatsd.md
+++ b/content/guides/dogstatsd.md
@@ -298,7 +298,7 @@ the packets:
 #### Fields
 - Mandatory:
   - `title` — Event title.
-  - `text` — Event text. Supports line breaks.
+  - `text` — Event text. Insert line breaks with an escaped slash (`\\n`)
 - Optional: `|[key]:[value]`
   - `|d:date_happened` — default: None — Assign a timestamp to the event. Default is the current Unix epoch timestamp when not supplied.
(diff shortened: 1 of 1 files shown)


content/guides/dogstatsd.md
```diff
@@ -298,7 +298,7 @@ the packets:
 #### Fields
 - Mandatory:
   - `title` — Event title.
-  - `text` — Event text. Supports line breaks.
+  - `text` — Event text. Insert line breaks with an escaped slash (`\\n`)
```
kmshultz: @technovangelist hmm, yes, it seems we're documenting in three different places what the two required fields are for events. This is totally unnecessary, I'll remove these soon in a followup to this PR.

kmshultz: @hush-hush what is the particular case @technovangelist is mentioning?

hush-hush: @kmshultz @technovangelist people can send multiple DogStatsD command in one UDP packet if they are split by line break. It's part of the StatsD protocol upon which DogStatsD is based.

For example instead of doing:
`some.metric.a:0.5|g`

then:
`some.metric.b:0.5|g`

People can do
`some.metric.a:0.5|g\nsome.metric.b:0.5|g`

see here: https://github.com/etsy/statsd/blob/master/docs/metric_types.md#multi-metric-packets

PR #1311, opened 2017-05-16
added way to specify platform specific metrics

technovangelist:
add platformmetrics to the frontmatter and then the metric name followed by a list of platforms you can find that metric on and the metric name in the table will be followed by **platform only where platform is the name of the platform. If you specify Windows and Solaris, it will say  **Windows and Solaris only. If you have more than 2 platform, it puts the commas in the right way.

I added the metrics for disk and network that were mentioned in the card. But I am sure there are others.

### DIFF
--- a/content/integrations/disk.md
+++ b/content/integrations/disk.md
@@ -3,6 +3,11 @@ title: Datadog-Disk Integration
 integration_title: Disk Check
 kind: integration
 newhlevel: true
+platformmetrics:
+  system.disk.read_time_pct:
+    - Windows
+  system.disk.write_time_pct:
+    - Windows
 ---
 
 # Overview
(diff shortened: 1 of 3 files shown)


technovangelist:
screen shot 2017-05-16 at 6 18 18 pm

kmshultz:
I like this idea but I don't think front matter is the place for it. This metadata needs to be in metadata.csv along with the rest of the metadata. Recently, I've just been cramming extra metadata into the description field, but at some point soon, we should coordinate with whomever owns metrics metadata to get some new fields (i.e. columns) added.

technovangelist:
getting it into metadata is a good next step, but there are lots of things that have been tried to get inther for a long time and its not there yet. I think we need a good interim step and this is a good place for it. Lets get something working to get past the support calls now and get something better for the long term later on.

kmshultz:
Fair enough, LGTM.

PR #1320, opened 2017-05-18
DogStatsD pages refactor

kmshultz:
Right now we've got two very similar pages about DogStatsD:

docs.datadoghq.com/guides/dogstatsd/
docs.datadoghq.com/guides/metrics/

When we do the Docs Redesign of the Future we may organize things differently than this PR proposes, but for now, it's irritating to have two DogStatsD references to maintain. I vote we merge them into one since there's quite a bit of duplicate content.

(I say reference**s** because while docs.datadoghq.com/guides/metrics/ presents itself as a Guide, IMO it's more of a reference, like docs.datadoghq.com/guides/dogstatsd/ (which currently calls itself a 'tutorial' (it's not))).

Even if we want to keep them separate, this PR updates, clarifies, and reorganizes the page. I have rewritten a lot based on my interpretation of what is there currently, and my interpretation may be wrong in some cases, so this DOES need general review for correctness.

Otherwise, if folks concur on merging this into one page, I'll remove the other page, track down pages that reference it and change their links, etc.


content/guides/dogstatsd.md
```diff
@@ -25,315 +25,333 @@ sidebar:
...
+* Tagging
 
 ## How It Works
 
-DogStatsD accepts custom application metrics points over
-<a href="http://en.wikipedia.org/wiki/User_Datagram_Protocol">UDP</a>,
-and then periodically aggregates and forwards the metrics to Datadog, where
-they can be graphed on dashboards. Here's a pretty standard DogStatsd setup:
+DogStatsD accepts your custom metrics, events, and service checks over UDP and periodically aggregates and forwards them to Datadog. Because it uses UDP, your application 
+can fire metrics at DogStatsD and resume its work without waiting for a response. If DogStatsD ever crashes
```
irabinovitch: can metrics via DogStatsD asynchronously, so that it may resume its work without waiting for a response.

content/guides/dogstatsd.md
```diff
@@ -25,315 +25,333 @@ sidebar:
...
 
 ## How It Works
 
-DogStatsD accepts custom application metrics points over
-<a href="http://en.wikipedia.org/wiki/User_Datagram_Protocol">UDP</a>,
-and then periodically aggregates and forwards the metrics to Datadog, where
-they can be graphed on dashboards. Here's a pretty standard DogStatsd setup:
+DogStatsD accepts your custom metrics, events, and service checks over UDP and periodically aggregates and forwards them to Datadog. Because it uses UDP, your application 
+can fire metrics at DogStatsD and resume its work without waiting for a response. If DogStatsD ever crashes 
+or begins to perform poorly, your application won't skip a beat.
```
irabinovitch: lets say "is unavailable" rather than talk about poor performance or crash.

content/guides/dogstatsd.md
kmshultz: > can metrics via

Did you leave out a word? Didn't like "fire"?

I thought about saying 'asynchronously' but opted for a layman description instead.

irabinovitch: fire just felt to casual to me.

content/guides/dogstatsd.md
```diff
@@ -25,315 +25,331 @@ sidebar:
...
+
 <p>
 <img src="/static/images/dogstatsd.png"/>
 </p>
 
-### Aggregation
+As it receives data, DogStatsD aggregates multiple data points for each unique metric into a single
```
irabinovitch: Is this accurate it? Doesn't it emit multiple quantiles/percentiles per datapoint?

irabinovitch: I guess it depends on the metric type

kmshultz: I'm not sure. The PR removes this statement:

> This means expect DogStatsD to produce one point per metric per flush interval while data is being submitted for that metric.

Which seems to be saying just that.

(I removed it because it's redundant with the early point, and is poorly worded, not because I thought it was incorrect.)

content/guides/dogstatsd.md
kmshultz: Agreed, changed.

kmshultz:
@hush-hush You recently commented on another DogStatsD related docs issue. Can you please review this rewrite for correctness? If you're not the best reviewer DogStatsD stuff, can you loop in someone who is? Thanks!

content/guides/dogstatsd.md
```diff
@@ -25,315 +25,330 @@ sidebar:
...
+EOF
+%>
 
-<%= python <<eof
+Before sending the metric to Datadog, DogStatsD uses the `sample_rate` to 
+correct the metric value, i.e. to estimate what it would have been without sampling.
 
-# Post a simple message
-statsd.event('There might be a storm tomorrow', 'A friend warned me earlier.')
+You can specify a sample rate for any metric type.
```
hush-hush: sample rates are only handle by Counter and Histogram.

content/guides/dogstatsd.md
```diff
@@ -25,315 +25,330 @@ sidebar:
...
-- Optional: `|[key]:[value]`
-  - `|d:date_happened` — default: None — Assign a timestamp to the event. Default is the current Unix epoch timestamp when not supplied.
-  - `|h:hostname` — default: None — Assign a hostname to the event.
-  - `|k:aggregation_key` — default: None — Assign an aggregation key to the event, to group it with some others.
-  - `|p:priority` — default: 'normal' — Can be “normal” or “low”.
-  - `|s:source_type_name` — default: None — Assign a source type to the event.
-  - `|t:alert_type` — default: 'info' — Can be “error”, “warning”, “info” or “success”.
-  - `|#tag1:value1,tag2,tag3:value3` — default: None. <strong><em><br/>
-  Note: The `:` in tags is part of the tag list string and has no parsing purpose like for the other parameters.</em></strong>
+- `_e` - the datagram must begin with `_e`
```
hush-hush: Could be nice to add an example.

content/guides/dogstatsd.md
```diff
@@ -25,315 +25,330 @@ sidebar:
...
-  - `|m:service_check_message` — A message describing the current state of the service check. <em>This field should always be
-positioned last among the metadata fields.</em>
-
-
-
-## Source
-
-DogStatsD is open-sourced under the BSD License. Check out the source
-[here](https://github.com/DataDog/dd-agent).
+- `_sc` — the datagram must begin with `_sc`
```
hush-hush: Here also an example could be nice

content/guides/dogstatsd.md
kmshultz: @hush-hush interesting, the client libraries support passing it for other types: http://datadogpy.readthedocs.io/en/latest/#datadog.dogstatsd.base.DogStatsd.gauge

And I see that DogStatsD parses the sample_rate field in any metric packet. I guess it just ignores it for e.g. gauge? https://github.com/DataDog/dd-agent/blob/master/aggregator.py#L63

content/guides/dogstatsd.md
kmshultz: 👍 I'll add some

content/guides/dogstatsd.md
kmshultz: Seems like we should remove the 'sample_rate' option from DogStatsD client libraries.

kmshultz:
@hush-hush I addressed your comments, please review again.

I just made a commit that removes /guides/metrics. Now, as I was changing all links that point to /guides/metrics, I just now realized that this guide is linked to from the front page: 'Dive into Metrics!' We should more carefully consider the content of DogStatsD because of this. For instance, the new page title I proposed is 'What is DogStatsD?', but even the current title 'Getting Started with DogStatsD' is bit disorienting to see when clicking the link text 'Dive into Metrics!'

**Do we want to consider tweaking the home page at this time? That is, changing step 2 to be something else, e.g. '[Enable some Integrations!](/integrations)' instead?**

Finally: There *is* some content from /guides/metrics that I didn't copy over to DogStatsD, as I don't think it belongs in either of these two docs:

* 'Metric names' (name requirements) - I'm linking to /faq/#api instead
* Some general description about what tagging is - maybe we link to /guides/tagging
* 'Other Submission Methods' - I link to the API docs in a new section 'Related Reading', but not to this Dropwizard Java metrics library or the metrics-datadog backend. It's weird to specifically call these out here. I *do* link to the libraries page in Related Reading.
* Seeing your Custom Metrics - this is general Datadog 101 stuff

**So, being that the last 3 have probably been useful to really green users who visit /guides/metrics from the homepage, and they don't make sense to include in /guides/dogstatsd, this *might* be a good time to reconsider the home page strategy.**

^ @irabinovitch @technovangelist @jyee

kmshultz:
Another option: we change the 'Dive into Metrics!' link to a new page /metrics that introduces what metric types Datadog accepts and _briefly_ explains—with links to relevant pages—the different ways one can get them into Datadog.

content/guides/dogstatsd.md
hush-hush: Yes, sample rate is ignored for most metrics but is parsed.

But, for example, a `sample_rate`  for gauge doesn't change anything to the final value.

irabinovitch:
Looks like we've got some merge conflicts here. Lets fix and merge into staging so we can review.

kmshultz:
@irabinovitch rebased and fixed merge conflicts. Pushed to staging.

kmshultz:
Closing in favor of the less disruptive #1372 for now.

PR #1324, opened 2017-05-22
Added list of deprecated metrics

masci:
The `metadata.csv` catalog was updated but we should list the metrics that are reported by the now deprecated `docker` check.


irabinovitch:
@xvello lets generate a list of customers still using the old integration. PM and Support can then work on communications.

masci:
Merging this in the meantime

PR #1325, opened 2017-05-22
Correct wording from disable to promote

Hasjamesn:
Correct wording from disable to promote in the "Promote Existing Members to Admin" documentation.


kmshultz:
Thank you @Hasjamesn!

PR #1327, opened 2017-05-23
Add note about disabling escaping in template variables

isaacdd:


content/monitoring.md
```diff
@@ -426,6 +426,10 @@ triggering scope into the message of multi alerts.
 
     ![short template var trigger](/static/images/monitor/templatevar_short.png)
 
+    Note that content of template variables is escaped by default. If your variable
```
kmshultz: 1.
>  Note that content of template variables is

Note that template variable content is

2.
> you would like to not escape

you would NOT like to be escaped

3. s/json/JSON

PR #1329, opened 2017-05-31
Add another R API client

yutannihilation:
Recently, I've released a simple R package, datadogr, for getting metrics. This PR adds it to the libraries page.

https://cran.r-project.org/package=datadogr

Though rdog is already neat, it is not officially released yet on CRAN. This is OK for casual people like me who loves GitHub, but some R users uses CRAN packages only. So, it would be great if my package will help those who loves R and Datadog :)

### DIFF
--- a/content/libraries.md
+++ b/content/libraries.md
@@ -107,6 +107,7 @@ Some great folks have written their own libraries to help interact with Datadog.
 #### R
 {: #community-r}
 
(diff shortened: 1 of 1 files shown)


kmshultz:
@yutannihilation we'll definitely add this - Thanks!

yutannihilation:
Thanks for merging!

PR #1333, opened 2017-06-02
updates to docker

irabinovitch:


irabinovitch:
@xvello were there other items you wanted updated here? We'll do a full revamp soon, but I wanted to get the specific out of date issues you mentioned quickly.

content/integrations/docker.md
```diff
@@ -19,7 +19,7 @@ Get metrics from Docker in real time to:
 
 # Installation
 
-There are three ways to setup the Docker integration: install the agent on the host, on a single privileged container, and on each individual container.
+There are two ways to setup the Docker integration: install the agent on the host or on a single sidecar container.
```
xvello: IIRC a sidecar container is tied to a "main" container, which is not the case here (dd-agent is running by itself).
I'd go with
> There are two ways to setup the Docker integration: running our docker-dd-agent container or installing the agent on the host.

xvello:
We could also add a link to the [up-to-date list of supported envvars](https://github.com/DataDog/docker-dd-agent#environment-variables) at the end of
https://github.com/DataDog/documentation/blob/7567b11c570cb8eeee1725a72e9cefd1b0dfd4a1/content/integrations/docker.md#environment-variables

content/integrations/docker.md
kmshultz: s/agent/Agent

Also, what is 'the host'? Many readers will know, but it's better to be explicit: 'Install the Agent on any container-running host, or run the docker-dd-agent container.'

technovangelist: but there is that 3rd way to install. in each container. There was a specific scenario where support was recommending that (certain metrics not being collected). you lose out on the container pricing since you have so many agents installed, but its an option.

xvello: The goal of this PR is specifically remove that way from official documentation, as this creates more issues than it solves, and several clients were mislead by that.
Indeed, some integrations need work, but we should keep that container-as-new-host method as a KB and not a public documentation.

irabinovitch: We definitely do not want to support in each container.

kmshultz:
@irabinovitch We also need to remove the **Each Container Installation** section beginning on line 105.

kmshultz:
Re: @xvello's comment:

> We could also add a link to the up-to-date list of supported envvars at the end of
https://github.com/DataDog/documentation/blob/7567b11c570cb8eeee1725a72e9cefd1b0dfd4a1/content/integrations/docker.md#environment-variables

Why don't we link directly to the docker-dd-agent repo? Or even better, we pull down parts of that repo's README to include here (more involved effort; that'd be for another PR).

kmshultz:
@xvello Please review the commit I just pushed and advise - Thanks!

PR #1340, opened 2017-06-05
Mattw/cleanagentusage

technovangelist:
There were some errors on build of the agent usage pages. this clears that up.


layouts/partials/BasicAgentUsage-nix.html
```diff
@@ -76,14 +76,13 @@
...
+<%= $sudo %><%= @item[:servicename] %> check &#91;integration&#93;
 
 Add the check_rate argument to get the most recent values for rates:
 
-    <%= $sudo %><%= @item[:servicename] %> check [integration] check_rate
+<%= $sudo %><%= @item[:servicename] %> check &#91;integration&#93; check_rate
 
-<%= $sudo %>
-  <%= @item[:servicename] %> check [integration]
+<%= $sudo %><%= @item[:servicename] %> check &#91;integration&#93;
```
kmshultz: This line is identical to line 79, why's it down here, too? Let's remove?

technovangelist: you pointed out the real issue. it wasn't the brackets. well it is when its not code. being indented 2 tabs (4 spaces) means the markdown is not interpreted and nanoc doesn't complain. this extra line was not indented. my fix fixed it, but the fix worked around the real issue.

PR #1341, opened 2017-06-05
Add a few examples of quoted and quote-less queries using special chars.

kmshultz:


adimusset:
I think this is a bit confusing - our special characters (eg they have a meaning in the query) are `(`, `)`, `"`, `*`, `:` so you can use them without quotes and they will mean something. Any special character (those included) that you want to make a string search on needs to be between quotes (`.` is an exception but it's not a big deal)

kmshultz:
@adimusset ok, so it seems like we need to distinguish between the two different "kinds" of special characters, because we seem to have been using two meanings for the term 'special character'. This is why I've been confused. 

I'll push an update soon.

kmshultz:
> (. is an exception but it's not a big deal)

And `*` is an exception too, correct?

adimusset:
not really because it does not mean `"*"`, it means the `*` wildcard. So it's not exactly like `.`. I think I have not been clear enough before, and some behaviors have changed so it can be hard to follow

kmshultz:
Hmm, maybe it'd be clearest if I just read the code? Can you link me (offline/Slac) to the relevant bits of code? Or is the code scattered about in various places?

kmshultz:
@adimusset I just clarified this (I think). Let me know if it reads better now.

PR #1344, opened 2017-06-07
Fix minor register-task-definition formatting issue

nickfishman:
Copying the command that's currently live results in only a single dash (`–cli-input-json`) which doesn't work when pasted into the terminal.


kmshultz:
Good catch @nickfishman — Thanks!

PR #1349, opened 2017-06-09
Sync APM FAQs from dogweb, other FAQ cleanup

kmshultz:


kmshultz:
@awang does this look accurate?

PR #1357, opened 2017-06-16
Autodiscovery guide updates

kmshultz:
TODO:

* Verify and finish list of caveats under #how-it-works/#different-execution.
* Clean up #service-identifiers section at the end
* Add more detail to Kubernetes instructions, i.e. actual commands


kmshultz:
I've introduced a few new terms here:

- **service identifier** — the document needs a convenient way to refer to image name/label. I considered 'container identifier' but that may not be so backend-agnostic (one day when we have non-Docker backends)
- **template source** — what files, KV stores, and K8s annotations are. It would seem we already had a term like this—Configuration Backend—except it only applies to KV stores. Calling them all configuration backends could create confusion that e.g. 'kubernetes' is a valid value for `SD_CONFIG_BACKEND`.

Voice any opposition to these terms, dear readers, or forever hold your peace.

cc @technovangelist if others concur on these terms, we should use them in the Autodiscovery video, too.

content/guides/autodiscovery.md
```diff
@@ -4,37 +4,75 @@ kind: guide
...
+
+As the Agent inspects each template and its service identifiers, it builds a mapping of service identifiers to template YAML (or JSON). Then, it inspects every running Docker container on the Agent's host, looking for containers that match any of the service identifiers (keys) in the identifier-to-template mapping. For each match, the Agent generates a static check configuration by substituting the matching container's IP address and port into the corresponding template. Then it enables the check.
+
+The Agent watches for Docker events—container creation, destruction, starts, and stops—and recompiles check configurations from the templates on such events. Any time Autodiscovery finds another instance of the some service it had previously generated configuration for—whether it's an additional container or the previous one redeployed—it builds and enables another static configuration.
+
+There are a few caveats:
+
+- The Agent caches templates it gets from KV stores. Changes to KV store templates require a restart of the Agent.
+- something about precedence of templates when multiple sources provided?
```
xvello: KV-store templates take precendence over conf.d/auto_conf files indeed

content/guides/autodiscovery.md
```diff
@@ -4,37 +4,75 @@ kind: guide
...
+
+As the Agent inspects each template and its service identifiers, it builds a mapping of service identifiers to template YAML (or JSON). Then, it inspects every running Docker container on the Agent's host, looking for containers that match any of the service identifiers (keys) in the identifier-to-template mapping. For each match, the Agent generates a static check configuration by substituting the matching container's IP address and port into the corresponding template. Then it enables the check.
+
+The Agent watches for Docker events—container creation, destruction, starts, and stops—and recompiles check configurations from the templates on such events. Any time Autodiscovery finds another instance of the some service it had previously generated configuration for—whether it's an additional container or the previous one redeployed—it builds and enables another static configuration.
+
+There are a few caveats:
+
+- The Agent caches templates it gets from KV stores. Changes to KV store templates require a restart of the Agent.
```
xvello: Reading the source, it looks like we are watching for changes in the KV-store and reloading templates when one is detected. @hkaj is that working on all three supported backends?

content/guides/autodiscovery.md
```diff
@@ -48,155 +86,169 @@ By default, the Datadog Agent includes Autodiscovery support for:
 - Redis
 - Riak
 
-These are provided by the configuration templates in the Datadog Agent `conf.d/auto_conf` directory.
+Storing templates as local files is easy to understand and doesn't require an external service. The downside is that you must redeploy the Agent container each time you change, add, or remove templates. You may also have to maintain your own docker-dd-agent container if you want to add your own templates.
```
xvello: Should we mention the /conf.d volume they can expose, instead of rebuilding the image? See https://github.com/DataDog/docker-dd-agent/blob/master/entrypoint.sh#L172

content/guides/autodiscovery.md
```diff
@@ -48,155 +86,169 @@ By default, the Datadog Agent includes Autodiscovery support for:
...
 
- When a list of values is expected for the variable and selecting a specific one is mandatory, you can specify the value from the list by appending an underscore followed by an index or key. For example `%%host_0%%` or `%%port_4%%`. Note that indexes begin at 0 and if no index is provided, the last value in the value list ordered increasingly will be used.
+### Template Source: Key-value Store
 
-Let's take the example of the port variable: a RabbitMQ container with the management module enabled has 6 exposed ports by default. The list of ports as seen by the agent is: `[4369, 5671, 5672, 15671, 15672, 25672]`. **Notice the order. The Agent always sorts values in ascending order.**
+Autodiscovery supports Consul, etcd, and Zookeeper as template sources. To use a KV store, configure its parameters in `datadog.conf` or in environment variables passed to docker-dd-agent when starting the container.
```
xvello: we could link `environment variables` to https://github.com/DataDog/docker-dd-agent/#environment-variables

content/guides/autodiscovery.md
```diff
@@ -48,155 +86,169 @@ By default, the Datadog Agent includes Autodiscovery support for:
...
+# If you wish otherwise, uncomment this option and modify its value.
+# sd_template_dir: /datadog/check_configs
 
-First you'll need to configure etcd or Consul as your Autoiscovery backend by either updating the `datadog.conf` file or passing the settings as environment variables when starting the Datadog Agent service.
+# If you Consul store requires token authentication for service discovery, you can define that token here.
+# consul_token: f45cbd0b-5022-samp-le00-4eaa7c1f40f1
+~~~
 
-### Configuring etcd or Consul in `datadog.conf`
+If you are using Consul and the Consul cluster requires token authentication, set `consul_token`.
```
xvello: FWI, 5.15 will support user/pass auth for etcd. See https://github.com/DataDog/dd-agent/pull/3357

content/guides/autodiscovery.md
```diff
@@ -48,155 +86,169 @@ By default, the Datadog Agent includes Autodiscovery support for:
...
+
+# Reference
+
+### Template Variable Indexes
+
+For containers that have many IP addresses or listens on many ports, you can tell Autodiscovery which ones to choose by appending an underscore to the template variable, followed by an index, e.g. `%%host_0%%`, `%%port_4%%`. After inspecting the container, Autodiscovery sorts the IPs and ports **numerically and in ascending order**. For a container that listens on ports 80, 443, and 8443, `%%port_0%%` refers to port 80. Unindexed template variables refer to the last item in the sorted list, so in this case, `%%port%%` means port 8443.
+
+You can also add a network name suffix to the `%%host%%` variable—`%%host_bridge%%`, `%%host_swarm%%`, etc—for containers attached to multiple networks. When `%%host%%` does not have a suffix, Autodiscovery picks the container's bridge network IP address.
+
+### Service Identifiers
```
xvello: This section should be expanded with examples and pushed further up the page.
See the ticket I sent you about user confusion on the matching

technovangelist:
re service identifiers. since services are a thing in k8s, is it worth/possible to choose a different word, esp since this service is different from that service?

content/guides/autodiscovery.md
```diff
@@ -4,37 +4,75 @@ kind: guide
...
 
-Datadog automatically keeps track of what is running where, thanks to its Autodiscovery feature. Autodiscovery allows you to define configuration templates that will be applied automatically to monitor your containers.
+The Datadog Agent can automatically track which services are running where, thanks to its Autodiscovery feature. Autodiscovery lets you define configuration templates for Agent checks and specify which container types each check should apply to. The Agent enables, disables, and recompiles static check configurations from the templates as containers come and go. When your NGINX container moves from 10.0.0.6 to 10.0.0.17, Autodiscovery helps the Agent update its NGINX check configuration with the new IP address so it can keep collecting NGINX metrics without any action on your part.
```
technovangelist: ...which container types...??? are they different types or just different containers?

The agent....recompiles.... ?? Recompiles? That sounds super impressive but is that accurate? maybe I am assuming too much from that word.

content/guides/autodiscovery.md
```diff
@@ -4,37 +4,75 @@ kind: guide
...
 
-## How it works
+<div class="alert alert-info">
+Autodiscovery was previously called Service Discovery. It is still known as Service Discovery in the Agent's code and in configuration options.
```
technovangelist: this sounds a bit awkward. It was previously called...and it still is...

content/guides/autodiscovery.md
```diff
@@ -4,37 +4,75 @@ kind: guide
...
+# How it Works
 
-The Autodiscovery feature watches for Docker events like when a container is created, destroyed, started or stopped. When one of these happens, the Agent identifies which service is impacted, loads the configuration template for this image, and automatically sets up its checks.
+In a traditional non-container environment, Datadog Agent configuration is, like the environment in which it runs, static. The Agent reads check configurations from disk when it starts, and as long as it's running, it continuously applies every configured check. The configuration files are static, and any network-related options configured within them serve to identify specific instances of a monitored service. When an Agent check cannot connect to such a service, it's probably not because the service was re-homed somewhere else; either the service is down, or the check configuration was unable to reach the service to begin with. In any case, the Agent continues to attempt connecting to the service until an administrator steps in to troubleshoot.
```
technovangelist: Thats a lot, of commas, in the, first, sentence.

for the second sentence, the way I read that is that if the config changes, the changes are applied right away without restarting the agent, but then you say the files are static. a bit confusing
 
When an Agent check cannot connect to such a service.... you aren't talking about a service and havent defined one.. 

either the service is down, or the check configuration was unable to reach the service to begin with. - the configuration cannot reach it, or the check cannot??

until an administrator steps in to troubleshoot... just the act of troubleshooting stops the attempt to connect? wont it still try. If the admin successfully solves the problem rather than just troubleshooting, i hope it will still attempt to connect.

content/guides/autodiscovery.md
```diff
@@ -4,37 +4,75 @@ kind: guide
...
 
-Configuration templates can be defined by simple template files or as single key-value stores using etcd or Consul.
+With Autodiscovery enabled, the Agent runs checks differently.
```
technovangelist: is this sentence necessary?

content/guides/autodiscovery.md
```diff
@@ -4,37 +4,75 @@ kind: guide
...
 
-## How to set it up
+### Different Configuration
 
-To use Autodiscovery, you'll first need to run the Datadog Agent as a service.
+First, Autodiscovery uses **templates** for check configuration, wherein two template variables—`%%host%%` and `%%port%%`—appear in place of any normally-hardcoded network option values. Because orchestration platforms like Docker Swarm deploy (and redeploy) containers on arbitrary hosts, static configuration files are not suitable for checks that collect data from network endpoints. For example: a template for the Agent's [Go Expvar check](https://github.com/DataDog/integrations-core/blob/master/go_expvar/conf.yaml.example) would contain the option `expvar_url: http://%%host%%:%%port%%`. For containers that have more than one IP or exposed port, Autodiscovery can pick the right one(s) using [template variable indexes](#template-variable-indexes).
```
technovangelist: normally-hardcoded? is it better to just say hardcoded?

%%port%%`—appear  .... missing a space?

content/guides/autodiscovery.md
```diff
@@ -4,37 +4,75 @@ kind: guide
...
 
-In Docker Swarm, you can do this by running the following command on one of your manager nodes (using your [API key](https://app.datadoghq.com/account/settings#api)):
+Second, because templates don't identify specific instances of a monitored service—which host? which port?—Autodiscovery needs one or more **service identifiers** for each template so it can find real values for the template variables. For Docker, service identifiers are [image names or container labels](#service-identifiers). With identifiers in hand, Autodiscovery can figure out which IP(s) and port(s) to substitute into the template.
```
technovangelist: a monitored service—which host? which port?—Autodiscovery needs... i don't understand the which host which port part.

For Docker, service identifiers are [image names or container labels](#service-identifiers).... should this be docker swarm? since k8s could be docker as well.

content/guides/autodiscovery.md
```diff
@@ -4,37 +4,75 @@ kind: guide
...
+
+As the Agent inspects each template and its service identifiers, it builds a mapping of service identifiers to template YAML (or JSON). Then, it inspects every running Docker container on the Agent's host, looking for containers that match any of the service identifiers (keys) in the identifier-to-template mapping. For each match, the Agent generates a static check configuration by substituting the matching container's IP address and port into the corresponding template. Then it enables the check.
+
+The Agent watches for Docker events—container creation, destruction, starts, and stops—and recompiles check configurations from the templates on such events. Any time Autodiscovery finds another instance of the some service it had previously generated configuration for—whether it's an additional container or the previous one redeployed—it builds and enables another static configuration.
```
technovangelist: Any time Autodiscovery finds another instance of the some service.... the some service???

whether it's an additional container or the previous one redeployed.... previous??? or current?

content/guides/autodiscovery.md
```diff
@@ -4,37 +4,75 @@ kind: guide
...
+
+There are a few caveats:
+
+- The Agent caches templates it gets from KV stores. Changes to KV store templates require a restart of the Agent.
+- something about precedence of templates when multiple sources provided?
+- others?
+
+---
+
+The next section clarifies this abstract description of Autodiscovery with some examples.
```
technovangelist: is this needed?

content/guides/autodiscovery.md
```diff
@@ -4,37 +4,75 @@ kind: guide
...
+
+No matter what container orchestration platform you use, you'll first need to run a single [docker-dd-agent container](https://hub.docker.com/r/datadog/docker-dd-agent/) on every host in your cluster. Then you'll write check templates, storing them in one or more template sources.
+
+## Running the Agent Container
+
+If you use Kubernetes, see the [Kubernetes integration page](http://docs.datadoghq.com/integrations/kubernetes/#installation) for instructions on running docker-dd-agent. If you use Amazon ECS, see [its integration page](http://docs.datadoghq.com/integrations/ecs/#installation).
+
+If you use Docker Swarm, run the following command on one of your manager nodes:
 
     docker service create \
```
technovangelist: shouldnt we be consistent here? if we tell folks to look elsewhere for k8s and ecs, can't we do the same for swarm?

content/guides/autodiscovery.md
```diff
@@ -4,37 +4,75 @@ kind: guide
...
       --mount type=bind,source=/var/run/docker.sock,target=/var/run/docker.sock \
       --mount type=bind,source=/proc/,target=/host/proc/,ro=true \
       --mount type=bind,source=/sys/fs/cgroup/,target=/host/sys/fs/cgroup,ro=true \
-      -e API_KEY=<YOUR API KEY> \
+      -e API_KEY=<YOUR_DATADOG_API_KEY> \
       -e SD_BACKEND=docker \
       datadog/docker-dd-agent:latest
 
-For Kubernetes, you can follow our [Kubernetes Integration](http://docs.datadoghq.com/integrations/kubernetes/) to create a DaemonSet. We also have [Amazon ECS integration instructions](http://docs.datadoghq.com/integrations/ecs/) available.
+Otherwise, see the docker-dd-agent documentation for detailed instructions and a comprehensive list of supported environment variables.
```
technovangelist: otherwise??? if not k8s, docker swarm, or ecs, what else  can we work with?

content/guides/autodiscovery.md
technovangelist: Autodiscovery supports.... key value based template sources

content/guides/autodiscovery.md
```diff
@@ -48,155 +86,169 @@ By default, the Datadog Agent includes Autodiscovery support for:
...
+  check_configs/
+    docker_image_1/                 # service identifier, e.g. httpd
+      - check_names: [<CHECK_NAME>] # e.g. apache
+      - init_configs: [<INIT_CONFIG>]
+      - instances: [<INSTANCE_CONFIG>]
+    ...
+~~~
 
-To pass the settings listed above as environment variables when starting the Datadog Agent in Docker Swarm, you would run the command:
+Each template is defined as a three-tuple: check name, `init_config`, and `instances`. The `docker_images` option from the previous section, which was used to provide service identifiers to Autodiscovery, is not required here; for KV store template sources, service identifiers appear as first-level keys under `check_config`. (Also note, the file-based template in the previous section didn't need a check name; the Agent infers it from the filename.)
```
technovangelist: would the term be three tuple, or just tuple?

content/guides/autodiscovery.md
```diff
@@ -48,155 +86,169 @@ By default, the Datadog Agent includes Autodiscovery support for:
...
+
+~~~
+etcdctl mkdir /datadog/check_configs/httpd
+etcdctl set /datadog/check_configs/httpd/check_names '["apache"]'
+etcdctl set /datadog/check_configs/httpd/init_configs '[{}]'
+etcdctl set /datadog/check_configs/httpd/instances '[{"apache_status_url": "http://%%host%%/server-status?auto"}]'
+~~~
 
-### Template structure in key-value stores
+Notice that each of the three values is a list. Autodiscovery assembles list items into check configurations based on shared list indexes. In this case, it composes the first (and only) check from `check_names[0]`, `init_configs[0]` and `instances[0]`.
```
technovangelist: is it worth saying its a YAML formatted list (or is it aJSON formatted list)?

content/guides/autodiscovery.md
```diff
@@ -48,155 +86,169 @@ By default, the Datadog Agent includes Autodiscovery support for:
...
-          - instances: [{instance_config_1a}, {instance_config_1b}]
-        ...
+~~~
+etcdctl set /datadog/check_configs/httpd/check_names '["apache", "http_check"]'
+etcdctl set /datadog/check_configs/httpd/init_configs '[{}, {}]'
+etcdctl set /datadog/check_configs/httpd/instances '[{"apache_status_url": "http://%%host%%/server-status?auto"},{"name": "My service", "url": "http://%%host%%", timeout: 1}]'
+~~~
 
-Note that in the structure above, you may have multiple checks for a single container. For example you may run a Java service that provides an HTTP API, using the HTTP check and the JMX integration at the same time. To declare that in templates, simply add elements to the `check_names`, `init_configs`, and `instances lists`. These elements will be matched together based on their index in their respective lists.
+Again, the list orders matter. The HTTP check will only work if all its elements have the same index (1) across the lists (they do).
```
technovangelist: list orders or list order?

technovangelist: The HTTP check will only work if all its elements have the same index (1) across the lists (they do).
  I don't understand this

content/guides/autodiscovery.md
```diff
@@ -48,155 +86,169 @@ By default, the Datadog Agent includes Autodiscovery support for:
...
-    etcdctl set /datadog/check_configs/httpd/check_names '["apache"]'
-    etcdctl set /datadog/check_configs/httpd/init_configs '[{}]'
-    etcdctl set /datadog/check_configs/httpd/instances '[{"apache_status_url": "http://%%host%%/server-status?auto"}]'
+~~~
+apiVersion: v1
+kind: Pod
+metadata:
+  name: apache
+  annotations:
+    service-discovery.datadoghq.com/apache.check_names: '["apache","http_check"]'
```
technovangelist: awesome, didn't know you could add multiple checks, but makes sense.

technovangelist:
this feels like a long document. @kmshultz you talked about splitting it up into autodisco on k8s, autodisco on docker, etc. any further ideas on this? If i just care about docker swarm, i probably don't want to be confused about k8s.

kmshultz:
Re: splitting this into smaller documents, I agree we should do that. But as the docs site is currently organized, there's no good way to do that, i.e. since the only semblance of organization we have now is `/guides`, `/references`, `/integrations`, and a few others. We could put AD usage instructions into the K8s integration page, the ECS page, etc (though there is no Docker Swarm integration page), but I'd rather have *those* pages link over to this guide using the relevant anchor, i.e. `/guides/autodiscovery#kubernetes`. Integrations pages should _only_ 1) describe how to configure the integration, and 2) list their published metrics, events, and service checks. 

When we redesign the docs site soon, we can organize it some other way, i.e. `/agent/install#kubernetes`, `/agent/usage#kubernetes`. But even then, I think some of the platform-agnostic info in this doc will remain here (and then, we'll no longer call it a "guide").

Thoughts @technovangelist @jyee @irabinovitch?

content/guides/autodiscovery.md
kmshultz: @xvello got it. And KV takes precedence over k8s annotations, which take precedence over files?

content/guides/autodiscovery.md
kmshultz: Ok, got it. But I think no matter what, users must restart Agent containers to get new check configs, correct? i.e. there is no SIGHUP or similar way to tell Agent to reload the files?

content/guides/autodiscovery.md
kmshultz: Will do, I hadn't addressed this section just yet. I did link to it from earlier in the page; you think it needs to be more prominent, though?

kmshultz:
@technovangelist 

> re service identifiers. since services are a thing in k8s and a different thing in docker swarm, is it worth/possible to choose a different word, esp since this service is different from that service? in fact 'service' is a very overloaded term in this article

Perhaps 'container identifier' is indeed best. Care to offer a suggestion?

content/guides/autodiscovery.md
kmshultz: Hmm, yes, 'containers' is probably better, since individual containers of the same type may be targeted via labels.

kmshultz: Agreed, recompiles is a little puffy. 're-generates'?

content/guides/autodiscovery.md
kmshultz: "Autodiscovery was previously called Service Discovery, and it's still called that in the Agent's code and in configuration options."

content/guides/autodiscovery.md
kmshultz: > Thats a lot, of commas, in the, first, sentence.

In a traditional non-container environment, Datadog Agent configuration is—like the environment in which it runs—static.

kmshultz: > for the second sentence, the way I read that is that if the config changes, the changes are applied right away without restarting the agent

Which words tripped you up, 'continuously applies'? If so, is 'continuously runs' better?

kmshultz: > When an Agent check cannot connect to such a service.... you aren't talking about a service and havent defined one..

The previous sentence says "any network-related options configured within them serve to identify specific instances of a monitored service". Would a parenthetical example have made it clear? "...instances of a monitored service (e.g. a redis instance)".

content/guides/autodiscovery.md
kmshultz: I intentionally left the space out. We haven't been consistent in the docs on whether we have a space on either side of an em-dash, but lately I've been leaving them out.

content/guides/autodiscovery.md
kmshultz: which host? which port? refers to the `%%host%%` and `%%port%%` template variables just referenced. Maybe `which %%host%%`? which `%%port%%` is better.

content/guides/autodiscovery.md
kmshultz: typo, s/some/same

content/guides/autodiscovery.md
kmshultz: That's a good question. Presumably any orchestration platform that runs Docker containers?

content/guides/autodiscovery.md
kmshultz: That's implied by the heading.

content/guides/autodiscovery.md
kmshultz: 3-tuple is most accurate.

content/guides/autodiscovery.md
xvello: In the case of host deployment, restarting the collector (`supervisorctl restart datadog-agent:collector`) is enough. But copying conf files from /conf.d is handled by the container's entrypoint, so a container restart is indeed needed

content/guides/autodiscovery.md
xvello: Well, I had two support cases that could've been solved by using container labels instead of toying around with the image names, so I'd vote for pushing it up. But that's probably sampling biais

content/guides/autodiscovery.md
xvello: k8s annotations take precedence over templates (KV then files)

hkaj:
re https://github.com/DataDog/documentation/pull/1357#issuecomment-310483053 I agree, but at the same time platform-specific instructions for auto discovery could be a separate `guide` for each platform with best practices, recommendations, etc. They could still point to a shared reference about the generics of auto discovery

hkaj:
I think it would also be interesting to have a paragraph and a (couple of) sample(s) for jmxfetch config with auto discovery.

kmshultz:
@xvello I've added some more content here. Mind giving it another look, please? 🙇

content/guides/autodiscovery.md
```diff
@@ -1,40 +1,75 @@
 ---
-title: Guide to using Autodiscovery with Docker
+title: Using Autodiscovery with Docker
 kind: guide
 listorder: 10
 ---
 
-Docker is being [adopted rapidly](https://www.datadoghq.com/docker-adoption/) and platforms like Docker Swarm, Kubernetes and Amazon's ECS make running services easier and more resilient by managing orchestration and replication across hosts. But all of that makes monitoring more difficult. How can you monitor a service which is dynamically shifting from one host to another?
+Docker is being [adopted rapidly](https://www.datadoghq.com/docker-adoption/). Orchestration platforms like Docker Swarm, Kubernetes, and Amazon ECS make running Docker-ized services easier and more resilient by managing orchestration and replication across hosts. But all of that makes monitoring more difficult. How can you reliably monitor a service which is unpredictably shifting from one host to another?
```
irabinovitch: s/Docker/Containers/ ?

irabinovitch: is unpredictably the right word? We didn't like dynamically?

would "dynamically being shifted" make more sense?

content/guides/autodiscovery.md
```diff
@@ -1,40 +1,75 @@
...
 
-Datadog automatically keeps track of what is running where, thanks to its Autodiscovery feature. Autodiscovery allows you to define configuration templates that will be applied automatically to monitor your containers.
+The Datadog Agent can automatically track which services are running where, thanks to its Autodiscovery feature. Autodiscovery lets you define configuration templates for Agent checks and specify which containers each check should apply to. The Agent enables, disables, and regenerates static check configurations from the templates as containers come and go. When your NGINX container moves from 10.0.0.6 to 10.0.0.17, Autodiscovery helps the Agent update its NGINX check configuration with the new IP address so it can keep collecting NGINX metrics without any action on your part.
 
-## How it works
+<div class="alert alert-info">
+Autodiscovery was previously called Service Discovery. It's still called Service Discovery in the Agent's code and in Autodiscovery configuration options.
+</div>
```
irabinovitch: Do we need to mention this?

content/guides/autodiscovery.md
```diff
@@ -1,40 +1,75 @@
...
+In a traditional non-container environment, Datadog Agent configuration is—like the environment in which it runs—static. The Agent reads check configurations from disk when it starts, and as long as it's running, it continuously runs every configured check. The configuration files are static, and any network-related options configured within them serve to identify specific instances of a monitored service (e.g. a redis instance at 10.0.0.61:6379). When an Agent check cannot connect to such a service, you'll be missing metrics until you troubleshoot the issue. The Agent check will retry its failed connection attempts until an administrator revives the monitored service or fixes the check's configuration.
 
-Configuration templates can be defined by simple template files or as single key-value stores using etcd or Consul.
+With Autodiscovery enabled, the Agent runs checks differently.
 
-## How to set it up
+### Different Configuration
```
irabinovitch: should this be plural? "Different Configurations"

content/guides/autodiscovery.md
xvello: I think so, because current client know and might search for the old term

content/guides/autodiscovery.md
xvello: As docker is the only supported runtime for now, I'd go with writing docker explicitly, but we'll transition to a generic "containers" once rkt is supported

content/guides/autodiscovery.md
xvello: We don't have agent install doc for swarm yet, we could [add it here](https://app.datadoghq.com/account/settings#agent/docker)

But the agent setup page need a revamp as the k8s and docker ones are getting more and more complex (different ways to deploy all mushed in the same page with no hierarchy). I'm torn between revamping these pages or just create doc.dd.com pages and link to them.

content/guides/autodiscovery.md
```diff
@@ -48,155 +83,253 @@ By default, the Datadog Agent includes Autodiscovery support for:
...
+1. You want to use options not used in the default template,
+1. the Apache containers run a different image than that specified in the default template (see the example below), OR
+1. the Apache containers expose more than one port (the default template doesn't use [template variable indexing](#template-variable-indexes)).
 
-## Configuration templates
+You can provide template files of your own in a few ways:
 
-The configuration templates in `conf.d/auto_conf` directory are nearly identical to the example YAML configuration files provided in [the Datadog `conf.d` directory](https://github.com/DataDog/dd-agent/tree/master/conf.d), but with one important field added. The `docker_images` field is required and identifies the container image(s) to which the configuration template should be applied.
+1. Add them to each host that runs docker-dd-agent and [mount the directory that contains them](https://github.com/DataDog/docker-dd-agent#configuration-files) into the docker-dd-agent container when starting it
+1. Package them into your own release of docker-dd-agent
```
xvello: I'd go with:

Build a custom docker image based on docker-dd-agent with your custom templates added in the `/etc/dd-agent/conf.d/auto_conf`

content/guides/autodiscovery.md
```diff
@@ -48,155 +83,253 @@ By default, the Datadog Agent includes Autodiscovery support for:
...
 
-As an example if the rabbitmq container mentioned above is available over two networks `bridge` and `swarm`, using `%%host_swarm%%` will pick the IP address from the swarm network.
-Note that for the `host` variable if several networks are found and no key is passed the agent attempts to use the default `bridge` network.
+_Any_ `httpd` image. Suppose you have one container running `library/httpd:latest` and another running `yourusername/httpd:v2`. Autodiscovery will apply the above template to both containers. When it's loading auto-conf files, Autodiscovery cannot distinguish between identically-named images from different sources or with different tags, and **you must provide short names for container images**, e.g. `httpd`, NOT `library/httpd:latest`.
 
-## Configuration templates with key-value stores
+If this is too limiting—if you need to apply different check configurations to different containers running the same image—use [labels](#container-labels) to identify the containers. Label each container differently, then add each label to any template file's `docker_images` list (yes, `docker_images` is where to put _any_ kind of container identifier, not just images).
```
xvello: 👍

content/guides/autodiscovery.md
xvello: IIRC that's 5.14 only (must check the milestone). We could document that to avoid support cases with old versions

content/guides/autodiscovery.md
```diff
@@ -48,155 +83,253 @@ By default, the Datadog Agent includes Autodiscovery support for:
...
+
+You can identify containers by label rather than container name or image. Just label any container `com.datadoghq.sd.check.id: <SOME_LABEL>`, and then use `<SOME_LABEL>` anywhere you'd normally use a container name or image. For example, if you label a container `com.datadoghq.sd.check.id: special-container`, Autodiscovery will apply to that container any auto-conf template that contains `special-container` in its `docker_images` list.
+
+Autodiscovery can only identify each container by label OR image/name—not both—and labels take precedence. For a container that has a `com.datadoghq.sd.check.id: special-nginx` label and runs the `nginx` image, the Agent will NOT apply templates that include only `nginx` as a container identifier.
+
+### Template Source Precedence
+
+If you provide a template for the same check type via multiple template sources, the Agent will prefer, in increasing order of preference:
+
+* Files
```
xvello: IIRC files are last in the precedence order. Annotations are first for sure, so order is Annotations -> K/V -> files

I think it's better to list them in order of lookup instead of increasing order of preference (as that's the opposite, right?)

content/guides/autodiscovery.md
kmshultz: @xvello at first I did have it at the top, just below the page title. Do you think that's better?

content/guides/autodiscovery.md
kmshultz: unpredictably, dynamically... I think it's a matter of taste. To me, "unpredictably" better evokes the problem. Do you think it's inaccurate, or you just don't like the tone/sound of it?

kmshultz:
Re: the Agent install instructions for each platform, I agree we need a separate space for this. With the docs redesign underway soon, we should have a dedicated section for the Agent (i.e. docs.dd.com/agent) where we comprehensively cover its architecture (different daemons), install instructions for different platforms (similar to dogweb), all configuration options and environment variables, etc.

I don't like having the install docs embedded in Integrations pages.

content/guides/autodiscovery.md
kmshultz: Ok, maybe this is confusing because I put it in ascending order of precedence. I thought another engineer told me KV takes highest precedence but perhaps I misunderstood.

content/guides/autodiscovery.md
xvello: yes, as in "just so you know... now let's get it over with"

content/guides/autodiscovery.md
kmshultz: > As docker is the only supported runtime for now, I'd go with writing docker explicitly, but we'll transition to a generic "containers" once rkt is supported

Indeed, and if we avoided saying Docker here, we'd want to remove it from the page's title, too.

PR #1359, opened 2017-06-19
Charly/composite monitor warning disclaimer

CharlyF:


CharlyF:
The description is not clear enough on the existence of the warning state:
In the section titled 'How we compute trigger conditions', it states 'When a composite monitor evaluates as alert-worthy, it inherits the most severe status among its individual monitors and triggers an alert.', and then shows a table listing composite statuses of warning and alert (critical).
One can understand that the alert status should be that of the most severe status among its individual monitors (i.e. warning in the 1st example). But the doc suggests that is not the case (automatically overriding the warning with a critical).

kmshultz:
I don't understand the comment or what's written in the pull request.

From the comment: 

> then shows a table listing composite statuses of warning and alert (critical).

Nowhere does this page use the word 'critical', it says 'triggered'. The point is that the status remains 'Warning', and this status does trigger an alert.

I was hoping this was made clear in the following statement, which appears before the table:

> Composite monitors consider Unknown, Warn and Alert to be alert-worthy (i.e. true)

kmshultz:
@CharlyF looking at this again, I think you're making two suggestions:

1. Clarify that when using the not operator (!), the _only_ possible monitor statuses are OK, Alert, and No Data.
2. Clarify in the first truth table that the status Warning does NOT get overridden somehow to Alert.

First, the pull request only addresses 1. For 2, would it be clearer if we added a final column, **Triggered**, populated with a checkbox for each row that generates an alert? Maybe the parenthetical `(triggered!)` is a little confusing.

On item 1, maybe a clear example with its own table would be best. I am not sure what you have written clarifies it enough. I'll think more on this.

CharlyF:
Yes, that is right! 
I made a PR to have this tracked, but this could indeed be better worded.

Let me know if you need anything else from me.

kmshultz:
Got it. Thanks for pointing it out!

kmshultz:
Closing in favor of #1376 (just merged)

PR #1369, opened 2017-06-21
Jyee/postgres check

jyee:
Trello: https://trello.com/c/x62OQct1/1568-documentation-postgresql-agent-check
Related: https://github.com/DataDog/integrations-core/issues/344

### DIFF
--- a/content/integrations/postgresql.md
+++ b/content/integrations/postgresql.md
@@ -47,47 +47,47 @@ When it prompts for a password, enter the one used in the first command.
 
 ## Configuration Options
 
-* `username` (Optional)
-* `password` (Optional)
-* `dbname` (Optional) - Name of the database you want to monitor
-* `ssl` (Optional) - Defaults to False. Whether to use SSL to connect.
-* `tags` (Optional) - Dictionary of tags
-* `relations` (Optional) - Dictionary of per-relation (table) metrics that you want to track. Each relation generates 10 metrics + 10 metrics per index.
-
-  By default all schemas are included. To track relations from specific schemas only, use the following syntax:
+* `username` (Optional) - The user account used to collect metrics, set in the [Installation section above](#installation)
+* `password` (Optional) - The password for the user account.
+* `dbname` (Optional) - The name of the database you want to monitor.
+* `ssl` (Optional) - Defaults to False. Indicates whether to use an SSL connection.
+* `use_psycopg2` (Optional) - Defaults to False. Setting this option to `True` will force the Datadog Agent to collect PostgreSQL metrics using psycopg2 instead of pg8000. Note that pyscopg2 does not support SSL connections.
+* `tags` (Optional) - A list of tags applied to all metrics collected. Tags may be simple strings or key-value pairs.
+* `relations` (Optional) - By default, all schemas are included. Add specific schemas here to collect metrics for schema relations. Each relation will generate 10 metrics and an additional 10 metrics per index. Use the following structure to declare relations:
 
       relations:
-        - relation_name: another_table
+        - relation_name: my_relation
           schemas:
-            - public
-            - prod
+            - my_schema_1
+            - my_schema_2
   {:.language-yaml}
+
 * `collect_function_metrics` (Optional) - Collect metrics regarding PL/pgSQL functions from pg_stat_user_functions
-* `collect_count_metrics` (Optional) - Collect count metrics, default value is True for backward compatibility but they migth be slow, suggested value is False.
+* `collect_count_metrics` (Optional) - Collect count metrics. The default value is `True` for backward compatibility, but this might be slow. The recommended value is `False`.
 
 
 ## Custom metrics
 
 PostgreSQL custom metrics can be collected by mapping SQL query results to metrics in Datadog. This mapping is expressed using four components: descriptors, metrics, query, and relation.
 
-* **Descriptors** are a sequence of tuples matching a column name or alias from the custom metric query to a tag in Datadog. In the example below, `relname` is mapped to `table`. Custom metrics in Datadog can then be filtered and aggregated by `table` tag.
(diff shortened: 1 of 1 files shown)


content/integrations/postgresql.md
```diff
@@ -47,58 +47,79 @@ When it prompts for a password, enter the one used in the first command.
...
-* `collect_count_metrics` (Optional) - Collect count metrics, default value is True for backward compatibility but they migth be slow, suggested value is False.
+* `collect_count_metrics` (Optional) - Collect count metrics. The default value is `True` for backward compatibility, but this might be slow. The recommended value is `False`.
 
 
 ## Custom metrics
- 
-Below are some examples of commonly used metrics, which are implemented as custom metrics. Uncomment them if you want to use them as is, or use as an example for creating your own custom metrics. 
 
-The format for describing custome metrics is identical with the one used for common metrics in `postgres.py`. Be extra careful with ensuring proper custom metrics description format. If your custom metric does not work after an agent restart, look for errors in the output of `/etc/init.d/datadog-agent info` command, as well as `/var/log/datadog/collector.log` file.
+PostgreSQL custom metrics can be collected by mapping SQL query results to metrics in Datadog. This mapping is expressed using four components: descriptors, metrics, query, and relation.
```
kmshultz: The first sentence sounds strange to me, I think because of the word "mapping". Don't the SQL query results _become_ metrics? "Mapping to metrics" makes it sound like there are existing metrics to which the query results will be mapped (and in that case, what does "mapping" mean?)

Then the second sentence gives the mappings components, but I still don't know what it is for.

content/integrations/postgresql.md
```diff
@@ -47,58 +47,79 @@ When it prompts for a password, enter the one used in the first command.
...
+
+* **Descriptors** are a sequence of tuples matching a column name or alias from the custom metric query to a tag in Datadog. In the example below, `relname` is mapped to `schema`. Custom metrics in Datadog can then be filtered and aggregated by `schema` tag.
```
kmshultz: s/sequence/list

Descriptors **is**? Since this refers to an option name. Otherwise, "are" does agree with "a sequence".

Again, the 'matching' and 'mapping' words are throwing me. 'matching to a tag in Datadog': what tag in Datadog? What we are actually doing is creating a tag and naming it what we like.

How about:

**Descriptors** let you tag the custom metric with any of the query's column names or aliases. Each descriptor is a pair: the first element is the column name or alias, and the second element is the tag name. In the example below, the metric will be tagged: `schema:<relname>`.

content/integrations/postgresql.md
```diff
@@ -47,58 +47,79 @@ When it prompts for a password, enter the one used in the first command.
...
+* **Metrics** are mappings from a query column or column function to a tuple containing the Datadog custom metric name and metric type (`RATE`, `GAUGE`, or `MONOTONIC`). In the example below, the sum of the `idx_scan` column is mapped to the Datadog custom metric name `postgresql.idx_scan_count_by_schema`.
+* **Query** is where you'll construct the SQL query to obtain your custom metrics. Each column name in your select query should have a corresponding tuple in the Descriptors section. Each column listed in the Metrics section will be substituted for the `%s` in the query.
```
kmshultz: s/select/SELECT

> **should** have a corresponding tuple in the Descriptors section

is it recommended, or required (i.e. "must")? (Asking for myself, I actually don't know :))

kmshultz: > a corresponding tuple in the Descriptors section

a corresponding `descriptor`.

content/integrations/postgresql.md
```diff
@@ -47,58 +47,79 @@ When it prompts for a password, enter the one used in the first command.
...
+* **Relation** indicates whether to include schema relations specified in the [configuration options above](#configuration-options). If set to `true`, the last `%s` in the Query will be set to the list of schema names specified in the Relations configuration option.
```
kmshultz: > specified in the Relations configuration option

`relations` configuration option

content/integrations/postgresql.md
hush-hush: the first argument is actually ignored by the check (see: https://github.com/DataDog/integrations-core/blob/master/postgres/check.py#L600). I think we should not they the first item in the list is use for matching since only the order matter (first element in `descriptors` is matched with the first column and so on).

here what I wrote:
```
descriptors: is used to add tags to you metrics. It's a list of lists each containing 2 strings. The second
             one will be the tag name. The first is only here for documentation purposes (usually used to
             make clear what we are getting from the query). Note that instance tags will be added
             to that list.
```

content/integrations/postgresql.md
hush-hush: `column name` does not matter here, only the order. Also `Each column listed in the Metrics section will be substituted for the `%s` in the query.` make it sound people should use as many `%s` in the query as `Metrics`. All metrics are join together and added, so only one `%s` is needed.

content/integrations/postgresql.md
hush-hush: "the last `%s`"  the second actually (first being the use for `Metrics`)

content/integrations/postgresql.md
```diff
@@ -47,58 +47,79 @@ When it prompts for a password, enter the one used in the first command.
...
-    collect_count_metrics: False
-{:.language-yaml}
+      # All index scans & reads
+      - descriptors:
+          - [relname, schema]
+        metrics:
+            SUM(idx_scan) as idx_scan_count: [postgresql.idx_scan_count_by_schema, RATE]
+            SUM(idx_tup_read) as idx_read_count: [postgresql.idx_read_count_by_schema, RATE]
+        query: SELECT relname, %s FROM pg_stat_all_indexes GROUP BY relname;
+        relation: false
```
hush-hush: I think that displaying the final query made to postgres make the example much more understandable.

jyee:
Closing with new work in #1409

PR #1375, opened 2017-06-27
redirect broken tracing API links

irabinovitch:
/tracing-api/ seems to be in the google search results for some reason but 404s. The correct page is /tracing-api/.


jeremy-lq:
LGTM

PR #1376, opened 2017-06-27
Address some concerns from CharlyF

kmshultz:
@CharlyF does this address your concerns from #1359 adequately?

I nearly included a sentence like:

"A monitor using the not operator cannot take on a status of `Skipped`, `Ignored`, `Unknown`, or `Warn`."

But it really seems obvious (to me) from what's already written.

Let me know what you think.


kmshultz:
screen shot 2017-06-27 at 10 01 49 am

^ Here's a screenshot of the checkboxes. Does this make it clear that that the `Warn` status does not get overridden to `Alert`?

PR #1380, opened 2017-06-28
[WIP] update lambda docs for clarity and display new features

DanielLanger:
@garnermccloud mind looking at this before baklava does to see if I left anything out?

### DIFF
--- a/content/integrations/awslambda.md
+++ b/content/integrations/awslambda.md
@@ -73,7 +73,7 @@ print('MONITORING|{0}|{1}|{2}|{3}|#{4}'.format(
 unix_epoch_timestamp = int(time.time())
 metric_type = 'histogram'
 metric_name = 'my.metric.name.hist'
-tags = ['tag1:value', 'tag2']2
+tags = ['tag1:value', 'tag2']
 
 for i in xrange(0,10):
 	print('MONITORING|{0}|{1}|{2}|{3}|#{4}'.format(
--- a/content/integrations/awslambda.md
+++ b/content/integrations/awslambda.md
@@ -81,7 +81,7 @@ for i in xrange(0,10):
 ))
 ~~~
 
-Note: Using the histogram metric type provides `avg, count, max, min 95p, and median` values
+Note: Using the histogram metric type provides `avg, count, max, min, 95p, and median` values
 
 ###Service Check
 
--- a/content/integrations/awslambda.md
+++ b/content/integrations/awslambda.md
@@ -97,4 +97,6 @@ print('MONITORING|{0}|{1}|{2}|{3}'.format(
 
 <%= get_metrics_from_git()%> 
 
-Each of the metrics retrieved from AWS will be assigned the same tags that appear in the AWS console, including but not limited to function name, security-groups, and more.
+The metrics above get tagged in Datadog with any tags from AWS, including (but not limited to) function name, security-groups, and more.
+
+Custom metrics only get tagged with function name.
--- a/content/integrations/awslambda.md
+++ b/content/integrations/awslambda.md
@@ -14,41 +14,39 @@ Enable this integration to begin collecting Cloudwatch & custom metrics from you
 
 # Installation
 
-If you haven't already, set up the [Amazon Web Services integration first](/integrations/aws).
+If you haven't already, set up the [Amazon Web Services integration first](/integrations/aws){:target="_blank"}.
 
 # Configuration
 
 In the Amazon Web Services integration tile, ensure that Lambda is checked under metric collection.
 
+The following permissions are also required to fully use the Lambda integration:
+
+* `logs:DescribeLogGroups`:  List available log groups.
+* `logs:DescribeLogStreams`: List available log streams for a group.
+* `logs:FilterLogEvents`:  Fetch specific log events for a stream to generate metrics.
+* `tag:getResources`: Used to get custom tags applied to Lambda functions.
+
 To send custom metrics to Datadog, you must print a log line from your Lambda, using the following format:
 
 ~~~
-MONITORING|unix_epoch_timestamp|value|metric_type|metric.name|#tag1:value,tag2
+MONITORING|<unix_epoch_timestamp>|<value>|<metric_type>|<metric_name>|#<tag_list>
 ~~~
 
-## Notes on each section
+Where: 
 
-### MONITORING
-* This is used to find the log statement within Cloudwatch Logs and pull it into Datadog
+* `MONITORING` signals to the Datadog integration that it should collect this log entry
 
-### unix_epoch_timestamp
-* Please ensure the `unix_epoch_timestamp` is in seconds (not milliseconds)
+* `<unix_epoch_timestamp>` is in seconds, not milliseconds
 
-### value
-* The value of the metric **must** be a number.
-* For metric_type `check` the options are, `'0': OK, '1': WARNING, '2': CRITICAL, '3': UNKNOWN`
 
-### metric_type
-* Supported metric types are `count, gauge, histogram, and check`
+* `<value>` MUST be a number (i.e. integer or float)
 
-### metric.name
-* A unique name to identify your metric
+* `<metric_type>` is `count`, `gauge`, `histogram`, or `check`
 
-### tags
-* Optionally add tags to apply to this metric
-* If you do not want to add any tags use the format: `MONITORING|unix_epoch_timestamp|value|metric_type|metric.name`
-* The tag `function_name` will automatically be applied to custom metrics
+* `<metric_name>` uniquely identifies your metric and adheres to the [metric naming policy](http://docs.datadoghq.com/faq/#api){:target="_blank"}
 
+* `<tag_list>` is optional, comma separated, and must be preceded by `#`. NOTE, The tag `function_name:<name_of_the_function>` will automatically be applied to custom metrics
 
 ## Sample snippets (in Python):
 
@@ -58,7 +56,6 @@ MONITORING|unix_epoch_timestamp|value|metric_type|metric.name|#tag1:value,tag2
 unix_epoch_timestamp = int(time.time())
 value = 42
 metric_type = 'count'
-#metric_type = 'gauge'
 metric_name = 'my.metric.name'
 tags = ['tag1:value', 'tag2']
 
@@ -81,29 +78,19 @@ for i in xrange(0,10):
 ))
 ~~~
 
-Note: Using the histogram metric type provides `avg, count, max, min, 95p, and median` values. These values are calculated at one second granularity. 
+Note: Using the histogram metric type provides `avg`, `count`, `max`, `min`, `95p`, and `median` values. These values are calculated at one second granularity. 
 
 ### Service Check
 
 ~~~
 unix_epoch_timestamp = int(time.time())
-value = 1
+value = 1 # WARNING
 metric_type = 'check'
 metric_name = 'my.metric.name.check'
 
 print('MONITORING|{0}|{1}|{2}|{3}'.format(
 	timestamp, value, metric_type, metric_name
 ))
-~~~
-
-
-## Required Permssions
-
-~~~
-logs:DescribeLogGroups
-logs:DescribeLogStreams
-logs:FilterLogEvents
-tag:getResources
 ~~~
 
 # Metrics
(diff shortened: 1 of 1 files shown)


DanielLanger:
screen shot 2017-06-28 at 3 49 03 pm
screen shot 2017-06-28 at 3 48 58 pm

DanielLanger:
NOTE DO NOT MERGE YET

content/integrations/awslambda.md
```diff
@@ -21,27 +21,93 @@ If you haven't already, set up the [Amazon Web Services integration first](/inte
...
 
 To send custom metrics to Datadog, you must print a log line from your Lambda, using the following format:
-`MONITORING|unix_epoch_timestamp|value|metric_type|my.metric.name|#tag1:value,tag2`
 
-Please ensure the `unix_epoch_timestamp` is in seconds (not milliseconds).
+~~~
+MONITORING|unix_epoch_timestamp|value|metric_type|metric.name|#tag1:value,tag2
+~~~
+
+##Notes on each section
```
garnermccloud: need spaces between #'s and headers for markdown

content/integrations/awslambda.md
```diff
@@ -21,27 +21,93 @@ If you haven't already, set up the [Amazon Web Services integration first](/inte
...
+logs:FilterLogEvents
+tag:getResources
+~~~
 
 # Metrics
 
 <%= get_metrics_from_git()%> 
 
-Each of the metrics retrieved from AWS will be assigned the same tags that appear in the AWS console, including but not limited to host name, security-groups, and more.
+Each of the metrics retrieved from AWS will be assigned the same tags that appear in the AWS console, including but not limited to function name, security-groups, and more.
```
garnermccloud: We won't be adding the security-groups tags to the custom metric, just the function name.

garnermccloud:
Can you add a note about the histograms being calculated at the second granularity.

content/integrations/awslambda.md
DanielLanger: This is about the AWS Cloudwatch metrics

content/integrations/awslambda.md
```diff
@@ -21,27 +21,93 @@ If you haven't already, set up the [Amazon Web Services integration first](/inte
...
+MONITORING|unix_epoch_timestamp|value|metric_type|metric.name|#tag1:value,tag2
+~~~
+
+## Notes on each section
+
+### MONITORING
+* This is used to find the log statement within Cloudwatch Logs and pull it into Datadog
+
+### unix_epoch_timestamp
+* Please ensure the `unix_epoch_timestamp` is in seconds (not milliseconds)
```
kmshultz: No need to say 'Please'.

content/integrations/awslambda.md
```diff
@@ -21,27 +21,93 @@ If you haven't already, set up the [Amazon Web Services integration first](/inte
...
 
-Please ensure the `unix_epoch_timestamp` is in seconds (not milliseconds).
+~~~
+MONITORING|unix_epoch_timestamp|value|metric_type|metric.name|#tag1:value,tag2
+~~~
+
+## Notes on each section
+
+### MONITORING
+* This is used to find the log statement within Cloudwatch Logs and pull it into Datadog
```
kmshultz: This is passive voice. Better to say:

"The Datadog Lambda integration watches Cloudwatch logs and pulls in any log entries that begin with `MONITORING`"

content/integrations/awslambda.md
```diff
@@ -21,27 +21,93 @@ If you haven't already, set up the [Amazon Web Services integration first](/inte
...
+
+### unix_epoch_timestamp
+* Please ensure the `unix_epoch_timestamp` is in seconds (not milliseconds)
+
+### value
+* The value of the metric **must** be a number.
```
kmshultz: The metric value MUST be a number (i.e. integer or float)

content/integrations/awslambda.md
```diff
@@ -21,27 +21,93 @@ If you haven't already, set up the [Amazon Web Services integration first](/inte
...
+* For metric_type `check` the options are, `'0': OK, '1': WARNING, '2': CRITICAL, '3': UNKNOWN`
```
kmshultz: This is a general fact about service checks; it's not specific to Lambda. Unless you've had customers specifically confused about this, I think we should omit this.

content/integrations/awslambda.md
```diff
@@ -21,27 +21,93 @@ If you haven't already, set up the [Amazon Web Services integration first](/inte
...
+
+### metric_type
+* Supported metric types are `count, gauge, histogram, and check`
```
kmshultz: The list of types are all in one code block (they shouldn't be)

This can be simple:

"metric_type: `count`, `gauge`, `histogram`, or `check`"

content/integrations/awslambda.md
```diff
@@ -21,27 +21,93 @@ If you haven't already, set up the [Amazon Web Services integration first](/inte
...
+
+### metric.name
+* A unique name to identify your metric
 
-For example, here is sample snippet for printing a valid custom metric, from your Lambda function (in Python):
+### tags
+* Optionally add tags to apply to this metric
+* If you do not want to add any tags use the format: `MONITORING|unix_epoch_timestamp|value|metric_type|metric.name`
```
kmshultz: I think this is already implied by "Optionally add tags"

content/integrations/awslambda.md
```diff
@@ -21,27 +21,93 @@ If you haven't already, set up the [Amazon Web Services integration first](/inte
...
+* The tag `function_name` will automatically be applied to custom metrics
```
kmshultz: is it `function_name`, or `function_name:<name_of_the_function>`?

content/integrations/awslambda.md
```diff
@@ -21,27 +21,93 @@ If you haven't already, set up the [Amazon Web Services integration first](/inte
...
+
+## Sample snippets (in Python):
+
+### Count/Gauge
 
 ~~~
 unix_epoch_timestamp = int(time.time())
 value = 42
 metric_type = 'count'
+#metric_type = 'gauge'
```
kmshultz: Is this commented-out line meant to show that the `print` on line 65 should have the same format for counts and gauges?

content/integrations/awslambda.md
```diff
@@ -21,27 +21,93 @@ If you haven't already, set up the [Amazon Web Services integration first](/inte
...
+))
+~~~
+
+Note: Using the histogram metric type provides `avg, count, max, min, 95p, and median` values. These values are calculated at one second granularity. 
+
+### Service Check
+
+~~~
+unix_epoch_timestamp = int(time.time())
+value = 1
```
kmshultz: maybe put:

`value = 1 # WARNING`

content/integrations/awslambda.md
```diff
@@ -21,27 +21,93 @@ If you haven't already, set up the [Amazon Web Services integration first](/inte
...
 
 To send custom metrics to Datadog, you must print a log line from your Lambda, using the following format:
-`MONITORING|unix_epoch_timestamp|value|metric_type|my.metric.name|#tag1:value,tag2`
 
-Please ensure the `unix_epoch_timestamp` is in seconds (not milliseconds).
+~~~
+MONITORING|unix_epoch_timestamp|value|metric_type|metric.name|#tag1:value,tag2
+~~~
+
+## Notes on each section
```
kmshultz: H2 and H3 headers for this section are a bit heavy. We can just turn this whole thing into one list, like the [DogStatsD guide does](http://docs.datadoghq.com/guides/dogstatsd/#metricformat), i.e.

```
MONITORING|unix_epoch_timestamp|value|metric_type|metric.name|#tag1:value,tag2
```

- `MONITORING` – The Datadog Lambda integration watches Cloudwatch logs and pulls in any log entries that begin with this string.
- `unix_epoch_timestamp` – the log timestamp in seconds (NOT milliseconds)

etc

content/integrations/awslambda.md
```diff
@@ -21,27 +21,93 @@ If you haven't already, set up the [Amazon Web Services integration first](/inte
...
+metric_type = 'check'
+metric_name = 'my.metric.name.check'
+
+print('MONITORING|{0}|{1}|{2}|{3}'.format(
+	timestamp, value, metric_type, metric_name
+))
+~~~
+
+
+## Required Permssions
```
kmshultz: This feels a little lost beneath all the code examples. Can we put it near the beginning of **Configuration** section?

(After all, it's needed for non-custom metrics and custom metrics alike, yes?)

Also, a short sentence or two of explanation would be nice, rather than just a header and then a list.

content/integrations/awslambda.md
DanielLanger: Just wanted to be extra clear on this, in case they don't notice that in the service check example

content/integrations/awslambda.md
DanielLanger: yes, what are your thoughts on that?

kmshultz: I'm not sure it needs to be pointed out that gauges and counters will have similar looking code.

In fact, the `print` line looks the same for histograms and service checks. The only thing that's different for histogram is the looping over a range to show that a histogram is an aggregation of values.

content/integrations/awslambda.md
```diff
@@ -21,27 +21,93 @@ If you haven't already, set up the [Amazon Web Services integration first](/inte
...
+metric_name = 'my.metric.name.hist'
+tags = ['tag1:value', 'tag2']
+
+for i in xrange(0,10):
+	print('MONITORING|{0}|{1}|{2}|{3}|#{4}'.format(
+    	unix_epoch_timestamp, i, metric_type, metric_name, ','.join(tags)
+))
+~~~
+
+Note: Using the histogram metric type provides `avg, count, max, min, 95p, and median` values. These values are calculated at one second granularity.
```
kmshultz: Good that we're pointing out the one second granularity, but it will make more sense if we give an example. The current example shows that ten log lines will be printed, but they all use the same timestamp, so it's trivial to understand that they will be aggregated together.

How about this (tell me if I'm wrong here, actually; this is my assumed understanding of this):

---

<code snippet>

The code above prints 10 log lines, each with a different metric value: 0 through 9. When the Datadog Lamba integration parses these log entries, it creates these metrics and values:

- my.metric.name.hist.avg: 4.5
- my.metric.name.hist.count: 10
- my.metric.name.hist.max: 9
- my.metric.name.hist.min: 0
- my.metric.name.hist.95p: ?
- my.metric.name.hist.median: 4.5

Each of the 10 log entries has the same timestamp, and Datadog aggregates them together because they all occurred _within the same second_.

If we sleep for a second and generate five more log entries with a different timestamp:

```
time.sleep()
unix_epoch_timestamp = int(time.time())

for i in xrange(0,5):
 	print('MONITORING|{0}|{1}|{2}|{3}|#{4}'.format(
     	unix_epoch_timestamp, i, metric_type, metric_name, ','.join(tags)
  ))
```

Then the histogram's values change accordingly (i.e. `count` is now 5, `max` is 4, etc).

---

@garnermccloud is this correct?

content/integrations/awslambda.md
kmshultz: Ok, we can keep it.

content/integrations/awslambda.md
DanielLanger: @garnermccloud correct me if I'm wrong but this appears correct

DanielLanger: @kmshultz I also think this is kind of like the service check, that it's not lambda specific. Maybe we link to https://help.datadoghq.com/hc/en-us/articles/205638045-What-is-the-histogram-metric-type- but mention that this flush rate is 1 second and not 20

kmshultz: Good point! Works for me.

content/integrations/awslambda.md
kmshultz: @DanielLanger even still, can we clarify? i.e.

"The metrics above get tagged in Datadog with any tags from AWS, including (but not limited to) function name, security-groups, and more.

Custom metrics only get tagged with function name."

irabinovitch:
@DanielLanger Just FYI that we're mid migration to a new docs website.  With that in mind our plan is to freeze/close PRs on Friday.  We may ask you to re-open this on the new website if we're not merged by Friday.

DanielLanger:
@irabinovitch yup kent mentioned

PR #1395, opened 2017-07-11
adding k8s for process agent

mikezvi:


content/guides/process.md
```diff
@@ -92,6 +93,22 @@ Installing in a container will run the dd-process-agent and the standard dd-agen
...
+
+Installing with the daemonset will deploy the dd-process-agent and the standard dd-agent on your Kubernetes cluster. These will run together in a single pod so there's no need to run a dd-agent daemonset separately.
+
+Refer to the standard [daemonset installation](http://docs.datadoghq.com/integrations/kubernetes/#installation-via-daemonsets-kubernetes-110) and the [docker-dd-agent](https://github.com/DataDog/docker-dd-agent) information pages for further documentation.
+
+1. Create a manifest from the template at [dd-process-agent.yml](https://github.com/DataDog/dd-process-agent-install/blob/master/kubernetes/dd-process-agent.yml) with your API key.
+2. (optional) Modify any other settings in the yml file based on the standard [daemonset installation](http://docs.datadoghq.com/integrations/kubernetes/#installation-via-daemonsets-kubernetes-110).
+3. Create the daemonset and get the agent running in the cluster:
+
+    ```$ kubectl create -f dd-process-agent.yml```
```
mikezvi: I don't know if the back-ticks will render the same as four spaced code, but I could not exit the scope of the list without intermediate text.

conorbranagan: This looked nice for me (wrapping in code so it doesn't get formatted)

```
3. Create the daemonset and get the agent running in the cluster:

        $ kubectl create -f dd-process-agent.yml

   You can confirm the Agent pod is running with 

        kubectl get daemonsets
```

content/guides/process.md
```diff
@@ -92,6 +93,22 @@ Installing in a container will run the dd-process-agent and the standard dd-agen
...
+
+You can confirm the agent pod is running with
```
conorbranagan: we actually cap Agent throughout our and other docs so I think we want to continue that here (and maybe fix up if needed)

content/guides/process.md
mikezvi: Yep, 7 spaces did it.  I left the tabbing for the last instruction because it's a note that is relevant generally, not only in series with the other instructions.

PR #1403, opened 2017-07-13
more documentation around creating preview sites

mstbbs:
Just some more context around how things work. This could all be built out much more, but I have been focusing on not dropping any balls on this migration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,20 @@ coming soon
 
 # Working on Docs
 
+## Datadog Staff
+
+* Always branch off of master
+* Create a branch `slack_handle/your_feature` if you would like to create a preview site and run tests
+* Never commit directly to master
+* When you are ready to commit, create a new pull request to master from your branch
+
+## Outside Contributors
+
+* Fork the master branch
+* When you are ready to commit make a pull request back to `DataDog/master`
+
+## A note about markdown
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,6 +63,20 @@ coming soon
 
 # Working on Docs
 
+## Datadog Staff
+
+* Always branch off of master
+* Create a branch `slack_handle/your_feature` if you would like to create a preview site and run tests
```
technovangelist: i would drop the 'if you would like....' - always create the branch. no ifs

README.md
```diff
@@ -63,6 +63,20 @@ coming soon
...
+* Never commit directly to master
```
kmshultz: This seems redundant with the first item. Maybe combine them both: "Always branch off of master; never commit directly to master"

README.md
kmshultz: I don't think the 'if' is meant to operate on 'Create'; it operates on 'you would like to create a preview site" 

Probably 'Name' is better: "**Name** your branch `slack_handle/your_feature` if you would like to..."

technovangelist: but i think its a good naming strategy and we should encourage it. It seems to be what most do anyway

README.md
technovangelist: yes

PR #1408, opened 2017-07-13
enabled external link checks;

mstbbs:
@irabinovitch - i have added a links.ignore file. you can add patterns that you want to be checked against found links. if the found link contains one of these patterns then it will be skipped.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -13,6 +13,7 @@ variables:
 # ================== copy scripts =============== #
 before_script:
   - source /usr/local/bin/helpers.sh  # source helpers so they are available in the container
+  - find local/etc/ -type f -exec cp {} /etc \;  # load configs
 
 # ================== templates ================== #
 .base_template: &base_template
@@ -114,7 +115,7 @@ test_live_images:
     URL: ${LIVE_DOMAIN}
   script:
     - pull_artifact_from_s3
-    - test_site_links "images" "${URL}" "True" "False" "True"
+    - test_site_links "images" "${URL}" "True" "True" "True"
   only:
     - master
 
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -13,6 +13,7 @@ variables:
 # ================== copy scripts =============== #
 before_script:
   - source /usr/local/bin/helpers.sh  # source helpers so they are available in the container
+  - find local/etc/ -type f -exec cp {} /etc \;  # load configs
```
mstbbs: load the links.ignore file into the expected location in the container.

.gitlab-ci.yml
```diff
@@ -114,7 +115,7 @@ test_live_images:
     URL: ${LIVE_DOMAIN}
   script:
     - pull_artifact_from_s3
-    - test_site_links "images" "${URL}" "True" "False" "True"
+    - test_site_links "images" "${URL}" "True" "True" "True"
```
mstbbs: These "True" args are: for check-all, check-external, verbose output. good to have on for the first little while

local/etc/links.ignore
```diff
@@ -0,0 +1,14 @@
...
+tel:
+irc:
+file:
+localhost:
+url.to.snpashot.com
+app.datadoghq.com
+app.datadog.com
+some-account.atlassian.net
+bitbucket.org/groupname/reponame
+alert.victorops.com/integrations/datadog/20140523/alert
```
mstbbs: the linkchecker will look skip any link that contains a pattern from this list

irabinovitch:
Thanks Michael. LGTM.

PR #1409, opened 2017-07-13
Update PostgreSQL integration doc

jyee:
Continuation of work from #1369 but against the new hugo docs.


content/integrations/postgresql.md
```diff
@@ -42,63 +40,87 @@ When it prompts for a password, enter the one used in the first command.
...
 * `collect_function_metrics` (Optional) - Collect metrics regarding PL/pgSQL functions from pg_stat_user_functions
-* `collect_count_metrics` (Optional) - Collect count metrics, default value is True for backward compatibility but they migth be slow, suggested value is False.
+* `collect_count_metrics` (Optional) - Collect count metrics. The default value is `True` for backward compatibility, but this might be slow. The recommended value is `False`.
+
+
+## Custom metrics
+
+Datadog will generate PostgreSQL metrics from custom query results. For each custom query, four components are required: `descriptors`, `metrics`, `query`, and `relation`.
+
+* `descriptors` is used to add tags to your custom metrics. It's a list of lists each containing 2 strings. The first string is for documentation purposes and usually used to make clear what we are getting from the query. The second string will be the tag name. Note that instance tags will automatically be added to that list.
```
kmshultz: > The first string is for documentation purposes and usually used to make clear what we are getting from the query.

So, what's it actually used for? i.e. what does the Agent do with it?

content/integrations/postgresql.md
```diff
@@ -42,63 +40,87 @@ When it prompts for a password, enter the one used in the first command.
...
+            - my_schema_2
 
 * `collect_function_metrics` (Optional) - Collect metrics regarding PL/pgSQL functions from pg_stat_user_functions
-* `collect_count_metrics` (Optional) - Collect count metrics, default value is True for backward compatibility but they migth be slow, suggested value is False.
+* `collect_count_metrics` (Optional) - Collect count metrics. The default value is `True` for backward compatibility, but this might be slow. The recommended value is `False`.
+
+
+## Custom metrics
+
+Datadog will generate PostgreSQL metrics from custom query results. For each custom query, four components are required: `descriptors`, `metrics`, `query`, and `relation`.
```
kmshultz: > Datadog will generate

What is 'Datadog' in this case? When I read that, normally I take it to mean the web app or backend.

I guess all the action of receiving and storing metrics technically does happen on the backend, but in docs for Agent checks, I normally refer to 'the Agent' as the actor. "The Agent submits", "The Agent emits", "The Agent collects", etc.

content/integrations/postgresql.md
```diff
@@ -42,63 +40,87 @@ When it prompts for a password, enter the one used in the first command.
...
+
+* `descriptors` is used to add tags to your custom metrics. It's a list of lists each containing 2 strings. The first string is for documentation purposes and usually used to make clear what we are getting from the query. The second string will be the tag name. Note that instance tags will automatically be added to that list.
+* `metrics` are key-value pairs where the key is the query column iname or column function and the value is a tuple containing the Datadog custom metric name and metric type (`RATE`, `GAUGE`, or `MONOTONIC`). In the example below, the results of the sum of the `idx_scan` column will appear in Datadog  with the metric name `postgresql.idx_scan_count_by_schema`.
+* `query` is where you'll construct the SQL query to obtain your custom metrics. Each column name in your SELECT query should have a corresponding item in the `descriptors` section. Each item listed in the Metrics section will be substituted for the first `%s` in the query.
```
kmshultz: > Each item listed in the Metrics section

nit: Each item listed in `metrics`

content/integrations/postgresql.md
kmshultz: Since the items in `metrics` and `descriptors` reference things that appear in `query`, would it be best to start the list with `query` and _then_ describe `metrics` and `descriptors`?

content/integrations/postgresql.md
```diff
@@ -42,63 +40,87 @@ When it prompts for a password, enter the one used in the first command.
...
+* `relation` indicates whether to include schema relations specified in the [`relations configuration option`](#configuration-options). If set to `true`, the second `%s` in the Query will be set to the list of schema names specified in the Relations configuration option.
```
kmshultz: > the second `%s` in the Query

nit: the second `%s` in `query`

content/integrations/postgresql.md
```diff
@@ -42,63 +40,87 @@ When it prompts for a password, enter the one used in the first command.
...
+          - [relname, schema]
+        metrics:
+            SUM(idx_scan) as idx_scan_count: [postgresql.idx_scan_count_by_schema, RATE]
+            SUM(idx_tup_read) as idx_read_count: [postgresql.idx_read_count_by_schema, RATE]
+        query: SELECT relname, %s FROM pg_stat_all_indexes GROUP BY relname;
+        relation: false
+
+The example above will run two queries in PostgreSQL:
+
+* `SELECT relname, SUM(idx_scan) as idx_scan_count FROM pg_stat_all_indexes GROUP BY relname;` which will generate the metric `postgresql.idx_scan_count_by_schema`.
```
kmshultz: worth mentioning its tag?

content/integrations/postgresql.md
```diff
@@ -42,63 +40,87 @@ When it prompts for a password, enter the one used in the first command.
...
+* `SELECT relname, SUM(idx_tup_read) as idx_read_count FROM pg_stat_all_indexes GROUP BY relname;` which will generate the metric `postgresql.idx_read_count_by_schema`.
 
-### Custom metrics
+Both metrics will use the tag `schema` with tag values from the results in the `relname` column.
```
kmshultz: for example? e.g. `schema:<relname`

Since we allow value and value-less tags, it's best to spell out exactly what an example tags looks like. I recall a few support tickets where customers got confused about tag formats and found our documentation on it lacking.

content/integrations/postgresql.md
```diff
@@ -42,63 +40,87 @@ When it prompts for a password, enter the one used in the first command.
...
 
-Below are some examples of commonly used metrics, which are implemented as custom metrics. Uncomment them if you want to use them as is, or use as an example for creating your own custom metrics.
+### Example 2
 
-The format for describing custome metrics is identical with the one used for common metrics in `postgres.py`. Be extra careful with ensuring proper custom metrics description format. If your custom metric does not work after an agent restart, look for errors in the output of `/etc/init.d/datadog-agent info` command, as well as `/var/log/datadog/collector.log` file.
+Included in the `postgres.yaml.example` file is an example for the SkyTools 3 Londoniste replication tool:
```
kmshultz: > Included in the `postgres.yaml.example` file is

nit: The `postgres.yaml.example` file includes

content/integrations/postgresql.md
```diff
@@ -42,63 +40,87 @@ When it prompts for a password, enter the one used in the first command.
...
+* `tags` (Optional) - A list of tags applied to all metrics collected. Tags may be simple strings or key-value pairs.
+* `relations` (Optional) - By default, all schemas are included. Add specific schemas here to collect metrics for schema relations. Each relation will generate 10 metrics and an additional 10 metrics per index. Use the following structure to declare relations:
 
       relations:
-        - relation_name: another_table
+        - relation_name: my_relation
           schemas:
-            - public
-            - prod
+            - my_schema_1
```
kmshultz: I tend to prefer this format for non-literal values:

<MY_SCHEMA_1>
<MY_SCHEMA_2>

Been meaning to put that in the style guide (if others concur).

content/integrations/postgresql.md
kmshultz: > Note that instance tags will automatically be added to that list

What list?

And isn't this true of nearly all metrics submitting with nearly all Agent checks? (i.e. maybe not worth mentioning)

content/integrations/postgresql.md
```diff
@@ -42,63 +40,87 @@ When it prompts for a password, enter the one used in the first command.
...
+* `collect_count_metrics` (Optional) - Collect count metrics. The default value is `True` for backward compatibility, but this might be slow. The recommended value is `False`.
+
+
+## Custom metrics
+
+Datadog will generate PostgreSQL metrics from custom query results. For each custom query, four components are required: `descriptors`, `metrics`, `query`, and `relation`.
+
+* `descriptors` is used to add tags to your custom metrics. It's a list of lists each containing 2 strings. The first string is for documentation purposes and usually used to make clear what we are getting from the query. The second string will be the tag name. Note that instance tags will automatically be added to that list.
+* `metrics` are key-value pairs where the key is the query column iname or column function and the value is a tuple containing the Datadog custom metric name and metric type (`RATE`, `GAUGE`, or `MONOTONIC`). In the example below, the results of the sum of the `idx_scan` column will appear in Datadog  with the metric name `postgresql.idx_scan_count_by_schema`.
```
kmshultz: > Datadog custom metric name

omit 'Datadog'? This name is what we the user are choosing, yet 'Datadog custom metric name' makes it sound somehow like some reserved Datadog-custom thing (whatever that would mean). I suppose the intent here was to say 'this will be the name of your metric when you look for it in the Datadog app'? If so, I think that's pretty obvious.

content/integrations/postgresql.md
kmshultz: > which will generate the metric

which will generate a rate metric

content/integrations/postgresql.md
jyee: Apparently nothing according to @hush-hush. I adapted what he'd initially wrote. I'm wondering in this case how transparent we should be that the first string does nothing vs. telling people to match it to a query column or something else more descriptive. i.e. do we make this phrasing more opaque to promote better custom metrics?

content/integrations/postgresql.md
jyee: Normally I think that's fine, especially in places where it'd be something like <YOUR_DATADOG_API_KEY>. In this case, basically all the values would be that format, so it's likely not useful.

content/integrations/postgresql.md
jyee: @hush-hush any thoughts on this? This is a part I pulled from your notes.

jyee: @kmshultz are you thinking just re-ordering these explanations or also for the examples? This is the order that we use for the postgres core integration, so I'd prefer to guide people to use the same order that we use.

jyee:
@kmshultz Updated per our discussion this morning.

content/integrations/postgresql.md
```diff
@@ -42,63 +40,87 @@ When it prompts for a password, enter the one used in the first command.
...
 * `collect_function_metrics` (Optional) - Collect metrics regarding PL/pgSQL functions from pg_stat_user_functions
-* `collect_count_metrics` (Optional) - Collect count metrics, default value is True for backward compatibility but they migth be slow, suggested value is False.
+* `collect_count_metrics` (Optional) - Collect count metrics. The default value is `True` for backward compatibility, but this might be slow. The recommended value is `False`.
+
+
+## Custom metrics
+
+The Agent generates PostgreSQL metrics from custom query results. For each custom query, four components are required: `descriptors`, `metrics`, `query`, and `relation`.
+
+* `query` is where you'll construct the SQL SELECT query to obtain your custom metrics. Each column name in your SELECT query should have a corresponding item in the `descriptors` section. Each item in `metrics` will be substituted for the first `%s` in the query.
```
kmshultz: nits: omit SQL; s/query/statement:

"query is where you'll construct the SELECT statement to generate your custom metrics"

content/integrations/postgresql.md
```diff
@@ -42,63 +40,87 @@ When it prompts for a password, enter the one used in the first command.
...
+* `metrics` are key-value pairs where the key is the query column name or column function and the value is a tuple containing the custom metric name and metric type (`RATE`, `GAUGE`, or `MONOTONIC`). In the example below, the results of the sum of the `idx_scan` column will appear in Datadog  with the metric name `postgresql.idx_scan_count_by_schema`.
```
kmshultz: suggestion:

`metrics` are key-value pairs: keys are column names or functions whose values you want to collect; values prescribe the metrics' names and types.

content/integrations/postgresql.md
```diff
@@ -42,63 +40,87 @@ When it prompts for a password, enter the one used in the first command.
...
+* `descriptors` is used to add tags to your custom metrics. It's a list of lists each containing 2 strings. The first string is for documentation purposes and should be used to make clear what you are getting from the query. The second string will be the tag name.
```
kmshultz: s/is/are? The previous bullet begins "`metrics` are", so we should be consistent about singular-ness/plural-ness of option names.

kmshultz: s/2/two.

content/integrations/postgresql.md
```diff
@@ -42,63 +40,87 @@ When it prompts for a password, enter the one used in the first command.
...
+            - my_schema_2
 
 * `collect_function_metrics` (Optional) - Collect metrics regarding PL/pgSQL functions from pg_stat_user_functions
-* `collect_count_metrics` (Optional) - Collect count metrics, default value is True for backward compatibility but they migth be slow, suggested value is False.
+* `collect_count_metrics` (Optional) - Collect count metrics. The default value is `True` for backward compatibility, but this might be slow. The recommended value is `False`.
+
+
+## Custom metrics
+
+The Agent generates PostgreSQL metrics from custom query results. For each custom query, four components are required: `descriptors`, `metrics`, `query`, and `relation`.
```
kmshultz: suggestion: "The Agent can generate custom metrics from any SELECT statements you write in `postgres.yaml`. For each custom query, you must provide four components:"

kmshultz:
@jyee I just looked over the latest commit and wrote an alternate version of the four-components section (feel free to pick from it, or ignore :))

***

The Agent generates one or more metrics from each custom query you provide. For each custom query, you must provide four components:

**`query`**: write a base SELECT statement to collect your custom metrics; the other three components let you fill out this base query and tag the metrics it generates.

In the SELECT clause, add any columns whose values you'd like to tag the metrics with (you'll add the tag names in `descriptors`). Add `%s` at the end of the clause (in `metrics`, you'll specify the columns to appear in its place)

If you configured the `relations` option and you'd like to use those relations in the query, add a second `%s` wherever you'd like them to appear (the Agent will substitute them in).

**`metrics`**: list any columns from the query results whose values you want to collect as metrics. Each metric is a key-value pair: the key is the column to collect; the value—a tuple—is the metric name and type.

The Agent will subtitute the keys in `metrics` into the first `%s` in `query`.

**`descriptors`**: provide a tag name corresponding to each column in `query`'s SELECT clause. Each descriptor is a pair: the second item is the tag name, and the first item—while once used to match tag name to column name—is no longer used for anything; put anything there you want.

Provide a descriptor for each column, and list them in the same order as the columns.

**`relation`**: specify whether or not (true or false) to substitue `relations` into `query`'s second `%s` (if you configured `relations` and added a second `%s`).

The examples below make this clearer.

***

PR #1439, opened 2017-07-21
Describe how to add tags manually to metrics obtained from JMX MBeans

sv3nd:
- the optional 'tags' entry was added to the configuration of jmxfetch in release 0.14 (see this PR: https://github.com/DataDog/jmxfetch/pull/117)
- this commit adds to the java integration how to use this new 'tags' option to append new tags manually to a metric
- the documented example is directly inspired from the unit test of the  feature: https://github.com/DataDog/jmxfetch/blob/21069333fa9c04df3a09e79211f36827935ddc2c/src/test/resources/jmx_additional_tags.yml

### DIFF
--- a/content/integrations/java.md
+++ b/content/integrations/java.md
@@ -54,6 +54,10 @@ Make sure you can open a [JMX remote connection](http://docs.oracle.com/javase/1
             conf:
               - include:
                   domain: my_domain
+                  tags:
+                      simple: $attr0
+                      raw_value: my_chosen_value
+                      multiple: $attr0-$attr1
                   bean:
                     - my_bean
                     - my_second_bean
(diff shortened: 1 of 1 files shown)


content/integrations/java.md
```diff
@@ -103,7 +103,7 @@ You can explicitly specify supplementary tags. For instance, assuming the follow
 
     mydomain:attr0=val0,attr1=val1
 
-It would create a metric called `mydomain` (or some variation depending on the attribute inside the bean) with tags: `attr0:val0, attr1:val1, domain:mydomain, simple:val0, raw_value:my_chosen_value, multiple:val0-val1`.
```
sv3nd: Thanks for the several corrections you added to the text. 

For this commit though, I believe those 3 tags should not be removed: they are part of the tags that would be added by the jmxfetcher in that case. Actually, they are the whole point of my pull request :) 

What happens is that since https://github.com/DataDog/jmxfetch/pull/117 got merged, we can add tags at the level  of each Mbean, as shown by the example lines I added to the example configuration: 

```
                  tags:
                         simple: $attr0
                         raw_value: my_chosen_value
                         multiple: $attr0-$attr1
```

And those tags get added to the tags of the Datadog metric: in our case they get translated into `simple:val0, raw_value:my_chosen_value, multiple:val0-val1`. I like this example since it also shows how `$attr0-$attr1` get replaced with values from the MBean. 

I tested that behaviour on my laptop with the release 0.15 of jmxfetcher and that's indeed what I observe.

l0k0ms: how I get it sorry... I thought it was a cp glitch on my side. 
@kmshultz is it ok for merging for you ?

sv3nd: Cool then. 
I added a commit to put the correct tags back.

l0k0ms:
Perfect, merging to master then. 

Thank you very much for this contribution :)

Best,

PR #1472, opened 2017-07-26
Adding missing "-" to example YAML

teeebs:
Looks like a "-" was dropped from this doc's example yaml at some point, added it back.


kmshultz:
Thanks @teeebs.

PR #1533, opened 2017-08-02
[GCP] updating gcp label collection

l0k0ms:


content/integrations/google_cloud_platform.md
```diff
@@ -11,7 +11,7 @@ Connect to Google Cloud Platform to:
 
 * See your Google Compute Engine hosts in the infrastructure overview
 * Import your Google Compute Engine host tags
-* Tag your Google Compute Engine hosts with additional compute-specific metadata (e.g. zone)
+* Tag your Google Compute Engine hosts with additional GCP labels (only `key:value` label pattern are collected and defined as tag)
```
irabinovitch: defined as a tag

content/integrations/google_cloud_platform.md
```diff
@@ -11,7 +11,7 @@ Connect to Google Cloud Platform to:
 
 * See your Google Compute Engine hosts in the infrastructure overview
 * Import your Google Compute Engine host tags
-* Tag your Google Compute Engine hosts with additional compute-specific metadata (e.g. zone)
+* Tag your Google Compute Engine hosts with additional GCP labels (only `key:value` label pattern are collected and defined as a tag)
```
kmshultz: > (only `key:value` label pattern are collected and defined as a tag)

(only key-value labels are added as tags, i.e. to tag by zone, add a label `zone:us-west-1a`, not `us-west-1a`)

kmshultz: Although, isn't this kind of detailed for the Overview section? Where else can we put it?

l0k0ms: Zone is already collected as tag (cf end of the doc page) I'm going to find an other example.

We could do a KB/FAQ about it, what do you think ?

kmshultz: Ahh, fair enough, let's use another example. What about adding it to the end of **Tags Assigned** section? I don't think this needs to be a KB.

PR #1537, opened 2017-08-03
Adding info for chat name usage with hipchat API V2 token

l0k0ms:


content/integrations/hipchat.md
```diff
@@ -28,7 +28,10 @@ Tick the checkbox if you want to be notified for every comment, in all configure
 
 You can now share graphs or send alerts to HipChat rooms using the syntax @hipchat-chat_name
 
-**NOTICE** : If your chat name contains special characters such as commas or brackets, they'll be escaped when sending notifications with the @ handle. You shouldn't have to worry about that, the autocomplete box automatically takes care of escaping chat names.
+**NOTICE** :
```
kmshultz: Instead of **NOTICE**, let's wrap this in a warning block, e.g.

If you are using a hip chat API v1 token...
...
</div>

content/integrations/hipchat.md
```diff
@@ -28,7 +28,10 @@ Tick the checkbox if you want to be notified for every comment, in all configure
 
 You can now share graphs or send alerts to HipChat rooms using the syntax @hipchat-chat_name
 
-**NOTICE** : If your chat name contains special characters such as commas or brackets, they'll be escaped when sending notifications with the @ handle. You shouldn't have to worry about that, the autocomplete box automatically takes care of escaping chat names.
+**NOTICE** : 
+If you are using a hipchat API V1 token, if your chat name contains special characters such as commas or brackets, they'll be escaped when sending notifications with the @ handle. You shouldn't have to worry about that, the autocomplete box automatically takes care of escaping chat names.
```
kmshultz: > If you are using a hipchat API V1 token, if your chat name contains...

If you are using a hipchat API V1 token and your chat handle contains...

> they'll be escaped when sending notifications with the @ handle. You shouldn't have to worry about that, the autocomplete box automatically takes care of escaping chat names.

What is the motivation for adding this? Are using trying to add, e.g. "@kent\,shultz" in the text box? If so, can we add an example, e.g. "if you username is "kent,shultz", do not enter "@kent\,shultz".

Also, we can just say once that the Datadog takes care of the escaping, e.g.

"If you are using a hipchat API V1 token and your chat handle contains special characters like commas or brackets, don't escape them when you enter the handle; the Hipchat integration does that for you."

kmshultz: also, hipchat -> HipChat (everywhere in the document please)

content/integrations/hipchat.md
l0k0ms: 👍

content/integrations/hipchat.md
l0k0ms: It's when you want to send a notification from a monitor, you would do @my_hipchat_handle , but some HipChat handles can contain some special characters that we escape automatically in-app, to not confuse our users we display this notice here. 

But apparently this auto-escaping is only working with API V1, hence the update

content/integrations/hipchat.md
```diff
@@ -28,12 +28,16 @@ Tick the checkbox if you want to be notified for every comment, in all configure
 
 You can now share graphs or send alerts to HipChat rooms using the syntax @hipchat-chat_name
 
-**NOTICE** : If your chat name contains special characters such as commas or brackets, they'll be escaped when sending notifications with the @ handle. You shouldn't have to worry about that, the autocomplete box automatically takes care of escaping chat names.
+**Notice**: If you are using a HipChat API V1 token, if your HipChat handle contains special characters such as commas or brackets, they'll be escaped when sending notifications with the @ handle. You shouldn't have to worry about that, the autocomplete box automatically takes care of escaping chat names.
 
-#### Hipchat Server
+<div class="alert alert-warning">
```
kmshultz: I meant to put this starting before "If you are using a HipChat API V1 token" and ending after "special characters '-' and '_'. 

And removing "**Notice**:".

content/integrations/hipchat.md
```diff
@@ -28,12 +28,16 @@ Tick the checkbox if you want to be notified for every comment, in all configure
...
+If you want to use a HipChat API V2 token, then your HipChat handle may ONLY  have lower case alphanumerical characters and the special characters '-' and '_'.
```
kmshultz: there are two spaces between "ONLY" and "have"

kmshultz: nit: lower case -> lowercase

(either is ok but one-word _lowercase_ is common nowadays)

kmshultz: s/alphanumerical/alphanumeric

PR #1544, opened 2017-08-04
adding log file tailing info

l0k0ms:
cf trello card for more info

### DIFF
--- a/content/guides/logs.md
+++ b/content/guides/logs.md
@@ -67,10 +67,16 @@ In this format, the agent will attempt to import a function called
 `custom_parser` from `/path/to/my/parsers_module.py`.
 
 If your custom log parser is not working, the first place to check is the
-Agent collector logs. If the Agent is unable to import your function, there will
-be a line with `Could not load Dogstream line parser`. (On the other hand, if
-all goes well you should see `dogstream: parsing {filename} with
-{function name} (requested {config option text})`.)
+Agent collector logs: 
+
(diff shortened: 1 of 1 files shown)


content/guides/logs.md
```diff
@@ -67,10 +67,16 @@ In this format, the agent will attempt to import a function called
...
-{function name} (requested {config option text})`.)
+Agent collector logs: 
+
+* If the Agent is unable to import your function, there will be a line with `Could not load Dogstream line parser`. 
+
+* If all goes well you should see `dogstream: parsing {filename} with
+{function name} (requested {config option text})`.
+
+<div class="alert alert-warning">
+The Datadog agent is tailing your log files for log parsing, updating a single line in your log file won't trigger your Dogstreams, instead you should add your log to your file.
```
kmshultz: I don't understand what this means.

(also, s/agent/Agent)

l0k0ms: Some users test their dogstream just by updating a single log line in their log file, but this doesn't trigger the dogstream since our Agent tail the file. Some user complained that we didn't specify this behavior on our doc, hence this PR :)

kmshultz: Ok, got it. Then how about:

To test that dogstreams are working, append a line—don't edit an existing one—to any log file you've configured the Agent to watch. The Agent only _tails the end_ of each log file, so it won't notice any changes you make elsewhere in the file.

l0k0ms: 👍

PR #1554, opened 2017-08-07
Pingdom FAQ added

l0k0ms:
https://trello.com/c/CSuFgnVq/

### DIFF
--- a/content/integrations/pingdom.md
+++ b/content/integrations/pingdom.md
@@ -28,10 +28,10 @@ Pingdom events can be added by configuring the relevant [Integration Status Moni
 {{< get-metrics-from-git >}}
 
 
-## FAQ
-### Why updating user/password and saving the changes returns an error ?
-If you have the following error while trying to save your pingdom integration: 
+## Troubleshooting
+### Why am I getting an error when updating user/password?
+You may have seen the following when saving your pingdom credentials:
 
 `“There was an issue while testing your Pingdom configuration: Not permitted for account type”`.
 
-Add the email address of your pingdom account owner in the **(Optional) Account to query** field, then save your integration.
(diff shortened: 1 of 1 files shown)


content/integrations/pingdom.md
```diff
@@ -28,3 +28,10 @@ Pingdom events can be added by configuring the relevant [Integration Status Moni
 {{< get-metrics-from-git >}}
 
 
+## FAQ
+### Why updating user/password and saving the changes returns an error ?
+If you have the following error while trying to save your pingdom integration: 
+
+`“There was an issue while testing your Pingdom configuration: Not permitted for account type”`.
```
kmshultz: Since the error is short, I think we can just make it the title (i.e. header) of the FAQ.

content/integrations/pingdom.md
```diff
@@ -28,3 +28,10 @@ Pingdom events can be added by configuring the relevant [Integration Status Moni
...
+
+Add the email address of your pingdom account owner in the "(Optional) Account to query" field, then save your integration.
```
kmshultz: use bold, not quotes, for **(Optional) Account to query**

l0k0ms: 👍

content/integrations/pingdom.md
l0k0ms: I would rather let it this way so that people can search for the error or their use case. what do you think?

kmshultz: Ok, but let's improve the header, then:

### Why am I getting an error when updating user/password?"

You may have seen the following when saving your pingdom credentials:

`“There was an issue while testing your Pingdom configuration: Not permitted for account type”`.

Add the email address of your kingdom account owner in the **(Optional) Account to query** field, then save.

l0k0ms:
@kmshultz PR updated

PR #1571, opened 2017-08-09
[statsD] wording update

l0k0ms:
a customer had an issue with metric metadata not getting set on our end (we don't have the `timing` type on custom metrics) so he wasn't seeing units.   
He thought this should be the case because in Etsy's StatsD docs https://github.com/etsy/statsd/blob/master/docs/metric_types.md#timing it mentions certain protocol that we don't follow.

### DIFF
--- a/content/guides/dogstatsd.md
+++ b/content/guides/dogstatsd.md
@@ -32,7 +32,7 @@ sidebar:
 This page explains what DogStatsD is, how it works, and what data it accepts.
 </p>
 
-The easiest way to get your custom application metrics into Datadog is to send them to DogStatsD, a metrics aggregation service bundled with the Datadog Agent. DogStatsD implements the [StatsD](https://github.com/etsy/statsd) protocol and adds a few Datadog-specific extensions:
+The easiest way to get your custom application metrics into Datadog is to send them to DogStatsD, a metrics aggregation service bundled with the Datadog Agent. DogStatsD implements most of the [StatsD](https://github.com/etsy/statsd) protocol and adds a few Datadog-specific extensions:
 
 * Histogram metric type
 * Service checks and Events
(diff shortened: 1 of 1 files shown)


kmshultz:
Perhaps we should call out the main things it doesn't support?

l0k0ms:
@kmshultz I don't know if there are other features that we do not support.

kmshultz:
@hush-hush thoughts? 

While it's honest and accurate to say "we support most of StatsD", it isn't particularly helpful.

Of course we don't need to list _everything_ DogStatsD doesn't support. If StatsD is still under active development (not super active, it seems), it'd be hard to keep such a list up to date. But listing some major things—things our users often _think_ are supported—would be good.

irabinovitch:
I would prefer to have an exceptions section rather than saying we support "most"

hush-hush:
We already point that the timing/timer isn't supported by dogstatsd protocol but only by our libs:
https://github.com/DataDog/documentation/blob/master/content/guides/dogstatsd.md#now-start-the-timer

```
Under the hood, DogStatsD actually treats timers as histograms; Whether you send timer data using the methods above, or send it as a histogram (see below), you’ll be sending the same data to Datadog.
```

I also think a list of differences is better than `most of `. We may want to add a section for that or just point to the timer section.

l0k0ms:
@hush-hush would you have a list of differences available?

hush-hush:
@l0k0ms everything we have is in this documentation page and I think it's all of them (@olivielpeau do you know more on this ?).

I just think that instead of adding `most of` we could add `except Timer, see Timer section`.

l0k0ms:
@hush-hush nice 👍 , I'm waiting for @olivielpeau feedback on this before updating the PR.

kmshultz:
If the list of unsupported is super short, we should say it in the Intro.

PR #1576, opened 2017-08-10
updating h2,h3,h4,h5 titles design

l0k0ms:
In order to quick fix the titles layout we can push this until the complete doc redesign.
https://d2cx6t4ssmwdrv.cloudfront.net/gus/css-update/

screen_2
screen_3


technovangelist:
generally i like the idea, but i think hyun could give some recommendations on specific colors to use. he has done that for me in the past pretty quick.

l0k0ms:
ping @hyun309

hyun309:
We generally try to avoid using purple type unless it's used as an accent but I think it's OK here in the H2. I think the lighter purple in the H3 isn't ideal though. I would minimize the amount of colors and just make everything except H2 black or dark gray.

l0k0ms:
@hyun309 Good idea 👍  what do you think about this:
screen_1

screen_2

hyun309:
H1 and H2 is too close in color and size. I think it would be better if you just used purple on H1 only and keep H2 in black without the underline.

hyun309:
Ah, ic. In that case I would just match the two purples. The difference in the two shades of purple are too subtle and can be mistaken as unintentional.

l0k0ms:
@hyun309 H1 and H2 have now the same color: https://d2cx6t4ssmwdrv.cloudfront.net/gus/css-update/

What do you think? :)

hyun309:
Looks fine for now. Thanks!

l0k0ms:
Perfect, @kmshultz can you validate and merge this PR? :)

irabinovitch:
👍

PR #1586, opened 2017-08-14
Updates Apache + HTTP check examples to clarify what the HTTP check does

cmatteri:
On some systems Apache is named httpd. Indeed the container in this example is named httpd. Thus, a user may assume that the HTTP check monitors Apache and be confused why there are two checks. This commit clarifies what the HTTP check does.


l0k0ms:
👍  good for me @kmshultz what do you think ?

Issue #1593, opened 2017-08-16
IRC is still referenced


richupshall:
IRC is still referenced in this page https://docs.datadoghq.com/guides/agent_checks/ and this page https://docs.datadoghq.com/hostnames/

l0k0ms:
Hello @richupshall, 

Thanks for the feedback! Fix is on its way.

Best regards,
Pierre

PR #1596, opened 2017-08-17
Removing irc references

l0k0ms:
https://github.com/DataDog/documentation/issues/1593


l0k0ms:
@jeremy-lq we could add the link to the slack channel in the footer of the doc? So it's available on all doc page.

l0k0ms:
@jeremy-lq  Slack link is already in the help section. https://docs.datadoghq.com/help/#slack

LGTM ?

jeremy-lq:
Possible we should just link to the help section in that case, but I'll leave that decision to you. Approving so you can merge whichever you decide.

l0k0ms:
@jeremy-lq good idea 👍  linking to help page instead.

PR #1602, opened 2017-08-18
Desc automatic update

l0k0ms:
This will allow to inline the short description in the manifest.json file of each integration as the link preview

to call the description from the manifest just use: {{< get-desc-from-git >}}

(sorry for the amount of commits.. I tried to rebase but It messed up.. will try again next time)

There is some tests inside this code to avoid dogweb integrations, this will be removed once the logic will be iso between integration core and dogweb.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -24,6 +24,29 @@ before_script:
   only:
     - branches
 
+# ================== Local Test ================== #
+build_local:
+  stage: build
+  environment: "local"
+  variables:
+    BUCKET: ${PREVIEW_BUCKET}
+    URL: ${PREVIEW_DOMAIN}
+    CONFIG: ${PREVIEW_CONFIG}
+    MESSAGE: ":gift_heart: Your preview site is available!\nNow running tests..."
+  script:
+    - version_static_assets
+    - placehold_translations
+    - update_pre_build
+    - sync_integration_metrics
+    - build_hugo_site
+    - minify_html
+    - collect_static_assets
+    - push_site_to_s3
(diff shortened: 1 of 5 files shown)


.gitlab-ci.yml
```diff
@@ -39,6 +39,7 @@ build_preview:
     - post_dd_event "documentation deploy ${CI_COMMIT_REF_NAME} started" "${CI_PROJECT_URL}/pipelines/${CI_PIPELINE_ID}" "info"
     - version_static_assets
     - placehold_translations
+    - update_pre_build
```
mstbbs: let's name this something more explicit, perhaps `- sync_integration_descriptions`?

local/bin/sh/helpers.sh
```diff
@@ -155,6 +155,9 @@ sync_integration_metrics() {
 	integrations_sync.py --token ${GITHUB_TOKEN}
 }
 
+update_pre_build() {
+    update_pre_build.py --token ${GITHUB_TOKEN}
+}
```
mstbbs: let's decorate this with the `start_step`, `pass_step`, and `|| fail_step` functions. These will standardize the bash calls and report to datadog. see _most_ of the other helpers as examples.

.gitlab-ci.yml
l0k0ms: 👍

local/bin/sh/helpers.sh
l0k0ms: 👍

l0k0ms:
This is the first step before the readme scrapper. logic might change soon

PR #1603, opened 2017-08-18
Adding containers to Process guide

mikezvi:


content/guides/process.md
```diff
@@ -1,23 +1,80 @@
...
 
-Checking “show summary graphs” at the top of the page will graph metrics for all processes that are shown in the table.
+**String Search**
 
-{{< img src="process/live_process_preview.png" >}}
+Processes and containers are by their nature extremely high cardinality objects.  Our fuzzy string search will match any facet of a process or container, giving you a view into exactly what you want.  Below is our Demo environment, filtered with the string `postgres /9.`.  Note that `/9.` has matched in the current working directory, and that `postgres` matches the command itself.
+
+{{< img src="process/postgres.png" >}}
+
+**Tagging**
```
mikezvi: Is there an exhaustive list of tags we're adding that might not be available elsewhere?

content/guides/process.md
```diff
@@ -1,23 +1,80 @@
...
+
+ECS Containers are tagged by `#task_name`, `#task_version`, and `#ecs_cluster`
+
+Kubernetes Containers are tagged by `#pod`, `#pod_ip`, `#service`, `#namespace`, `#cluster-name`, `#replica_set`, `#daemon_set`, and `#deployment`.
+
+**Filtering and Pivoting**
+
+Making sense of hundreds of thousands or millions of processes and containers can seem overwhelming!  Using tagging, described in the previous section, makes navigation easy.
+
+In the below, we have filtered down to a Kubernetes cluster of 9 nodes.  RSS and CPU utilization on containers is reported compared to the hard limits set on the containers, when they exist.  Here, we see that the containers in this cluster are way overprovisioned, and that we could use tighter limits and bin packing to acheive better utilization of resources.
```
mikezvi: Are metrics reported compared to hard or soft limits?

content/guides/process.md
```diff
@@ -1,23 +1,80 @@
...
-## Configure the Datadog Agent to collect Live Process data
+It would be easy to pivot by ECS `#task_name` and `#task_version` and understand changes to resource utilization between updates.
+
+{{< img src="process/tasksxversion.png" >}}
+
+Below, we have searched for ssh processes and pivoted by `#user` to understand who is logged into which hosts.
+
+{{< img src="process/sshusers.png" >}}
+
+Ok, so I guess that last one is less exciting after redaction!
```
mikezvi: too unprofessional or cute?

content/guides/process.md
```diff
@@ -1,23 +1,80 @@
...
+{{< img src="process/live_process_preview.png" >}}
 
-Each row in the table can be expanded for more contextual information about an individual process.
+## Searching, Filtering, and Pivoting
 
-Checking “show summary graphs” at the top of the page will graph metrics for all processes that are shown in the table.
+**String Search**
 
-{{< img src="process/live_process_preview.png" >}}
+Processes and containers are by their nature extremely high cardinality objects.  Our fuzzy string search will match any facet of a process or container, giving you a view into exactly what you want.  Below is our Demo environment, filtered with the string `postgres /9.`.  Note that `/9.` has matched in the current working directory, and that `postgres` matches the command itself.
```
conorbranagan: > search will match any facet of a process or container

This is not actually true. For processes we only search on cmdline and user, for containers it's just name and image (and soon id). in your example both parts are in the full cmdline.

content/guides/process.md
```diff
@@ -1,23 +1,80 @@
...
+
+*Processes* are tagged by `#user`
+
+*Containers* are tagged by `#container_image`
+
+Additionally, we include integrations with popular orchestrators, such as ECS and Kubernetes, which provide further container-level tags.  We also decorate each container with Docker, ECS, or Kubernetes icons so you can tell which are being orchestrated at a glance.
+
+ECS Containers are tagged by `#task_name`, `#task_version`, and `#ecs_cluster`
+
+Kubernetes Containers are tagged by `#pod`, `#pod_ip`, `#service`, `#namespace`, `#cluster-name`, `#replica_set`, `#daemon_set`, and `#deployment`.
```
conorbranagan: also `job` and `replication_controller` (although I think this is deprecated in k8s, maybe not worth mentioning)

conorbranagan: both `cluster-name` and `ecs_cluster` come from host tags - does it matter?

content/guides/process.md
conorbranagan: I'm not sure which the cgroup limits will map to, I'll need to check

content/guides/process.md
conorbranagan: I might be the wrong one to ask but I like it

content/guides/process.md
```diff
@@ -31,22 +88,20 @@ and adding the following line to the `[Main]` section
     
 After configuration is complete, restart the Agent.
 
-## Docker container
+**Docker container**
 
-**Update to the Datadog Agent image version 5.16.0 or above**
+Update to the Datadog Agent image version 5.16.0 or above:
```
conorbranagan: maybe for the next pass but we could call out 5.17 which will:

* have docker collection on by default
* won't need the HOST_PROC/HOST_SYS environment if you're running inside a container

5.17 should go on Wednesday

content/guides/process.md
```diff
@@ -1,23 +1,80 @@
...
+{{< img src="process/containerinspect.png" >}}
+
+## Real-time monitoring
+
+While actively working with the Process and Containers page, metrics are collected at 2s resolution.  This is very important for highly volatile metrics such as CPU.
+
+In the background, for historical context, metrics are collected at 10s resolution.  To disable realtime mode, edit your `datadog.conf` file as described in the installation section below and append:
+
+    [process.config]
+    realtime_interval: 10
```
conorbranagan: this flag might change in the next version unless it's used right now. we can keep backwards compatibility if needed. (this would be the last time we'd change any flag names like this)

content/guides/process.md
mikezvi: It's slightly inconsistent but I want to highlight so ECS feels more complete.

content/guides/process.md
mikezvi: Yeah, we'll add some things then and also release note the Tx/Rx issue.

PR #1604, opened 2017-08-21
Update content to make it iso with integrations core

l0k0ms:
Adding automatic desc + Adding/rename files to make it iso with integrations core repo: 
activemq_xml.md
cassandra_nodetool.md
couch.md
dns_check.md
docker_daemon.md
elastic.md
gearmand.md
go-metro.md
go_expvar.md
hdfs_datanode.md
hdfs_namenode.md
http_check.md
kafka_consumer.md
kube_dns.md
kubernetes_state.md
mcache.md
mongo.md
php_fpm.md
postgres.md
redisdb.md
ssh_check.md
system_core.md
tcp_check.md
twemproxy.md
windows_service.md
zk.md


content/integrations/activemq.md
```diff
@@ -105,7 +105,7 @@ else echo -e "\e[031mAgent is not running\e[0m"; fi
 ## Data Collected
 ### Metrics
 
-{{< get-metrics-from-git "activemq_xml" >}}
+{{< get-metrics-from-git >}}
```
mstbbs: activemq does not have metrics here: https://github.com/DataDog/integrations-core/blob/master/activemq/metadata.csv

this may cause the build to abort

mstbbs:
@l0k0ms - this is all looking good. We should decide on a different pattern for front-matter placeholders. The pattern you are using now `{{< get-desc-from-git >}}` is currently being used by Hugo to designate shortcodes, and so may be confusing to other developers. In the python script you are using this pattern for substitution. Maybe something like `<< desc-from-git >>` ?

content/integrations/activemq.md
l0k0ms: 👍  good catch

l0k0ms: The git integration title param is activemq_xml, so  it's pulling the right metadata.csv :)

l0k0ms:
@kmshultz yes :) It also fix lots of missing metrics from integration core repo

PR #1605, opened 2017-08-21
removing tracing enable

l0k0ms:
As of 5.13 its enabled by default and so step 2 on https://docs.datadoghq.com/tracing/#installing-the-agent could be misleading:

Enable the APM in the Datadog Agent configuration file

```
[Main]
# Enable the trace agent.
apm_enabled: true
```


jeremy-lq:
Do we want to completely remove it, or note that as of 5.13 it's enabled by default but can be disabled with `apm_enabled: no`?

PR #1609, opened 2017-08-22
Adding private disclaimer

mikezvi:


content/guides/process.md
```diff
@@ -5,6 +5,8 @@ listorder: 16
 beta: true
 ---
 
+As this feature is in private beta, please do not discuss in public forums.
```
conorbranagan: Should we tell folks to reach out if they stumble upon this and want access? Or do we not expect that to happen?

mikezvi: We should, good thinking.

PR #1618, opened 2017-08-25
Add note that lambda function has to be in same region as kms key

DanielLanger:


content/integrations/awsrds.md
```diff
@@ -70,7 +70,7 @@ The command output will include two parts: a ciphertext blob followed by the key
 
 ##### Create your Lambda function
 
-12. From the Lambda Management Console, create a new Lambda Function.
+12. From the Lambda Management Console, create a new Lambda Function. NOTE, your Lambda function must be in the same region as the KMS key you created.
```
kmshultz: Instead of "NOTE", I prefer either to bold the entire statement (if it's warning about a really common or dire mistake), or, less severely, to capitalize "must", i.e.:

Your Lambda function MUST be in the same region as the KMS key you created.

content/integrations/awsrds.md
```diff
@@ -70,7 +70,7 @@ The command output will include two parts: a ciphertext blob followed by the key
 
 ##### Create your Lambda function
 
-12. From the Lambda Management Console, create a new Lambda Function. NOTE, your Lambda function must be in the same region as the KMS key you created.
+12. From the Lambda Management Console, create a new Lambda Function. **Your Lambda function MUST be in the same region as the KMS key you created.**
```
kmshultz: Not sure we need both bold AND a MUST. Does this point need to be that severely made?

PR #1623, opened 2017-08-28
Add Environments to Navbar, add help page (unlinked)

awang:


config.yaml
```diff
@@ -135,29 +135,33 @@ languages:
           url: tracing/terminology/
           weight: 30
 
+        - identifier: customnav_tracingnav_environments
```
mstbbs: @awang - the identifier here is a reference to an i18n entry. You'll notice that it returns blank for your new menu item: https://d2cx6t4ssmwdrv.cloudfront.net/awang/apm_mac_windows/tracing/environments/
-
add an entry for the translation here: https://github.com/DataDog/documentation/blob/master/i18n/en.yaml#L184

(no need to add a Japanese entry as the rest of the menu is not translated and Hugo will fall back to english, _unless_ you want to add for consistency)

awang: Thanks!  Does that work?  https://github.com/DataDog/documentation/pull/1623/commits/7658ecd0534fb470a90761bd33f0475aba85704a

mstbbs: Looks like it worked for the environments ( i see the link now ). Did you also want a link to the new help page?

PR #1642, opened 2017-09-18
fixing broken link

l0k0ms:
Since we split mesos article we need to update the link


l0k0ms:
I'm merging this in order to avoid broken links ticket creation

PR #1648, opened 2017-09-18
Sync integrations core repo

l0k0ms:
Almost all contents have been ported to the integration core repo:
https://github.com/DataDog/integrations-core/pull/764

We need to wait for this PR to be merged before merging this.


l0k0ms:
The build fails because we do not handle the case were metadata.csv is empty.
We need to update the integrations core repo to specify that no metrics are collected

irabinovitch:
Fixed a merge conflict.

PR #1656, opened 2017-09-20
updating json configuration

l0k0ms:
Customer pointed out two changes that need to be made to docs:

1) Wrong type of dash used
From https://docs.datadoghq.com/integrations/ecs/:

The doc site has the wrong dash at the beginning of the command line, so it's not just a copy/past - however the dashes in the middle of the command line flag are the right type:

"aws ecs register-task-definition –cli-input-json" should be aws ecs register-task-definition --cli-input-json

https://cl.ly/3S311M3u2P41

Though I think the longer dash is the "--"...

2) Need to make this consistent:

The memory value in the JSON (128) file doesn't match up with the by-hand configuration instructions (256) - I got an OOM on the datadog agent container before updating this value.

In dd-agent-ecs.json:
https://cl.ly/3m1H0Y412t1N

From https://docs.datadoghq.com/integrations/ecs/:
https://cl.ly/3O1E0f0Y1l3F


l0k0ms:
adding a card for this.

PR #1659, opened 2017-09-21
Correct mistake where text doesnt match image

Tishwa:
The images shows `role:cassandra` and the text says `role:sobotka`. Corrected it so that the text would say `role:cassandra` too.


irabinovitch:
Thanks @Tishwa

PR #1665, opened 2017-09-26
rename into system CPU core

l0k0ms:


irabinovitch:
@l0k0ms Do we need a redirect here?

l0k0ms:
@irabinovitch nop already handled :)

PR #1704, opened 2017-10-04
load order fixes

mstbbs:
fix clean-all bugs;
have hugo render from mem


mstbbs:
prior to merging, checkout this branch and run:
`make clean-all; make start`
if it works, merge away. if not, let me know

PR #1707, opened 2017-10-05
removing V2 reference + ja version

l0k0ms:


content/integrations/hipchat.md
```diff
@@ -30,8 +30,7 @@ Tick the checkbox if you want to be notified for every comment, in all configure
 You also share graphs or send Monitor alerts to HipChat rooms using `@hipchat-<chat_name>`.
 
 <div class="alert alert-warning">
-"If you are using a HipChat API V1 token and your chat handle contains special characters like commas or brackets, you don't need to escape them when you enter the handle; the autocomplete box does that for you.<br><br>
-If you are using a HipChat API V2 token, your chat handle may ONLY include lowercase alphanumeric characters and the special characters '-' and '_'.
```
jeremy-lq: If we don't support the HipChat API V2, we may want to mention so in the docs, to avoid unnecessary support tickets.

PR #1729, opened 2017-10-16
only run algolia search on search page.

mstbbs:
if nothing submitted, then run "*" blank query

### DIFF
--- a/src/js/datadog-docs.js
+++ b/src/js/datadog-docs.js
@@ -171,57 +171,60 @@ $(document).ready(function () {
     headings.init();
   }
 
-  var client = algoliasearch("EOIG7V0A2O", 'bf60de88836cb62a73509ef075542065');
-  var results = new RegExp('[\?&]' + "s" + '=([^&#]*)').exec(window.location.href);
-  var $pagination = $('#tipue_search_content');
-  var query = results[1] || "*";
+  // algolia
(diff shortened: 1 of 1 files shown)


src/js/datadog-docs.js
```diff
@@ -171,57 +171,60 @@ $(document).ready(function () {
...
-        num_page_links_to_display : 9,
-        items_per_page : 7,
-        wrap_around: false,
-        show_first_last: false
-      });
-
-    }
-  });
+  // algolia
+  if (window.location.href.indexOf('/search/') > -1) {
```
mstbbs: only execute this code on the search page.

src/js/datadog-docs.js
```diff
@@ -171,57 +171,60 @@ $(document).ready(function () {
...
+
+      var client = algoliasearch("EOIG7V0A2O", 'bf60de88836cb62a73509ef075542065');
+      var results = new RegExp('[\?&]' + "s" + '=([^&#]*)').exec(window.location.href);
+      var $pagination = $('#tipue_search_content');
+      var query = "";
+      try {query = results[1];} catch (e) {}
```
mstbbs: catch errors where the result from algolia is not valid (200)

src/js/datadog-docs.js
l0k0ms: 👍

PR #1742, opened 2017-10-23
Scraping integrations from dogweb

l0k0ms:
Preview: https://d2cx6t4ssmwdrv.cloudfront.net/gus/redesign-scrap-dogweb/integrations/amazon_api_gateway/


local/bin/py/update_pre_build.py
```diff
@@ -11,7 +11,7 @@
 import fileinput
 import json
 import re
-
+import os
```
mstbbs: not a blocker, but i notice we're importing the entire `os` module here, and importing large chunks of it above in two different statements. In the future it would make sense to either import the entire module and refactor the function calls to be namespaced, or continue to pull specific classes/functions out from a single call.

local/bin/py/update_pre_build.py
```diff
@@ -285,15 +283,30 @@ def sync(*args):
...
     
-    print('Updating integrations description...')
-    update_integration_pre_build(options.integrations, INTEGRATION_FOLDER)
+    if options.integrations:
+        print('Updating integrations description for {}...'.format(options.integrations))
+        update_integration_pre_build(options.integrations, INTEGRATION_FOLDER)
+
+    if options.dogweb:
+        print('Updating integrations description for {}...'.format(options.dogweb))
+        update_integration_pre_build(options.dogweb + 'integration' + sep, INTEGRATION_FOLDER)
```
mstbbs: I like this feedback 👍

local/bin/py/update_pre_build.py
l0k0ms: indeed, I just did this as a quick fix, but at some point I will refactor the script.. I find it too slow to run. and too dependent of the file name

PR #1751, opened 2017-10-26
Redesign code enhancment

l0k0ms:


mstbbs:
@l0k0ms do we need the transformation? why not just use the shortcode directly?

l0k0ms:
@mstbbs can we? I thought we could only use {{< highlitght >}} for code block display

PR #1759, opened 2017-10-27
skip beta pages in algolia object aggregation

mstbbs:
checks to see if the robots meta tag is present and skips if its

### DIFF
--- a/local/bin/py/index_algolia.py
+++ b/local/bin/py/index_algolia.py
@@ -55,46 +55,51 @@ def index_algolia(app_id, api_key, content_path=None):
 
                             html = BeautifulSoup(myfile, "html.parser")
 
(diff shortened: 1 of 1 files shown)


local/bin/py/index_algolia.py
```diff
@@ -55,46 +55,51 @@ def crawl_site(path, exclusions):
...
-                                    language = 'japanese'
-                                else:
-                                    language = 'english'
-
-                                article['language'] = language
-
-                                articles.append(article)
+                            # no crawl?
+                            private = html.find_all('meta', {'name': 'robots', 'content': 'noindex'})
+                            if private:
```
mstbbs: these lines are the only real change. everything else is nested, existing code

PR #1765, opened 2017-11-01
Update anomalies.md

mikhail:
Include information about Historical Context


githomin:
I'd rather fix the language than add an extra disclaimer.

mikhail:
I specifically wanted the word context and the pair "historical context" to be in the docs because the monitor page directs to this documentation:
image

Naturally I didn't read every word after opening that link - I searched the page for "context" and didn't find it.

githomin:
Yeah, good point. We'll update the tooltip above to be more accurate as well.

irabinovitch:
@mikhail Thanks you for raising this and for your detailed feedback. We have updated the documentation in #1766 and are working on updating the tooltip as well.  T If you are happy with Homin's changes in #1766 would it be ok to close this ticket?

Thanks again for your efforts.  The Datadog docs and our community are greatly improved by contributions such as yours.

PR #1770, opened 2017-11-02
add documentation explaining the sliding window

dd-caleb:
We received a support ticket:

> Customer has a composite monitor that stayed in alert state 5 minutes longer than necessary

This was actually behavior that we intended, so we decided to provide documentation explaining why it happens.

DataDog/yuzu

### DIFF
--- a/content/guides/composite_monitors.md
+++ b/content/guides/composite_monitors.md
@@ -8,7 +8,7 @@ listorder: 14
 If you're unfamiliar with the basics of Datadog Monitors, first read the <a href="https://docs.datadoghq.com/guides/monitors/">Guide to Monitors</a>
 </div>
 
-Composite monitors let you combine many individual monitors into one so that you can define more specific alert conditions. 
+Composite monitors let you combine many individual monitors into one so that you can define more specific alert conditions.
 You can choose up to 10 existing monitors—monitor A and monitor B, say—and then set a trigger condition using boolean operators (e.g. “A && B”). The composite monitor will trigger when its individual monitors' statuses simultaneously have values that cause the composite's trigger condition to be true.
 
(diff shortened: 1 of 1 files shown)


content/guides/composite_monitors.md
```diff
@@ -131,6 +131,21 @@ Consider a composite monitor that uses three individual monitors—A, B, and C
 
 Two of the four scenarios will trigger an alert, even though not all of the individual monitors have the most severe status, `Alert` (and in row 1, none do). But how _many_ alerts might you potentially receive from the composite monitor? That depends on the individual monitors' alert types.
 
+### How a component monitor's status is determined
```
jeremiefourbil: `composite` ?

content/guides/composite_monitors.md
```diff
@@ -131,6 +131,21 @@ Consider a composite monitor that uses three individual monitors—A, B, and C
...
+
+Rather than periodically sampling the current state of component monitors, composite monitors are evaluated by using a sliding window of monitor results for each component monitor. For example, if you have a composite monitor defined as `A && B`, and the component results look like this:
```
platinummonkey: same here I think

content/guides/composite_monitors.md
dd-caleb: I think `component` is correct? I'm explaining how their status is determined. Open to suggestions for different wording. Maybe:

    ### How a composite monitor's component monitors' status is determined

But that's kinda long-winded

jeremiefourbil: oh i see it makes sense. forget my comment :)

rachelwin:
Confirming that you mean the sliding window in the new status page?

dd-caleb:
Sorry I'm not familiar with the new status page. Does this terminology have a meaning in that context? Should I re-phrase to avoid confusion?

rachelwin:
No, we will have docs in time for GA release (currently alpha for org 2
only). The design is slightly changing - so will update once that has been
applied.

For general context, the evaluation window design is similar to the
evaluation window used for anomaly detection - it shows the graph limited
to the query frame (for bars it's a bit different - where each bar would
represent the query time frame). Hopes this helps in the meanwhile.

Best,

Rachel

PR #1790, opened 2017-11-10
Clarifies how composite monitors are evaluated

cmatteri:


l0k0ms:
@cmatteri @dd-caleb Many thanks for this PR and the review, I have updated only the new documentation with this content since this guide is removed on the new doc tree.

New doc is coming in prod soon :)

PR #1803, opened 2017-11-16
update agent 6 installation

NBParis:
### What does this PR do?
Update the documentation to recommend a fresh install of the agent 6 instead of the upgrade.

### Motivation
Almost all raised issue by log users are linked to the upgrade of the agent 6


ToddRizley:
👍

PR #1828, opened 2017-12-05
Add new section on automatic aws log collection

DanielLanger:


DanielLanger:
Preview: https://docs-staging.datadoghq.com/daniel/awslogs/logs/aws/

content/logs/aws.md
```diff
@@ -47,10 +47,37 @@ Push your AWS log information to Datadog using a Lambda function bound to an S3
 
 ## Collection
 
-Your lambda function is now ready to send logs to your Datadog platform.
-Setup the relevant triggers for each AWS service you want to monitor.
-
-### S3, Cloudwatch, API Gateway, Kinesis and SNS
+Your lambda function is now ready to send logs to your Datadog platform. Setup the relevant triggers for each AWS service you want to monitor.
```
technovangelist: Your platform??? They own datadog??? Shouldn't it be 'the' datadog platform

content/logs/aws.md
```diff
@@ -47,10 +47,37 @@ Push your AWS log information to Datadog using a Lambda function bound to an S3
...
+
+**The Lambda function you just created must be in the same region as the S3 bucket you are using as a trigger. If you have logs across multiple regions you must create additional Lambda funcions**
```
technovangelist: here you refer to it as a trigger (correct) but above in the overview its a binding.

content/logs/aws.md
```diff
@@ -47,10 +47,37 @@ Push your AWS log information to Datadog using a Lambda function bound to an S3
...
+
+### Automatically set up triggers
+If you are storing logs in many S3 buckets, Datadog can automatically manage triggers for you.
```
technovangelist: we are 'managing' the triggers? or monitoring them?

content/logs/aws.md
```diff
@@ -47,10 +47,37 @@ Push your AWS log information to Datadog using a Lambda function bound to an S3
...
+
+1. Add the required permissions to your Datadog role in the [IAM Console](https://console.aws.amazon.com/iam/home#/roles). You may already have some of these permissions from our other AWS integrations. Information on how these permissions are used can be found in the [permissions](#permissions) section below
+
+{{< highlight json>}}
```
technovangelist: this is just a list of things, nothing JSON'y about it. so the commas are being highlighted in a wierd way, but i think it would be worth adding a complete valid IAM policy rather than just the ALLOW clause

content/logs/aws.md
```diff
@@ -47,10 +47,37 @@ Push your AWS log information to Datadog using a Lambda function bound to an S3
...
+2. Navigate to the "Collect Logs" tab in the [AWS Integration tile](https://app.datadoghq.com/account/settings#integrations/amazon_web_services)
+3. Select the AWS Account from where you want to collect logs, and enter the ARN of the Lambda created in the previous section.
+{{< img src="logs/aws/AWSLogStep1.png" alt="Enter Lambda">}}
+4. Check off the services from which you'd like to collect logs and hit save. To stop collecting logs from a particular service, simply uncheck it.
+{{< img src="logs/aws/AWSLogStep2.png" alt="Select services">}}
+5. If you have logs across multiple regions, you must create additional Lambda functions in those regions and enter them in this tile.
+6. To stop collecting all AWS logs, press the "x" next to each Lamdba ARN. All triggers for that function will be removed. 
+
+### Manually set up triggers
 In your lambda, go in the triggers tab and select `Add Trigger`:
```
technovangelist: Lambda is capitalized

content/logs/aws.md
```diff
@@ -17,40 +17,71 @@ Push your AWS log information to Datadog using a Lambda function bound to an S3
...
+3. Select the AWS Account from where you want to collect logs, and enter the ARN of the Lambda created in the previous section.
+{{< img src="logs/aws/AWSLogStep1.png" alt="Enter Lambda">}}
+4. Check off the services from which you'd like to collect logs and hit save. To stop collecting logs from a particular service, simply uncheck it.
+{{< img src="logs/aws/AWSLogStep2.png" alt="Select services">}}
+5. If you have logs across multiple regions, you must create additional Lambda functions in those regions and enter them in this tile.
+6. To stop collecting all AWS logs, press the "x" next to each Lamdba ARN. All triggers for that function will be removed. 
+7. Within a few minutes you should see your AWS Logs appear in our [logging platform](https://app.datadoghq.com/logs).
+
+### Manually set up triggers
 In your lambda, go in the triggers tab and select `Add Trigger`:
```
technovangelist: Lambda is capitalized

content/logs/aws.md
```diff
@@ -110,6 +142,20 @@ Store the log files for multiple distributions in the same bucket. When enabling
 
 You can then collect the log from the s3 bucket thanks to the lambda function.
```
technovangelist: Lambda, not lambda. probably worth searching everywhere, except the permissions

DanielLanger:
@NBParis Thoughts on Matt's comments:

Why do we need 3GB of memory when less than 1gb is probably good enough?

Does it really need 120 seconds. You should set this to the min needed.

NBParis:
@technovangelist @DanielLanger Regarding the limits (max mem and timeout) this is because the lambda might have to deal with huge log files as we are talking about potentially TB of data.

So the recommended setup is indeed to have both those parameters at max.

NBParis:
@technovangelist Indeed the Lambda interface has changed recently and the screenshot are not up to date anymore. Did you update them or should I ? 
As it seems that you removed some of them but i cannot see new ones.

content/logs/aws.md
```diff
@@ -8,50 +8,81 @@ beta: true
...
 
-Your lambda function is now ready to send logs to your Datadog platform.
-Setup the relevant triggers for each AWS service you want to monitor.
-
-### S3, Cloudwatch, API Gateway, Kinesis and SNS
-In your lambda, go in the triggers tab and select `Add Trigger`:
+Your Lambda function is now ready to send logs to the Datadog platform. Setup the relevant triggers for each AWS service you want to monitor.
+
+**The Lambda function you just created must be in the same region as the S3 bucket you are using as a trigger. If you have logs across multiple regions you must create additional Lambda funcions.**
+
```
NBParis: @DanielLanger I would add here a small paragraph that explains that there is two way to configure the triggers:
automatic (which requires some permissions ) and manual.

With a link on automatic and manual to the corresponding section (and maybe a link on permissions as well).
This way it might be easier when we will split this document and clearer for users as they are directly aware of the two choices.

NBParis: And i would add the exact same paragraph on the onboarding with the same links.

content/logs/aws.md
```diff
@@ -8,50 +8,81 @@ beta: true
...
+{{< /highlight >}}
+
+2. Navigate to the *Collect Logs* tab in the [AWS Integration tile](https://app.datadoghq.com/account/settings#integrations/amazon_web_services)
+3. Select the AWS Account from where you want to collect logs, and enter the ARN of the Lambda created in the previous section.
+{{< img src="logs/aws/AWSLogStep1.png" alt="Enter Lambda">}}
+4. Check off the services from which you'd like to collect logs and hit save. To stop collecting logs from a particular service, simply uncheck it.
+{{< img src="logs/aws/AWSLogStep2.png" alt="Select services">}}
+5. If you have logs across multiple regions, you must create additional Lambda functions in those regions and enter them in this tile.
+6. To stop collecting all AWS logs, press the *x* next to each Lamdba ARN. All triggers for that function will be removed. 
+7. Within a few minutes you should see your AWS Logs appear in our [logging platform](https://app.datadoghq.com/logs).
```
NBParis: The `few minutes` might be confusing here. 

It can indeed take a couples minutes for the first logs to show up in the platform but then it should take only a couples seconds. 
Maybe we can try to be clearer to avoid users to think there is a delay for all logs as we have for Cloudwatch metrics (for instance).

PR #1831, opened 2017-12-06
[forecast] add use-case

githomin:
### What does this PR do?
Adds a use-case.

### Motivation
Nowhere in our docs or blogposts is the prototypical use-case mentioned, so we should point users in the right direction here.


StephenKappel:
Why is "use-case" hyphenated? Otherwise, seem good 👍

githomin:
er... because it seems like its on its way to becoming a compound word?

PR #1832, opened 2017-12-07
Splitting Containers from Processes

mikezvi:
### What does this PR do?
Updating Live Processes docs for GA

### Motivation
Live Processes docs used to be "Live Processes and Containers", but we already have a containers page.  We can unify them in the future, if that makes sense, but I needed to do one or the other, and since they have different installation flows and some process users are not Docker users, I took the safe route for now, even though the pages are pretty similar.


content/infrastructure/process.md
```diff
@@ -61,57 +61,38 @@ Refer to the standard [daemonset installation](http://docs.datadoghq.com/integra
...
-Ok, so I guess that last one is less exciting after redaction!
-
-## Broad Inspection, Deep Inspection
+Ok, so I guess that one is less exciting after redaction!
 
-Everyone's workflow differs.  Initially the table is displayed at the finest grain, but with the group-by field, you should start your investigation where it's appropriate for you: Grouping by Availability Zone, Host, Cluster, Pod, or wherever.
+## Enriched Live Containers view
 
-From there, you can dig down into finer grains, or inspect each group to see individual processes or containers.  In the below screenshot, you can see an investigation that started by indexing by pod and service, dug into one pod to see the containers, and then expanded a container to see the process tree inside.  In the container inspect tray, we also have some recent context for these metrics.
+Live processes adds extra visibility to your container deployments.  In addition to aggregating on orchestrator metadata, like ECS `task` or Kubernetes `service`, and drilling down to the container level, when Live Processes is enabled, you can continue to drill down into the process tree inside each container.
```
conorbranagan: Second sentence in this paragraph feels maybe too long

PR #1839, opened 2017-12-11
[forecast] take docs out of beta

githomin:
### What does this PR do?
Takes Forecast docs out of beta


StephenKappel:
Still looks good to me!

PR #1849, opened 2017-12-14
Jyee/redesign acct mgmt

jyee:
### What does this PR do?
Completely updates the account management section.

### Motivation
The account management section was just a compilation of previous docs and was incomplete.

### Additional Notes
We still need to add Py and Rb code examples in the api org docs. This will involve checking the Py and Rb libraries (not certain if they have org functionality).


jyee:
This still needs work. I need to add a basic Org settings page that covers renaming an org, usage reports (merge in the existing doc) and billing history.

PR #1876, opened 2017-12-26
Adding a note about /passwd

mikezvi:

### DIFF
--- a/content/graphing/infrastructure/process.md
+++ b/content/graphing/infrastructure/process.md
@@ -38,7 +38,7 @@ and adding the following line to the `[Main]` section:
 ```
 
 After configuration is complete, [restart the Agent](/agent/faq/start-stop-restart-the-datadog-agent).  
-**Note**: Collecting Container information in the standard install, the dd-agent user needs to have permissions to access docker.sock.
+**Note**: To Collect container information in the standard install, the dd-agent user needs to have permissions to access docker.sock.
 
 ### Docker container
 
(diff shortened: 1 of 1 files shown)


content/graphing/infrastructure/process.md
```diff
@@ -120,6 +120,8 @@ While actively working with the Live Processes, metrics are collected at 2s reso
 
 - The Process Agent is available for the default Debian docker-dd-agent image only.  It is not included in the Alpine image.
 
+- The `/etc/passwd` file mounted into the docker-dd-agent is necessary to collect usernames for each process.  This is a public file and the Process Agent does not use any fields except the username.  All features except the `user` metadata field will function without access to this file.
```
mikezvi: I originally had "All users have read permissions for this file by default", but I changed it to "public" because I felt like it was less sterile and more likely to put people at ease.  Let me know if you think the former is better.

mikezvi: Also I tested the assertion "All features except the `user` metadata field will function without access to this file" but "all" is doing some heavy lifting there.  Let me know if you don't think it's true.

irabinovitch: Would we show user id if /etc/passwd is missing?

mikezvi: The current behavior is not to; the field is just blank.  I think it is a good idea, though.

irabinovitch: OK.  LGTM but we should clarify this is a tip/faq for container based deployments.

PR #1881, opened 2017-12-28
adding list of source attribute values

l0k0ms:
### What does this PR do?

Adding a list of source attribute values depending of their respective integration name
### Motivation

It was missing
### Preview link
https://docs-staging.datadoghq.com/gus/api-tag/integrations/faq/list-of-api-source-attribute-value/

### DIFF
--- a/content/integrations/faq/list-of-api-source-attribute-value.md
+++ b/content/integrations/faq/list-of-api-source-attribute-value.md
@@ -1,0 +1,206 @@
+---
+title: List of API source attribute value
+kind: faq
+---
+
+|Integration name | API source attribute|
+|:---|:---|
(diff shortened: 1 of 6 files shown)


content/integrations/faq/list-of-api-source-attribute-value.md
```diff
@@ -0,0 +1,206 @@
+---
```
irabinovitch: This feels like it should be generated from dogweb AND postgres.  
Keep in mind new sourcetypes are primarily added in the database.

l0k0ms: @irabinovitch agreed, I have added a card in the trello board. 
But can we merge this in the meantime?

PR #1906, opened 2018-01-03
remove empty instances

NBParis:
### What does this PR do?
We do not need anymore to have [{}] in the list of instances to have the logs working.
Therefore the documentation should not advise to do so.

Especially as the status of the agent do not return error for the status command if the list of instances is empty (for logs file).

### Motivation
Having the correct information in the agent status command is very important


l0k0ms:
@NBParis should we updated docker configuration as well ?

NBParis:
@l0k0ms Yes we should update everywhere in the doc whith yaml example, so even languages.

PR #1928, opened 2018-01-09
adding unique hostname disclaimer

l0k0ms:
### What does this PR do?

Adding disclaimer for unique hostnames within Datadog 

### Motivation

If hostnames are not unique, users can see gaps in their graphs

### Preview link

 https://docs-staging.datadoghq.com/gus/unique-hostname/graphing/infrastructure/#agent-host-names

### DIFF
--- a/content/graphing/infrastructure/_index.md
+++ b/content/graphing/infrastructure/_index.md
@@ -69,8 +69,8 @@ the current rule fails and passes through to the next.
 **Note**: If an EC2 instance is an ECS host, Datadog uses the `instance-id` as the hostname, [even if the DNS hostname isn't an EC2 default](https://github.com/DataDog/dd-agent/blob/5.14.1/utils/hostname.py#L104). If you don't wish to use the `instance-id`, set the hostname in the Agent configuration file.
 
 <div class="alert alert-warning">
-Hostnames must be unique within a Datadog account.<br> 
-For example: you can't have two hosts with the hostname <code>10.8.0.1</code> reporting to the same Datadog account. Otherwise you may see gaps in your graphs.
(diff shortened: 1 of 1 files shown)


content/graphing/infrastructure/_index.md
```diff
@@ -66,10 +66,18 @@ match is selected.
 If name is recognized as obviously non-unique (e.g. `localhost.localdomain`),
 the current rule fails and passes through to the next.
 
+**Note**: If an EC2 instance is an ECS host, Datadog uses the `instance-id` as the hostname, [even if the DNS hostname isn't an EC2 default](https://github.com/DataDog/dd-agent/blob/5.14.1/utils/hostname.py#L104). If you don't wish to use the `instance-id`, set the hostname in the Agent configuration file.
+
+<div class="alert alert-warning">
+Hostnames must be unique within a Datadog account.<br> 
+For example: you can't have two hosts with the hostname <code>10.8.0.1</code> reporting to the same Datadog account. Otherwise you may see gaps in your graphs.
```
jyee: Is it better to phrase this as "hostnames should be unique" and "For example, if you have... account, you may see gaps in your graphs."

Also, I wonder if this should have more explanation than just "gaps". I could be wrong, but gaps in a graph make it sound like metrics, then no metrics, then metrics again in a graph. Depending on the metric, I think the bigger issue might be data inconsistency or faulty data (e.g. cpu on host 1 reports solid 10% use and host 2 reports 90% use, but both are reporting as the same host, so you'd have a graph that flaps between 10 and 90%)

l0k0ms: Metric gaps can actually be a consequence duplicated hostnames..

PR #1939, opened 2018-01-12
Functions desc update

l0k0ms:
### What does this PR do?

Explaining what a span is for query functions

### Motivation

It was confusing with the trace span definition

### Preview link

* https://docs-staging.datadoghq.com/gus/functions-ewma/graphing/miscellaneous/functions/#smoothing

### DIFF
--- a/content/graphing/miscellaneous/functions.md
+++ b/content/graphing/miscellaneous/functions.md
@@ -206,17 +206,25 @@ Delta value between points for a given metric
 
 Exponentially weighted moving average with a span of 3
 
+The span value is the number of data points. So ewma_3() would use the last 3 data points to calculate the average.
+
 ### ewma_5()
 
 Exponentially weighted moving average with a span of 5
+
+The span value is the number of data points. So ewma_5() would use the last 5 data points to calculate the average.
 
 ### ewma_10()
 
 Exponentially weighted moving average with a span of 10
 
+The span value is the number of data points. So ewma_10() would use the last 10 data points to calculate the average.
+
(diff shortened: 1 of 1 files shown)


l0k0ms:
the link that beaks the test_preview_links test is valid

AndrewMcBurney:
There's are some similar lines in the descriptions which could be cut down.

I would consider changing the descriptions as follows:

```md
## Smoothing

The span value refers to the number of data points used in the calculation.

**Example:**

* `ewma_5()` uses the _last 5_ data points to calculate the average.
* `median_9()` uses the _last 9_ data points to calculate the median.

### ewma_3()

Exponentially weighted moving average with a span of 3.

### ewma_5()

Exponentially weighted moving average with a span of 5.

### ewma_10()

Exponentially weighted moving average with a span of 10.

### ewma_20()

Exponentially weighted moving average with a span of 20.

### median_3()

Rolling median with a span of 3.

...
```

This way we're cutting down on the number of similar lines.

l0k0ms:
@AndrewMcBurney I agree with you, but most of the time this page is used by cake team as a reference in a support ticket discussion to give more details on a specific function. 

Even if this layout is super heavy it allows each function description to be self-explanatory.  For example, if a SE links this: https://docs-staging.datadoghq.com/gus/functions-ewma/graphing/miscellaneous/functions/#median-9 then we want the span description within the function `median_9()` section. Otherwise, the reader might miss it. :/

AndrewMcBurney:
@l0k0ms sounds good to me 👍

PR #1982, opened 2018-01-24
Adding confirmation steps for flare prior upload

l0k0ms:
### What does this PR do?

Add info about the confirmation steps prior to upload for flare command

### Motivation

The datadog-agent prompts a confirmation before uploading the flare to Datadog support. This is not documented and some users forget to confirm so their flare doesn't get updated.


Martiflex:
👍

PR #1990, opened 2018-01-25
David.jones/service checks

davidejones:
### What does this PR do?
This pulls in service_checks.json from the repos and adds the data to read me files where needed.

### Motivation
https://trello.com/c/1l6HYZOH/148-automate-inclusion-of-service-checks-in-docs

### Preview link
https://docs-staging.datadoghq.com/david.jones/service-checks/integrations/haproxy/#service-checks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ static/css/
 static/js/
 static/images/
 data/integrations/
+data/service_checks/
 data/manifests/
 content/integrations/*.md
 !content/integrations/_index.md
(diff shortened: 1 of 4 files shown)


l0k0ms:
Do you know why the formatting isn't displayed?  
```
`CRITICAL` for a given service if HAProxy is reporting it `down`. `maint`, `ok` and any other state will result in `OK`.
```

davidejones:
@l0k0ms Ah yes this works like the metrics where it just inserts a shortcode into the readme with python and then hugo is actually doing the output. 

I've wrapped that part in a markdownify so that should be working now.

PR #2025, opened 2018-02-06
Detailed taxonomy for Application log attributes

NBParis:
### What does this PR do?
Explain what attributes to use in order to get most of the application logs

### Motivation
We have a specific treatment for application log attributes such as thread name, error message, stack traces, ... and this was not yet documented.


NBParis:
@l0k0ms Could you take the cloudapp link and put that as an image directly ? Thanks

jyee:
LGTM too

PR #2031, opened 2018-02-07
[Kafka] Sync with zendesk

l0k0ms:
### What does this PR do?

Sync the doc KB with zendesk KB: https://datadog.zendesk.com/hc/en-us/articles/115002655346

### Motivation

It was out of date

### Additional Notes

Content on zendesk should be removed once merged

### DIFF
--- a/content/integrations/faq/troubleshooting-and-deep-dive-for-kafka.md
+++ b/content/integrations/faq/troubleshooting-and-deep-dive-for-kafka.md
@@ -63,7 +63,7 @@ Additionally, there are cases where users are using custom Producer and Consume
 ```
 if you are running Producers and Consumers from other languages, this isn't an option, and you have to use another way to submit these metrics from your code, for instance through dogstatsd.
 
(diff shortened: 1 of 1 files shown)


content/integrations/faq/troubleshooting-and-deep-dive-for-kafka.md
```diff
@@ -62,7 +63,7 @@ Additionally, there are cases where users are using custom Producer and Consume
 ```
 if you are running Producers and Consumers from other languages, this isn't an option, and you have to use another way to submit these metrics from your code, for instance through dogstatsd.
 
-4. This issue is specifically for the Kafka_Consumer check. If you specify a partition in your Kafka_Consumer.yaml file that doesn't exist in your environment, you see the following error in info.log:
+4. This issue is specifically for the Kafka_Consumer check. If you specify a partition in your `Kafka_Consumer.yaml` file that doesn't exist in your environment, you see the following error in info.log:
```
AndrewMcBurney: Here you put backticks around `Kafka_Consumer.yaml`, because it's a file, but not `info.log`, which is also a file.

Could you please put backticks around `info.log`? 🙂

PR #2086, opened 2018-02-27
Adding process monitoring with ansible KB

l0k0ms:
A customer couldn't get process agent to work for an Agent 6 beta deployment on ansible so a support member tested and created an ansible playbook that finally got it to work. The customer suggested it should be in our docs.


jeremy-lq:
We should get proper support for the Process Agent in the [Ansible Role](https://github.com/DataDog/ansible-datadog).

l0k0ms:
@jeremy-lq should I open an issue on https://github.com/DataDog/ansible-datadog ?

PR #2096, opened 2018-02-28
Changing font for code highlights to enhance table readability

l0k0ms:
### What does this PR do?

It updates the font used for code highlight in order to allow a better display within table an on doc pages.

This was the old font:

image

This is the new font: 

image

### DIFF
--- a/content/api/checks/checks_post.md
+++ b/content/api/checks/checks_post.md
@@ -9,13 +9,13 @@ external_redirect: /api/#post-a-check-run
 
 ##### ARGUMENTS
 
-* `check` *[required]*:  
+* **`check`** *[required]*:  
     The text for the message
 
-* `host_name`*[required]*:  
+* **`host_name`** *[required]*:  
     The name of the host submitting the check
 
-* `status` *[optional]*:   
+* **`status`** *[optional]*:   
     An integer for the status of the check:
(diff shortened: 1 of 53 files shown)


AndrewMcBurney:
Would it be possible to use a smaller monospace font? The only concern I have is that poeple might not associate the sans-serif font with code or commands. If we have a smaller monospace font, that would be :100:

AndrewMcBurney:
I personally really like [Source Code Pro](https://fonts.google.com/specimen/Source+Code+Pro), which is a free google font we could import.

AndrewMcBurney:
So the font size is currently `16px` for the code in the tables. This seems pretty big when compared with the sans-serif font in the tables, like you mentioned.

Another recommendation is potentially changing the monospace font size to `12px` or `14px`. I've attached some screenshots previewing the changes for both suggested font sizes:

screen shot 2018-02-28 at 11 16 24 am

> `12px` monospace

screen shot 2018-02-28 at 11 16 37 am

> `14px` monospace

l0k0ms:
@AndrewMcBurney really good ideas! I took your font, added back monospace and reduce the size.. let's watch the preview: https://docs-staging.datadoghq.com/gus/code-layout-update/getting_started/

PR #2128, opened 2018-03-05
Adding log graph doc

l0k0ms:
### What does this PR do?

Add doc for log graph

For now, there is only the Timeseries log graph, hence this doc is in `/logs/explore` section when new log graph will be available, they all will be stored in a different page under `/logs/explore/log_graphs`

### Motivation
Prepare Log GA

### Preview link

https://docs-staging.datadoghq.com/gus/analytics/logs/explore/#log-graph


config.yaml
```diff
@@ -438,7 +438,7 @@ languages:
           weight: 90
         # LOGS - Subnav start
         - identifier: customnav_lognav_explore
-          name: "Explore"
+          name: "Search & Graph"
```
technovangelist: search and graph seems odd. whats here?

content/logs/explore.md
```diff
@@ -179,6 +181,26 @@ If an attribute was called `user.first name`, perform a search on this attribute
...
+{{< img src="logs/explore/graph/log_graph_switch.png" alt="Log graph switch" responsive="true" popup="true" style="width:80%;">}}
+
+The only available Log Graph (for now) is the timeseries:
+
+{{< img src="logs/explore/graph/basic_timeseries_log_graph.png" alt="basic timeseries log graph" responsive="true" popup="true" style="width:80%;">}}
+
+To start using it:
+
+1. Choose the Measure to graph
+2. Split by [Tag](/getting_started/tagging) or [Facet](#facets) to split your graph over the wanted dimension.
```
technovangelist: hmmm, wanted? or is desired better. they mean the same thing

content/logs/explore.md
```diff
@@ -179,6 +181,26 @@ If an attribute was called `user.first name`, perform a search on this attribute
...
+{{< img src="logs/explore/graph/split_by_log_graph.png" alt="split by log graph" responsive="true" popup="true" style="width:80%;">}}
+
+3. See logs related to a section of the graph:  
+    Select or click on a section of the graph to either zoom in the graph or see the list of logs corresponding to your selection:
```
technovangelist: what is the difference between selecting and clicking?

AndrewMcBurney:
I would remove "The only available Log Graph (for now) is the timeseries" and the corresponding image:

screen shot 2018-03-05 at 10 08 30 am

config.yaml
technovangelist: so there is a lot of stuff in search and graph. can it be split out? also the other menu items are processing and parsing. to keep it consistent, can it be searching and graphing? or change the others to process and parse

l0k0ms: Yup that's the end-goal of this, but since there is no real "Graph" i.e it's only the log graph timeseries that's available for now. We thought that it wouldn't be relevant to have one page for graphing almost empty.

But as soon as a new log graph is available, this content will be moved into `/logs/explore/log_graphs`

content/logs/explore.md
l0k0ms: Selecting: You select a time range and see associated logs
Clicking: you click on an aggregation result to see the associated logs

content/logs/explore.md
l0k0ms: desired 👌

PR #2131, opened 2018-03-05
add border to content images

davidejones:
### What does this PR do?
This PR adds borders around images to help clarify what is an image and what is part of the page.

### Motivation
https://trello.com/c/cOUhtrp5/171-add-border-to-doc-images

### Preview link
https://docs-staging.datadoghq.com/david.jones/border-images/


kdobesh:
the docs look so much more structured now! 

@davidejones do you think we could push this with  a `1px #d6d6d6` border instead though? it matches what we do on the tables and I am not trying to rock the boat with new styles here... Thanks!

PR #2152, opened 2018-03-08
Fix cmd.exe flare command

cmatteri:
It can't have an ampersand and it must use double quotes


cmatteri:
Thanks for the quick reviews on these!

PR #2161, opened 2018-03-10
Removing process metrics

l0k0ms:
### What does this PR do?

Remove system processes metrics from the integrations system page, since those metircs are displayed on the process integration page

### Motivation

CR

### Preview

https://docs-staging.datadoghq.com/gus/system-metric/integrations/system/#data-collected


Martiflex:
Looks good
👍

PR #2199, opened 2018-03-19
Changing handle param to id

l0k0ms:
### What does this PR do?

Change each 'handle' mention on api docs to 'id'.

### Motivation

When user explicitly specifies the 'handle' argument for any of the user API python methods, the call fails.

Error if including 'handle=' in API call:
Traceback (most recent call last):
File "test.py", line 11, in <module>
print(api.User.get(handle='sroman@gmail.com'))
TypeError: get() missing 1 required positional argument: 'id'

Every link that may be useful: 
potentially related source code

https://github.com/DataDog/datadogpy/blob/master/datadog/api/users.py

https://github.com/DataDog/datadogpy/blob/master/datadog/api/resources.py

https://github.com/DataDog/dogweb/blob/prod/dogweb/model/user.py

### Preview link
* https://docs-staging.datadoghq.com/gus/user-api-param-update/api/?lang=python#users

### DIFF
--- a/content/api/users/code_snippets/api-user-create.rb
+++ b/content/api/users/code_snippets/api-user-create.rb
@@ -6,4 +6,4 @@ app_key = '87ce4a24b5553d2e482ea8a8500e71b8ad4554ff'
 
 dog = Dogapi::Client.new(api_key, app_key)
 
-dog.create_user(: handle => 'test@datadoghq.com', : name => 'test user')
+dog.create_user(: id => 'test@datadoghq.com', : name => 'test user')
--- a/content/api/users/code_snippets/api-user-create.rb
+++ b/content/api/users/code_snippets/api-user-create.rb
@@ -6,4 +6,4 @@ app_key = '87ce4a24b5553d2e482ea8a8500e71b8ad4554ff'
 
 dog = Dogapi::Client.new(api_key, app_key)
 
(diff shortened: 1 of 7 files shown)


content/api/users/code_snippets/api-user-create.rb
```diff
@@ -6,4 +6,4 @@
 
 dog = Dogapi::Client.new(api_key, app_key)
 
-dog.create_user(: handle => 'test@datadoghq.com', : name => 'test user')
\ No newline at end of file
+dog.create_user(: id => 'test@datadoghq.com', : name => 'test user')
```
AndrewMcBurney: Could you change this to:

```ruby
dog.create_user(:id => 'test@datadoghq.com', :name => 'test user')
```

or even better:

```ruby
dog.create_user(id: 'test@datadoghq.com', name: 'test user')
```

Ruby has two hash syntaxes which is kinda weird, but the second one is preferred for symbol keys as of [Ruby 1.9](http://www.rubydoc.info/gems/rubocop/0.45.0/RuboCop/Cop/Style/HashSyntax).

technovangelist: oooo, nice comment. i did not know this and just got confused when i saw the two

PR #2220, opened 2018-03-20
duplicate no-indexing code, consolidate and and make it noindex when …

davidejones:
### What does this PR do?
This pr will consolidate the noindex calls to one new partial called "noindex.html" 
This meta tag will now be included if any of the following is true

- front matter param beta is true
- front matter param private is true
- front matter param placeholder is true
- front matter param noindex is true
- site config environment is preview

### Motivation
For seo reasons so that we aren't indexing fr pages that are placeholders and have english content which is a duplicate of the english site. Also to make maintenance a little easier next time.

### Preview link
https://docs-staging.datadoghq.com/david.jones/no-index/

### DIFF
--- a/layouts/partials/noindex.html
+++ b/layouts/partials/noindex.html
@@ -1,0 +1,3 @@
+{{ if (or (eq .Params.beta true) (eq .Params.private true) (eq .Params.placeholder true) (eq .Site.Params.environment "preview") (eq .Params.noindex true) ) }}
+<meta name="robots" content="noindex, nofollow">
+{{ end }}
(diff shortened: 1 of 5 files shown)


layouts/partials/noindex.html
```diff
@@ -0,0 +1,3 @@
+{{ if (or (eq .Params.beta true) (eq .Params.private true) (eq .Params.placeholder true) (eq .Site.Params.environment "preview") (eq .Params.noindex true) ) }}
+<meta name="robots" content="noindex, nofollow">
+{{ end }}
```
mstbbs: you should set that pref :p

PR #2236, opened 2018-03-23
Switching between search results pages creates browser history entries

amalik12:
### What does this PR do?
Switching to a new page in the search results creates a new entry in the browser's history, and appends a query parameter to the url corresponding to the page, allowing users to link to specific search result pages.

Before:
* User enters a search query from docs homepage
* User is on page 1 of results
* User navigates to page 5
* User hits back button
* User is on docs homepage

After:
* User enters a search query from docs homepage
* User is on page 1 of results
* User navigates to page 5
* User hits back button
* User is on page 1 of results

### Motivation

Switching between pages in search results currently does not create new history entries, so if a user 
changes pages and hits the back button on the browser, they're taken to the page visited before the search results. This makes navigation within search results harder, and can be very disorienting for users.

### Preview link

All search results pages

### DIFF
--- a/src/js/datadog-docs.js
+++ b/src/js/datadog-docs.js
@@ -155,6 +155,7 @@ $(document).ready(function () {
                             if(current_page < page_nums[0]) {
                                 less(null);
                             } else {
+                                addHistory(current_page);
                                 changePage(current_page);
                             }
                         }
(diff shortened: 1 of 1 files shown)


l0k0ms:
https://docs-staging.datadoghq.com/adisa/search-results-history-nav/search/?s=docker

PR #2265, opened 2018-03-30
Ddog docs/user api create

jpinkham:
### What does this PR do?

Corrects an issue with the "Curl" example for user creation. Documents the 'icon' response field after user creation.

### Motivation

When I tried to create a user via curl, using the provided example, I received an error stating I needed to supply a 'handle' value, not 'id' as the docs stated.

### Additional Notes

The way the documentation is structured, there is only one list of required fields and their description.  This makes it impossible to update the documentation to correct an issue with 'handle' being a required field, only for the curl version.  I could only update the code example.


content/api/users/code_snippets/result.api-user-create.sh
```diff
@@ -6,6 +6,7 @@
     "disabled": false,
     "role": null,
     "is_admin": false,
-    "email": "test@datadoghq.com"
+    "email": "test@datadoghq.com",
+    "icon": /path/to/matching/gravatar/icon
```
l0k0ms: shouldn't it be `"icon": "/path/to/matching/gravatar/icon"` ?

jpinkham: Whoops, yes it should be a string, even though it's just a placeholder

PR #2272, opened 2018-04-02
add language entry for search and fix issue with toc on initial load

davidejones:
### What does this PR do?
This PR fixes an issue with the table of contents appearing under the header on page load. It also adds in a language string for "search"

### Motivation
I found the visual bug with the toc while testing
I also saw that the search placeholder text wasn't a string translation so added it

### Preview link
https://docs-staging.datadoghq.com/david.jones/minor-fixes/

### Additional Notes
Visit https://docs-staging.datadoghq.com/david.jones/minor-fixes/getting_started/ when the page has loaded click the book icon in the top right to toggle the table of contents. You will see it shows correctly. If you follow the same actions on production you will see that it appears slightly under the header but will correct itself when you scroll down.

### DIFF
--- a/i18n/fr.json
+++ b/i18n/fr.json
@@ -218,5 +218,9 @@
 
   "questions": {
     "other": "Questions?"
+  },
+
+  "search": {
+    "other": "Search"
   }
 }
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -218,5 +218,9 @@
 
   "questions": {
     "other": "Questions?"
+  },
+
+  "search": {
+    "other": "Search"
   }
 }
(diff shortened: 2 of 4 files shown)


i18n/fr.json
```diff
@@ -218,5 +218,9 @@
 
   "questions": {
     "other": "Questions?"
+  },
+
+  "search": {
+    "other": "Search"
```
mstbbs: @davidejones - is this a file that gets sent to transifex?

davidejones: @mstbbs no it won't get sent to transifex we only send source files, so the `en.json` one will get sent. 

The reason that I've modified both files is that i was under the impression if i don't add it to the fr.json then hugo errors because its trying to lookup strings that don't exist when building the french site. 

I just tested this theory though and it gave me no issues so I must have been thinking of something else :confused:

PR #2378, opened 2018-04-20
clarify downtime monitor behavior

rsheyd:
### What does this PR do?
There was leftover language that didn't make sense with the updated language about monitor behavior during scheduled downtime. "This is not the case..." was referring to text that was removed in February. This PR attempts to clarify the behavior of notifications after downtime expires.

### Motivation
Customer requests and updated monitor behavior.


rsheyd:
Language that was removed and referenced by currently existing language: https://github.com/DataDog/documentation/pull/2066

PR #2423, opened 2018-04-30
deal with package*.json artefacts

phrawzty:
### What does this PR do?
Removes the `package.json` and `package-lock.json` files from the repo and adds entries for same in `.gitignore`.

### Motivation
Desire to keep things tidy.

### Additional notes
Need 👀 to ensure that this is safe and makes sense.


phrawzty:
image

NEVER MIND

PR #2438, opened 2018-05-03
Fixing table layout

l0k0ms:
### What does this PR do?

Fix table layout in files
Fix table links

### Preview
https://docs-staging.datadoghq.com/gus/table-fix/agent/basic_agent_usage/

### DIFF
--- a/content/agent/_index.md
+++ b/content/agent/_index.md
@@ -113,26 +113,26 @@ To keep backwards compatibility, the Agent will still pick up configuration file
 
 The new command line interface for the Agent is sub-command based:
 
-| Command         | Notes
+| Command         | Notes                                                                      |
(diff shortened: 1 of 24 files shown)


phrawzty:
It looks like most (all?) of the modifications can be summed up as:
* Remove `{{% table responsive="true" %}}` from the tables
* Add whitespace to cells

Could you please add some details as to what these two modifications do?

…on the plus side, the tables are indeed "fixed". :)

l0k0ms:
The partial `{{% table responsive="true" %}}` was messing with the links inside of tables since we switch to footnote link recently.
It seems that instead of inlining the correct link we were inlining the placeholder. 

We add whitespace to cells in order to make them more readable when editing the raw file. (linter)

PR #2460, opened 2018-05-07
Fix 3 small typos in the tracing visualisation index page

tanguylebarzic:
### What does this PR do?
Fix 3 small typos / strange wordings


tanguylebarzic:
@jeremy-lq Thanks for the review! I don't understand your nit though - In the sentence "Therefore any custom process or logging library able to forward logs through TCP can be used.", there can only be one 'verb' - `can be used`. I don't think it's possible to also have `is able` - I'm not a native English speaker though, I may be missing something.

PR #2470, opened 2018-05-09
Wrap placeholder <YOUR_API_KEY> with quotes

samueldg:
### What does this PR do?

This prevents the yaml syntax highlighting from going wild
after the `<` / `>` characters.

### Motivation

Was looking in https://docs.datadoghq.com/tracing/setup/kubernetes/ and noticed the syntax highlighting for the example was weird.

dd-kubernetes-docs-before

### Preview link

dd-kubernetes-docs-after

### Additional Notes

Not sure why the syntax highlighting doesn't happen on this page which has a similar example: https://docs.datadoghq.com/agent/basic_agent_usage/kubernetes/, but I went ahead and updated it regardless, for consistency.

### DIFF
--- a/content/agent/basic_agent_usage/kubernetes.md
+++ b/content/agent/basic_agent_usage/kubernetes.md
@@ -65,7 +65,7 @@ spec:
             protocol: TCP
         env:
           - name: DD_API_KEY
-            value: <YOUR_API_KEY>
+            value: "<YOUR_API_KEY>"
           - name: DD_COLLECT_KUBERNETES_EVENTS
             value: "true"
           - name: DD_LEADER_ELECTION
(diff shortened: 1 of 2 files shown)


technovangelist:
The quotes make sense, but why using the angle brackets? do we do that anywhere else in the docs. been a while since i last looked for that...

technovangelist:
yup, looks like we do tend to use them in most places

technovangelist:
You mention that this was done to make it look better, have you verified that the quotes work? its just a string, so there shouldn't be any reason not to work, but one should always verify

samueldg:
Thanks for the review, I just noticed this is already live 😄 

To answer @technovangelist , the angle brackets are the suggested placeholder markers: https://github.com/DataDog/documentation/blob/master/CONTRIBUTING.md#conventions : "Code substitution"

And in YAML you can always put quotes around to indicate they are strings, so this is equivalent.

PR #2479, opened 2018-05-15
Fix build

l0k0ms:
### What does this PR do?

The file `datadog-checks-tests-helper.md` has been renamed `datadog_checks_tests_helper.md`  hence the fail on build since the action couldn't be applied to this.

This PR improve the build by avoiding exception break since integrations generation is "best effort"


phrawzty:
I'm not sure how this PR addresses the problem you're describing. If a file exists that should not, than removing it is exactly what we want to do, no? Why would the `remove()` fail at all?

local/bin/py/update_pre_build.py
```diff
@@ -263,7 +263,10 @@ def merge_integrations(self):
                         else:
                             content = re.sub(self.regex_h1_replace, r'##\2', content, count=0)
                         target_file.write(content)
-                    remove(input_file)
+                    try:
+                        remove(input_file)
+                    except:
```
phrawzty: Catching `*` is bad form. Suggest specifying the error you're excepting on:
```python
except OSError:
```

local/bin/py/update_pre_build.py
```diff
@@ -275,7 +278,10 @@ def merge_integrations(self):
                     else:
                         open(output_file, 'w').close()
                 elif action == 'discard':
-                    remove(input_file)
+                    try:
+                        remove(input_file)
+                    except:
```
phrawzty: As above.

local/bin/py/update_pre_build.py
phrawzty: Catching `*` is bad form. Suggest specifying the error you're excepting on:
```python
except OSError:
```

local/bin/py/update_pre_build.py
phrawzty: As above.

PR #2526, opened 2018-05-22
content/tracing/setup: update agent configuration

gbbr:
This change updates the "/tracing/setup" page to contain all available
agent configurations. It additionally adds an extra column specifying
the type of values that these configurations accept.


gbbr:
Thanks guys! Merged.

PR #2536, opened 2018-05-23
Add faq about docker and kubernetes metrics

dabcoder:
### What does this PR do?
Add description of the current situation regarding `docker.containers.running` and `kubernetes.pods.running` metrics in real time. 

### Motivation
Discussion about the topic.

### DIFF
--- a/content/agent/faq/docker-kubernetes-metrics-real-time.md
+++ b/content/agent/faq/docker-kubernetes-metrics-real-time.md
@@ -7,4 +7,4 @@ When displaying the `docker.containers.running` and/or `kubernetes.pods.running`
 
 The root cause is that zero values for those metrics are not sent when some contexts expire (such as a Redis pod that may be terminated on a node and recreated on another one), so you need to add the `fill-by` last-value to force contexts to expire sooner.
 
-**Note**: The `docker.container.runing.total` metric does not exhibit this symptom (since it is a client-computed host sum).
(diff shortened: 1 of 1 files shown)


content/agent/faq/docker-kubernetes-metrics-real-time.md
```diff
@@ -0,0 +1,10 @@
+---
+title: docker.containers.running and kubernetes.pods.running values
+kind: faq
+---
+
+When displaying the `docker.containers.running` and/or `kubernetes.pods.running` metrics in your metric explorer, they may not display the containers/pods situation accurately. 
+
+The root cause is that zero values for those metrics are not sent when some contexts expire (such as a Redis pod that may be terminated on a node and recreated on another one), so you need to add the `fill-by` last-value to force contexts to expire sooner.
+
+**Note**: The `docker.container.runing.total` metric does not exhibit this symptom (since it is a client-computed host sum).
```
mfpierre: s/runing/running

ethomatos: Hello MFPierre, docker.containers.running is supposed to have a value of what containers are actually running, correct? Isn't the check being executed on the agent running for example inside a pod? If there are no containers running then the value should be 0. If a pod runs then dies, then it goes from 0 to 1 back to 0. What is meant by metrics not being sent? Is it that the agent does not send docker.containers.running=0 to the platform?

PR #2544, opened 2018-05-24
Move agent checks and prometheus checks docs to the devtools section

phrawzty:
### What does this PR do?
Moves the _Agent Checks_ and _Prometheus Checks_ docs to the Developer Tools section, as these are development materials (i.e. not just config & usage).

### Motivation
Moving all the developer documentation into one spot will make it easier to iterate and improve the whole over time.

### Preview link
image


phrawzty:
> The first portion of the Prometheus Check page does seem like it belongs with the Agent docs.

Fair enough. Closing this PR to re-structure.

PR #2570, opened 2018-05-30
Make the description of the count_no* functions more specific

cmatteri:
### Motivation
The existing description is not very specific. A more detailed description has been requested. Feel free to make suggestions about the wording. It makes sense to me as it is now, but I'm not the target audience.

### Preview link
https://docs-staging.datadoghq.com/chrism/count-functions/graphing/miscellaneous/functions/#count-nonzero

### Additional Notes

### DIFF
--- a/content/graphing/miscellaneous/functions.md
+++ b/content/graphing/miscellaneous/functions.md
@@ -17,16 +17,21 @@ aliases:
 ---
 
 ## Arithmetic
+
 ### abs()
+
 Absolute value for a given metric.
 
 ### log2()
+
 Base-2 logarithm for a given metric.
 
 ### log10()
+
 Base-10 logarithm for a given metric.
 
 ### cumsum()
+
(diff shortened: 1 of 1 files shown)


content/graphing/miscellaneous/functions.md
```diff
@@ -299,11 +299,11 @@ For example, an offset of 2 would start graphing at the number 3 ranked series,
 
 ### count_nonzero()
 
-Count all the non-zero values for a given metric
+For a query grouped by some tag key, count the number of tag values with nonzero metric values at each point. E.g. for `count_nonzero(system.load.1{*} by {host})` return a time series representing the number of hosts with nonzero system load at each point.
```
phrawzty: The word "some", when used as a determiner, indicates plurality, therefore the object needs to be plural as well. That said, note that the usage here is zero, one, or more tags, which should be expressed clearly. Consider:
> For a query grouped by one or more tag keys, count the […]

PR #2582, opened 2018-06-01
Remove Service instructions from Kubernetes Agent Docs

burningion:
### What does this PR do?

Removes the instructions to create a Kubernetes Service for the Datadog APM Agent, and replaces with instructions for using the Downward API.

### Motivation

Instructions to set the Datadog Agent as a Service meant requests were being load balanced instead of directed to the host Node.

### Preview link

https://docs-staging.datadoghq.com/kirk.kaiser/tracing-kubernetes-fixes/tracing/setup/kubernetes/

### DIFF
--- a/content/tracing/setup/kubernetes.md
+++ b/content/tracing/setup/kubernetes.md
@@ -12,7 +12,7 @@ further_reading:
   text: "Explore your services, resources and traces"
 ---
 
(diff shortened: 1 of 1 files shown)


l0k0ms:
@DataDog/kenafeh should we also update the app and the manifest on datadog-agent repo ?
https://github.com/DataDog/datadog-agent/blob/master/Dockerfiles/manifests/agent.yaml

hkaj:
@l0k0ms the manifest is good in datadog-agent. We shouldn't add this additional env var there especially because we have one already that has the same content: https://github.com/DataDog/datadog-agent/blob/master/Dockerfiles/manifests/agent.yaml#L30-L33

If anything, the trace agent should use the existing one instead cc @palazzem

PR #2621, opened 2018-06-11
Add notes on include/exclude for v5/6

ToddRizley:
### What does this PR do?
Add notes about hot to include/exclude containers from live collection and nuances of agent 5 set up

### Motivation
Support ticket confusion

### DIFF
--- a/content/graphing/infrastructure/livecontainers.md
+++ b/content/graphing/infrastructure/livecontainers.md
@@ -28,18 +28,23 @@ After deploying the [Docker Agent][7], no other configuration is necessary.
 
 Note that for collecting Container information in the standard install rather than with the [Docker Agent][7], the `dd-agent` user needs to have permissions to access **docker.sock**.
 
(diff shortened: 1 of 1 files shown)


content/graphing/infrastructure/livecontainers.md
```diff
@@ -17,17 +17,35 @@ further_reading:
...
+Both arguments take an **image name** as value; regexp are also supported.
+
+For example, to exclude all debian images except containers with a name starting with *frontend*, add these two configuration lines in your `datadog.yaml`file:
+
+```
+ac_exclude: ["image:debian"]
+ac_include: ["name:frontend.*"]
+```
+
+**Note:** For Agent v5, instead of including the above in the `datadog.conf` main configuration file, You have to explicitly add a `datadog.yaml` file to `/etc/datadog-agent/datadog.yaml` as the Process Agent relies on configuration options set here. This configuration only excludes containers from real-time collection **not** from Autodiscovery.
```
jeremy-lq: Pushed a couple nit fixes, but I believe this section is incorrect. Can we verify?

ToddRizley: @jeremy-lq lines  31-46 are pulled straight from the agent's datadog.yaml. Line 47+ was confirmed by Burrito and tested on my end as the result of confusion as to set up for agent 5. it's all agent 5 specific.

PR #2650, opened 2018-06-18
skip service checks for now

davidejones:
### What does this PR do?
Skips service checks json files for now as hugo is having a hard time reading them

### Motivation
Errors in a build


davidejones:
Cancelling for now

PR #2665, opened 2018-06-21
Update parsing documentation with Keyvalue examples

NBParis:
### What does this PR do?
Add more examples on how to use the keyvalue parser

### Motivation
There was a lot of questions about the keyvalue parser and it needed to be improved with more examples.


NBParis:
It seems that the example with `<valueStr>` is not showing up in the preview. It is blank instead.
Any idea how to force it to show up (double quotes?)? https://cl.ly/2C0n1D3k3y2s

jeremy-lq:
@NBParis Pushed a commit to fix the angle bracket issue.

NBParis:
Works like a charm. Seems that those can be escaped not only in the Grok parsers but also in the documentation :)

PR #2674, opened 2018-06-25
Update Fluentd link to latest version

NBParis:
### What does this PR do?

Link was referencing an old version.

### DIFF
--- a/content/logs/log_collection/fluentd.md
+++ b/content/logs/log_collection/fluentd.md
@@ -10,24 +10,55 @@ As long as you can forward your FluentD logs over tcp/udp to a specific port, yo
 
 In order to get the best use out of your logs in Datadog, it is important to have the proper metadata associated with your logs (including hostname and source). By default, the hostname and timestamp should be properly remapped thanks to our default [remapping for reserved attributes][2]. 
 
-### Source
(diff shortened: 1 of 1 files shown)


content/logs/log_collection/fluentd.md
```diff
@@ -10,28 +10,60 @@ As long as you can forward your FluentD logs over tcp/udp to a specific port, yo
...
+  dd_sourcecategory '<SOURCE_CATEGORY>'
+
+</match>
 ```
 
-### Custom tags
+### Kubernetes and Docker tags
 
-[Host tags][5] are automatically set on your logs if there is a matching hostname in your [infrastructure list][4]. Use the `ddtags` attribute to add custom tags to your logs:
+Tags in Datadog are critical to be able to jump from one part of the product to the other. Having the right metadata associated to your logs is therefore important to jump from the container view or any container metrics to the most related logs.
```
cswatt: some grammar/syntax:
Datadog tags are critical to be able to jump from one part of the product to another. Having the right metadata associated with your logs is therefore important in jumping from a container view or any container metrics to the most related logs.

I also think there should be more explicit language around the "jumping" — is that referring to something in a UI view?

PR #2702, opened 2018-06-29
nest s3 under archives and update path of archives in s3

estib:
### What does this PR do?
2 changes:
1. nest the `AWS S3 Archives` page under a more general `Archives` category.
2. Update the directory path of the archive contents

### Motivation
For...
1. We do not want to suggest that the only kind of archiving that we are planning for is the s3 archiving. We want to suggest that more kinds of archives are to come. 
2. A heartfelt love of humankind and the distribution of correct information. 

### Preview link
https://docs-staging.datadoghq.com/estib/update-s3-archive-path/logs/s3/

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -672,10 +672,15 @@ languages:
           weight: 90
           parent: menu_references_logs
         - identifier: customnav_lognav_archives
-          name: "Archives on AWS S3"
+          name: "Archives"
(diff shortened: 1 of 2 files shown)


estib:
also, the beta is going from "private" to "public", so removed the reference to "limited" and kept "beta".

PR #2708, opened 2018-07-03
add glossary entries

phrawzty:
### What does this PR do?
Adds a number of glossary entries.

### Motivation
It should be easier to know what words mean.

### Additional Notes
This document is currently `draft`, meaning that it's not displayed publicly. Before we move to prod, we should decide whether we want to populate the other terms or just remove them (for now).


content/glossary.md
```diff
@@ -3,19 +3,52 @@ title: Glossary
 kind: Documentation
 draft: true
 ---
+
 **Agent**
 
+**API**
+
+Datadog provides an HTTP API in order to interact programmatically with the platform. Every feature, resource, mechanism, and bit of functionality can be accessed via the API. Interacting with the API manually is a great way to learn how Datadog works under the hood! In production however, you will use a purpose-built tool, library, or interface — such as the Datadog product itself — as an abstraction layer between you and the API.
```
technovangelist: and bit of functionality feels weird. can we leave it at '... , and mechanism can be accessed...'

technovangelist: don't think manually is needed. 'interacting with' i think implies manually

technovangelist: And maybe not ending with an !

content/glossary.md
```diff
@@ -3,19 +3,52 @@ title: Glossary
...
+
+See the [API documentation][1] for more information.
+
 **Check**
 
+Checks are small Python programs run periodically by the Agent. A Check performs an action and then gathers the result, which the Agent then stores and reports to the Datadog platform. These programs are freeform and are generally used to collect metrics from custom environments or applications.
```
technovangelist: periodically is a bit vague. what is the period, 10-15 seconds. Periodically could mean every few hours

technovangelist: 'A Check...' : does check need to be capitalized? if refering to the word being defined, maybe **bolding it is better**

content/glossary.md
```diff
@@ -3,19 +3,52 @@ title: Glossary
...
+
+Note that the word “check” — when not capitalized — refers to the generic act of taking a measurement.
```
technovangelist: or maybe this is a change that was made in the last few years that I never noticed.

content/glossary.md
```diff
@@ -3,19 +3,52 @@ title: Glossary
...
+
+**Client Library**
+
+There are a number of libraries in a variety of languages in order to help you instrument your applications with Datadog directly. Datadog provides official libraries for C#, Golang, Java, PHP, Python, and Ruby. Libraries in other languages have been contributed by the community, and are supported on a best-effort basis.
```
technovangelist: change 'in order to' to that or which

content/glossary.md
```diff
@@ -3,19 +3,52 @@ title: Glossary
...
+
+See the [Integrations documentation][5] for more information.
+
 **Monitor**
 
-**Multiple-organization accounts (multi-org accounts)**. Multi-org accounts allow a parent-organization to create multiple child-organizations. This is most often used with Managed Service Providers (MSPs) that have many customers. Each customer requires their own organization that is not accessible by other customers. All billing for multi-org accounts will be performed on the parent-organization and usage reports for the parent-organization will include usage information for all child-organizations.
+**Multiple-organization accounts (multi-org accounts)**
+
+Multi-org accounts allow a parent-organization to create multiple child-organizations. This is most often used with Managed Service Providers (MSPs) that have many customers. Each customer requires their own organization that is not accessible by other customers. All billing for multi-org accounts will be performed on the parent-organization and usage reports for the parent-organization will include usage information for all child-organizations.
```
technovangelist: should parent-org... and child-org... have the hyphen?

content/glossary.md
```diff
@@ -33,4 +66,16 @@ draft: true
 
 **Timeboard**
 
-**Trace**
\ No newline at end of file
+**Tracer**
```
technovangelist: have we changed the term to tracer from trace?

content/glossary.md
```diff
@@ -33,4 +66,16 @@ draft: true
...
+
+Datadog supports distributed tracing via the APM Integration. This allows you to instrument your existing code and see requests as they pass through all your systems. It works across different languages, databases, and RPC frameworks. Many languages are [supported natively][6]; however, developers can add their own language support by creating a custom shipper.
```
technovangelist: worth pointing to how to do the custom shipper???

content/glossary.md
phrawzty: > and bit of functionality feels weird. can we leave it at '... , and mechanism can be accessed...'

Agree

> don't think manually is needed. 'interacting with' i think implies manually

Disagree. There are lots of ways to interact with the API that aren't cURL.

> And maybe not ending with an !

Party-pooper. :P

content/glossary.md
phrawzty: We don't want to get into the weeds in terms of details. The actual period is an implementation detail that's covered in the in-depth documentation.

Check, as in the abstraction describing the thing the Agent does, is a proper name in the Datadog glossary. This "fact" causes weird writing problems and I think we should re-visit it in any case.

content/glossary.md
phrawzty: 👍

content/glossary.md
phrawzty: Idk - this was in the original file. I just added carriage returns.

content/glossary.md
phrawzty: See the dedicated documentation for more details.

content/glossary.md
jeremy-lq: We use them both. A Tracer submits traces.

content/glossary.md
technovangelist: then you are interacting with the framework/library that is interacting with the api.

PR #2711, opened 2018-07-04
prepare glossary for publish

phrawzty:
### What does this PR do?

This (re-)adds the Glossary section. It is not complete, but it is a good start and we can iterate from here.

### Motivation

Improve the developer experience by making things easier to find and learn about.

### Preview link

https://docs-staging.datadoghq.com/phrawzty/prep_glossary_for_publish/glossary

### Additional Notes

This does *not* add the Glossary to the navigation menu. We must decide where we'd like this nav item to be positioned. I'm open to discussion on this. 😄 
*Edit*: It already had a nav bar config, so I uncommented it.

### DIFF
--- a/content/glossary.md
+++ b/content/glossary.md
@@ -1,72 +1,51 @@
 ---
 title: Glossary
 kind: Documentation
-draft: true
 ---
 
-**Agent**
+#### Agent
 
-**API**
+The Agent is a small application that runs on hosts. It executes checks and manages the flow of information from the host to the Datadog platform. The Agent is open source and compiled binaries are made available for Windows, OS X, and many flavors of Linux.
+
+See the [Agent documentation][8] for more information.
+
+#### API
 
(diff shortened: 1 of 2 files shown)


l0k0ms:
plop, why are you using H4 for the titles? are your going to nest those words under an other H3/H2 section ?

phrawzty:
@l0k0ms No particular reason, I suppose. I like the way `####` is rendered on the docs site. What would you recommend?

l0k0ms:
whatever really :) I do like it this way too. my only concern here is that the table of content only pics up H2/H3 you can disable it with `disable_toc: true` as a param

content/glossary.md
```diff
@@ -1,72 +1,52 @@
...
 See the [DogStatsD documentation][4] for more information.
 
-**Downtime**
-
-**Embed**
-
-**Integration**
+#### Integration
 
 An Integration is a way to get data from your systems into Datadog. Integrations gather data from a given source, ensure that those data are classified correctly, and provide some other assets to assist with configuration and usage. Data sources can be pretty much anything from daemons on a server, to cloud services, to third-party APIs, and more.
```
irabinovitch: s/pretty much//

PR #2722, opened 2018-07-05
add Validate Monitor to API docs

cswatt:
### What does this PR do?
Adds a Validate Monitor section to the API docs. Also changes the ordering of the "Unmute a monitor" section so that it makes more sense — previously it was right after "Create a monitor," I changed it to be right after "Mute a monitor"

### Preview link
https://docs-staging.datadoghq.com/cecilia/api_validate_monitor/api/?lang=python#validate-a-monitor


l0k0ms:
@cswatt there is no code snippet for this endpoint?

PR #2733, opened 2018-07-06
Yassine.benazzou/simultaneous alert recovery

byassine93:
### What does this PR do?

This PR provides clarification when both the alert and alert recovery states are triggered.

### Motivation

[Trello Card](https://trello.com/c/RDGAAYnH/50-anomaly-monitor-error-on-edit-page-cant-reproduce-on-my-side)

### Preview link

[Monitor Types - Anomaly](https://docs.datadoghq.com/monitors/monitor_types/anomaly/)


content/monitors/monitor_types/anomaly.md
```diff
@@ -36,6 +36,8 @@ To create an anomaly detection monitor, navigate to the [New Monitor][1] page an
...
 
+Setting different windows for the alert and alert recovery periods might lead to an ambiguous state. The alert and alert recovery window sizes should be set such that both can not be satisfied at the same time. For example, setting an alert threshold at 50% for a 2-hour window (i.e., 1 hour has to be anomalies to trigger the alert) and the recovery threshold at 50% for a 10-minute window (i.e., 5 minutes have to be non-anomalous to recover) might result to trigger the alert and the alert recovery states simultaneously. If the last 5 minutes were not anomalous but the 1 hour before that was anomalous, both the alert and the alert recovery will be triggered.
```
cswatt: `Setting different windows for the alert and alert recovery periods might lead to an ambiguous state. The alert and alert recovery window sizes should be set such that both cannot be satisfied at the same time. For example, setting an alert threshold at 50% for a 2-hour window (i.e., 1 hour has to be anomalous to trigger the alert) and the recovery threshold at 50% for a 10-minute window (i.e., 5 minutes have to be non-anomalous to recover) might result in triggering the alert and the alert recovery states simultaneously. If the last 5 minutes are not anomalous but the 1 hour before that _was_ anomalous, both the alert and the alert recovery will be triggered.`

StephenKappel:
@byassine93 -- Let's make this a FAQ. That will allow us to directly link to it this text from the app. Also, the title of the question can make it clear that this is not allowed. (The current text doesn't make it clear that the ambiguous states are prevented by validation - only that some criteria lead to them.) The question could be something like, "Why am I prevented from saving a monitor with a message like ''alert and alert recovery criteria are such that the monitor can be simultaneously in alert and alert recovery states?"

PR #2734, opened 2018-07-06
Add GIF for View in context

NBParis:
### What does this PR do?
Update the documentation with the latest change on the view in context functionality 

### Motivation
Feature was updated.

### Preview link
https://docs-staging.datadoghq.com/nils/view-in-context-update/logs/explore/#view-in-context


phrawzty:
I'm concerned about usability here. Even on a large desktop screen the resolution makes this animation hard to follow. If I right-click and view image it's OK, but what are the chances of a user actually doing that?

PR #2748, opened 2018-07-11
asap function documentation

byassine93:
### What does this PR do?

Adds a description to the ASAP smoothing function created in prod (enabled only for Datadog for now).

### Motivation

Asap has been released to prod : [Trello Card](https://trello.com/c/w5aTMqr9/2845-implement-asap-in-dogweb)

### Preview link

[Functions](https://docs.datadoghq.com/graphing/miscellaneous/functions/)

### DIFF
--- a/content/graphing/miscellaneous/functions.md
+++ b/content/graphing/miscellaneous/functions.md
@@ -225,7 +225,7 @@ Delta value between points for a given metric
 
 ### asap()
 
-The `asap()` function applies the [ASAP algorithm][5] with the same window size for all the metrics in a graph. 
+The `asap()` function applies [ASAP algorithm][5] with the same window size for all the metrics in a graph. 
 
 ### ewma_3()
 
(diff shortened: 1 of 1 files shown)


content/graphing/miscellaneous/functions.md
```diff
@@ -223,6 +223,10 @@ Delta value between points for a given metric
 
 ## Smoothing
 
+### asap()
+
+The `asap()` function applies [ASAP algorithm][5] with the same window size for all the metrics in a graph.
```
StephenKappel: I think we should indicate what it is that ASAP does (even if we don't describe how it does it). For example,

> The `asap()` function applies a moving average to the time series, using the [ASAP algorithm][5] to choose the optimal span for the moving average. When applied to a group-by query,`asap()` will use the same span for all groups.

I recommend that we use "span" instead of "window size" for consistency with the documentation for ewma below.

PR #2769, opened 2018-07-13
Fixing variable name

l0k0ms:
The network specification as described in that linked section is not clear enough. This is the line in question:

```
%%host_network%%: specify the network name to use, when attached to several networks
```

The confusing part of that statement is that `host_network` seems like a reserved word and the reader may get the question of "how do you specify a network with a dynamic variable such as `%%host_network%%`. 

In truth, `host_network` is not a reserved word. It was meant to indicate that they can append the network name to `host_` to specify the network, for example:

`%%host_NETWORKNAME%%`
`%%host_bridge%%`
`%%host_mynewnetwork%%`
`%%host_myredisnetwork%%`

### DIFF
--- a/content/agent/autodiscovery.md
+++ b/content/agent/autodiscovery.md
@@ -320,11 +320,13 @@ The following template variables are currently handled by the Agent:
 
 - Container IP: `host`
   - `%%host%%`: autodetect the network (use `bridge` or, if only one network is attached, this one)
-  - `%%host_network%%`: specify the network name to use, when attached to several networks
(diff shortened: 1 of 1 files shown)


content/agent/autodiscovery.md
```diff
@@ -320,11 +320,13 @@ The following template variables are currently handled by the Agent:
 
 - Container IP: `host`
   - `%%host%%`: autodetect the network (use `bridge` or, if only one network is attached, this one)
-  - `%%host_network%%`: specify the network name to use, when attached to several networks
+  - `%%host_<NETWORK NAME>%%`: specify the network name to use, when attached to several networks (e.g `%%host_bridge%%`, `%%host_myredisnetwork%%`, ...)
```
cswatt: missing a fullstop after the g. `e.g.` not `e.g`

PR #2786, opened 2018-07-16
Add note about Pod annotation

dabcoder:
### What does this PR do?
Pod Annotation is not yet supported for logs.

### Motivation
To reflect what is described in app. 

### Preview link
https://github.com/DataDog/documentation/blob/davidb/pod-annotation-logs/content/logs/log_collection/docker.md

### DIFF
--- a/content/logs/log_collection/docker.md
+++ b/content/logs/log_collection/docker.md
@@ -152,7 +152,7 @@ LABEL "com.datadoghq.ad.instances"='[{"nginx_status_url": "http://%%host%%:%%por
 LABEL "com.datadoghq.ad.logs"='[{"source": "nginx", "service": "webapp"}]'
 ```
 
-Pod Annotation is not yet supported for logs. Check our [Autodiscovery Guide][8] for more information about Autodiscovery setup and examples.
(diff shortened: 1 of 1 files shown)


content/logs/log_collection/docker.md
```diff
@@ -152,7 +152,7 @@ LABEL "com.datadoghq.ad.instances"='[{"nginx_status_url": "http://%%host%%:%%por
 LABEL "com.datadoghq.ad.logs"='[{"source": "nginx", "service": "webapp"}]'
 ```
 
-Check our [Autodiscovery Guide][8] for more information about Autodiscovery setup and examples.
+Pod Annotation is not yet supported for logs. Check our [Autodiscovery Guide][8] for more information about Autodiscovery setup and examples.
```
NBParis: Not very keen on documenting all the feature that we do not have.

We have this earlier in the page `Since version 6.2 of the Datadog Agent, you can configure log collection directly in the container labels.` 

We could add something like:  `If you are running in Kubernetes and do not have access to container labels, pod annotation will soon be supported.`

dabcoder: I'd suggest `do not use container labels` instead of `do not have access to container labels` but otherwise 👍.

PR #2788, opened 2018-07-16
K8s correction

themsquared:
hostPort should be added here. Customers will copy this, try to enable APM, and fail because the host port isn't called out clearly. By giving this default, we'll avoid those challenges.

### What does this PR do?
Fixes the K8s template to correctly open ports.

### Motivation
Frustrated customers.

### DIFF
--- a/content/agent/basic_agent_usage/kubernetes.md
+++ b/content/agent/basic_agent_usage/kubernetes.md
@@ -58,9 +58,11 @@ spec:
         name: datadog-agent
         ports:
           - containerPort: 8125
+            hostPort: 8125
             name: dogstatsdport
             protocol: UDP
           - containerPort: 8126
+            hostPort: 8126
             name: traceport
             protocol: TCP
         env:


l0k0ms:
@themsquared you might also want to update the doc here: https://github.com/DataDog/datadog-agent/blob/master/Dockerfiles/manifests/agent.yaml

hkaj:
@themsquared I'm fine with documenting this, but it should be commented out by default. This has non-trivial security and networking implications. This should be disabled by default with a warning in comments about the gotchas around this. Namely:

- not all network plugins support host ports as of now (this is the cause of many support tickets actually). When using those, the workaround (it is not a solution) is to use `hostNetwork` (it also has networking and security implications)
- not all variants of k8s support host ports (openshift disables them by default for example) so this will fail
- this opens a port on their hosts, which may have different isolation/security rules as their pod network
- in the end they will still need to pass `status.hostIP` to their app from the downward API for this to work, so there's still some changes to make before it works.

Generally, feel free to ping @DataDog/container-integrations for any container-related documentation.

PR #2790, opened 2018-07-16
commenting default host port

l0k0ms:
Per comment here: https://github.com/DataDog/documentation/pull/2788


hkaj:
thanks @l0k0ms 
we could add these comments in the DogStatsD section: https://github.com/DataDog/documentation/pull/2790/files#diff-ebc1cf67e51dd668fcb5edd150cc68aaL162 , and add a part for trace as well.
Something like
```
To send custom metrics via dogstatsd from your application pods, uncomment the `hostPort` line. This will expose the dogstatsd port on each of your Kubernetes nodes. **Warning**: this opens a port on your hosts. Make sure your firewall covers that correctly. Another word of caution: some network plugins don't support hostPorts yet, so this will not work. The workaround in this case is to add `hostNetwork: true` in your agent pod spec. This will share the network namespace of your host with the Datadog agent. Again, make sure this is okay with your security policies.
```
WDYT?

PR #2795, opened 2018-07-17
Adding / modifying content for Distribution type

phrawzty:
### What does this PR do?

Adds more detail around the new Distribution type, specifically as regards the Developer Tools section.

### Motivation

Making the new Distribution type easier to find and understand.

### Preview link

* https://docs-staging.datadoghq.com/phrawzty/distribution_is_a_thing/developers/metrics
* https://docs-staging.datadoghq.com/phrawzty/distribution_is_a_thing/developers/metrics/distributions/
* https://docs-staging.datadoghq.com/phrawzty/distribution_is_a_thing/developers/dogstatsd/data_types/#distributions
* https://docs-staging.datadoghq.com/phrawzty/distribution_is_a_thing/graphing/metrics/distributions/

### DIFF
--- a/content/developers/metrics/distributions.md
+++ b/content/developers/metrics/distributions.md
@@ -2,9 +2,9 @@
 title: Distributions
 kind: documentation
 further_reading:
-- link: "developers/metrics"
+- link: "developers/graphing/metrics/distributions"
   tag: "Documentation"
-  text: Learn more about Metrics
+  text: Learn more about the dedicated UI for distribution metrics 
 - link: "developers/libraries"
   tag: "Documentation"
   text: Official and Community-contributed API and DogStatsD client libraries
--- a/content/developers/metrics/distributions.md
+++ b/content/developers/metrics/distributions.md
@@ -22,16 +22,6 @@ Because these aggregations are global in nature, by default, Datadog only applie
 
 ## Submission
 
-<!--
-### Agent check
-
-{{% table responsive="true" %}}
-| Method | Overview |
-| :----- | :----- |
-| self.distribution(...) | Track the statistical distribution of a set of values over one or more hosts. |
-{{% /table %}}
--->
-
 ### DogStatsD
 
 {{% table responsive="true" %}}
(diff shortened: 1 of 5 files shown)


content/developers/metrics/distributions.md
```diff
@@ -0,0 +1,42 @@
+---
+title: Distributions
+kind: documentation
+further_reading:
```
l0k0ms: We should add a link to https://docs.datadoghq.com/graphing/metrics/distributions/ on this page to encourage them to learn more about the dedicated UI for distribution metrics.

mikezvi:
Graphing > Metrics > distributions seems very uncoupled from the others, but also doesn't have a lot to do with graphing.

I think we should link to here from "Data Types and Tags" under "Tagging", if nothing else.  Otherwise the refactor is much cleaner.

mikezvi:
Now that we've split things out a bit, should we provide a toy example of why Distributions are so important, or is it clear in the phrasing `Because averages of percentile aggregations are not statistically valid...`

Extremely simple example I've been using:

Say host foo reports a metric with the values [1,1,1,2,2,2,3,3] and host bar reports the same metric with the values [1,1,2] during a flush interval.   

Here, the p50 (median) for foo is 2 and the p50 for bar is 1.  Aggregating by the average value would result in 1.5.  

In reality, the global p50 (median) is the median of the combined set, [1,1,1,1,1,2,2,2,2,3,3], which is 2.

content/graphing/metrics/distributions.md
```diff
@@ -30,6 +34,10 @@ With the [distribution UI][3], create additional aggregate timeseries by applyin
 
 {{< img src="graphing/metrics/distributions/distribution_metric.png" alt="Distribution metric" responsive="true" >}}
 
+When creating your own graph, distribution metrics automatically have additional space aggregations available in the UI:
```
cswatt: capitalize Distribution for consistency across the doc

phrawzty:
@mikezvi 
> I think we should link to here from "Data Types and Tags" under "Tagging", if nothing else. Otherwise the refactor is much cleaner.

Could you please provide full nav paths or URLs? I'm not sure what sections you're referring to here.

phrawzty:
@mikezvi 
> Extremely simple example I've been using: […]

That's a great example and I have 100% nicked it for the documentation. 😄

mikezvi:
Sorry!  There is "Tagging" documentation here: https://docs-staging.datadoghq.com/phrawzty/distribution_is_a_thing/developers/dogstatsd/data_types/#tagging

This page includes the Distribution type, but tagging for Distributions is not as simple as this model.  I don't think we need to put a ton of words explaining that right here, but we should have a simple "See advanced tagging operations for Distributions here: https://docs-staging.datadoghq.com/phrawzty/distribution_is_a_thing/graphing/metrics/distributions/" or something that reads better.

PR #2796, opened 2018-07-17
Updating desc for is_match param

l0k0ms:
### What does this PR do?
Updates description of the is_match param.

### Motivation

A customer reached out since they assumed that there was no need to append the value of a template variable like `tag_variable.name`, thinking that using `tag_variable.name` was only to check if a value was empty.

### Preview link

* https://docs-staging.datadoghq.com/gus/notifications/monitors/notifications/#is-match-or-is-exact-match


content/monitors/notifications.md
```diff
@@ -134,41 +134,57 @@ The @ notification inside the template variables follows the same rules.
...
 #### `{{is_match}}` or `{{is_exact_match}}`
 
-The `{{is_match}}` conditional allows you to match the triggering context to some string to display a different message.
-For example, you might want to notify your db team if a triggering host has `role:db` but notify your app team if the host has `role:app`.
+##### {{is_match}} 
 
-You can use any of the available tag variables in your condition. A match is made if the comparison string is anywhere in the resolved variable.
-
-The variable uses the following format:
+The `{{is_match}}` conditional allows you to match the triggering context to some string to display a different message. You can use any of the available tag variables in your condition. A match is made if the comparison string is anywhere in the resolved variable. The variable uses the following format:
```
cswatt: The `{{is_match}}` conditional allows you to match the triggering context to some string to display a different message. You can use any of the available tag variables in your conditional statement. A match is made if the comparison string is anywhere in the resolved variable. Variables use the following format:

content/monitors/notifications.md
```diff
@@ -134,41 +134,57 @@ The @ notification inside the template variables follows the same rules.
...
 ```
-{{#is_match "tag_variable" "comparison_string"}}
-  This shows if comparison_string is in tag_variable.
+{{#is_match "<TAG_VARIABLE>.name" "<COMPARISON_STRING>"}}
+  This shows if <COMPARISON_STRING> is in <TAG_VARIABLE>.
 {{/is_match}}
 ```
 
-Here is an example of how you can give a different message depending on the triggering context:
+For example, if you want to notify your db team if a triggering host has the `role:db` tags but notify your web team if the host has the `role:web` tags.
```
cswatt: For example, if you want to notify your database team if a triggering host has the `role:db` tags, but notify your web team if the host has the `role:web` tags, use the following:

content/monitors/notifications.md
phrawzty: "some string" is sort of weird, no? Possibly consider:
* a string
* any string
* a given string

PR #2806, opened 2018-07-18
Remove mention of https connections to proxies in Agent6 config

olivielpeau:
### What does this PR do?

Removes mention of https connections to proxies in Agent6 config

### Motivation

Although the Agent 6 (since 6.3.0) technically supports connecting to
proxies using HTTPS, all known Datadog customers that use proxies
connect to their proxies using HTTP.

Specifying `https://` also leads to a silent breaking behavior when
upgrading from v6 versions earlier than 6.3.0 and 6.3+ versions.

To avoid confusion, let's simply remove the mention of the `https://`
scheme for the proxy URL.

### DIFF
--- a/content/agent/proxy.md
+++ b/content/agent/proxy.md
@@ -35,8 +35,8 @@ Edit the `datadog.yaml` file with your proxy information. Use the `no_proxy` lis
 
 ```
 proxy:
-    http: http(s)://user:password@proxy_for_http:port
-    https: http(s)://user:password@proxy_for_https:port
+    http: http://user:password@proxy_for_http:port
+    https: http://user:password@proxy_for_https:port
 #   no_proxy:
 #     - host1
 #     - host2


hush-hush:
This would imply we don't support HTTPS proxy. I would rather leave the `(s)` and add a note about `upgrading from v6 versions earlier than 6.3.0 and 6.3+ versions`.

olivielpeau:
@hush-hush In my experience, there's a lot of confusion around proxies in general, and  HTTPS proxies in particular, so there's no benefit in leaving the suggestion of `https://` in our docs unless we explain what the difference actually is.

I forgot to update `config_template.yaml`, but I would happily also remove `(s)` from there.

olivielpeau:
https://github.com/DataDog/datadog-agent/pull/2028 reflects this change on the `datadog.yaml.example` file. We have a card in our sprint to improve the proxy docs with more details on possible options, which will allow mentioning the different supported values without risking confusion from customers

PR #2818, opened 2018-07-20
add glossary entry for Dashboards

phrawzty:
### What does this PR do?

add glossary entry for Dashboards

### Motivation

make the glossary more better

### Additional Notes


phrawzty:
> On a side note, do you think we should add links to corresponding doc pages when we speak about screenboard, timeboard or other concepts here?

Yes. :)

PR #2822, opened 2018-07-23
To fix a example mail address to correct address.

yamatoya:
### What does this PR do?

You should use 'example.com' which is described RFC 2606
https://tools.ietf.org/html/rfc2606

### Motivation

Recentlly, I tray to use your service and read documents.

### Preview link
* https://docs.datadoghq.com/developers/faq/what-do-notifications-do-in-datadog/
* https://docs.datadoghq.com/ja/faq/


l0k0ms:
Hello @yamatoya, 

Many thanks for this contribution 🙇!

Best,
Pierre

PR #2825, opened 2018-07-23
Clarify syntax for user group notifications to Slack

NoisomePossum:
Adds some clarification to the user group notification syntax since it wasn't very clear what "handle" and "ID" were referring too. 

Also adds a '!' inside the beginning of the brackets. The notification won't work without this.


NoisomePossum:
Those changes look great to me!

PR #2856, opened 2018-07-30
Editing example for log-rotate

l0k0ms:
### What does this PR do?
Fixes log rotate example

### Motivation
customer feedback

### Preview link

https://docs-staging.datadoghq.com/gus/logrotate/logs/faq/setting-file-permissions-for-rotating-logs/#setting-permissions-for-log-file-rotation


NBParis:
The example is correct but it seems strange to have nothing in the log rotate configuration (and only a post rotate).

Also it is possible to set permission using the reguar chmod (without using facl):

```
/tmp/output.log {
        size 1k
        create 644 group1 group2
        rotate 4
}
```

if appropriate feel free to add this.

PR #2857, opened 2018-07-30
Yaxis, fixing parameter values

l0k0ms:
## What does this PR do?

Filtering top and bottom can be done via string (percentage) or number, using:

```
- filter: "5%" // shorthand notation
- filter: {
 top: "5%"
}
```
- or -

```
- filter: 19000   // shorthand notation
- filter: {
 top: 19000
}
```

Notations `above` or `below` are just alias of `top` and `bottom`.

### DIFF
--- a/content/graphing/miscellaneous/graphingjson.md
+++ b/content/graphing/miscellaneous/graphingjson.md
@@ -284,12 +284,12 @@ Filter configuration allows you to automatically change y-axis bounds based on a
 
 For y-axis filtering, there are two ways to set up the configuration.
 
-To begin, there is a simple configuration where you specify an absolute value or a percentage and all values above the value or all values that sit within the top ##% are cutoff.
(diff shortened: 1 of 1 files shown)


content/graphing/miscellaneous/graphingjson.md
```diff
@@ -284,12 +284,12 @@ Filter configuration allows you to automatically change y-axis bounds based on a
 
 For y-axis filtering, there are two ways to set up the configuration.
 
-To begin, there is a simple configuration where you specify an absolute value or a percentage and all values above the value or all values that sit within the top ##% are cutoff.
+To begin, there is a simple configuration where you specify an absolute value or a percentage and all top values or all values that sit within the top `X%` are cutoff.
```
cswatt: To begin, there is a simple configuration where you specify an absolute value or a percentage. All top values or all values that sit within the top `X%` are cut off.

PR #2862, opened 2018-07-30
Add rollup video and caveats on rollup use in monitors

johnaxel:
### What does this PR do?
Adds a video about how the rollup function works, as well as some caveats about the function's use, as suggested by @chrismdd 

### Motivation
We created a cool new video to help users understand rollups, and we want them to see it!

### Preview link
https://docs-staging.datadoghq.com/john/add_rollups_video/graphing/miscellaneous/functions/#rollup

### DIFF
--- a/content/graphing/miscellaneous/functions.md
+++ b/content/graphing/miscellaneous/functions.md
@@ -341,9 +341,13 @@ The span value is the number of data points. So `median_9()` uses the last 9 dat
 
 ### .rollup()
 
+{{< vimeo 264998150 >}}
+
 Recommended for expert users only. Appending this function to the end of a query allows you to control the number of raw points rolled up into a single point plotted on the graph. The function takes two parameters, method and time: `.rollup(method,time)`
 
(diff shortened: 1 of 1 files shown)


benjaminwood:
@johnaxel or @chrismdd I know this is an ancient PR - but I'm trying to understand this bit which is still in the docs today:

> Note that rollups should usually be avoided in [monitor][7] queries, because of the possibility of misalignment between the rollup interval and the evaluation window of the monitor. The start and end of rollup intervals are aligned to UNIX time, not to the start and end of monitor queries, which means that a monitor may evaluate (and trigger on) an incomplete rollup interval containing only a small sample of data. To avoid this issue, users should delay the evaluation of the monitor by (at least) the length of the rollup interval.

I'm using a query which includes the rollup function to scale workloads in Kubernetes (with the [Keda Datadog scaler](https://keda.sh/docs/2.7/scalers/datadog)). I *think* I am up against what is described in the above paragraph, though it is Keda that is polling for updates and not a datadog monitor. Could this be the case? If so, could you further explain the last bit:

> To avoid this issue, users should delay the evaluation of the monitor by (at least) the length of the rollup interval.

Is there something I can do to ensure I'm not getting an incomplete/small sample of data?

goldsky4:
hey @benjaminwood rather than working through this older PR request would you mind opening a support ticket with our team so we can help answer any questions you have regarding rollup in monitors? You can open a ticket via our support platform here: https://help.datadoghq.com/

PR #2895, opened 2018-08-02
[Agent 6.4] Udpate the article for Agent log only collection

NBParis:
### What does this PR do?
Update the article to send only logs with a Datadog Agent.

### Motivation
In the version 6.4, there is a new clean way to remove all the metrics, traces and other payloads and this needed to be documented as the old way was not very clean.

### Preview link

* https://docs-staging.datadoghq.com/nils/agent-collect-logs-only/logs/faq/can-the-datadog-agent-be-used-to-send-only-logs/


olivielpeau:
late review, but LGTM too 👍

PR #2904, opened 2018-08-02
Add information for why DB services aren't visible in trace search list

burningion:
### What does this PR do?

Add clarification for why DB / Cache services aren't visible in the list of Trace Search services.

DB and Cache services don't create their own traces, they only generate traces from the upstream service applications that call them. Thus, adding them to the list of services doesn't really make sense, as they're not really services themselves (from an instrumentation viewpoint).

### Motivation

Customer confusion about "missing" services available in Trace Search.

### Preview link

https://docs-staging.datadoghq.com/kirk.kaiser/clarify-trace-search-other-services/

### DIFF
--- a/content/tracing/search.md
+++ b/content/tracing/search.md
@@ -33,9 +33,11 @@ further_reading:
 
 Use Trace Search & Analytics to filter application performance metrics and [APM events][8] by user-defined tags. It allows deep exploration of the web requests flowing through your service.
 
-Trace Search & Analytics can be enabled per APM service and per host. A service on which it is enabled exposes all its APM Events to Datadog.
+Trace Search & Analytics can be enabled per APM service and per host. A service on which it is enabled exposes all its APM Events to Datadog. 
 
-In this view you can:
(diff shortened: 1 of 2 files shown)


content/tracing/search.md
```diff
@@ -33,9 +33,11 @@ further_reading:
 
 Use Trace Search & Analytics to filter application performance metrics and [APM events][8] by user-defined tags. It allows deep exploration of the web requests flowing through your service.
 
-Trace Search & Analytics can be enabled per APM service and per host. A service on which it is enabled exposes all its APM Events to Datadog.
+Trace Search & Analytics can be enabled per APM service and per host. A service on which it is enabled exposes all its APM Events to Datadog. 
 
-In this view you can:
+Downstream services like databases and cache layers won't be in the list of available services (as they don't generate traces on their own), but their information will be picked up by the top level services that call them.
```
l0k0ms: r/`but their information will be picked up by the top level services that call them.`/but their information is picked up by the top level services that call them.`

PR #2905, opened 2018-08-02
big kubernetes refactor

cswatt:


l0k0ms:
This is a great PR sadly we do not have a 4 layers logic on the documentation. Maximum is 3


1. Agent
2. | - Basic Agent Usage
3. | - - Kubernetes
4. | - - - _index.md

I think we should take out Kubernetes from Basic Agent Usage and have it on the master navigation since its use-case is orthogonal to "classic" platform installations:

1. Agent
2. | - Basic Agent Usage
2. | - Kubernetes
3. | - - - _index.md

PR #2912, opened 2018-08-03
David.jones/optimize

davidejones:
### What does this PR do?
This PR is to help increase build times. It does this by

Using a new CI build image, this image will
- Only build the ja folder site in previews if a ja file changed in there (live deploy will always build it)
- Remove the translation artifact from being built when it isn't a scheduled run as it isn't needed.

Docs site changes that make hugo build the site faster
- partial cache the footer and sidenav based on language views
- remove the recursive navigation and switch to something more standard. Use js to open and highlight nav items.
- move api to a partial that can be cached for some minor improvement

These changes make the build step of deployment take around 4 mins instead of around 5 or 6

### Motivation
Slow build times

### Preview link
https://docs-staging.datadoghq.com/david.jones/optimize/

### Additional Notes
Please check the left navigation opens and highlights correctly it has been completely reworked. it should work in multiple languages.


.gitlab-ci.yml
```diff
@@ -18,7 +18,7 @@ before_script:
 
 # ================== templates ================== #
 .base_template: &base_template
-  image: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/corp-ci:master
+  image: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/corp-ci:david.jones_optimize
```
davidejones: @mstbbs I'm assuming this is ok to merge to master like this and then after live release we can update the master image

mstbbs: If the image exists, we cool.

davidejones:
Ok I've made some changes so it will fade in, what do you think @mstbbs 
https://docs-staging.datadoghq.com/david.jones/optimize/

davidejones:
Closing PR for now, we will revisit this

PR #2930, opened 2018-08-07
Add info about automatic rollup settings

johnaxel:
### What does this PR do?
Adds a description of the rollups that Datadog does invisibly in your graphs.

### Motivation
@chrismdd suggested we add it

### Preview link
https://docs-staging.datadoghq.com/john/extend_rollup_info/graphing/miscellaneous/functions/


chrismdd:
@johnaxel can we use "in-app metric type" instead of "metric type" to avoid confusion with the agent and dogstatsd metric types that have similar names? Otherwise LGTM.

johnaxel:
Good point @chrismdd — just added. Thanks for the comment!

PR #2935, opened 2018-08-08
Add beta notice for distribution metrics

jeremy-lq:
### What does this PR do?
Add beta notice for distributions

### Motivation
Clarity for customers


mikezvi:
thanks for jumping on this right away, all

PR #2963, opened 2018-08-13
add clarity to the agent commands and install process

phrawzty:
### What does this PR do?

The Agent commands and maintenance instructions, across many (all?) platforms, could use some love. This is sample modification - if we like it, I'll go ahead and apply it through the other platforms as well.

### Motivation

Correct information > Incorrect information

### Preview link

https://docs-staging.datadoghq.com/phrawzty/agent_instructions/agent/basic_agent_usage/amazonlinux/

### DIFF
--- a/content/agent/basic_agent_usage/amazonlinux.md
+++ b/content/agent/basic_agent_usage/amazonlinux.md
@@ -22,9 +22,7 @@ This page outlines the basic features of the Datadog Agent for Amazon Linux. If 
 
 ## Commands
 
-The service manager provided by the operating system is responsible for the Agent lifecycle: depending on which one is installed, `sudo service`, `sudo initctl` or `sudo systemctl` should be used to `start`, `stop` or `restart` the Agent, respectively.
-
-Other functionality is provided by the Agent binary directly. For example, `datadog-agent status` or `flare`.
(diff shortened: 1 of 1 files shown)


content/agent/basic_agent_usage/amazonlinux.md
```diff
@@ -37,10 +35,10 @@ Other functionality is provided by the Agent binary directly. For example, `data
 | `sudo service datadog-agent`                      | `sudo datadog-agent --help`                            | Display command usage              |
 | `sudo -u dd-agent -- dd-agent check <check_name>` | `sudo -u dd-agent -- datadog-agent check <check_name>` | Run a check                        |
 
-**NB**: If `service` is not available on your system, use:
+**Note**: If `service` is not available on your system, use:
 
-* `upstart`-based systems: `sudo initctl start/stop/restart datadog-agent`
-* `systemd`-based systems: `sudo systemctl start/stop/restart datadog-agent`
```
olivielpeau: I think the reason we mentioned the full `systemctl/initctl start/stop/restart datadog-agent` command here is that the order between `start/stop/restart` (the command) and `datadog-agent` (the service name) is different from `service`, i.e.:

* upstart/systemd: `command service_name`
* service: `service_name command`

olivielpeau: I agree that documenting how to use OS-specific service managers is a bit out-of-scope of these docs, however many customers (and DD engineers btw) don't know much about how to use them so I think it's still worth including these details.

PR #3034, opened 2018-08-27
David.jones/code tabs design

davidejones:
### What does this PR do?
This PR changes the design of the code tabs and improves support for various content within the tabs.

### Motivation
https://trello.com/c/u4cer90f/2562-docs-code-tab

### Preview link
A test page was created to verify how this works and looks but has been removed as part of the PR commits. See the card for more details on the back and forth on this item.

### Additional Notes
This should have no effect on prod until we actually start using the code tabs in content files.


src/js/datadog-docs.js
```diff
@@ -503,30 +503,81 @@ $(document).ready(function () {
...
-      tabPanel.find('.active').removeClass('active');
-      tab.addClass('active');
-      tabPane.addClass('active');
-      tabPane.addClass('show');
-    });
+        // clicking a tab open them all
+        $('.code-tabs .nav-tabs a').click(function(e){
+          e.preventDefault();
+
+          // prepare
```
mstbbs: @davidejones Thanks for commenting the code. Small thing (non-blocking), I find it easier to follow the comments if they are formatted as complete sentences. So, instead of `// prepare`, perhaps `// Calculate the top offset of the element so the value can be used to place it in front of the user after tab switch` (or whatever the actual calculation is for)

PR #3051, opened 2018-08-29
Removing stale FAQs

l0k0ms:
### What does this PR do?
Removes FAQs page that had less than 2 visitors over the last 6 months

### Motivation
Doc clean


phrawzty:
Is this something that we do? I mean, I get that the F in FAQ is "frequently", but is this information invalid?

l0k0ms:
The information should be still valid, but nobody reads it.

PR #3058, opened 2018-08-30
Update Docker and Container Docs

vatteh:
### What does this PR do?
This pr adds the DD_PROCESS_AGENT_CONTAINER_SOURCE option in the docker.md page and a `health` explanation in the livecontainers.md page

### Motivation
Request by @xvello for next agent release

### Preview link
https://docs-staging.datadoghq.com/victor.atteh/update_docker_container_docs/
`DD_PROCESS_AGENT_CONTAINER_SOURCE` edit - https://docs-staging.datadoghq.com/victor.atteh/update_docker_container_docs/agent/basic_agent_usage/docker/#misc
`health` edit - https://docs-staging.datadoghq.com/victor.atteh/update_docker_container_docs/graphing/infrastructure/livecontainers/#notes-known-issues

### DIFF
--- a/content/agent/basic_agent_usage/docker.md
+++ b/content/agent/basic_agent_usage/docker.md
@@ -93,8 +93,7 @@ Exclude containers from the metrics collection and Autodiscovery, if these are n
 
 #### Misc
 
-- `DD_PROCESS_AGENT_CONTAINER_SOURCE`: Overrides container source auto-detection to force a single source. e.g `"docker", "ecs_fargate", "kubelet"`
-
+- `DD_PROCESS_AGENT_CONTAINER_SOURCE`: Overrides container source auto-detection to force a single source. e.g `"docker"`, `"ecs_fargate"`, `"kubelet"`
 ### Configuration files
 
 You can also mount YAML configuration files in the `/conf.d` folder. They will automatically be copied to `/etc/datadog-agent/conf.d/` when the container starts. The same can be done for the `/checks.d` folder: any Python files in the `/checks.d` folder will automatically be copied to `/etc/datadog-agent/checks.d/` when the container starts.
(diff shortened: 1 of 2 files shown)


content/agent/basic_agent_usage/docker.md
```diff
@@ -91,6 +91,10 @@ Exclude containers from the metrics collection and Autodiscovery, if these are n
 
 **Note**: The `docker.containers.running`, `.stopped`, `.running.total` and `.stopped.total` metrics are not affected by these settings and always count all containers. This does not affect your per-container billing.
 
+#### Misc
+
+- `DD_PROCESS_AGENT_CONTAINER_SOURCE`: Overrides container source auto-detection to force a single source. e.g `"docker"`, `"ecs_fargate"`, `"kubelet"`
```
conorbranagan: I think we'd want to put these in the docs for each specific integration vs here. If I want to install for fargate, for example, I end up https://docs.datadoghq.com/integrations/ecs_fargate/ or https://www.datadoghq.com/blog/monitor-aws-fargate/. The latter actually has the config file where we'd want to override the container source (although I'm curious if this is auto-detected cc @xvello?)

For kubelet I'd put it in our kubernetes setup docs.

@l0k0ms Any other ideas here?

content/graphing/infrastructure/livecontainers.md
```diff
@@ -109,6 +109,8 @@ While actively working with the Containers page, metrics are collected at 2s res
 
 - RBAC settings can restrict Kubernetes metadata collection. Refer to the [RBAC entites for the Datadog Agent][2].
 
+- The `health` value is the containers' readiness probe, not it's liveness probe.
```
conorbranagan: In Kuberentes the `health` value is...

content/agent/basic_agent_usage/docker.md
conorbranagan: cc @mikezvi also for ideas

xvello: Autodetection works for docker, ecs_fargate and kubelet, so I think we should not bother users for the usual case.
The edge case we need this option for is to force kubelet if docker is also available (as we give priority on the runtime over the orchestrator for continuity). I think that should just be listed on a "if the autodiscovered source is not what you want, you can override" basis

PR #3091, opened 2018-09-04
rename http.py to check_http.py

lgh2:
### What does this PR do?
This PR renames the Python example http check script in the Agent Check documentation.

### Motivation
When I was initially setting up the Agent on a VM, I ran into the issue of `pip`  (which depends on the `http` module) not working because I was running it in a directory with the example `http.py` script in it. 

The `http.py` name conflicts with the http module in the Python standard library ([docs](https://docs.python.org/3/library/http.html)). Python's [module search](https://docs.python.org/3.7/tutorial/modules.html#the-module-search-path) first looks in the directory where the script is run, and imports the first script it finds with a matching name. 

### Preview link
https://docs-staging.datadoghq.com/laura.hampton/rename-http-example-script/developers/agent_checks/

### DIFF
--- a/content/developers/agent_checks.md
+++ b/content/developers/agent_checks.md
@@ -18,7 +18,7 @@ further_reading:
 
 ## Overview
 
-This page first looks at the `AgentCheck` interface, and then proposes a simple Agent check that collects timing metrics and status events from HTTP services.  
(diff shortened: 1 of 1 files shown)


phrawzty:
Good catch @lgh2! Thanks for cleaning up the trailing whitespace, too. 😄

PR #3098, opened 2018-09-06
removing redondant aliases

l0k0ms:
### What does this PR do?

Removes aliases that are an alias of the same page.

### Motivation
bug


phrawzty:
good catch!

PR #3101, opened 2018-09-06
keeping only private param

l0k0ms:
keeping only private param as an non-enabler of links in sitemap


jiboong:
@l0k0ms perfect, thanks

PR #3124, opened 2018-09-11
Removing api Titles + Placeholders update

l0k0ms:
### What does this PR do?

* Remove useless titles from the API section and transform them into bold text.
* Update placeholders in signatures to match contributing guidelines

### Motivation
The new search is sensitive to titles, in order to avoid weird results when it comes to the API pages, this PR removes useless titles and put instead bold text.


phrawzty:
This might do what we want, but I think the "better" option would be to alter the Docsearch config to match our structure: https://github.com/algolia/docsearch-configs/blob/master/configs/datadoghq.json#L69-L81

PR #3125, opened 2018-09-11
Adding service and name characters limitations

l0k0ms:
### What does this PR do?
Service name and name mustn't have more than 100 characters.

### Motivation

Customer feedback


LotharSee:
FYI there is also a limit of 5000 on the resource name.

PR #3174, opened 2018-09-17
Adding relevant FAQ as FR

l0k0ms:
### What does this PR do?
Adds a relevant entry as further reading section

### Motivation
Support feedback

### DIFF
--- a/content/graphing/functions/timeshift.md
+++ b/content/graphing/functions/timeshift.md
@@ -1,6 +1,10 @@
 ---
 title: Timeshift
 kind: documentation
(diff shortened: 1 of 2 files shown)


phrawzty:
Should `content/integrations/faq/list-of-api-source-attribute-value.md` also be in this?

l0k0ms:
@phrawzty yes. when you run the doc site locally it checks if this file has been updated on DD side. If it's the case then the build updates it hence it then counts as an edited file hence his presence with the PR.

PR #3178, opened 2018-09-18
Inaccurate description

ncracker:
http.status_code description doesn't makes sense. It's not displaying IP addresses

### What does this PR do?
A fix

### Motivation
Inaccuracy in the docs


l0k0ms:
Nicely spotted, many thanks for this contribution 🙇

PR #3198, opened 2018-09-19
Update PHP tracer docs with PHP versions

chuck:
Update PHP tracer docs with PHP versions, more info on library/framework versions, and slightly updated lists of libraries and frameworks.


chuck:
Note to @DataDog/baklava - please don't merge until someone from APM has signed off as well. Thanks!

jeremy-lq:
Thanks, @palazzem. With APM sign-off, we'll merge.

PR #3201, opened 2018-09-20
Agent send logs only with env var

NBParis:
### What does this PR do?
Update the documentation on how to use the agent to only send logs with the associate environment variables.

### Motivation
It was not clear for container user how to set those variables without going through the Datadog.yaml file

### DIFF
--- a/content/logs/faq/can-the-datadog-agent-be-used-to-send-only-logs.md
+++ b/content/logs/faq/can-the-datadog-agent-be-used-to-send-only-logs.md
@@ -42,7 +42,7 @@ The Agent sends metrics and other payloads to Datadog. To make sure that the Age
 
 {{% tab "Environment Variables" %}}
 
-Set the following environment variables to false:
+Set the following environment variables to `false`:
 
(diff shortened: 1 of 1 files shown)


content/logs/faq/can-the-datadog-agent-be-used-to-send-only-logs.md
```diff
@@ -35,6 +38,24 @@ The Agent sends metrics and other payloads to Datadog. To make sure that the Age
 3. Configure the Agent to collect logs as explained in the [log documentation page][2].
 4. [Restart the Agent][4]
 
+{{% /tab %}}
+
+{{% tab "Environment Variables" %}}
+
+Set the following environment variables to false:
```
cswatt: I would personally like to backtick `false`

NBParis: not sure what `backtick` means in that case.
Feel free to commit directly on the PR if you want :)

PR #3209, opened 2018-09-24
Fixing attribtue -> attribute typo.

dhermes:
Done via:

```
git grep -l attribtue | xargs sed -i s/attribtue/attribute/g
```

I originally discovered this while reading [1].

[1]: https://docs.datadoghq.com/logs/log_collection/python/


l0k0ms:
Many thanks for spotting this 🙇 ! 
I'm going to update the other repo with this fix.

PR #3225, opened 2018-09-26
Add log samples in monitor notification

NBParis:
### What does this PR do?
Add documentation on how to enable log samples in the monitor notification message

### Motivation
This was recently added and needed to be documented.


NBParis:
@cswatt can we revert this ? There is an issue with the image and all my blurring is gone.

PR #3268, opened 2018-10-02
Laura.hampton/subprocess checks

lgh2:
### What does this PR do?
Adds information on custom checks that run command line programs as subprocesses

### Motivation
Customer request for a check that runs a command line subprocess, and finding out that subprocess checks run with `subprocess` can affect the Agent: https://github.com/DataDog/datadog-agent/blob/master/docs/dev/checks/python/check_api.md#running-subprocesses 

It seemed like this information should be moved to the main documentation. 

### Preview link
I didn't get a preview link for this.

### DIFF
--- a/content/developers/agent_checks.md
+++ b/content/developers/agent_checks.md
@@ -463,16 +463,17 @@ The command and its arguments are passed to `get_subprocess_output()` in the for
 $ vgs -o vg_free
 
 ```
-should be passed to `get_subprocess_output()` like this:
+
+must be passed to `get_subprocess_output()` like this:
 
 ```python
 ["vgs", "-o", "vg_free"]
 
 ```
 
-When the command line program is run, the check will capture the same output as if it were run on the command line in the terminal. It is important to do string processing on the output and call `int()` or `float()` on the result, so that it returns a numerical type.
+When the command line program is run, the check captures the same output as if it were run on the command line in the terminal. It is important to do string processing on the output and call `int()` or `float()` on the result, so that it returns a numerical type.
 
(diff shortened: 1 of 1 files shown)


content/developers/agent_checks.md
```diff
@@ -442,6 +442,52 @@ class HTTPCheck(AgentCheck):
         })
 ```
 
+## Writing checks that run command line programs
+
+It's possible to create a custom check that runs a command line program and captures its output as a custom metric. For example, a check can run the `vgs` command to report information about volume groups. Since the Python code that runs the checks is embedded in the multi-threaded Go runtime, using the Python `subprocess` or `multithreading` modules from the standard library can cause the Agent to crash or processes to get stuck in a zombie state.
```
l0k0ms: `can cause the Agent to crash` or `causes the Agent to crash`?

content/developers/agent_checks.md
```diff
@@ -442,6 +442,52 @@ class HTTPCheck(AgentCheck):
...
+    out, err, retcode = get_subprocess_output(cmd, self.log, raise_on_empty_output=True)
+```
+
+The command and its arguments are passed to `get_subprocess_output()` in the form of a list, with the command and each of its arguments as a string within the list. For instance, a command that is entered at the command prompt like this:
+
+```
+$ vgs -o vg_free
+
+```
+should be passed to `get_subprocess_output()` like this:
```
l0k0ms: r/`should be passed`/`must be passed` per [RFC 2119] (https://github.com/DataDog/documentation/blob/master/CONTRIBUTING.md#rfc-2119)

PR #3274, opened 2018-10-03
Update PHP tracer link to our own library

chuck:
### What does this PR do?
Update the PHP tracing library data table entry to point to our own library

### Motivation
Noticed the outdated link

### DIFF
--- a/data/libraries.yaml
+++ b/data/libraries.yaml
@@ -240,8 +240,8 @@ Tracing:
       notes: OpenTracing API implementation in JavaScript for Node.js.
   - PHP:
     - name: dd-trace-php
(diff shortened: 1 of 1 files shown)


data/libraries.yaml
```diff
@@ -240,8 +240,8 @@ Tracing:
       notes: OpenTracing API implementation in JavaScript for Node.js.
   - PHP:
     - name: dd-trace-php
-      link: https://github.com/jcchavezs/dd-trace-php/
-      authors: José Carlos Chávez
+      link: https://github.com/DataDog/dd-trace-php/
```
lucaspimentel: I think you need to add `official: true` so it's marked as an official library.

PR #3432, opened 2018-11-06
separating out azure metrics

cswatt:
### What does this PR do?
Adds two static metrics pages for Azure for MySQL and Azure for Postgres.

### Motivation
These metrics are already listed in the main Azure integration, these two pages provide redundant information. This is happening because Azure is featuring us in a blog post and we want broken-out docs metrics pages for Azure's MySQL and Postgres solutions.


l0k0ms:
you need to update the Makefile accordingly https://github.com/DataDog/documentation/blob/cecilia/azure-metrics/Makefile#L60 otherwise a local build will delete your integrations doc pages

cswatt:
> you need to update the Makefile accordingly https://github.com/DataDog/documentation/blob/cecilia/azure-metrics/Makefile#L60 otherwise a local build will delete your integrations doc pages
> 
> Do we have logos for those integrations?

PR #3469, opened 2018-11-14
[monitors] fix stray default value text and spacing

StephenKappel:
What does this PR do?
Cleans up some text that was accidentally left over from somewhere else.

### Motivation
I was using the docs and came across this non-sensical clause


l0k0ms:
Many thanks for this fix 🙇

PR #3481, opened 2018-11-15
Remove outdated note about monitor time aggregators

chrismdd:
### What does this PR do?

It removes information that is out of date. I believe at one point we implicitly enabled "require a full window" for `on average` and `at all times`, but this is no longer the case. Users can now configure "require a full window" independently of the monitor's time aggregation method.


chrismdd:
Please see the second commit (2b150fb5ed26ae4af7ca5905853a19027dc0783b). The first one just removes trailing whitespace.

PR #3491, opened 2018-11-17
Update docs to use the preferred setting for agent port

realark:
Starting in 0.18.0 we want to encourage use of the DD_AGENT_PORT
setting which is standard across all language tracers.


l0k0ms:
@jeremy-lq @realark yes.

PR #3502, opened 2018-11-19
Clarify service and operation names in configuring Trace Search

burningion:
### What does this PR do?

Clarify the service and operation names used to enable trace search.

### Motivation

Default value was set to be the default operation name of a Java application.


burningion:
Waiting on a build before merging.

PR #3530, opened 2018-11-27
Update heroku with log collection instruction

NBParis:
### What does this PR do?
Add log collection instruction for Heroku.

### Motivation
Heroku will soon be released for the EU region and the documentation must be ready for it.

### Preview link
https://docs-staging.datadoghq.com/nils/heroku-logs/agent/basic_agent_usage/heroku/#heroku-log-collection

### Additional Notes
https://trello.com/c/bEbzWxhh/2242-logs-add-a-heroku-integration


content/agent/basic_agent_usage/heroku.md
```diff
@@ -63,6 +63,38 @@ Depending on your use case, you may want to set your hostname so that hosts are
 - The Datadog Agent configuration files are at `/app/.apt/etc/datadog-agent`
 - The Datadog Agent logs are at `/app/.apt/var/log/datadog`
 
+## Heroku Log Collection
+
+** This integration is currently available for the EU region only**
```
cswatt: we generally want to avoid use of "currently" and lowkey promising that something will eventually happen. what do you think @l0k0ms

NBParis: In this case, something will happen as it will be released in the US region as well :)

jyee:
@NBParis We're having them manually set a hostname that will be applied to all the logs? Is this because we can't get a hostname from logs?

Ideally we'd allow them to set host as a combo of appname.dynoname (appname they can set in the log drain url, dyno name we can get from the logs). That way it would match up to the metrics and traces they're sending via the buildpack.

NBParis:
> @NBParis We're having them manually set a hostname that will be applied to all the logs? Is this because we can't get a hostname from logs?
> 
> Ideally we'd allow them to set host as a combo of appname.dynoname (appname they can set in the log drain url, dyno name we can get from the logs). That way it would match up to the metrics and traces they're sending via the buildpack.

That's definitely the goal of this. To make sure that we get the same hostname on the metrics, traces and logs.
Maybe I missed something but it seemed that we recommended to set a specific hostname to avoid a huge granularity of hostname ([explained here](https://docs.datadoghq.com/agent/basic_agent_usage/heroku/#hostname)).

If I miss something I'm more than happy to have a chat on how this can be better setup.

jyee:
@NBParis So we actually recommend to *not* set a hostname, but to enable DD_DYNO_HOST (heroku buildpack specific env var) which will dynamically set the host to appname.dynoname.  See https://github.com/DataDog/heroku-buildpack-datadog/blob/master/extra/datadog.sh#L88

Is there a way to combine the host passed from the ingestion URL params with the dyno name in the log line? I'm not too familiar with pipelines and whether that's possible.

NBParis:
@jyee so unfortunately not, the hostname cannot be remapped in the pipeline, it is available for JSON logs only and is done before any other processing.

Example of heroku log:

```
at=info method=POST path="/event" host=datadog-slack-channel-manager.herokuapp.com request_id=cb038d6f-12e3-417b-911a-1e913540837d fwd="100.26.222.54" dyno=web.1 connect=1ms service=44ms status=200 bytes=195 protocol=http
```

The dyno name is actually included in the log itself and I assume that it will not change that much so that's why we would recommend to set it the hostname in the url.
Do you think that will be a big problem?

jyee:
@NBParis I think that could work. The biggest issue we faced before when users set hostname explicitly was huge cardinality. I think a good place we can meet in the middle is to tell users to set the hostname to their Heroku app name in both the buildpack and for logs. We already add the dyno name to the metrics and trace tags, so everything should align with logs.

content/agent/basic_agent_usage/heroku.md
```diff
@@ -63,6 +63,66 @@ Depending on your use case, you may want to set your hostname so that hosts are
...
+```
+heroku drains:add https://http-intake.logs.datadoghq.eu/v1/input/<API_KEY>?ddsource=heroku&service=<SERVICE>&host=<HOST> -a <APPLICATION_NAME>
+```
+
+{{% /tab %}}
+{{< /tabs >}}
+
+* Replace `<API_KEY>` with your [Datadog API Key][2]
+* Replace `<APPLICATION_NAME>` and `<SERVICE>` with your application name¨
+* Replace `<HOST>` by the wanted hostname. Refer to the [host section](#hostname) to make sure that this is the same value set for the metrics as well.
```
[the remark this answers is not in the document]
NBParis: @irabinovitch made some changes. let me know if that works for you now.

content/agent/basic_agent_usage/heroku.md
```diff
@@ -100,8 +100,8 @@ heroku drains:add https://http-intake.logs.datadoghq.eu/v1/input/<API_KEY>?ddsou
 {{< /tabs >}}
 
 * Replace `<API_KEY>` with your [Datadog API Key][2]
-* Replace `<APPLICATION_NAME>` and `<SERVICE>` with your application name¨
-* Replace `<HOST>` by the wanted hostname. Refer to the [host section](#hostname) to make sure that this is the same value set for the metrics as well.
+* Replace `<APPLICATION_NAME>` and `<SERVICE>` with your application name
+* Replace `<HOST>` by the wanted hostname. Refer to the [host section](#hostname) to see how it is set for metrics. Note that it is not yet possible to set the dyno name as the hostname for logs.
```
irabinovitch: s/by/with/

irabinovitch:
Could use a grammar pass, but I'm going to remove my blocking review so others can help with grammar.

irabinovitch:
Could use a grammar pass, but I'm going to remove my blocking review so others can help with grammar.

content/agent/basic_agent_usage/heroku.md
```diff
@@ -101,7 +101,7 @@ heroku drains:add https://http-intake.logs.datadoghq.eu/v1/input/<API_KEY>?ddsou
 
 * Replace `<API_KEY>` with your [Datadog API Key][2]
 * Replace `<APPLICATION_NAME>` and `<SERVICE>` with your application name
-* Replace `<HOST>` by the wanted hostname. Refer to the [host section](#hostname) to see how it is set for metrics. Note that it is not yet possible to set the dyno name as the hostname for logs.
+* Replace `<HOST>` by the wanted hostname. Refer to the [host section](#hostname) to see how it is set for metrics. Note that it is not yet possible to dynamically set the dyno name as the hostname for logs.
```
irabinovitch: " Replace `<HOST>` with the desired hostname. Note that per [host section](#hostname) metrics and traces will set the default host name to dyno name. It is not yet possible to dynamically set the dyno name as the hostname for logs.  In the mean time to correlate between metrics, traces and logs the ____ tag can be used.

NBParis: ___ can be filled with `dyno` and `dynotype`

PR #3541, opened 2018-11-28
clarify port usage

rsheyd:
### What does this PR do?
Clarify network port usage by agent.

### Motivation
Questions by solutions engineers and customers asking for clarification, as well as some contradictions about why we list inbound ports even though we say all traffic is outbound.


rsheyd:
Trello card with some background:
https://trello.com/c/Gz2fWc3x/2251-clarify-network-doc-page

PR #3594, opened 2018-12-10
Fix broken link

svmhdvn:
### What does this PR do?
Fixes broken link in troubleshooting page.

### Motivation
The link is broken due to a constraint with the Hugo markdown renderer. Each shortcode gets its own markdown context, so the link is invisible in that context.

### Preview link
http://docs-staging.datadoghq.com/svmhdvn/typo/agent/troubleshooting

### DIFF
--- a/content/agent/troubleshooting.md
+++ b/content/agent/troubleshooting.md
@@ -148,6 +148,7 @@ In the commands below, replace `<CASE_ID>` with your Datadog support case ID, if
 | Windows      | Consult the dedicated [Windows documentation][10]       |
 | Heroku       | Consult the dedicated [Heroku documentation][13]        |
 
+[8]: /agent/#using-the-gui
 [10]: /agent/basic_agent_usage/windows/#agent-v6
 [13]: /agent/basic_agent_usage/heroku/#send-a-flare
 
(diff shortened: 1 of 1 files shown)


content/agent/troubleshooting.md
```diff
@@ -306,7 +307,6 @@ sudo journalctl -u dd-agent.service
 [5]: /help
 [6]: /agent/faq/agent-commands/
 [7]: https://github.com/DataDog/dd-agent/blob/master/utils/flare.py
-[8]: /agent/#using-the-gui
```
ruthnaebeck: I'd like to keep this link at the bottom too. It helps when re-using links to be able to see the complete list in one place.

svmhdvn: Wouldn't this lead to unnecessary duplication? Especially because having the link at the bottom could cause writers to incorrectly think that each hugo shortcode context has a view of the global page's markdown context

l0k0ms: I agree with @svmhdvn, we shouldn't duplicate links if possible. Not only this would impact our translation costs (since it's priced per word). But also it would confuse writers and could create corner cases.  The real issue here is that shortcodes don't support bottom links. idk if it's possible to fix without doing a pre-build script or smtg.

ruthnaebeck: Ok. It does make it harder to keep track of the numbers that have been used. I don't like the idea of re-using numbers at the bottom of the document that are already used in the tabs.

svmhdvn: The automatic link formatting script I will be uploading will take care of numbering for you, as long as each link you reference has an associated link definition. You don't have to manually keep track yourself.

PR #3614, opened 2018-12-13
Update ServiceNow documentation

rpelliard:
### What does this PR do?

Update the ServiceNow documentation to provide instructions for the new installation workflow that is using update sets.

Also moves the `servicenow.md` file to this repository (from dogweb `integrations/servicenow/README.md`). Please let me know if this is not a good idea !

### Motivation

This installation workflow is complex and not yet documented.

### Preview link

https://docs-staging.datadoghq.com/ronan/update-servicenow-docs/integrations/servicenow


rpelliard:
I can't see the ServiceNow tile on the https://docs.datadoghq.com/integrations/ page, is this normal ? I'm pretty sure it was here before, could it be related to me moving the file from dogweb?

jeremy-lq:
@rpelliard The /integration relies on proper metadata and images being present. We'll get it updated.

PR #3630, opened 2018-12-16
Add C++ tracing documentation

willgittoes-dd:
### What does this PR do?
Adds C++ APM tracing documents.

### Motivation
The C++ tracer is now released GA, so users will want documentation.

### Preview link
https://docs-staging.datadoghq.com/will.gittoes/cpp-apm/tracing/setup/cpp
https://docs-staging.datadoghq.com/will.gittoes/cpp-apm/tracing/advanced_usage/?tab=c


content/tracing/advanced_usage/_index.md
```diff
@@ -277,6 +277,20 @@ if (($span = GlobalTracer::get()->getActiveSpan()) !== null){
 
 **Note**: `GlobalTracer::get()->getActiveSpan()` returns `null` if there is no active span.
 
+{{% /tab %}}
+{{% tab "C++" %}}
+
+Add tags directly to a Span object by calling Span.SetTag(). For example:
```
cswatt:
```suggestion
Add tags directly to a span object by calling `SetTag()`. For example:
```

I want to lowercase "span". Also, is it `Span.SetTag()` here or just `SetTag()`? (forgive my lack of familiarity with c++)

content/tracing/advanced_usage/_index.md
```diff
@@ -515,6 +529,24 @@ $span->finish();
 
 [php compatibility]: /tracing/setup/php/#compatibility
 
+{{% /tab %}}
+{{% tab "C++" %}}
+
+To manually instrument your code, install the tracer as in the Setup examples, and then use the tracer object to create Spans.
```
cswatt:
```suggestion
To manually instrument your code, install the tracer as in the setup examples, and then use the tracer object to create spans.
```

content/tracing/advanced_usage/_index.md
```diff
@@ -830,6 +862,11 @@ public void ConfigureServices(IServiceCollection services)
 
 [dotnet opentracing]: https://www.nuget.org/packages/Datadog.Trace.OpenTracing/
 
+{{% /tab %}}
+{{% tab "C++" %}}
+
+The Datadog C++ tracer currently can only be used through the OpenTracing API. The usage instructions in this document all describe generic OpenTracing functionality.
```
cswatt:
```suggestion
The Datadog C++ tracer can only be used through the OpenTracing API. The usage instructions in this document all describe generic OpenTracing functionality.
```

We don't like to imply promises of future functionality, so we try to avoid words like "currently"

content/tracing/advanced_usage/_index.md
```diff
@@ -1018,6 +1055,47 @@ Coming Soon. Reach out to [the Datadog support team][contact support] to be part
 
 [contact support]: https://docs.datadoghq.com/help
 
+{{% /tab %}}
+{{% tab "C++" %}}
+
+Distributed tracing can be accomplished by [using the Inject and Extract methods on the tracer](https://github.com/opentracing/opentracing-cpp/#inject-span-context-into-a-textmapwriter), which accept [generic `Reader` and `Writer` types](https://github.com/opentracing/opentracing-cpp/blob/master/include/opentracing/propagation.h). Priority sampling (enabled by default) should be on to ensure uniform delivery of spans.
```
cswatt:
```suggestion
Distributed tracing can be accomplished by [using the `inject` and `extract` methods on the tracer](https://github.com/opentracing/opentracing-cpp/#inject-span-context-into-a-textmapwriter), which accept [generic `reader` and `writer` types](https://github.com/opentracing/opentracing-cpp/blob/master/include/opentracing/propagation.h). Priority sampling (enabled by default) should be on to ensure uniform delivery of spans.
```

content/tracing/advanced_usage/_index.md
```diff
@@ -1210,6 +1288,19 @@ Coming Soon. Reach out to [the Datadog support team][contact support] to be part
 
 [contact support]: https://docs.datadoghq.com/help
 
+{{% /tab %}}
+{{% tab "C++" %}}
+
+Priority sampling is enabled by default, and can be disabled in the TracerOptions. You can mark a span to be kept or discarded by setting the tag `sampling.priority`. A value of `0` means reject/don't sample and any value greater than 0 means keep/sample.
```
cswatt:
```suggestion
Priority sampling is enabled by default, and can be disabled in the TracerOptions. You can mark a span to be kept or discarded by setting the tag `sampling.priority`. A value of `0` means reject/don't sample. Any value greater than 0 means keep/sample.
```

content/tracing/advanced_usage/_index.md
```diff
@@ -1324,6 +1415,13 @@ Coming Soon. Reach out to [the Datadog support team][contact support] to be part
 
 [contact support]: https://docs.datadoghq.com/help
 
+{{% /tab %}}
+{{% tab "C++" %}}
+
+Coming Soon. Reach out to [the Datadog support team][contact support] if you're interested in this feature.
```
cswatt: Do we need this? Can we just not have the tab?

content/tracing/setup/cpp.md
```diff
@@ -0,0 +1,161 @@
...
+
+<div class="alert alert-warning">
+The APM tracer for C++ applications is in Open Public Beta.
+</div>
+
+## Getting Started
+
+To begin tracing applications written in any language, first [install and configure the Datadog Agent](https://docs.datadoghq.com/tracing/setup).
+
+You will need to compile against [OpenTracing-cpp](https://github.com/opentracing/opentracing-cpp).
```
cswatt:
```suggestion
Compile against [OpenTracing-cpp](https://github.com/opentracing/opentracing-cpp).
```

content/tracing/setup/cpp.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Compatibility
+
+dd-opentracing-cpp needs C++14 to build, but if you use [dynamic loading](#dynamic-loading) then you are instead only limited by OpenTracing's requirement for [C++11 or later](https://github.com/opentracing/opentracing-cpp/#cc98).
```
cswatt:
```suggestion
`dd-opentracing-cpp` requires C++14 to build, but if you use [dynamic loading](#dynamic-loading) then you are instead only limited by OpenTracing's requirement for [C++11 or later](https://github.com/opentracing/opentracing-cpp/#cc98).
```

content/tracing/setup/cpp.md
```diff
@@ -0,0 +1,163 @@
...
+
+Supported platforms are: Linux & Mac. If you need Windows support, please let us know.
```
cswatt:
```suggestion
Supported platforms include Linux and Mac. If you need Windows support, [contact Datadog support][1].
```

content/tracing/setup/cpp.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Installation
+
+Datadog tracing can be enabled in one of two ways:
+
+* Compile against dd-opentracing-cpp, where the Datadog lib is compiled in and configured in code
```
cswatt:
```suggestion
* Compile against `dd-opentracing-cpp`, where the Datadog lib is compiled and configured in code
```

content/tracing/setup/cpp.md
```diff
@@ -0,0 +1,163 @@
...
+* Dynamic loading, where the Datadog OpenTracing library is loaded at run-time and configured via JSON
```
cswatt:
```suggestion
* Dynamic loading, where the Datadog OpenTracing library is loaded at runtime and configured via JSON
```

content/tracing/setup/cpp.md
```diff
@@ -0,0 +1,163 @@
...
+    auto span_b = tracer->StartSpan("B", {opentracing::ChildOf(&span_a->context())});
+    span_b->SetTag("tag", "value");
+  }
+
+  tracer->Close();
+  return 0;
+}
+```
+
+Just link against libdd_opentracing and libopentracing (making sure that they are both in your LD_LIBRARY_PATH):
```
cswatt:
```suggestion
Link against `libdd_opentracing` and `libopentracing`, making sure that they are both in your `LD_LIBRARY_PATH`:
```

content/tracing/setup/cpp.md
```diff
@@ -0,0 +1,163 @@
...
+cd opentracing-cpp-1.5.0/.build
+cmake ..
+make
+make install
+# Install dd-opentracing-cpp shared plugin.
+wget https://github.com/DataDog/dd-opentracing-cpp/releases/download/v0.3.7/linux-amd64-libdd_opentracing_plugin.so.gz
+gunzip linux-amd64-libdd_opentracing_plugin.so.gz -c > /usr/local/lib/libdd_opentracing_plugin.so
+```
+
+Include `<opentracing/dynamic_load.h>` and load the tracer from libdd_opentracing_plugin.so:
```
cswatt:
```suggestion
Include `<opentracing/dynamic_load.h>` and load the tracer from `libdd_opentracing_plugin.so`:
```

content/tracing/setup/cpp.md
```diff
@@ -0,0 +1,163 @@
...
+    auto span_b = tracer->StartSpan("B", {opentracing::ChildOf(&span_a->context())});
+    span_b->SetTag("tag", "value");
+  }
+
+  tracer->Close();
+  return 0;
+}
+```
+
+Just link against libopentracing (making sure that libopentracing.so is in your LD_LIBRARY_PATH):
```
cswatt:
```suggestion
Just link against `libopentracing`, making sure that `libopentracing.so` is in your `LD_LIBRARY_PATH`:
```

content/tracing/advanced_usage/_index.md
willgittoes-dd: So here I use a capitalised Span because I'm referring to a code object. I suppose in that case I should say `Span` or span (but not Span or `span`). 

I think given how many times I have to say "span", we don't need to make it an inline code block and we should go with "span".

For SetTag, either `SetTag()` or `Span::SetTag` is fine, the latter is probably better.

I've made both these changes in this file, but I'm happy to change it further if you have a different suggestion :)

content/tracing/advanced_usage/_index.md
willgittoes-dd: The methods are capitalised. So `Inject` and `Extract` are correct. Same thing for Reader and Writer.

content/tracing/advanced_usage/_index.md
willgittoes-dd: Cargo-culted this, since it seems very common on this page. Removing the tab entirely causes no errors, it shows the 'Java' tab by default. Perhaps that's a bit confusing, since changing one tab changes all the tabs on a page. So once 'C++' is clicked once, the user will expect all blocks of text to be pertinent to C++. I think an empty tab would be better? 

I'll put up a change with an empty tab, but if you still prefer no tab/show Java instead that's also fine :)

willgittoes-dd: I changed my mind! I saw some more cases where the tab is just missing entirely (PHP often), so since that's already a pattern I'll use it too.

content/tracing/setup/cpp.md
willgittoes-dd: Maybe "compiled-in, and configured in code"? What I mean to say is that the library is included in compilation. Similar to saying that something is "baked in".

Entirely different phrasing if we want to work around it entirely:

 * Compile against `dd-opentracing-cpp`; including Datadog headers and linking against the library, and configuring the tracer in code

content/tracing/setup/cpp.md
willgittoes-dd: Committing your suggestion, but would you also like to remove the "Just"? :D

willgittoes-dd:
Thanks so much for your help! :)

config.yaml
```diff
@@ -1747,6 +1753,13 @@ languages:
           url: "tracing/languages/php/"
           weight: 170
           parent: customnav_tracingnav_languages
+        - identifier: customnav_tracingnav_cpp
```
palazzem: Let's be sure C++ is at the bottom on the Language / Setup list

content/tracing/languages/cpp.md
```diff
@@ -0,0 +1,166 @@
...
+- link: "tracing/visualization/"
+  tag: "Documentation"
+  text: "Explore your services, resources and traces"
+- link: "tracing/advanced_usage/?tab=c"
+  tag: "Advanced Usage"
+  text: "Advanced Usage"
+---
+
+<div class="alert alert-warning">
+The APM tracer for C++ applications is in Open Public Beta.
```
palazzem: Let's change this disclaimer. Now we are GA with the C++ Tracer so no reason to say anything. What we should add here (I don't know if an alert is good or not) is something like: "C++ does not provide integrations for OOTB instrumentation, but it's used by Proxy tracing such as Envoy and nginx."

The idea is to set the right expectations for users, because in the case of C++ we don't have integrations, made exception of Envoy and nginx.

l0k0ms: noted

PR #3661, opened 2018-12-20
Activating log and processes collection by default + EU support

l0k0ms:
### What does this PR do?
Activates Log and processes collection by default for AWS ECS task description + EU support

### Motivation
Better product, easier config.

### DIFF
--- a/static/json/datadog-agent-ecs-eu.json
+++ b/static/json/datadog-agent-ecs-eu.json
@@ -39,6 +39,14 @@
           "value": "<YOUR_API_KEY>"
         },
         {
+          "name": "DD_SITE",
+          "value": "datadoghq.eu"
+        },
+        {
+          "name": "DD_LOGS_CONFIG_LOGS_DD_URL",
+          "value": "agent-intake.logs.datadoghq.eu:443"
+        },
+        {
           "name": "DD_LOGS_ENABLED",
           "value": "true"
         },
@@ -84,7 +92,7 @@
     },
     {
       "host": {
-        "sourcePath": "/cgroup/"
+        "sourcePath": "/sys/fs/cgroup/"
       },
       "name": "cgroup"
     }
--- a/static/json/datadog-agent-ecs-eu.json
+++ b/static/json/datadog-agent-ecs-eu.json
@@ -57,10 +57,6 @@
         {
           "name": "DD_PROCESS_AGENT_ENABLED",
           "value": "true"
-        },
-        {
-          "name": "SD_BACKEND",
-          "value": "docker"
         }
       ]
     }
--- a/static/json/datadog-agent-ecs-eu.json
+++ b/static/json/datadog-agent-ecs-eu.json
@@ -48,7 +48,7 @@
         },
         {
           "name": "DD_LOGS_ENABLED",
-          "value": "true"
+          "value": "false"
         },
         {
           "name": "DD_LOGS_CONFIG_CONTAINER_COLLECT_ALL",
@@ -56,7 +56,7 @@
         },
         {
           "name": "DD_PROCESS_AGENT_ENABLED",
-          "value": "true"
+          "value": "false"
         }
       ]
     }
--- a/static/json/datadog-agent-ecs.json
+++ b/static/json/datadog-agent-ecs.json
@@ -40,7 +40,7 @@
         },
         {
           "name": "DD_LOGS_ENABLED",
-          "value": "true"
+          "value": "false"
         },
         {
           "name": "DD_LOGS_CONFIG_CONTAINER_COLLECT_ALL",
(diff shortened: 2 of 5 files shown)


static/json/datadog-agent-ecs.json
```diff
@@ -26,11 +36,23 @@
...
+        {
+          "name": "DD_LOGS_CONFIG_CONTAINER_COLLECT_ALL",
+          "value": "true"
+        },
+        {
+          "name": "DD_PROCESS_AGENT_ENABLED",
+          "value": "true"
+        },
+        {
+          "name": "SD_BACKEND",
```
xvello: this option is not used for agent6 anymore, we can remove it from all agent6 specs

static/json/datadog-agent-ecs-eu.json
```diff
@@ -0,0 +1,101 @@
+{
+  "containerDefinitions": [
+    {
+      "name": "datadog-agent",
+      "image": "datadog/agent:latest",
+      "cpu": 10,
+      "memory": 256,
```
xvello: too bad json does not support comments, I'd love to add a "customize according to observed usage". Is that something we could add to the doc that links to it?

static/json/datadog-agent-ecs1-eu.json
```diff
@@ -42,8 +55,12 @@
           "value": "true"
         },
         {
-      	  "name": "SD_BACKEND",
-	  "value": "docker"
+          "name": "DD_PROCESS_AGENT_ENABLED",
+          "value": "true"
+        },
+        {
+          "name": "SD_BACKEND",
```
hkaj: ditto

static/json/datadog-agent-ecs1.json
```diff
@@ -26,11 +36,23 @@
...
+        {
+          "name": "DD_LOGS_CONFIG_CONTAINER_COLLECT_ALL",
+          "value": "true"
+        },
+        {
+          "name": "DD_PROCESS_AGENT_ENABLED",
+          "value": "true"
+        },
+        {
+          "name": "SD_BACKEND",
```
hkaj: ditto

static/json/datadog-agent-ecs.json
l0k0ms: fixed :)

static/json/datadog-agent-ecs-eu.json
l0k0ms: we could indeed, I'm going to add it.

NBParis:
I think we should rename the PR as we do not enable them by default. We prepare all the configuration and let them enable everything with one single flag.

static/json/datadog-agent-ecs1-eu.json
```diff
@@ -33,17 +38,25 @@
...
+          "value": "datadoghq.eu"
+        },
+        {
+          "name": "DD_LOGS_CONFIG_LOGS_DD_URL",
+          "value": "agent-intake.logs.datadoghq.eu:443"
+        },
         {
           "name": "DD_LOGS_ENABLED",
-          "value": "true"
+          "value": "false"
```
NBParis: Well the `ecs-logs.json` should have log enabled.

static/json/datadog-agent-ecs1-eu.json
```diff
@@ -33,17 +38,25 @@
           "name": "DD_API_KEY",
           "value": "<YOUR_API_KEY>"
         },
+        {
+          "name": "DD_SITE",
+          "value": "datadoghq.eu"
+        },
+        {
+          "name": "DD_LOGS_CONFIG_LOGS_DD_URL",
+          "value": "agent-intake.logs.datadoghq.eu:443"
```
NBParis: I'm not seeing the us version anymore. Am I missing it somewhere else?

irabinovitch: Didn't 6.8 make this easier with a single SITE variable? or is that 6.9? Can we document based on that?  Theres far too many options here.

irabinovitch:
Rather than having 4  json files can we just have 1 with instructions on how to update them for your needs?

static/json/datadog-agent-ecs1-eu.json
l0k0ms: The dd_site variable works with 6.8 but doesn't support the log endpoint yet. This support will come with 6.9. Hence the need to have it set up independently for now.

l0k0ms:
@irabinovitch with 6.9 it will be easier since we could by default set dd_site to `datadoghq.com` and then just tell people to update the variable.

I think we should keep the Agent 5 configuration file since it's froozen

But I can remove the logic for amazon linux 1.0 and just tell people to update the volume path indeed.

What do you think?

PR #3670, opened 2018-12-21
[doc] Fix run command example in Docker APM doc

henrydoan:
### What does this PR do?
This section of documentation explains how to collect traces from a host (or any hosts) via a Docker Agent, stating to use option `-p 8126:8126/tcp`.  But the example `run` command doesn't include this option.  Fixing example to include this.

### Motivation
`run` command example doesn't match what's being explained in this documentation

### Additional Notes
Confirmed with @Simwar and @DylanLovesCoffee during Support Containers office hours that this example code block should be updated.

### DIFF
--- a/content/agent/docker/apm.md
+++ b/content/agent/docker/apm.md
@@ -21,17 +21,17 @@ Enable the [datadog-trace-agent][1] in the `datadog/agent` container by passing 
 
 ## Tracing from the host
 
-Tracing is available on port `8126/tcp` from *any host* by adding the option `-p 8126:8126/tcp` to the `docker run` command.
+Tracing is available on port `8126/tcp` from *your host only* by adding the option `-p 127.0.0.1:8126:8126/tcp` to the `docker run` command.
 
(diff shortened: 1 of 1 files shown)


content/agent/docker/apm.md
```diff
@@ -31,6 +31,7 @@ For example, the following command allows the Agent to receive traces from anywh
 docker run -d -v /var/run/docker.sock:/var/run/docker.sock:ro \
               -v /proc/:/host/proc/:ro \
               -v /sys/fs/cgroup/:/host/sys/fs/cgroup:ro \
+              -p 8126:8126/tcp \
```
palazzem: Correct. Probably we should update this section to `-p 127.0.0.1:8126:8126/tcp` saying:

"to receive traces from anywhere" (->) "to receive traces from your host only"

I think this should be enough for most of the users, while exposing the port anywhere is only for advanced users with specific use cases. Other than that, all good to me!

PR #3675, opened 2018-12-21
Java tracer MDC autoinjection docs.

realark:
Docs for Java Tracer's upcoming MDC autoinjection feature.

This can be merged once java tracer 0.21.0 is released.

### DIFF
--- a/content/tracing/advanced_usage/_index.md
+++ b/content/tracing/advanced_usage/_index.md
@@ -1408,15 +1408,16 @@ The purpose of this section is to explain how the correlation between traces and
 {{< tabs >}}
 {{% tab "Java" %}}
 
-The simplest way to inject trace identifiers is to leverage the MDC Frameworks ([Map Diagnostic Context][1]) to automatically add the trace and span ids into the logs. Datadog MDC keys may be injected automatically with a tracer integration, or manually injected with Datadog APIs.
+Leverage the MDC Frameworks ([Map Diagnostic Context][1]) to automatically add the trace and span ids into your logs. Datadog MDC keys may be injected automatically with a tracer integration, or manually injected with Datadog APIs.
 
 **How to enable Log Injection***
+
 1. Enable injection in the [Java Tracer's configuration][3]. Note: Currently only slf4j is supported for MDC autoinjection.
-1. Alternatively, Use Datadog `CorrelationIdentifier#getTraceId()`, and `CorrelationIdentifier#getSpanId()` APIs to inject identifiers at the beginning and end of each span to log (see examples below).
-2. Configure MDC to use the injected Keys
+2. Alternatively, Use Datadog `CorrelationIdentifier#getTraceId()`, and `CorrelationIdentifier#getSpanId()` APIs to inject identifiers at the beginning and end of each span to log (see examples below).
+3. Configure MDC to use the injected Keys:
   * `dd.trace_id` Active trace id during the log statement (or `0` if no trace)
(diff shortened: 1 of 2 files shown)


content/tracing/languages/java.md
```diff
@@ -224,6 +225,33 @@ Now add `@Trace` to methods to have them be traced when running with `dd-java-ag
 Datadog's Java Tracer provides support for 'in-process' JMX metrics collection. This is enabled with `jmxfetch.enabled` configuration parameter. Additional JMX metrics are configured using configuration files that are passed to `jmxfetch.metrics-configs`. Contents of those configuration files are equivalent to contents of the `conf` section for external jmxfetch. See [JMX Integration][8] for further details on configuration.
 By default, when JMX metrics collection is enabled it monitors JVM heap memory, thread count, and garbage collection. Use it in conjunction with APM for a broader view into your Java application's performance.
 
+## Logs Slf4j+MDC Autoinjection
```
NBParis: Those instructions are a duplicate from https://docs.datadoghq.com/tracing/advanced_usage/?tab=java#logging.
Having the same content in different place is rarely a good practice.

I would suggest to just add a link to the logging java section so if one day we need to make a change or an update, we don't need to change 3 different parts of the documentation.

So the documentation that actually should be updated is https://docs.datadoghq.com/tracing/advanced_usage/?tab=java#logging with the flag you have there.

realark: Sure I'm all for reduced duplication. I would like to keep the flag in the config section because customers expect to find all tracer configurations there. I added a link to the advanced logging section from there.

Also, I found that `advanced_usage` section confusing so I updated the wording and format a bit. Please let me know what you think.

realark:
@NBParis I believe all feedback has been addressed. Can you please review again? Thanks!

content/tracing/advanced_usage/_index.md
```diff
@@ -1408,12 +1408,18 @@ The purpose of this section is to explain how the correlation between traces and
...
+The simplest way to inject trace identifiers is to leverage the MDC Frameworks ([Map Diagnostic Context][1]) to automatically add the trace and span ids into the logs. Datadog MDC keys may be injected automatically with a tracer integration, or manually injected with Datadog APIs.
 
-- To inject those identifier into application logs using MDC frameworks, use the following method:
+**How to enable Log Injection***
+1. Enable injection in the [Java Tracer's configuration][3]. Note: Currently only slf4j is supported for MDC autoinjection.
+1. Alternatively, Use Datadog `CorrelationIdentifier#getTraceId()`, and `CorrelationIdentifier#getSpanId()` APIs to inject identifiers at the beginning and end of each span to log (see examples below).
+2. Configure MDC to use the injected Keys
+  * `dd.trace_id` Active trace id during the log statement (or `0` if no trace)
+  * `dd.span_id` Active trace id during the log statement (or `0` if no trace)
+  * Log format must contain a snippet of both keys. Example: `dd.trace_id=%X{dd.trace_id} dd.span_id=%X{dd.span_id}`
```
NBParis: I would remove this sentence (`Log format must....`) as it is covered in the Log documentation part and the described format is actually not the correct one to use.

NBParis:
@realark yep looks much better. One slight change I would recommend (a sentence to remove) and we should be good to go.

content/tracing/advanced_usage/_index.md
realark: Removed!

realark:
@NBParis Thanks! That bullet point has been removed.

PR #3758, opened 2019-01-10
Removing below references

l0k0ms:
This feature won't be available before the end of this week, let's remove it from the doc and merge it back when it's actually released


NBParis:
A new bug was found during the last testing phase so the release is postponed to next week.
So let's indeed remove the reference in the doc.

PR #3800, opened 2019-01-18
Update Java's MDC format defaults

realark:
Update MDC formatter to log `0` when no trace/span id is present.


NBParis:
@palazzem fine with me.

PR #3839, opened 2019-01-24
Adding below monitor desc

l0k0ms:
### What does this PR do?
Adds the below monitor desc now that's GA.

* https://docs-staging.datadoghq.com/gus/below-the-return/monitors/monitor_types/log/#no-data-alerts-and-below-conditions

### DIFF
--- a/content/monitors/monitor_types/log.md
+++ b/content/monitors/monitor_types/log.md
@@ -36,6 +36,8 @@ Log monitors alert when a specified type of log exceeds a user-defined threshold
 
 * Above
 * Above or equal
+* Below 
+* Below or equal
 
 Then configure the Alert and/or Warning threshold depending on the chosen condition:
     {{< img src="monitors/monitor_types/log/set_alert_conditions.png" alt="Set alert conditions" responsive="true" style="width:50%;" >}}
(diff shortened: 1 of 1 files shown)


content/monitors/monitor_types/log.md
```diff
@@ -71,6 +73,18 @@ For Slack and email notifications you can enable the top 10 of breaching values
 
  {{< img src="monitors/monitor_types/log/slack-log-multi-sample.png" alt="Slack notification example" responsive="true" style="width:50%;" >}}
 
+## No Data alerts and Below Conditions  
+
+To be notified if a specific set of logs are not received anymore, set the condition `below 1`. This notifies when no logs match the monitor query on the given timeframe. 
+
+However, note that when splitting the monitor by any dimension (tag or facet) and using a `below` condition, the alert is triggered **if and only if** there are logs for a given group, and the count is below the threshold—or if there are no logs at all for all of the groups.
```
ruthnaebeck:
```suggestion
However, note that when splitting the monitor by any dimension (tag or facet) and using a `below` condition, the alert is triggered **if and only if** there are logs for a given group, and the count is below the threshold—or if there are no logs for **all** of the groups.  
```

PR #3840, opened 2019-01-24
Updating list of api source value

l0k0ms:
Updates list of API source value


ruthnaebeck:
@l0k0ms - Isn't this file built programmatically?

l0k0ms:
@ruthnaebeck it is, but if you build the doc locally then if might mess up your commit history

PR #3855, opened 2019-01-25
Correct "from the query to the graph" example

xlucas:
### What does this PR do?

It changes the aggregator type from sum to average in the spatial aggregation example, as illustrated by the embedded graph.

### Preview link

https://docs-staging.datadoghq.com/xavier.lucas/graphing-functions-aggregator-example/graphing/functions/#proceed-to-space-aggregation


cswatt:
thanks @xlucas !

PR #3866, opened 2019-01-28
Changing region into site

l0k0ms:
### What does this PR do?
Moving forward we should use the terminology "Datadog site" when referring to Datadog EU.


ckelner:
@l0k0ms what's the reasoning behind "site" vs "region" -- is that to lessen confusion around our service being "multi-region aware" (as it _is not today_)? Should we be telling SEs to shy away from this language?

PR #3872, opened 2019-01-28
Update python log injection documentation

majorgreys:
### What does this PR do?

Update documentation for python logs injection to make it consistent with the depth of our other supported languages and clarify pitfalls developers might experience.


andrewsouthard1:
Would fix this one typo, other than that looks good!

https://github.com/DataDog/documentation/pull/3872/files#diff-5b7c8d2ab3d8623aad1508ed296564fdR1507 `auto-injected` needs to be `auto-injection`.

PR #3915, opened 2019-02-04
[Fastly] Enforcing status code type to int

l0k0ms:
### What does this PR do?

According to DD attribute naming convention the type for http.status_code is int. 

### Motivation

Customer feedback.


irabinovitch:
Anything we can do to make the formatting more obvious in the doc?  I missed it when looking the first 2 times.

phrawzty:
For reference, this is the change (it's hard to see):

image

l0k0ms:
I opened a PR in dogweb to make this more obvious in the doc.

PR #3945, opened 2019-02-07
Add Uptime percentage widget guide

ruthnaebeck:
### What does this PR do?
- Add guide index
- Add Uptime percentage widget guide from KB: https://help.datadoghq.com/hc/en-us/articles/360001219143-Uptime-percentage-widget-for-SLO-SLA-use

### Motivation
KB site migration

### Preview link
https://docs-staging.datadoghq.com/ruth/uptime-widget/graphing/guide/


ruthnaebeck:
Yep. I brought the same point up to the article's author @ncracker. He said:

> ...this will persist as an option.
It's an option that is agnostic of monitors unlike the new upcoming widgets

ckelner:
Hi @ruthnaebeck -- I had also sent this over to @meghanelizabeth earlier, trying to take the conversation out of DMs and remove myself from being a middle man, she shared with me:

> Meghan Jordan [2:35 PM]
> I mentioned this to cody, he's in a meeting we didnt get to finish convo
> we could rebrand to host uptime but in the further reading just recommend the monitor uptime widget docs
> ok let's rebrand it to host uptime instead
> not SLA/SLOs and add monitor uptime widget doc in further reading
> for synthetics, the monitor uptime widget docs can also be added in the further reading section

To me it sounds like changing two things:
- Adding links to the monitor uptime widget (SLA/SLI) and synthetics (endpoint monitoring) and denoting this is an alternative approach to those tools
- Renaming the guide (more below)

On renaming, it isn't clear to me what is being suggested @meghanelizabeth -- are we stating we want to name it "Host uptime percentage widget" or "Host based uptime percentage widget"? Or how would you title it (can you give the full working title you are thinking of)?

I ask because I think "host" might not be accurate or correct in this case (though "Host based" gets a little closer), because we are leveraging the HTTP Check to look at some endpoint (e.g. `google.com`) -- which is not a single host (we're doing a hacky job of endpoint monitoring, not host-specific monitoring) -- but the check itself is indeed "host based" (you need somewhere the agent is installed to be running the HTTP Check).

I'll defer to @ncracker on the removal of SLA/SLO and let him make an argument for it if he has one (perhaps further thoughts on this topic also)

I am now in favor of keeping both, but I think we should point out we have more "proper" (for lack of a better word) tools to achieve this sort of thing.

ruthnaebeck:
The main reasons to keep this article:
1) You don't need to have a monitor to use it
2) In the case of Synthetics, you cannot ping an internal site / endpoint. With this approach, the Agent can be installed on a host that has access to the internal site / endpoint. This allows you to get uptime metrics for that site / endpoint.

meghanelizabeth:
The title can be something like, "Visualizing host uptime", "Using the query value widget for host uptime", "Calculating host uptime" 

Within Further Reading for both the host uptime article and synthetics we could add a link to the monitor uptime widget docs.

ruthnaebeck:
Sounds good. I spoke with @ncracker and he is ok with removing the SLA / SLO from the article title.

PR #3986, opened 2019-02-13
make the forms default go to the search results page.

davidejones:
### What does this PR do?
This PR makes the sidebar search form go to the search results page by default if JS fails.

### Motivation
https://cl.ly/9a99ed910673
https://cl.ly/0d06dca2f23c

If a JS bug happens or intermittent connection happens causing a js library to fail to load (like docsearch). The html forms defaults/fallback is to go to the search results page anyway. At which point we would hope it will load on that attempt.

### Preview link
https://docs-staging.datadoghq.com/david.jones/search-update/

### DIFF
--- a/layouts/partials/sidenav/sidenav-m.html
+++ b/layouts/partials/sidenav/sidenav-m.html
@@ -19,9 +19,9 @@
     <div class="row sticky">
         <div class="col">
             <a href="{{ "" | absLangURL }}"><h4>{{ i18n "datadog_docs" }}</h4></a>
-            <form class="docssearch">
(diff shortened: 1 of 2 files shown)


davidejones:
Need to fix a link checker issue before merging this

davidejones:
I've put in a fix for the issue the link checker found.. however the task seems to be failing still because of another reason (it can't get a status code from some urls).

I'm going to wait a while here and try running it again, do not merge this until link checker task passes.

davidejones:
This passed link checker and can be merged 👍

PR #3988, opened 2019-02-13
removed click event causing console error

zbayoff:
### What does this PR do?
Fixes console log error happening when user click link on side nav. 

### Trello link
https://trello.com/c/Vw46SU1X/3016-error-when-using-nav-on-doc-site

### Preview link
https://docs-staging.datadoghq.com/zach/docs-nav-bug-fix/


davidejones:
@jiboong do you remember why you added this js to begin with? This commit https://github.com/DataDog/documentation/commit/340e9465661e6c6c8b356952728c1a53b41e5de3

davidejones:
@zbayoff We might need another solution, Won added that code you are removing just 6 months ago to handle enabling side menu items to link to in-page anchors (we think).

Just a guessing here but perhaps the code just needs to be updated to check its an anchor before running moveToAnchor on it. Most of the side nav links aren't anchors

l0k0ms:
@davidejones I'm going to remove the anchor logic from the side nav. I think we introduced it for the advanced instrumentation section. But now this page is growing to big and we are going to break it down into separate pages (during this sprint). I don't see other places where it's used.

ruthnaebeck:
FYI - My PR accounts for all types - https://github.com/DataDog/documentation/pull/3991

davidejones:
Thanks @ruthnaebeck those changes are a great! However I've noticed this still doesn't quite resolve this issue we are looking at. I think this PR is more related to how we are using the Javascript with anchor tags than the hugo output of it.

The below js functionality is to allow a side nav link that refers to an anchor on the same page you are on to animate to that position on the page. Removing this js will remove this functionality.

```javascript
$(".sidenav-nav a").click(function(){
  moveToAnchor($(this).attr('href').substr(1), true);
});
```

**Here is an example of this functionality in action**
- in Chrome keep the console open and visit this page https://docs.datadoghq.com/tracing/advanced_usage/
- now click in the side nav on say `OpenTracing`
- you won't see the js error in the console and the page will animate scroll to the right place
- Now click another url without an anchor on the same page for example `Alerting`, you will see the js error then it redirects to the page

Note: if you try these same steps on Zach's branch it will still work but is no longer doing the animated slide to the anchor just an abrupt jump. The error has stopped showing though.

**What we want to achieve**

- The error to be gone when clicking regular sidenav links
- To still slide animate to sidenav links that have anchors on the same page you are on
- add some comment to the code so we know what its doing

@jiboong @zbayoff Does that make sense / sound correct? Zach You think you can do the above steps in this PR? Let me know if you want me to talk through what i mean.

ruthnaebeck:
What about adding an additional class to links that are just anchors? The code I updated is already determining this.

davidejones:
@ruthnaebeck that sounds like a better idea

zbayoff:
I'm not sure if this is an ideal fix, but since this is a special section, I just targeted the Advanced Instrumentation section, and applied the animation function on clicking the anchors.

PR #3996, opened 2019-02-14
improve announcement banner loading, added inline js

zbayoff:
### What does this PR do?
Adds inline js to load up the announcement banner immediately, so it doesn't 'flash' when the page loads.

### Preview link
https://docs-staging.datadoghq.com/zach/announce-banner-improve
Load the page, announcement banner should be visible immediately. 

### Trello Card
https://trello.com/c/d9kylnEf/3041-docs-announcement-banner-opens-after-page-load-causing-page-to-jump-down-on-refreshes

### DIFF
--- a/layouts/partials/announcement_banner/announcement_banner.html
+++ b/layouts/partials/announcement_banner/announcement_banner.html
@@ -4,4 +4,11 @@
     <a href="{{.Site.Params.corp_url}}{{$announce.link}}">{{$announce.text | safeHTML }}</a>
     <i class="icon icon-small-x pull-right"></i>
 </div>
+<script>
+    var announcement_banner;
+    announcement_banner = sessionStorage.getItem('announcement_banner');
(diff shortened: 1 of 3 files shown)


layouts/partials/announcement_banner/announcement_banner.html
```diff
@@ -4,4 +4,11 @@
     <a href="{{.Site.Params.corp_url}}{{$announce.link}}">{{$announce.text | safeHTML }}</a>
     <i class="icon icon-small-x pull-right"></i>
 </div>
+<script>
+    var announcement_banner;
```
anagio: Let's change this. The banner by default is hidden with css which causes it to display then hide.

Remove this scss. The banner should be hidden if local/session storage have it set to be hidden. See corpsites method for checking

.announcement_banner {
    display: none;
}

PR #4016, opened 2019-02-18
Remove dd-opentracing-cpp version numbers

willgittoes-dd:
### What does this PR do?
Make docs reference only the latest version of the C++ and NGINX tracer. Version number-specific instructions have been removed, and replaced with a quick call to Github to reference the latest version. 

### Motivation
Some customers were using old versions of the tracer, because they just copied the docs blindly. The updated docs can still be blindly copied (so it's easy for the user), but no-longer reference a specific version.

### Preview link
https://docs-staging.datadoghq.com/willgittoes-dd/proxies-latest-version/tracing/languages/cpp/
https://docs-staging.datadoghq.com/willgittoes-dd/proxies-latest-version/tracing/proxies/nginx/

### Additional Notes
Until https://github.com/DataDog/dd-opentracing-cpp/pull/85 is merged and released, this is WIP. Since these instructions will not work until we're updated to use the latest dependencies.


willgittoes-dd:
OK, the prerequisite changes in dd-opentracing-cpp have now been merged, so this documentation is ready to go (pending approval from @andrewsouthard1  😊 )

PR #4018, opened 2019-02-18
Adding log + process conf instructions

l0k0ms:
### What does this PR do?
Adds Log + process collection instructions with helm

### Motivation
Better doc.

### Preview link

[Helm doc](https://docs-staging.datadoghq.com/gus/helm-log/agent/kubernetes/helm/?tab=macoshomebrew#installing-the-datadog-helm-chart)


content/agent/kubernetes/helm.md
```diff
@@ -111,39 +111,52 @@ To install the chart with the release name `<RELEASE_NAME>`, retrieve your Datad
...
-
 This chart adds the Datadog Agent to all nodes in your cluster via a DaemonSet. It also optionally deploys the [kube-state-metrics chart][5] and uses it as an additional source of metrics about the cluster. A few minutes after installation, Datadog begins to report hosts and metrics.
 
-### Configuration
+**Note**: For a full list of the Datadog chart's configurable parameters and their default values, refer to the [Datadog Helm repository README][6]. 
+
+## Configuring the Datadog Helm chart
 
-For a full list of the Datadog chart's configurable parameters and their default values, refer to the [Datadog Helm repository readme][6].
+As a best practice, a YAML file that specifies the values for the chart parameters should be provided to configure the chart:
```
ajacquemot: not sure it's a best practice, I think this is a required one ?

content/agent/kubernetes/helm.md
```diff
@@ -111,39 +111,52 @@ To install the chart with the release name `<RELEASE_NAME>`, retrieve your Datad
...
+
+## Configuring the Datadog Helm chart
 
-For a full list of the Datadog chart's configurable parameters and their default values, refer to the [Datadog Helm repository readme][6].
+As a best practice, a YAML file that specifies the values for the chart parameters should be provided to configure the chart: 
 
-Specify each parameter using the `--set key=value[,key=value]` argument to `helm install`. For example:
+1.  **Copy the default [`datadog-values.yaml`][7] value file.**
+2.  Set the `apiKey` parameter with your [Datadog API key][4].
+3. Upgrade the Datadog Helm chart with the new `datadog-values.yaml` file:
```
ajacquemot:
```suggestion
3.  Upgrade the Datadog Helm chart with the new `datadog-values.yaml` file:
```

content/agent/kubernetes/helm.md
```diff
@@ -111,39 +111,52 @@ To install the chart with the release name `<RELEASE_NAME>`, retrieve your Datad
...
-Use the `helm upgrade` command to upgrade to a new version of the chart, or to change the configuration of your release. 
+Update your [datadog-values.yaml][7] file with the process collection configuration, then upgrade your Datadog Helm chart:
 
-```bash
-helm upgrade -f my-values.yaml <RELEASE_NAME> stable/datadog --recreate-pods
+```
+datadog:
+  (...)
+  processAgentEnabled: true
+  volumes:
```
ajacquemot: wondering whether we should do the same as for logs, i.e. do we really need to have this in the `values.yaml` ? I think we could hardcode those values directly in the deamonset ? cc @xvello

content/agent/kubernetes/helm.md
xvello: It's basically required if you want to know what has been running (so you re-apply the same settings when upgrading) but in theory one could have a bash script running `helm` with three lines of `--set`.

content/agent/kubernetes/helm.md
xvello: Live processes work without these files, and will show uids instead of usernames, but indeed that'd be good to just add the volumes automatically when `processAgentEnabled` is true

content/agent/kubernetes/helm.md
l0k0ms: I hesitated to document that.. but moving forward with all the config options I think we shouldn't advocate for it. 

Having a one simple line bash command to first launch the Agent is fine, it makes onboarding easier, testing easier...  But then using the `value.yaml` logic is more heathy and allows for more consistency over the configuration options documentation. what do you think?

xvello: Perfect, we keep the one-liner for easy testing, but move people towards a `values.yaml` file when they want to customize

content/agent/kubernetes/helm.md
xvello: Actually, it's already the case, see https://github.com/helm/charts/commit/fa9cf124fa7f0ad91754c37920de3bfdc850da87

This means only setting processAgentEnabled is enough

l0k0ms: ping @mikezvi for tracking

l0k0ms:
@xvello so the default `value.yaml` is not up to date then.. we should update it.

xvello:
@l0k0ms you mean https://github.com/helm/charts/pull/11529/files 😇

content/agent/kubernetes/helm.md
ajacquemot: sounds good to me

PR #3994, opened 2019-02-14
[Log] adding list log API endpoint

l0k0ms:
### What does this PR do?
Adds log list API endpoint

### Motivation
PM request

### Preview link

https://docs-staging.datadoghq.com/gus/log-search-api/api/?lang=python#list-logs

PR #4026, opened 2019-02-19
Log search api

l0k0ms:
Follow up on: https://github.com/DataDog/documentation/pull/3994

### What does this PR do?
Adds log list API endpoint

### Motivation
PM request

### Preview link

https://docs-staging.datadoghq.com/gus/log-search-api-2/api/?lang=python#list-logs


content/en/api/logs/list_logs.md
```diff
@@ -0,0 +1,37 @@
+---
+title: List Logs
+type: apicontent
+order: 16.2
+external_redirect: /api/#list-logs
+---
+
+## List Logs
+
+List endpoint returns logs that match a log search query. Results are paginated.
```
epot: 👋 don't we want to document the response payload as well?

epot: I mean on top of the example provided. Maybe this is how we document all REST APIs, sorry if it's a dummy question :)

content/en/api/logs/code_snippets/list_logs.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+# Make sure you replace the <DD_API_KEY> and <DD_APP_KEY> key below
+# with the ones for your account
+
+api_key=<DD_API_KEY>
+app_key=<DD_APP_KEY>
+
+curl -X GET https://http-intake.logs.datadoghq.com/api/v1/logs-queries/list?api_key=${api_key}&application_key=${app_key}
+     -H "Content-Type: application/json" \
+     -d '{query: "service:nginx query -method:POST Windows",time: {offset: -3600, from: "2001-01-01 12:00:00", to: "2001-01-01 14:00:00"} sort: "desc", limit: 50}'
```
epot: time must be something like `2019-03-02T09:42Z` or `2019-03-02T09:50:36.320Z` to work properly

content/en/api/logs/list_logs.md
l0k0ms: plop, result payload are provided: https://github.com/DataDog/documentation/pull/4026/files#diff-0d2646143c4ec52a12bff5d13d59a500 but maybe you had something different in mind?

epot: Yes I would have thought we could specifically say what each field is and what value it can contain when that makes sense (for instance the different status values, how the "nextLogId" can be used...). And this on top of one or several example(s).

l0k0ms: That makes sense, but it would mean doing it for all API endpoints :/ I think we are going to wait for API v2 before documenting all returned fields. 
Regarding `nextLogid`, there is now a small doc in `how to use pagination` that explains how to do it. For the status, idk what it refers to... is it the request status? ping @pcariou-dd

PR #4095, opened 2019-02-28
Improve PHP docs consistency

labbati:
### What does this PR do?

This PR slightly changes the PHP trace documentation and makes it more consistent with the Java tracer.

Furthermore, it tries to simplify on-boarding, focusing on automatic installation rather then manual installation.

### Preview link
[Preview](https://docs-staging.datadoghq.com/labbati/docs-consistency/tracing/languages/php/)

### DIFF
--- a/content/en/tracing/languages/php.md
+++ b/content/en/tracing/languages/php.md
@@ -126,17 +126,15 @@ The PHP tracer can be configured using environment variables.
 
 ### Apache
 
-Set using [`SetEnv`][12] from the server config, virtual host,
-directory, or **.htaccess** file.
+Set using [`SetEnv`][12] from the server config, virtual host, directory, or **.htaccess** file.
 
 ```
 SetEnv DD_TRACE_DEBUG true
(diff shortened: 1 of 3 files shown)


content/en/tracing/languages/php.md
```diff
@@ -123,99 +84,39 @@ PHP APM supports the following PHP versions:
...
-});
-```
+### Integrations
 
-### Zend Framework 1 integration
+#### Web Framework Compatibility
 
-Zend Framework 1 is automatically instrumented by default, so you are not required to modify your ZF1 project. However, if automatic instrumentation is disabled, enable the tracer manually.
+Please not that if the web framework that you use is not listed below you will still see traces for your web requests
+in the UI. The only caveat is that some metadata and spans which are very specific to a particular web fr
```
cswatt: Looks like this is unfinished

labbati: Good catch! Thanks :)

PR #4124, opened 2019-03-05
fixing some stuff

cswatt:
removes some information that may not be true


content/en/graphing/graphing_json/widget_json.md
```diff
@@ -14,70 +14,6 @@ This page covers how to use JSON to build a widget in a Datadog dashboard, as yo
 
 {{< img src="graphing/graphing_json/references-graphing-jsoneditor.png" alt="references graphing jsoneditor" responsive="true" style="width:80%;">}}
```
iddl: For now we should also remove the text  above mentioning 
```
This page covers how to use JSON to build a widget in a Datadog dashboard, as you would see in the *JSON* tab of a widget. To learn more about the GUI editor, visit the main [Graphing documentation Page][1]
```

we could remove the image as well

PR #4138, opened 2019-03-07
[Agent Proxy] updates

ruthnaebeck:
### What does this PR do?
- Add warning messages about using the metric, proxy port for logs.
- Add information on `use_port_443` for proxy logs.

### Motivation
Trello request

### Preview links
- https://docs-staging.datadoghq.com/ruth/proxy-ports/agent/proxy/#using-a-web-proxy-as-proxy
- https://docs-staging.datadoghq.com/ruth/proxy-ports/agent/proxy/#port-443

### DIFF
--- a/content/en/agent/proxy.md
+++ b/content/en/agent/proxy.md
@@ -31,7 +31,11 @@ Traditional web proxies are supported natively by the Agent. If you need to conn
 {{% tab "Agent v6" %}}
 
 Set different proxy servers for `https` and `http` requests in your Agent `datadog.yaml` configuration file.
-The Agent uses `https` to send data to Datadog, but integrations might use `http` to gather metrics. No matter the proxied requests, you can activate SSL on your proxy server. Below are some configuration examples for your `datadog.yaml` file:
+The Agent uses `https` to send data to Datadog, but integrations might use `http` to gather metrics. No matter the proxied requests, you can activate SSL on your proxy server. Below are some configuration examples for your `datadog.yaml` file.
+
+<div class="alert alert-warning">
+The <code>&ltHOST&gt;:&ltPORT&gt;</code> used to proxy metrics can NOT be used to proxy logs.
+</div>
 
 Setting an HTTP proxy for all `https` requests:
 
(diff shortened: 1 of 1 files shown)


content/en/agent/proxy.md
```diff
@@ -31,7 +31,11 @@ Traditional web proxies are supported natively by the Agent. If you need to conn
 {{% tab "Agent v6" %}}
 
 Set different proxy servers for `https` and `http` requests in your Agent `datadog.yaml` configuration file.
-The Agent uses `https` to send data to Datadog, but integrations might use `http` to gather metrics. No matter the proxied requests, you can activate SSL on your proxy server. Below are some configuration examples for your `datadog.yaml` file:
+The Agent uses `https` to send data to Datadog, but integrations might use `http` to gather metrics. No matter the proxied requests, you can activate SSL on your proxy server. Below are some configuration examples for your `datadog.yaml` file.
+
+<div class="alert alert-warning">
+The <code>&ltHOST&gt;:&ltPORT&gt;</code> used to proxy metrics can NOT be used to proxy logs.
```
NBParis: Should we explain why? The reason being that logs are forwarded through TCP.

content/en/agent/proxy.md
```diff
@@ -90,6 +94,10 @@ The Agent uses the following values in order of precedence:
 {{% /tab %}}
 {{% tab "Agent v5" %}}
 
+<div class="alert alert-warning">
+The <code>&ltHOST&gt;:&ltPORT&gt;</code> used to proxy metrics can NOT be used to proxy logs.
```
NBParis: Should we explain why? The reason being that logs are forwarded through TCP.

content/en/agent/proxy.md
ruthnaebeck: @NBParis - This is already explained in the log proxy section. This is an extra warning message in the metrics section. Should I still add an explanation here?

content/en/agent/proxy.md
ruthnaebeck: Same

content/en/agent/proxy.md
NBParis: Well with this logic we could say that the warning is not necessary as there is a log proxy section :)

I believe that just stating why it's not possible or adding a link to the log proxy section would be helpful. As it would not be just a `not for logs, pass your way` but more a `for logs, please go there`.

What do you think?

PR #4179, opened 2019-03-14
[Forecasts Monitors] Adding note about full window of data for evalua…

byassine93:
…tion

### What does this PR do?

This PR adds a short paragraph about full window of data for evaluation for forecast monitors.

### Motivation

[Trello Support ticket](https://trello.com/c/d7OKg3w0/90-forecast-monitor-is-not-showing-all-the-groups
@DataDog/butter 
)

### DIFF
--- a/content/en/monitors/monitor_types/forecasts.md
+++ b/content/en/monitors/monitor_types/forecasts.md
@@ -62,6 +62,8 @@ Datadog automatically sets the **Advanced** options for you by analyzing your me
 * We recommend using larger intervals between points to avoid having noise influence the forecast too much.
 * The number of deviations controls the width of the range of forecasted values. A value of 1 or 2 should be large enough to accurately forecast most "normal" points.
 
+Then, you can choose if the monitor requires a full window of data for evaluation. Setting it to "Requires" might result in evaluations getting skipped. Indeed, evaluations will get skipped if the data is sparse or if the past data in the evaluation window has some missing values. 
+
 Complete all steps in the New Monitor form (**Say what's happening**, etc.) and click **Save** to create the Forecast monitor.
 
 Both the Monitor Edit page and the Monitor Status pages provide "Historical Context" so that you can see how the metric behaved in the past. This should provide some insight into what the forecast algorithm takes into account when predicting future values.
--- a/content/en/monitors/monitor_types/forecasts.md
+++ b/content/en/monitors/monitor_types/forecasts.md
@@ -62,7 +62,7 @@ Datadog automatically sets the **Advanced** options for you by analyzing your me
 * We recommend using larger intervals between points to avoid having noise influence the forecast too much.
 * The number of deviations controls the width of the range of forecasted values. A value of 1 or 2 should be large enough to accurately forecast most "normal" points.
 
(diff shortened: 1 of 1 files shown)


StephenKappel:
Ug. It looks like this docs page got somewhat butchered when it was moved from its old location in the docs. It's not sure if it's a page about forecasts (in general) or a page about forecast monitors. Can someone from @DataDog/baklava help us get this re-organized? The text under https://docs.datadoghq.com/monitors/monitor_types/forecasts/#forecast-monitor isn't even about monitors -- it's about dashboards until the "Forecast alerts" heading.  The screenshot for the advanced option section is also out of date (but this is going to be updated again in a few weeks, so we might not want to change it now).

content/en/monitors/monitor_types/forecasts.md
```diff
@@ -62,6 +62,8 @@ Datadog automatically sets the **Advanced** options for you by analyzing your me
 * We recommend using larger intervals between points to avoid having noise influence the forecast too much.
 * The number of deviations controls the width of the range of forecasted values. A value of 1 or 2 should be large enough to accurately forecast most "normal" points.
 
+Then, you can choose if the monitor requires a full window of data for evaluation. Setting it to "Requires" might result in evaluations getting skipped. Indeed, evaluations will get skipped if the data is sparse or if the past data in the evaluation window has some missing values.
```
StephenKappel:
```suggestion
Then, you can choose if the monitor requires a full window of data for evaluation. Setting it to "Requires" will force the monitor to ignore (i.e., show "No Data" as the monitor state) any series that doesn't have data going back all the way to the start of the time range shown in the Evaluation Window graph.
```

"skipped" is a backend term that I don't think is exposed to the user anywhere. We just show "No Data" on the status page.

Let's not talk about sparsity here because it's a bit confusing. If only every other interval has a point, for example, the monitor will evaluate, skip, evaluate, skip, etc. based on whether the first point in the window has data or not. I think the starting point of the series is key point. That is, new hosts will be ignored.

apigirl:
@StephenKappel - sure, I'll add a card to take a look at this page

byassine93:
The dashboard content for forecasts should go on this page : https://docs.datadoghq.com/graphing/functions/algorithms

content/en/monitors/monitor_types/forecasts.md
```diff
@@ -62,6 +62,8 @@ Datadog automatically sets the **Advanced** options for you by analyzing your me
 * We recommend using larger intervals between points to avoid having noise influence the forecast too much.
 * The number of deviations controls the width of the range of forecasted values. A value of 1 or 2 should be large enough to accurately forecast most "normal" points.
 
+Then, you can choose if the monitor requires a full window of data for evaluation. Setting it to "Requires" will force the monitor to ignore (i.e., show "No Data" as the monitor state) any series that doesn't have data going back all the way to the start of the time range shown in the Evaluation Window graph.
```
apigirl:
```suggestion
Then, you can choose if the monitor requires a full window of data for evaluation. Setting it to "Requires" will force the monitor to ignore (i.e., show "No Data" as the monitor state) any series that doesn't have data going back to the start of the time range shown in the Evaluation Window graph.
```

PR #4265, opened 2019-03-25
clarifying what the agent does with more than one configuration file

apigirl:
### What does this PR do?
clarifying what the agent does with more than one configuration file

### Motivation
customer confusion

### Preview link
Replace the branch name and add the complete path:
https://docs-staging.datadoghq.com/kaylyn/log-configs/agent/guide/agent-configuration-files/?tab=agentv6#checks-configuration-files-for-agent-6

### DIFF
--- a/content/en/agent/guide/agent-configuration-files.md
+++ b/content/en/agent/guide/agent-configuration-files.md
@@ -75,7 +75,7 @@ In order to provide a more flexible way to define the configuration for a check,
 
 `/datadog-agent/conf.d/<CHECK_NAME>.d/`.
 
-To prevent duplicate logs from being sent to Datadog, the agent does not accept multiple YAML files that point to the same log source. In the case where there is more than one YAML file that points to the same log source, the agent considers the files in alphabetical order and uses the first file.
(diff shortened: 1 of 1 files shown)


content/en/agent/guide/agent-configuration-files.md
```diff
@@ -75,6 +75,8 @@ In order to provide a more flexible way to define the configuration for a check,
 
 `/datadog-agent/conf.d/<CHECK_NAME>.d/`.
 
+To prevent duplicate logs from being sent to Datadog, the agent does not accept multiple YAML files that point to the same log source. In the case where there is more than one YAML file that points to the same log source, the agent considers the files in alphabetical order and uses the first file.
```
l0k0ms:
```suggestion
Note: For log collection, to prevent duplicate logs from being sent to Datadog, the agent does not accept multiple YAML files that point to the same log source. In the case where there is more than one YAML file that points to the same log source, the agent considers the files in alphabetical order and uses the first file.
```

l0k0ms: I think this note should go below, just above the array, otherwise it seems weird imho.

PR #4279, opened 2019-03-29
Tags alias

l0k0ms:
This PR aims to introduce an alias for: https://docs.datadoghq.com/tags/

In order to redirect to: https://docs.datadoghq.com/tagging/ but there seems to be an issue coming from a file or a partial.. 

https://docs-staging.datadoghq.com/gus/tags-aliases/tags/


ruthnaebeck:
@l0k0ms - When I navigate to `/tags` on your staging branch, it's a blank page.

I'm not sure what you are doing with the other files. I added `CorpWeb` for review.

l0k0ms:
There was a file generating an error, I removed it, and now we are ending up in this blank page.. I'm still investigating. Along the way i'm removing all outdated files that have no more usage.. any help from @DataDog/corpweb is more than welcome on this journey :)

nsollecito:
@l0k0ms `tags` is a reserved word in Hugo to generate tag listing pages. While configured by default, it doesn't appear this is currently being used in the docs site. Let me know if you want us to remove this from the site config. 

I've confirmed removing this will allow the redirect to work properly... We can rename the tags taxonomy as an option as well in case we want to reserve it's use for something in the future

l0k0ms:
@nsollecito many thanks for this 🙇 i'm going to look into this !

PR #4294, opened 2019-04-01
Bcon/arithmetic monitor evaluation

unclebconnor:

### DIFF
--- a/content/en/monitors/guide/division-in-monitor-evaluations.md
+++ b/content/en/monitors/guide/division-in-monitor-evaluations.md
@@ -1,13 +1,13 @@
 ---
-title: Arithmetic in Monitor Evaluations
+title: Division in Monitor Evaluations
 kind: guide
 aliases:
-  - /monitors/faq/arithmetic-in-Monitor-Evaluations
+  - /monitors/faq/Division-in-Monitor-Evaluations
 ---
 
 ## Overview
 
-You may want to create an alert based on a formula involving multiple metric queries. There are some tools and behaviors that you should consider to ensure your monitor settings are appropriate for evaluating queries as you intend. The scenarios discussed below will mostly pertain to division, but many of these principles can generally be applied to any combination of operations, except where explicitly stated.
+Creating an alert based on a query with division is a common practice. There are some tools and behaviors that should be considered to ensure monitor settings are appropriate for evaluating these queries as intended.
 
 ## Error Rate Example
 
@@ -62,7 +62,7 @@ Compare the result of the evaluation depending of the path:
 | **`as_count_eval_path`** | Aggregation function applied _before_ division | **(1+2+...+5)/(5+5+...+5)** | **0.6** |
 | **`classic_eval_path`**  | Aggregation function applied _after_ division  | **(1/5 + 2/5 + ... + 5/5)** | **3**   |
 
-You could also use `avg` and `as_rate()` to get the same result as `sum` and `as_count()`:
+You could also use `avg` and `as_rate()` to get the same result as `sum` and `as_count()` :
 
 ```
 avg(last_5m): sum:requests.error{*}.as_rate() / sum:requests.total{*}.as_rate()
@@ -74,13 +74,13 @@ avg(last_5m): sum:requests.error{*}.as_rate() / sum:requests.total{*}.as_rate()
 
 _Note that all evaluations above are mathematically correct. Choose a method to suit your intention._
 
-\*If the metric is a rate, you can omit `.as_rate()`. Include this modifier to be explicit.
+\*If the metric is a rate, you can optionally omit `.as_rate()`. Include this modifier to be explicit.
 
 ## Sparse Metrics
 
 In case of sparse or _0_ metrics in the denominator some results are rejected in case of the `classic_eval_path`.
 
-Let's have following metrics:
+Let's consider the following metric values:
 
 - `A = (10, 10, 10)`
 - `B = (0, 1, -)`
@@ -100,7 +100,7 @@ If the evaluation window includes many "null" buckets (**10/NaN + 10/Nan + ... +
 
 ## `.rollup()`
 
-You can apply a `.rollup()` function to ensure all time buckets being evaluated have valid values.
+You can apply a `.rollup()` function to ensure all time buckets being evaluated have valid values. This function can be similarly useful for any metric with gaps, regardless of whether the query involves artithmetic.
 
 **Original**: `sum:my_metric.is.sparse{*}`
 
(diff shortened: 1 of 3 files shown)


content/en/monitors/guide/division-in-monitor-evaluations.md
```diff
@@ -0,0 +1,211 @@
...
+| 2018-03-13 11:00:30   | 5        |
+| 2018-03-13 11:01:30   | 5        |
+| 2018-03-13 11:02:40   | 5        |
+| 2018-03-13 11:03:30   | 5        |
+| 2018-03-13 11:04:40   | 5        |
+```
+
+### Division with `as_count()`
+
+Monitors involving division and at least 1 `as_count()` modifier use a separate evaluation path than other monitors. This impacts queries of the format `queryA.as_count() / queryB.as_count()`. Queries without division (`+`, `-`, or `*`) are not impacted.
```
chrismdd: The current logic only checks whether the query has `as_count()`, so it affects queries with multiplication too. Evaluation order doesn't matter for addition and subtraction unless functions are applied (you get the same result regardless of the order).

content/en/monitors/guide/division-in-monitor-evaluations.md
```diff
@@ -0,0 +1,211 @@
...
+```
+
+Compare the result of the evaluation depending of the path:
+
+| Path                     | Behavior                                       | Expanded expression         | Result  |
+| :----------------------- | :--------------------------------------------- | :-------------------------- | :------ |
+| **`as_count_eval_path`** | Aggregation function applied _before_ division | **(1+2+...+5)/(5+5+...+5)** | **0.6** |
+| **`classic_eval_path`**  | Aggregation function applied _after_ division  | **(1/5 + 2/5 + ... + 5/5)** | **3**   |
+
+You could also use `avg` and `as_rate()` to get the same result as `sum` and `as_count()` :
```
chrismdd: The result would be similar but it wouldn't be the same in general. Consider the following two time series:

```
a = [1, 2, 3, 4, 5]
b = [5, 4, 3, 2, 1]
```

The sum of both series is the same, so the quotient of the sums is 1, but average of the quotients isn't quite 1:

```
const _ = require('lodash');
_.mean([1/5, 2/4, 3/3, 4/2, 5/1])
1.74
```

Average time aggregation with `as_rate()` is reasonable, but it sum/`as_count()` is better for error rates.

content/en/monitors/guide/division-in-monitor-evaluations.md
```diff
@@ -0,0 +1,211 @@
...
+avg(last_5m): sum:requests.error{*}.as_rate() / sum:requests.total{*}.as_rate()
+```
+
+| Path                    | Behavior                                      | Expanded expression           | Result  |
+| :---------------------- | :-------------------------------------------- | :---------------------------- | :------ |
+| **`classic_eval_path`** | Aggregation function applied _after_ division | **(1/5 + 2/5 + ... + 5/5)/5** | **0.6** |
+
+_Note that all evaluations above are mathematically correct. Choose a method to suit your intention._
+
+\*If the metric is a rate, you can optionally omit `.as_rate()`. Include this modifier to be explicit.
```
chrismdd: For both `COUNT` and `RATE` type metrics, either `.as_count()` or `.as_rate()` is required for the query to evaluate correctly. The UI query editor will maintain this invariant for you, but if you're creating a monitor via the API you need to add it to the query manually.

chrismdd: Actually this isn't completely accurate. You can construct a valid query without either of these modifiers, but we don't support doing that. I'll add more details in another comment below.

content/en/monitors/guide/division-in-monitor-evaluations.md
```diff
@@ -0,0 +1,211 @@
...
+
+You would see the following
+
+| Path                | Evaluation                          | Result |
+| :------------------ | :---------------------------------- | :----- |
+| `classic_eval_path` | **1/1 + 1/Nan + 1/1 + 1/Nan + 1/1** | 3      |
+
+By adding interpolation, you can ensure that metrics at every time interval
+
+**Modified**: `sum:my_metric.misaligned.groups.rate{env:a} by {timer,env}.fill(last)`
```
chrismdd: When `RATE` metrics are queried with either `.as_count()` or `.as_rate()`, they are essentially treated like counts during time aggregation (even though they are stored as a per second rate). If the query has `.as_rate()`, they are converted back to a per second rate after time aggregation. If you query a `RATE` metric without either of these modifiers, it's essentially treated like a gauge, and that works since it's stored as a per second rate, but we don't officially support doing that, and it may not work in the future.

I would recommend always using one of these modifiers for `COUNT`/`RATE` metrics (like the UI query editor does), and not using interpolation.

content/en/monitors/guide/division-in-monitor-evaluations.md
```diff
@@ -0,0 +1,211 @@
...
+sum(last_5m):sum:my_metric.misaligned.groups.rate{timer:1min,env:a} / sum:my_metric.misaligned.groups.rate{timer:2min,env:a}.fill(last)
+```
+
+With `.fill(last)` the new result would be:
+
+| Path                | Evaluation                      | Result |
+| :------------------ | :------------------------------ | :----- |
+| `classic_eval_path` | **1/1 + 1/1 + 1/1 + 1/1 + 1/1** | 5      |
+
+## Short Evaluation Windows
```
chrismdd: We probably don't want to go into too many implementation details here, but I definitely agree that we should recommend against small monitor windows for error rate queries. Make we can just say something about avoiding discrepancies if one metric is submitted slightly earlier than another?

content/en/monitors/guide/division-in-monitor-evaluations.md
```diff
@@ -0,0 +1,211 @@
+---
+title: Division in Monitor Evaluations
+kind: guide
+aliases:
+  - /monitors/faq/Division-in-Monitor-Evaluations
+---
+
+## Overview
+
+Creating an alert based on a query with division is a common practice. There are some tools and behaviors that should be considered to ensure monitor settings are appropriate for evaluating these queries as intended.
```
chrismdd: To reduce confusion, I think it would be best to focus on count division queries (e.g. error rates) in this article and to move the other things to separate articles.

l0k0ms:
hello just a small ping since no activity happen over the last 7 days. Is this PR still active?

unclebconnor:
Hey Gus,

Yeah I was waiting on some feedback from the team.  I've got what I need
and will do another draft in the next day or 2.  Thanks for following up.

Brian Connor
Datadog Solutions Engineer - Seattle, WA

content/en/monitors/guide/division-in-monitor-evaluations.md
unclebconnor: Changing this to just be more general.  I pulled this more or less from the old doc.

unclebconnor:
Updated on a different branch.  Closing this.

PR #4312, opened 2019-04-03
Allow doc to build even if one integration has an error

l0k0ms:
### What does this PR do?

This Prevent the whole documentation build to fail if a manifest is not there for a given integration.
It also removes the `tqdm` logic and some not so necessary print.
 
If there is no manifest, no documentation page will be build, hence we can remove the discard action we had previously for "non-real" integrations.

### Motivation

More robust doc.

### DIFF
--- a/local/etc/pull_config.yaml
+++ b/local/etc/pull_config.yaml
@@ -25,10 +25,10 @@
     - action: integrations
       branch: master
       globs:
-      - "*[!}]/metadata.csv"
-      - "*[!}]/manifest.json"
-      - "*[!}]/service_checks.json"
-      - "*[!}]/README.md"
+      - "*/metadata.csv"
+      - "*/manifest.json"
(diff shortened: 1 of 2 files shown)


local/etc/pull_config.yaml
```diff
@@ -44,10 +44,10 @@
...
       branch: master
       globs:
-      - "**/metadata.csv"
-      - "**/manifest.json"
-      - "**/service_checks.json"
-      - "**/README.md"
+      - "*[!}]/metadata.csv"
+      - "*[!}]/manifest.json"
+      - "*[!}]/service_checks.json"
+      - "*[!}]/README.md"
```
davidejones: is this correct? i'm not familiar with `*[!}]`

l0k0ms: it was correct.. but too complicated for no reason. fixing it

PR #4351, opened 2019-04-09
remove unsupported raw log correlation from node documentation

rochdev:
### What does this PR do?

Remove the documentation about manual ID injection for raw formatted logs in Node APM.

### Motivation

The feature was never implemented in the Logs backend. Users are following the documentation and the feature doesn't work.

### Preview link

https://docs-staging.datadoghq.com/rochdev/remove-raw-log-correlation/tracing/advanced/connect_logs_and_traces/?tab=nodejs#manual-trace-id-injection

### Additional Notes

### DIFF
--- a/content/en/tracing/advanced/connect_logs_and_traces.md
+++ b/content/en/tracing/advanced/connect_logs_and_traces.md
@@ -405,43 +405,9 @@ module.exports = Logger
 
 **Manual Trace ID Injection for Raw Formatted Logs**
 
-To ensure proper log correlation, verify the following is present in each log entry:
+Coming Soon. Reach out to [the Datadog support team][1] to learn more.
 
-- `dd.trace_id=<TRACE_ID>`: Where `<TRACE_ID>` is equal to `tracer.scope().active().context().toTraceId()`.
-- `dd.span_id=<SPAN_ID>`: Where `<SPAN_ID>` is equal to `tracer.scope().active().context().toSpanId()`.
-
-You should append or prepend these 2 strings directly to the message part of the log entry. This allows you to correlate trace and logs without having to alter your parsing rules.
-
-Example using `console` as the underlying logger:
-
-```javascript
-const tracer = require('dd-trace').init()
-
-class Logger {
-  log (level, message) {
-    const span = tracer.scope().active()
-    const time = (new Date()).toISOString()
-    const format = '[%s] [%s] - dd.trace_id=%s dd.span_id=%s %s'
-
-    let traceId = ''
-    let spanId = ''
-
-    if (span) {
-      traceId = span.context().toTraceId()
-      spanId = span.context().toSpanId()
-    }
-
-    console.log(format, time, level.toUpperCase(), traceId, spanId, message)
-  }
-}
-
-module.exports = Logger
-```
-
-**Note**: If you are not using a [Datadog Log Integration][1] to parse your logs, custom log parsing rules need to ensure that `trace_id` and `span_id` are being parsed as a string. More information can be found in the [FAQ on this topic][2].
-
-[1]: /logs/log_collection/nodejs
-[2]: /tracing/faq/why-cant-i-see-my-correlated-logs-in-the-trace-id-panel
+[1]: /help
 {{% /tab %}}
 {{% tab "PHP" %}}
 


andrewsouthard1:
@NBParis What needs to get done in the logs backend for this to work?

NBParis:
@andrewsouthard1 If you have samples of NodeJS logs that are generated from both those formats (JSON and raw) that would be great so we can build the appropriate processors automatically.

rochdev:
@NBParis JSON is already supported using the following format:

```json
{
  "dd": {
    "trace_id": 1234,
    "span_id": 1234
  }
}
```

For raw logs, the approach I've used is to add `dd.trace_id=` and `dd.span_id=` in the message directly to make it generic since we usually don't know that the source is Node in for example dockerized environment without explicit configuration for ever service, which we cannot expect every user to do. These could then be extracted and removed from the message. This would work with any source.

For example `[%s] [%s] - dd.trace_id=%s dd.span_id=%s %s` where the format is `[date] [level] - message`.

NBParis:
@rochdev do you have actual logs to share? As the date can be formatted in so many different ways so that an actual example would be really helpful.

Also for JSON logs if you have the default attributes that the logger is adding to the logs (like logger name and things like that) that can be very helpful to build an nodejs integration that would have remappers so that logs follow the official naming convention.

rochdev:
> do you have actual logs to share? As the date can be formatted in so many different ways so that an actual example would be really helpful.

The format is controlled 100% by the user, so we cannot know the format. However, a log entry will for sure always have at least a message part, which is why I'm relying on the message to not require altering the format.

> Also for JSON logs if you have the default attributes that the logger is adding to the logs (like logger name and things like that) that can be very helpful to build an nodejs integration that would have remappers so that logs follow the official naming convention.

There are many different loggers with different fields, but in general they are still similar (for example `message` and `msg`). As mentioned above, we cannot reliably rely on the source since in most cases the user will not have set it. However, fields like `msg` are generic enough that this could be applied to every logs. The biggest issue right now is the `hostname` field that overrides the actual hostname with the hostname reported by the container, which breaks host/log correlation.

I can provide a full list if you think it can be implemented in a generic way.

NBParis:
@rochdev if you could just have generic examples of logs that are generated without any config change from users that would be great.

Knowing that `msg` is used for instance is great value. Knowing how is called the severity is also something important, etc...
Which is why if you have log samples, that would be a great help for us to build the log integration.

rochdev:
@NBParis Here are examples from the most popular logging libraries:

#### console, morgan, log4j

These emit raw logs with user-defined formats (no default).

#### winston

The oldest popular JSON logger.

* **level**: The severity level. Uses syslog levels.
* **message**: The log message.
* **timestamp**: The time at which the log was generated. Format is ISO 8601.

#### bunyan

This has been the recommended JSON logger for Node projects for a few years.

* **name**: Name of the logger. We probably don't need to map this.
* **hostname**: This one is problematic because it overrides the hostname set by the agent with the container hostname.
* **pid**: The process ID.
* **level**: The severity level. 10 is trace, 20 is debug, 40 is warn, 50 is error, 60 is fatal.
* **msg**: The log message.
* **time**: The time at which the log was generated. Format is ISO 8601.
* **err**: Contains the error object if any.
  * **message**: The error message.
  * **name**: The error type.
  * **stack**: The stack trace.

#### pino

`pino` is basically using an exact subset of `bunyan` but runs faster.

* **hostname**: This one is problematic because it overrides the hostname set by the agent with the container hostname.
* **pid**: The process ID.
* **level**: The severity level. Same levels as bunyan.
* **msg**: The log message.
* **time**: The time at which the log was generated. Format is ISO 8601.

NBParis:
I'm surprised Log4j exists for NodeJS as well. They are everywhere.
Thanks for this, I'll create a nodeJS pipeline accordingly.

rochdev:
@NBParis As mentioned above, we can't rely on a Node specific pipeline. This should be handled in a generic pipeline that applies regardless of the source.

rochdev:
@NBParis Is this something we can fix quickly or should we merge this PR and re-add the documentation later? Did you get a change to look at the internal document that describes the correlation issues in more details?

irabinovitch:
I'd vote for merging and correcting when @NBParis has time to address the gaps.

PR #4356, opened 2019-04-10
update dogstatsd example to be valid go

JordanSussman:
### What does this PR do?

Modifes the example code to be valid go code. Currently, the code is invalid since:

- `dogstatsdgo` is never declared
- invalid indirect of statsd.Event literal (type statsd.Event)

### Motivation

Had to modify the example before I could utilize it for testing.

### Additional Notes

I also removed the `var err error` line since it is redundant.


l0k0ms:
Many thanks for this contribution and fix 🙇

PR #4506, opened 2019-04-30
Removing unused images

l0k0ms:
### What does this PR do?
Removes un-used images from the doc repo.

Strategy:

1. Get all image path in the `src/images/` folder and store them into a file

```
find . -name '*.png' -o -name '*.gif'-o -name '*.jpg' -type f >> image.txt
```

2. Run a small script that looks if an image is used in the documentation content: 

```
#!/bin/bash
for image_path in $(cut -d, -f1 < ./image.txt)
    do
        if grep -rsq $image_path ./documentation/content/en/ ; then
            echo "" > /dev/null
        else
            echo $image_path
        fi
    done
```

We get as an output the list of images that are not used across the whole doc content, we then just need to remove them manually to avoid any mistakes. 

### Motivation

A cleaner repo is a better repo, a better repo is better doc, a better doc is a better world.

### Preview link

https://docs-staging.datadoghq.com/gus/images-clean/

### Note

Since this script can take some time since we have to look across all doc files, idk if it's a good idea to implement it in the CI, but a spring clean-up can't hurt.


phrawzty:
> `cut -d, -f1 < ./image.txt`

That sets the field delimiter to `,`, but that character is never present in the output file:

```
$ grep , image.txt | wc -l
       0
```

Therefore, the output of that `cut` is just the lines of the file (on BSD `cut`, at least), so I'm not sure what you're trying to do there.

Also, why are you only looking in `./documentation/content/en/`? What about `developers/`, `fr/`, and `integrations/`?

l0k0ms:
@phrawzty indeed, didn't think about it since I found this code online and it just worked as expected. It allowed me to take the file line by line, each line being a image path, and pass this as a parameter for the grep.

.`/documentation/content/en/` includes all the content folders: https://github.com/DataDog/documentation/tree/master/content/en

the `fr/` one doesn't need to be screened since it's just a mirror of the `en/` for now, so if an image is not in `en/` it won't be in `fr/`

phrawzty:
> the `fr/` one doesn't need to be screened since it's just a mirror of the `en/` for now, so if an image is not in `en/` it won't be in `fr/`

Ok, but what about the other two directories?

l0k0ms:
Are you talking about:
https://github.com/DataDog/documentation/tree/master/content/en/integrations
https://github.com/DataDog/documentation/tree/master/content/en/developers

? if so they are included

If you are talking about content that lives outside of this repo if you build the documentation locally then all remote content is pulled in the `documentation/content/en/` folder so image are checked against those file too :)

phrawzty:
```
$ tree -L 1 content
content
├── developers
├── en
├── fr
└── integrations

4 directories, 0 files
```

I am talking about the things that are in the same directory level as `content/en`.

l0k0ms:
I think those are the left over from the repo before we moved all content to the `content/en` folder.

phrawzty:
> I think those are the left over from the repo before we moved all content to the `content/en` folder.

Oh. Well they should be removed then. ;)

ruthnaebeck:
Looks like I should wait on this until @l0k0ms has resolved the image checker conversation with #websites

l0k0ms:
closing PR until the CI job is fixed.

PR #4509, opened 2019-04-30
ADM Integrations Icon

apigirl:
### What does this PR do?

~Creates a way to see the ADM PR in Dogweb on Staging. This PR will be closed when the pipeline is no longer needed. This PR should never get merged in.~

Updating this PR to add the icon for ADM

### Motivation

PM and other stakeholder review.


steveharrington1:
In step 2, it should probably be made explicit that if a different kind of monitor is used other than a composite monitor, the regex in the template for defining the match will need to be updated by the user.

PR #4525, opened 2019-05-02
[APM Tag] Removing misleading wording

l0k0ms:
### What does this PR do?

Clarify wording:

The way the DD UI is right now: 

one can add `metadata` to span in order to use them in the trace search/analytics view: https://cl.ly/812a56763b2e

But in the OpenTracing reference, those information are not called `metadata` but `tags`, this is why all functions to set a span metadata with DD tracers are called `setTags` or something equivalent.

**This is confusing for users since the `tag` notion is already in use within Datadog. And there is no reference to `metadata` in the OpenTracing API norm.**

This PR makes it square with the current UI, a.k.a we add a span metadata with a `setTag` function... 

**But the correct way to go could be to rename span metadata in app by `Span Tags` ** 
But idk if this would be less/more confusing.. Product should make this decision.

### Motivation
Support feedback

### DIFF
--- a/content/en/tracing/advanced/adding_metadata_to_spans.md
+++ b/content/en/tracing/advanced/adding_metadata_to_spans.md
@@ -16,18 +16,18 @@ further_reading:
   text: "Explore your services, resources, and traces"
 ---
 
-Adding metadata in the form of key-value pairs to a span allows to correlate traces with other Datadog products to provide more details about specific spans. Metadata can be either added [to a single span](#adding-metadata-to-a-span) or [globally to all spans](#adding-metadata-globally-to-all-spans)
+Adding metadata in the form of key-value pairs to a span allows to correlate traces with other Datadog products to provide more details about specific spans. Metadata can be either added [to a single span](#adding-metadata-to-a-span-via-tags) or [globally to all spans](#adding-metadata-globally-to-all-spans-via-tags)
 
-## Adding Metadata to a span
+**Note**: Tracing has the idea of `tags` with OpenTracing, but tags having a specific meaning in the [Datadog ecosystem][1], tracing span *tags* are called **metadata**.
+
+## Adding Metadata to a Span via Tags
 
 {{< tabs >}}
 {{% tab "Java" %}}
 
-The Datadog UI uses specific metadata to set UI properties, such as an application's service name. A full list of these metadata can be found in the [Datadog][1] and [OpenTracing][2] APIs.
+The Datadog UI uses tags to set span level metadata. A full list of these metadata can be found in the [Datadog][1] and [OpenTracing][2] APIs.
 
-**Note**: Custom metadata are set using the OpenTracing API.
-
-Custom metadata may be set for auto-instrumentation by grabbing the active span out of the global tracer.
+Custom metadata may be set for auto-instrumentation by grabbing the active span out of the global tracer and setting a tag with `setTag`.
 
 ```java
 import io.opentracing.Tracer;
@@ -137,7 +137,8 @@ func main() {
 }
 ```
 
-Datadog's integrations make use of the `Context` type to propagate the current active span. If you want to add a metadata to a span attached to a `Context` via automatic instrumentation, call the `SpanFromContext` function:
+Datadog's integrations make use of the `Context` type to propagate the current active span.
+If you want to add span metadata via a tag attached to a `Context`, call the `SpanFromContext` function:
 
 ```go
 package main
@@ -239,7 +240,7 @@ if (null !== $span) {
 {{% /tab %}}
 {{< /tabs >}}
 
-## Adding Metadata globally to all spans
+## Adding Metadata globally to all spans via tags
 
 {{< tabs >}}
 {{% tab "Java" %}}
@@ -336,3 +337,4 @@ Values are of [variable type][1] and can be complex objects. Values are serializ
 ## Further Reading
 
 {{< partial name="whats-next/whats-next.html" >}}
+[1]: /tagging
--- a/content/en/tracing/advanced/adding_metadata_to_spans.md
+++ b/content/en/tracing/advanced/adding_metadata_to_spans.md
@@ -16,7 +16,7 @@ further_reading:
   text: "Explore your services, resources, and traces"
 ---
 
-Adding metadata in the form of key-value pairs to a span allows to correlate traces with other Datadog products to provide more details about specific spans. Metadata can be either added [to a single span](#adding-metadata-to-a-span-via-tags) or [globally to all spans](#adding-metadata-globally-to-all-spans-via-tags)
+Adding metadata in the form of key-value pairs to a span allows to correlate traces with other Datadog products to provide more details about specific spans. Metadata can be either added [to a single span](#adding-metadata-to-a-span-via-tags) or [globally to all spans](#adding-metadata-globally-to-all-spans-via-tags).
 
 **Note**: Tracing has the idea of `tags` with OpenTracing, but tags having a specific meaning in the [Datadog ecosystem][1], tracing span *tags* are called **metadata**.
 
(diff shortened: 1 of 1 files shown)


content/en/tracing/advanced/adding_metadata_to_spans.md
```diff
@@ -16,26 +16,18 @@ further_reading:
   text: "Explore your services, resources, and traces"
 ---
 
-Adding metadata in the form of key-value pairs to a span allows to correlate traces with other Datadog products to provide more details about specific spans. [Read more about tagging][1]
+Adding metadata in the form of key-value pairs to a span allows to correlate traces with other Datadog products to provide more details about specific spans. Metadata can be either added [to a single span](#adding-metadata-to-a-span) or [globally to all spans](#adding-metadata-globally-to-all-spans)
```
burningion: I'd mention the difference here between Datadog `tags` and Tracing `tags`. 

Something along the lines that Datadog has the idea of `tags`, and point to the definition. And then say that tracing has the idea of `tags` too, but in this document and around the product, we differentiate between the two by calling them `metadata` when we mean the Span `tags`.

content/en/tracing/advanced/adding_metadata_to_spans.md
```diff
@@ -16,26 +16,18 @@ further_reading:
...
 
-Metadata can be:
-
-- [Adding tags to a span](#adding-tags-to-a-span)
-- [Adding tags globally to all spans](#adding-tags-globally-to-all-spans)
-- [Further Reading](#further-reading)
-
-## Adding tags to a span
+## Adding Metadata to a span
```
burningion: Adding Metadata to a Span via Tags

content/en/tracing/advanced/adding_metadata_to_spans.md
```diff
@@ -16,26 +16,18 @@ further_reading:
...
 
 {{< tabs >}}
 {{% tab "Java" %}}
 
-Tags are key-value pairs attached to spans. All tags share a single namespace.
+The Datadog UI uses specific metadata to set UI properties, such as an application's service name. A full list of these metadata can be found in the [Datadog][1] and [OpenTracing][2] APIs.
```
burningion: Use tags to set span level metadata. A full list of the metadata...

content/en/tracing/advanced/adding_metadata_to_spans.md
```diff
@@ -16,26 +16,18 @@ further_reading:
...
 
-The Datadog UI uses specific tags to set UI properties, such as an application's service name. A full list of these tags can be found in the [Datadog][1] and [OpenTracing][2] APIs.
+**Note**: Custom metadata are set using the OpenTracing API.
 
-Custom tags are set using the OpenTracing API.
-
-Custom tags may be set for auto-instrumentation by grabbing the active span out of the global tracer.
+Custom metadata may be set for auto-instrumentation by grabbing the active span out of the global tracer.
```
burningion: Custom metadata may be set for auto-instrumentation by grabbing the active span out of the global tracer and setting a tag.

content/en/tracing/advanced/adding_metadata_to_spans.md
```diff
@@ -145,7 +137,7 @@ func main() {
 }
 ```
 
-Datadog's integrations make use of the `Context` type to propagate the current active span. If you want to add a tag to a span attached to a `Context` via automatic instrumentation, call the `SpanFromContext` function:
+Datadog's integrations make use of the `Context` type to propagate the current active span. If you want to add a metadata to a span attached to a `Context` via automatic instrumentation, call the `SpanFromContext` function:
```
burningion: ...If  you want to add span metadata via a tag, ...

content/en/tracing/advanced/adding_metadata_to_spans.md
```diff
@@ -247,7 +239,7 @@ if (null !== $span) {
 {{% /tab %}}
 {{< /tabs >}}
 
-## Adding tags globally to all spans
+## Adding Metadata globally to all spans
```
burningion: Adding Metadata globally to all spans via tags

content/en/tracing/advanced/adding_metadata_to_spans.md
```diff
@@ -16,18 +16,18 @@ further_reading:
   text: "Explore your services, resources, and traces"
 ---
 
-Adding metadata in the form of key-value pairs to a span allows to correlate traces with other Datadog products to provide more details about specific spans. Metadata can be either added [to a single span](#adding-metadata-to-a-span) or [globally to all spans](#adding-metadata-globally-to-all-spans)
+Adding metadata in the form of key-value pairs to a span allows to correlate traces with other Datadog products to provide more details about specific spans. Metadata can be either added [to a single span](#adding-metadata-to-a-span-via-tags) or [globally to all spans](#adding-metadata-globally-to-all-spans-via-tags)
 
-## Adding Metadata to a span
+**Note**: Tracing has the idea of `tags` with OpenTracing, but tags having a specific meaning in the [Datadog ecosystem][1], tracing span *tags* are called **metadata**.
```
burningion: **Note** Tracing metadata is added via `tags`, but tags already have a specific meaning throughout [Datadog][1], and so are called **metadata** within the Trace UI.

content/en/tracing/advanced/adding_metadata_to_spans.md
```diff
@@ -16,26 +16,18 @@ further_reading:
   text: "Explore your services, resources, and traces"
 ---
 
-Adding metadata in the form of key-value pairs to a span allows to correlate traces with other Datadog products to provide more details about specific spans. [Read more about tagging][1]
+Adding metadata in the form of key-value pairs to a span allows to correlate traces with other Datadog products to provide more details about specific spans. Metadata can be either added [to a single span](#adding-metadata-to-a-span-via-tags) or [globally to all spans](#adding-metadata-globally-to-all-spans-via-tags)
```
ruthnaebeck:
```suggestion
Adding metadata in the form of key-value pairs to a span allows to correlate traces with other Datadog products to provide more details about specific spans. Metadata can be either added [to a single span](#adding-metadata-to-a-span-via-tags) or [globally to all spans](#adding-metadata-globally-to-all-spans-via-tags).
```

content/en/tracing/advanced/adding_metadata_to_spans.md
```diff
@@ -75,7 +67,7 @@ def handle_customer(customer_id):
     span.set_tag('customer.id', customer_id)
 ```
 
-The current span can be retrieved from the context in order to set tags. This way, if a span was started by the instrumentation, you can retrieve the span and add custom tags. Note that if a span does not exist, `None` is returned:
+The current span can be retrieved from the context in order to set its metadata. This way, if a span was started by the instrumentation, you can retrieve the span and add custom metadata. Note that if a span does not exist, `None` is returned:
```
ruthnaebeck:
```suggestion
The current span can be retrieved from the context in order to set its metadata. This way, if a span was started by the instrumentation, you can retrieve the span and add custom metadata. **Note**: If a span does not exist, `None` is returned:
```

content/en/tracing/advanced/adding_metadata_to_spans.md
```diff
@@ -104,7 +96,7 @@ get '/posts' do
 end
 ```
 
-Access the current active span from any method within your code. Note, however, that if the method is called and there is no span currently active, `active_span` is nil.
+Access the current active span from any method within your code. Note, however, that if the method is called and there is no span currently active, `active_span` is `nil`.
```
ruthnaebeck:
```suggestion
Access the current active span from any method within your code. **Note**: If the method is called and there is no span currently active, `active_span` is `nil`.
```

content/en/tracing/advanced/adding_metadata_to_spans.md
```diff
@@ -247,7 +240,7 @@ if (null !== $span) {
 {{% /tab %}}
 {{< /tabs >}}
 
-## Adding tags globally to all spans
+## Adding Metadata globally to all spans via tags
```
ruthnaebeck:
```suggestion
## Adding metadata globally to all spans via tags
```

andrewsouthard1:
Per my note in #documentation, we want to remove the wording of "tags" here to metadata as it's much clearer. 

1. We should remove `via tags` from the titles here: https://docs-staging.datadoghq.com/gus/apm-tags-vs-metadata/tracing/advanced/adding_metadata_to_spans/?tab=java#adding-metadata-to-a-span-via-tags 

2. We also should link to Trace Search in both docs as adding custom metadata provides the most value with trace search, and also greatly enhances trace search. They should be linked in both.

PR #4609, opened 2019-05-15
Update tracing docs re: meta field string requirements

sneako:
The trace agent will respond with `json: cannot unmarshal number into Go struct field Span.meta of type string\n` if you try to send an integer as a meta field value.

### What does this PR do?

Adds a note about what types you can pass for Span.meta key/values

### Motivation

I got the following error when exporting a span with an integer in a meta value: `json: cannot unmarshal number into Go struct field Span.meta of type string\n`

### DIFF
--- a/content/en/api/tracing/send_trace.md
+++ b/content/en/api/tracing/send_trace.md
@@ -38,7 +38,7 @@ and each span is a dictionary with a `trace_id`, `span_id`, `resource`..
 *   **`duration`** - _required_ The duration of the request in nanoseconds.
 *   **`parent_id`** - _optional_ The span integer ID of the parent span.
 *   **`error`** - _optional_ Set this value to 1 to indicate if an error occured. If an error occurs, you should pass additional information, such as the error message, type and stack information in the `meta` property.
-*   **`meta`** - _optional_ A dictionary of key-value metadata. e.g. tags.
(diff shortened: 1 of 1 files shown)


content/en/api/tracing/send_trace.md
```diff
@@ -38,7 +38,7 @@ and each span is a dictionary with a `trace_id`, `span_id`, `resource`..
 *   **`duration`** - _required_ The duration of the request in nanoseconds.
 *   **`parent_id`** - _optional_ The span integer ID of the parent span.
 *   **`error`** - _optional_ Set this value to 1 to indicate if an error occured. If an error occurs, you should pass additional information, such as the error message, type and stack information in the `meta` property.
-*   **`meta`** - _optional_ A dictionary of key-value metadata. e.g. tags.
+*   **`meta`** - _optional_ A dictionary of key-value metadata. e.g. tags. NOTE: all keys and values must be strings
```
gbbr: I find the capitalised "NOTE" a bit aggressive. Maybe we can simply rephrase the original sentence, like:
```
**`meta`** - _optional_ A set of key-value metadata. Keys and values must be strings.
```
Or something like that.

You should also add:
```
**`metrics`** - _optional_ A set of key-value metadata. Keys must be strings and values must be 64-bit floating point numbers.
```

sneako:
Thank you!

PR #4631, opened 2019-05-17
Prometheus Autodiscovery: l5d is a container name

alanjcastonguay:
### What does this PR do?

Make clear that `l5d` is a container name inside the pod definition. This may be obvious to users of linkerd, but not easily discoverable for someone reading this page for the first time. 

### Motivation

Confused user: https://datadoghq.slack.com/archives/C4JREERCY/p1557932737130200


apigirl:
thanks for the PR @alanjcastonguay!

PR #4649, opened 2019-05-21
Update monitor integration type to show service check

nmuesch:
### What does this PR do?
Updates the API documentation for the monitor type table to reflect that an integration monitor could be either `query alert` or `service check`

### Motivation
Depending on what the monitor is looking at, it could either be a service check or query alert. Service check if you're monitoring an integration service check, or a query alert if you're using a metric/arithmetic. 

The Monitor tab for `Integration` in the UI allows you to make either, so this mapping should show both to avoid confusion. 

### Preview link

https://docs-staging.datadoghq.com/nick/monitor_type/api/monitors/monitors_create/

### DIFF
--- a/content/en/api/monitors/monitors_create.md
+++ b/content/en/api/monitors/monitors_create.md
@@ -22,7 +22,7 @@ If you manage and deploy monitors programmatically, it's easier to define the mo
 | event        | `event alert`                   |
 | forecast     | `query alert`                   |
 | host         | `service check`                 |
-| integration  | `query alert or service check`  |
+| integration  | `query alert` or `service check`  |
 | live process | `process alert`                 |
 | logs         | `log alert`                     |
 | metric       | `query alert`                   |
--- a/content/en/api/monitors/monitors_create.md
+++ b/content/en/api/monitors/monitors_create.md
@@ -13,23 +13,23 @@ If you manage and deploy monitors programmatically, it's easier to define the mo
 *   **`type`** [*required*]:
     The [type of the monitor][2], chosen from:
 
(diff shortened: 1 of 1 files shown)


content/en/api/monitors/monitors_create.md
```diff
@@ -13,23 +13,23 @@ If you manage and deploy monitors programmatically, it's easier to define the mo
...
+| Monitor Type | type attribute value            |
+| :--------    | :-------                        |
+| anomaly      | `query alert`                   |
+| apm          | `query alert`                   |
+| composite    | `composite`                     |
+| custom       | `service check`                 |
+| event        | `event alert`                   |
+| forecast     | `query alert`                   |
+| host         | `service check`                 |
+| integration  | `query alert or service check`  |
```
apigirl:
```suggestion
| integration  | `query alert` or `service check`  |
```

content/en/api/monitors/monitors_create.md
```diff
@@ -13,23 +13,23 @@ If you manage and deploy monitors programmatically, it's easier to define the mo
...
+| Monitor Type | type attribute value            |
+| :--------    | :-------                        |
+| anomaly      | `query alert`                   |
+| apm          | `query alert`                   |
+| composite    | `composite`                     |
+| custom       | `service check`                 |
+| event        | `event alert`                   |
+| forecast     | `query alert`                   |
+| host         | `service check`                 |
+| integration  | `query alert` or `service check`  |
```
gzussa: Shouldn't we have two rows instead and have more details on the monitor type?

PR #4658, opened 2019-05-22
synthetics: Update minimum interval

dcarley:
### What does this PR do?
Update the minimum interval for synthetics checks to 1min.

### Motivation
According to the API and UI the minimum interval is one minute, rather
than five minutes. I nearly discounted using this service because we
needed 1min and this was the first page that I read.

### Preview link
N/A: This currently only works if you are in the Datadog organization and working off of a branch - it will not work with a fork. 

### Additional Notes
Maybe the actual intervals values shouldn't be listed here in future if
they are likely to change again?


l0k0ms:
Hello @dcarley, 

Many thanks for this contribution and this glitch fix 🙇. 

The error might come from a cp/cv glitch since Browser tests have indeed a minimum collection interval of 5mins. I don't know if this minimum collection interval will change in the future, but we will definitively pay more attention to its value now.

Best regards,
Pierre

PR #4663, opened 2019-05-23
Check main page URL modif

MargotLepizzera:
### What does this PR do?
Adding details on the way the `Check main page URL's content` assertion works.

### Motivation
https://trello.com/c/F6sturfG/244-browser-test-failing-to-recognize-redirect


jyee:
`This assertion is made on the page the previous step was about.`

This does not make sense. The existing text says the assertion is made on the **current** page. It does not matter what the previous step was, it only matters what the current url is when the step runs (note that the previous step may or may not set a new URL).

MargotLepizzera:
We had cases of users leveraging this assertion on a new page that was opened without interacting with that new page first. The assertion was consequently performed on the first page, and not on the second one, which resulted in errors.

> This takes the URL of the current page, and asserts whether a specific value (`string`, `number`, `regex`) is present within it.
> If you want to use that assertion on a freshly opened tab, make sure to first create a step to interact with the page of that new tab.

PR #4667, opened 2019-05-23
C++ example for Trace Search and Analytics

cgilmour:
### What does this PR do?
Adds a C++ example for Trace Search and Analytics, to go with existing language examples.

### Motivation
Completeness.

### Preview link

https://docs-staging.datadoghq.com/cgilmour/cpp-analytics-example/tracing/trace_search_and_analytics/?tab=c#

### DIFF
--- a/content/en/tracing/trace_search_and_analytics/_index.md
+++ b/content/en/tracing/trace_search_and_analytics/_index.md
@@ -448,6 +448,24 @@ $span->setTag(Tag::ANALYTICS_KEY, true);
 
 ```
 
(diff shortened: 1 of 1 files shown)


cgilmour:
Good spotting.
For the C++ tracer, there's no equivalent to those integrations or DB services, so I didn't have anything to add for them.
I could add that, but it'd mostly say "this is not available for C++, and you must use custom instrumentation".
Happy to do another PR for that if it's more complete and consistent to add it.

PR #4671, opened 2019-05-24
Aws ssm

l0k0ms:
### What does this PR do?
Removes AWS SSM FAQ in order to merge it to the main AWS EC2 documentation page.

### Motivation
Those instructions are valid metrics collection setup and as such should be in the integrations page directly.


l0k0ms:
It's because it's used as resources for the update i made in dogweb on AWS EC2.

PR #4702, opened 2019-05-31
Add missing information to status processor

NoisomePossum:
### What does this PR do?
While testing some logs configs for a customer I noticed that logs with the status "out" were being assigned the status "OK". After checking with the logs team I found that this is because the status string begins with "o" which is logical based on the other rules listed here but it isn't currently documented.

### Motivation
Hoping to avoid confusion/tickets from customers looking at the documentation and not seeing this bit of information.

### Preview link

<!-- This is the base preview link. This currently only works if you are in the Datadog organization and working off of a branch - it will not work with a fork. 

Replace the branch name and add the complete path:
https://docs-staging.datadoghq.com/possum/status-remapper/content/en/logs/processing/processors.md


NoisomePossum:
Preview link: https://docs-staging.datadoghq.com/possum/status-remapper/content/en/logs/processing/processors.md

PR #4723, opened 2019-06-04
various updates for speed improvements during preview builds

davidejones:
### What does this PR do?
This PR makes a number of changes to the preview site build in an attempt to increase the speed of the build.

So far with these changes the build takes roughly 3mins versus the 6mins it has been lately.

The changes involve:
- downloading the automated files from the last master artifact instead of building from scratch
- removing some duplicate s3 calls
- removing creating artifacts during previews (not needed)
- reduce hugo build time by moving nav generation for mobile

### Motivation
To speed up preview site builds

### Preview link
https://docs-staging.datadoghq.com/david.jones/updatedeploy/

### Additional Notes
This just needs some QA to make sure the preview still works as normal

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -35,21 +35,23 @@ build_preview:
     URL: ${PREVIEW_DOMAIN}
     CONFIG: ${PREVIEW_CONFIG}
     MESSAGE: ":gift_heart: Your preview site is available!\nNow running tests..."
-    UNTRACKED_EXTRAS: "data,content/en/integrations"
   script:
     - post_dd_event "documentation deploy ${CI_COMMIT_REF_NAME} started" "${CI_PROJECT_URL}/pipelines/${CI_PIPELINE_ID}" "info"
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -35,21 +35,23 @@ build_preview:
...
+    #- sync_integration_descriptions
+    - sync_integration_descriptions_cached
     - placehold_translations
     - build_hugo_site
     - build_hugo_site_ja
     - minify_html
-    - collect_static_assets
+    #- collect_static_assets
+    # remove service_checks json as we don't need to s3 push that..
+    - rm -rf data/service_checks
```
l0k0ms: should we remove the the `data/metrics.*` as well?

davidejones: So i only removed the service checks as they are `.json` files and the s3 upload command sends json files but we don't need to send these so it was actually slowing the upload by uploading files we didn't need to.

PR #4724, opened 2019-06-04
PHP APM - explain agent connection timeout settings, and clarify how agent hostname can be correctly set

pawelchcki:
### What does this PR do?
List and explain agent connection timeout settings.

Clarify how agent hostname can be correctly set - since the order of execution matters and we already have configuration documentation that covers that topic - so instead of adding incomplete example. Lets link to that relevant part of PHP tracer documentation.


pawelchcki:
Thanks @cswatt! 🙇

PR #4768, opened 2019-06-11
David.jones/config updates

davidejones:
### What does this PR do?

This PR
- updates hugo version to the same as corp site
- removes the dependency on the user having to install hugo as its now installed via npm and called from there
- updates the config from being a single file to instead being in `config/` directory as multiple files. This allows us to have configs per language per environment e.g `params.fr.yaml` or `menu.ja.yaml` etc.
- fixes the heroku link that keeps coming up in link checker
- updates the placeholder generation to understand new config setup
- update `og:site_name` to pull from a variable now so we can change per lang
- update `meta_description` and `meta_title` variable name to not be nested under `home`
- moves `disclaimer` and `announcement_banner` to `params.en.yaml` so we can push it through transifex
- updates placeholder generation to pull disclaimer from new location

### Motivation

It started off with updating meta / opengraph tags but i discovered updating the config in this way gives us a number of benefits for doing language specific configuration and sending specific pieces of config to be translated via transifex.

### Preview link
https://docs-staging.datadoghq.com/david.jones/config-updates/


config/_default/params.en.yaml
```diff
@@ -0,0 +1,9 @@
+site_name: "Datadog Infrastructure and Application Monitoring"
+meta_image: datadog_logo_share_tt.png
+meta_title: Getting Started with Datadog
+meta_description: Datadog, the leading service for cloud-scale monitoring.
+disclaimer: ""
```
l0k0ms:
```suggestion
disclaimer: "This page is not yet available in English, we are working on its translation. <br> May you have any questions or feedback about our current translation project, <ahref=\"https://docs.datadoghq.com/fr/help/\">feel free to reach out to us!</a>"
```

PR #4807, opened 2019-06-16
Michaelw/browser logs

mstbbs:
### What does this PR do?
Adds browser logging to the docs site. The reason here is two-fold:
1. Dogfood the product to see how/what it would be like to capture all browser errors in Datadog (can we replace Sentry?) as well as track general performance and resource load times. This is a manual first pass based on Google's performance tracking recommendations. From what I can tell this API is well supported across browsers (back to ie10)

2. Build up enough logs to find a compelling story for the product announcement at Dash

If I were a good citizen I would have squashed this, but please leave feedback if you want this moved or done differently.

### Preview link
[site where this is installed](https://docs-staging.datadoghq.com/michaelw/browser-logs)

[browser logs in app](https://dd-corpsite.datadoghq.com/logs?cols=core_host%2Ccore_service&event=AQAAAWtlnM2y7hLfNwAAAABBV3RsbmJWaUFjV3pST05KZEl2Nw&from_ts=1560775416536&index=main&live=true&query=source%3Abrowser&stream_sort=desc&to_ts=1560776316536)


l0k0ms:
Hello, why removing the `yarn.lock` and the `makefile.config.example`  files in this PR ?

nsollecito:
@mstbbs does this new script need to be blocking? Can we use async/defer? 
Does it need to be loaded from a 3rd party domain? 
Can any/all of this be done at the bottom of the page load?

layouts/_default/baseof.html
```diff
@@ -1,6 +1,91 @@
...
+          };
+          // send the page stats
+          DD_LOGS.logger.info('Page loaded', pageStats);
+        } catch (e) {
+          DD_LOGS.logger.error('Page stats collection error', {'message': e});
+        }
+        try {
+          let pageResources = performance.getEntriesByType("resource");
+          if (pageResources) {
+            pageResources.forEach(el => {
```
davidejones: Can we avoid `forEach` as inline js isn't run through babel and its less compatible with older browsers.

mstbbs:
Resolved most of this. 
We want this to load early to capture errors in subsequent scripts. 
I have moved to async
I have moved the window.onload portion to the footer to reduce head congestion

PR #4834, opened 2019-06-19
Missing metrics script

ruthnaebeck:
### What does this PR do?
- Add missing metrics script to build pipeline

### Motivation
- Automate script

### Preview link
N/A

### Additional Notes

N/A

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -38,8 +38,8 @@ build_preview:
   script:
     - post_dd_event "documentation deploy ${CI_COMMIT_REF_NAME} started" "${CI_PROJECT_URL}/pipelines/${CI_PIPELINE_ID}" "info"
     - version_static_assets
-    #- sync_integration_descriptions
-    - sync_integration_descriptions_cached
+    - sync_integration_descriptions
+    #- sync_integration_descriptions_cached
     - placehold_translations
     - update_preview_baseurl
     - build_hugo_site_new
@@ -52,7 +52,7 @@ build_preview:
     #- remove_static_from_repo
     #- create_artifact
     #- create_artifact_untracked
-    #- ./local/bin/py/missing_metrics.py -k $(get_secret 'dd-demo-api-key') -p $(get_secret 'dd-demo-app-key') $(get_secret 'dd-corp-api-key') $(get_secret 'dd-corp-app-key')
+    - ./local/bin/py/missing_metrics.py -k $(get_secret 'dd-demo-api-key') -p $(get_secret 'dd-demo-app-key') -a $(get_secret 'dd_api_key') -b $(get_secret 'dd_app_key')
   only:
     - /.+?\/[a-zA-Z0-9_-]+/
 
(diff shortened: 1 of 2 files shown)


local/bin/py/missing_metrics.py
```diff
@@ -0,0 +1,73 @@
...
+def post_dd_metrics(metrics):
+# Post to Corpsite Datadog account
+  options = {
+      'api_key': 'x', # dd-corp-api-key
+      'app_key': 'x' # dd-corp-app-key
+  }
+  datadog.initialize(**options)
+  print(datadog.api.Metric.send(metrics))
+
+if __name__ == '__main__':
```
davidejones: in here add something like this

```
from optparse import OptionParser
parser = OptionParser(usage="usage: %prog [options]")
parser.add_option("-k", "--apikey", help="api key", default=None)
parser.add_option("-p", "--appkey", help="app key", default=None)
options, args = parser.parse_args()
```

Then just pass the `options` variable to the 3 different functions below.

In those functions you should be able to reference the data by `options.appkey` or `options.apikey` etc.

local/bin/py/missing_metrics.py
```diff
@@ -0,0 +1,73 @@
+import datadog
```
davidejones: can you add `#!/usr/bin/env python3` at the top too?

.gitlab-ci.yml
```diff
@@ -52,6 +52,7 @@ build_preview:
     #- remove_static_from_repo
     #- create_artifact
     #- create_artifact_untracked
+    #- ./local/bin/py/missing_metrics.py -k $(get_secret 'dd-demo-api-key') -p $(get_secret 'dd-demo-app-key') $(get_secret 'dd-corp-api-key') $(get_secret 'dd-corp-app-key')
```
davidejones: So then i think this would update to be like this. Only thing is i'm not sure we have dd_app_key setup i'll have to try look into that

```
./local/bin/py/missing_metrics.py -k $(get_secret 'dd-demo-api-key') -p $(get_secret 'dd-demo-app-key') -a $(get_secret 'dd_api_key') -b $(get_secret 'dd_app_key')
```

local/bin/py/missing_metrics.py
```diff
@@ -0,0 +1,73 @@
...
+from datadog import initialize, api
+import time
+import os
+import glob
+import csv
+
+def get_csv_metrics(env):
+  csv_files = glob.glob(env + 'integration/*/*metadata.csv')
```
davidejones: It should be in temp so something like this works

```
#!/usr/bin/env python3

import glob
import platform

tempdir = (
   "/tmp"
   if platform.system() == "Darwin"
   else tempfile.gettempdir()
)
for file in glob.glob(tempdir + "/extracted/**/*/*metadata.csv", recursive=True):
        print(file)
```

.gitlab-ci.yml
davidejones: also a bit further up we have

```
#- sync_integration_descriptions
- sync_integration_descriptions_cached
```

you might want to switch the comment the other way around so cached is off and the top one is on. This makes it so it pulls all the files from github on each preview site build instead of using an artifact. That way all the csv will be there in /tmp

PR #4870, opened 2019-06-25
make sure there is a default card id to help avoid issues with corp u…

davidejones:
### What does this PR do?
This PR adds a default id to integration cards using the url name should the `Params.name` field be empty or non existent.

### Motivation
Helps avoid issues with corp site being able to share some of this data.

### Preview link
https://docs-staging.datadoghq.com/david.jones/default-card-id/integrations/


l0k0ms:
for further context here, corpsite is pulling doc integration page for its own integration page, and there was some id issues. Which is now fixed with this PR

PR #4879, opened 2019-06-26
Update logs.md

mikelaning:
### What does this PR do?

Adds the necessary environment variable for configuring logs to hipaa endpoint

### DIFF
--- a/content/en/security/logs.md
+++ b/content/en/security/logs.md
@@ -44,6 +44,8 @@ logs_config:
   logs_no_ssl: false
 ```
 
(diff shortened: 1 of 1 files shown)


l0k0ms:
@mikelaning There is the DD EU site endpoint missing. We should add it.

@NBParis what's the endpoint for DD EU.

NBParis:
I pinged @DataDog/logs-intake and will update with the value.

NBParis:
Answers: We don't officially support HIPAA in EU yet which is why there no documented endpoint.

PR #4897, opened 2019-06-28
added images for dogweb gcp integrations

anshumgargdd:
### What does this PR do?
Added images so that they can be referenced in GCP integrations doc in dogweb

### Motivation
Images

### Additional Notes

Just added images. Shouldn't impact anything by itself


anshumgargdd:
Done

PR #4949, opened 2019-07-09
Fix documentation inconsistency with Trace Search env var in Python

gml3ff:
### What does this PR do?
Fixes a documentation inconsistency with Python. `ddtrace` documentation suggests using `DD_TRACE_ANALYTICS_ENABLED=true` for enable trace search whereas "regular" APM docs suggest using the older `DD_ANALYTICS_ENABLED=true` environment variable. 

The former is the correct environment variable to use in this context. This PR corrects the confusion.


gml3ff:
Link to relevant Python `ddtrace` docs for clarification: http://pypi.datadoghq.com/trace/docs/advanced_usage.html#trace-search-analytics

PR #5013, opened 2019-07-18
New kube-state-metrics cluster checks faq

cswatt:
### What does this PR do?
Adds an FAQ for running the `kubernetes_state` check as a cluster check.

### Motivation
customer questions

### Preview link

https://docs-staging.datadoghq.com/cswatt/kube-state-metrics-cluster-check/agent/faq/kubernetes-state-cluster-check

### DIFF
--- a/content/en/agent/faq/_index.md
+++ b/content/en/agent/faq/_index.md
@@ -29,4 +29,4 @@ aliases:
 [8]: /agent/faq/why-should-i-install-the-agent-on-my-cloud-instances
 [9]: /agent/faq/kubernetes-secrets
 [10]: /agent/faq/windows-agent-ddagent-user
-[11]: 
+[11]: /agent/faq/kubernetes-state-cluster-check
--- a/content/en/agent/faq/_index.md
+++ b/content/en/agent/faq/_index.md
@@ -17,6 +17,7 @@ aliases:
 * [Why should I install the Datadog Agent on my cloud instances?][8]
 * [How do I use Kubernetes secrets to set my API key?][9]
 * [Datadog Windows Agent user][10]
+* [How do I run the kubernetes_state check as a cluster check?][11]
 
 [1]: /agent/faq/host-metrics-with-the-container-agent
 [2]: /agent/faq/downgrade-datadog-agent
@@ -28,3 +29,4 @@ aliases:
 [8]: /agent/faq/why-should-i-install-the-agent-on-my-cloud-instances
 [9]: /agent/faq/kubernetes-secrets
 [10]: /agent/faq/windows-agent-ddagent-user
+[11]: 
--- a/content/en/agent/faq/kubernetes-state-cluster-check.md
+++ b/content/en/agent/faq/kubernetes-state-cluster-check.md
@@ -34,7 +34,18 @@ The `kubernetes_state` check, which monitors `kube-state-metrics`, is enabled au
 	    ]
 	```
 
-2. Deploy the Datadog Cluster Agent with `DD_CLUSTER_CHECKS_ENABLED` set to `true`.
+	**Notes**: 8080 is the metrics port. The `prometheus_timeout` and `min_collection_interval` are set to 30 because this setup assumes very large payloads, so these metrics are being collected at a lower granularity. Set `send_pod_phase_service_checks` to `true` if you want the pod phase as a service check. Telemetry is only activated in v4.6.1+ of the Datadog `kube-state-metrics` check.
+
+2. Deploy the Datadog Cluster Agent with the following variables:
+
+	```
+		- name: DD_CLUSTER_CHECKS_ENABLED
+		  value: true
+		- name: DD_EXTRA_CONFIG_PROVIDERS
+		  value: "kube_services"
+		- name: DD_EXTRA_LISTENERS
+		  value: "kube_services" 
+	```          
 
 3. Mount an empty file to override `/etc/datadog-agent/conf.d/kubernetes_state.d/auto_conf.yaml` in the Agent DaemonSet. Otherwise, the Autodiscovered check still runs and may still consume a significant amount of RAM. Configure Agent deployment to mount an empty directory in `/etc/datadog-agent/conf.d/kubernetes_state.d`
 
--- a/content/en/agent/faq/kubernetes-state-cluster-check.md
+++ b/content/en/agent/faq/kubernetes-state-cluster-check.md
@@ -1,0 +1,33 @@
+---
+title: How do I run the kubernetes_state check as a cluster check?
+kind: faq
+further_reading:
+- link: "agent/kubernetes"
+  tag: "Documentation"
+  text: "Kubernetes with Datadog"
+- link: "agent/autodiscovery"
+  tag: "Documentation"
+  text: "Using Autodiscovery with Kubernetes and Docker"
+- link: "agent/autodiscovery/clusterchecks"
+  tag: "Documentation"
+  text: "Running Cluster Checks with Autodiscovery"
+---
+
+The `kubernetes_state` check, which monitors `kube-state-metrics`, is enabled automatically by Autodiscovery. In larger Kubernetes clusters, this check may consume too much RAM. In this case, consider running the `kubernetes_state` check as a [cluster check][1].
+
+## Configuration
+
+Refer to the [Running Cluster Checks with Autodiscovery][1] documentation for more information. See options that begin with `clusterchecksDeployment` in the Helm chart [readme][2].
+
+**Note**: Mount an empty file to override `/etc/datadog-agent/conf.d/kubernetes_state.d/auto_conf.yaml` in the Agent DaemonSet. Otherwise, the Autodiscovered check still runs and may still consume a significant amount of RAM.
+
+## Background
+
+If you use Autodiscovery from the DaemonSet, one of your Agents (the one running on the same node as `kube-state-metrics`) runs the check and uses a significant amount of memory, while other Agents in the DaemonSet use far less. To prevent the outlier Agent from getting killed, you could increase the memory limit for all Agents, but this could be a waste. The alternative is to use the DaemonSet for lightweight checks to keep the general memory usage low, and use a small (e.g. 2-3 pods) dedicated deployment for heavy checks, where each pod has a large amont of RAM and only runs cluster checks.
+
+## Further Reading
+
+{{< partial name="whats-next/whats-next.html" >}}
+
+[1]: /agent/autodiscovery/clusterchecks
+[2]: https://github.com/helm/charts/blob/master/stable/datadog/README.md
--- a/content/en/agent/faq/kubernetes-state-cluster-check.md
+++ b/content/en/agent/faq/kubernetes-state-cluster-check.md
@@ -55,6 +55,40 @@ The `kubernetes_state` check, which monitors `kube-state-metrics`, is enabled au
 
 Refer to the [Running Cluster Checks with Autodiscovery][2] documentation for more information. See options that begin with `clusterchecksDeployment` in the Helm chart [README.md][3].
 
+### Verify check dispatching
+
+Run:
+
+```
+kubectl exec -it <datadog-cluster-agent-leader> agent clusterchecks
+```
+
+This command displays something like:
+
+```
+===== Checks on gke-my-cluster-default-pool-b969f074-npsw =====
+
+=== kubernetes_state check ===
+Source: kubernetes-services
+Instance ID: kubernetes_state:4a95df9b407f14d7
+empty_default_hostname: true
+kube_state_url: http://10.59.249.74:8080/metrics
+min_collection_interval: 30
+prometheus_timeout: 30
+send_pod_phase_service_checks: false
+tags:
+- kube_service:kube-state-metrics
+- kube_namespace:kube-system
+- cluster_name:test-cluster
+telemetry: true
+~
+Init Config:
+{}
+===
+```
+
+The Agent running on the node returned above should also have a `kubernetes_state` check in its `agent status`
+
 ## Background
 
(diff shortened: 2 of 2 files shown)


content/en/agent/faq/_index.md
```diff
@@ -28,3 +29,4 @@ aliases:
 [8]: /agent/faq/why-should-i-install-the-agent-on-my-cloud-instances
 [9]: /agent/faq/kubernetes-secrets
 [10]: /agent/faq/windows-agent-ddagent-user
+[11]:
```
l0k0ms:
```suggestion
[11]: /agent/faq/kubernetes-state-cluster-check
```

content/en/agent/faq/kubernetes-state-cluster-check.md
```diff
@@ -0,0 +1,33 @@
...
+  text: "Kubernetes with Datadog"
+- link: "agent/autodiscovery"
+  tag: "Documentation"
+  text: "Using Autodiscovery with Kubernetes and Docker"
+- link: "agent/autodiscovery/clusterchecks"
+  tag: "Documentation"
+  text: "Running Cluster Checks with Autodiscovery"
+---
+
+The `kubernetes_state` check, which monitors `kube-state-metrics`, is enabled automatically by Autodiscovery. In larger Kubernetes clusters, this check may consume too much RAM. In this case, consider running the `kubernetes_state` check as a [cluster check][1].
```
l0k0ms:
```suggestion
The `kubernetes_state` check, which monitors `kube-state-metrics`, is enabled automatically by [Autodiscovery](/agent/autodiscovery/). In larger Kubernetes clusters, this check may consume too much RAM. In this case, consider running the `kubernetes_state` check as a [cluster check][1].
```

content/en/agent/faq/kubernetes-state-cluster-check.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Configuration
+
+Refer to the [Running Cluster Checks with Autodiscovery][1] documentation for more information. See options that begin with `clusterchecksDeployment` in the Helm chart [readme][2].
```
l0k0ms:
```suggestion
Refer to the [Running Cluster Checks with Autodiscovery][1] documentation for more information. See options that begin with `clusterchecksDeployment` in the Helm chart [README.md][2].
```

content/en/agent/faq/kubernetes-state-cluster-check.md
```diff
@@ -0,0 +1,33 @@
...
+
+**Note**: Mount an empty file to override `/etc/datadog-agent/conf.d/kubernetes_state.d/auto_conf.yaml` in the Agent DaemonSet. Otherwise, the Autodiscovered check still runs and may still consume a significant amount of RAM.
+
+## Background
+
+If you use Autodiscovery from the DaemonSet, one of your Agents (the one running on the same node as `kube-state-metrics`) runs the check and uses a significant amount of memory, while other Agents in the DaemonSet use far less. To prevent the outlier Agent from getting killed, you could increase the memory limit for all Agents, but this could be a waste. The alternative is to use the DaemonSet for lightweight checks to keep the general memory usage low, and use a small (e.g. 2-3 pods) dedicated deployment for heavy checks, where each pod has a large amont of RAM and only runs cluster checks.
```
hkaj: `this could be a waste` --> `this would be a waste`, no doubt there 😄

hkaj:
@cswatt here are more detailed instructions in case you want to list that.

## install ksm check as a CLC

1. Deploy KSM with cluster check annotations on the service:
```yaml
ad.datadoghq.com/service.check_names: '["kubernetes_state"]'
ad.datadoghq.com/service.init_configs: '[{}]'
ad.datadoghq.com/service.instances: |
    [
    {
        "kube_state_url": "http://%%host%%:8080/metrics",
        "prometheus_timeout": 30,
        "min_collection_interval": 30,
        "send_pod_phase_service_checks": false,
        "telemetry": true
    }
    ]
```
2. Deploy the Datadog Cluster Agent with `DD_CLUSTER_CHECKS_ENABLED` set to `true`
3. Configure agent deployment to mount empty dir in `/etc/datadog-agent/conf.d/kubernetes_state.d`:
```yaml
...
    volumeMounts:
    ...
      - name: empty-dir
      mountPath: /etc/datadog-agent/conf.d/kubernetes_state.d
    ...
  volumes:
    - name: empty-dir
    emptyDir: {}
    ...
```
4. Deploy the agent

## verify check dispatching

`kexec -it <datadog-cluster-agent-leader> agent clusterchecks` should display something like:

```
===== Checks on gke-my-cluster-default-pool-b969f074-npsw =====

=== kubernetes_state check ===
Source: kubernetes-services
Instance ID: kubernetes_state:4a95df9b407f14d7
empty_default_hostname: true
kube_state_url: http://10.59.249.74:8080/metrics
min_collection_interval: 30
prometheus_timeout: 30
send_pod_phase_service_checks: false
tags:
- kube_service:kube-state-metrics
- kube_namespace:kube-system
- cluster_name:test-cluster
telemetry: true
~
Init Config:
{}
===
```

And the agent running on the node returned above should also have a kubernetes_state check in its `agent status`.

hkaj:
@CharlyF has written an even more complete guide on this GH issue https://github.com/DataDog/datadog-agent/issues/3923

content/en/agent/faq/kubernetes-state-cluster-check.md
```diff
@@ -0,0 +1,102 @@
...
+
+1. Deploy `kube-state-metrics` with cluster check annotations on the service:
+
+	```
+	ad.datadoghq.com/service.check_names: '["kubernetes_state"]'
+	ad.datadoghq.com/service.init_configs: '[{}]'
+	ad.datadoghq.com/service.instances: |
+	    [
+	    {
+	        "kube_state_url": "http://%%host%%:8080/metrics",
```
CharlyF: We should have a disclaimer of the port. It's worth mentionning that 8080 is the metrics port.

content/en/agent/faq/kubernetes-state-cluster-check.md
```diff
@@ -0,0 +1,102 @@
...
+	        "prometheus_timeout": 30,
```
CharlyF: the timeout and min collection interval are set to 30 here because we have very large payloads, but it's worth a disclaimer as it means they would only collect their metrics at a lower granularity.
I would suggest mentioning them or pointing to the doc for the ksm check, but no need to lower the defautl granularity without explanation.

content/en/agent/faq/kubernetes-state-cluster-check.md
```diff
@@ -0,0 +1,102 @@
...
+	        "min_collection_interval": 30,
+	        "send_pod_phase_service_checks": false,
+	        "telemetry": true
```
CharlyF: same for those two options. They might want the pod phase as a service check.
Telemetry is only activated in the latest version of the Datadog KSM check

content/en/agent/faq/kubernetes-state-cluster-check.md
```diff
@@ -0,0 +1,102 @@
...
+	    }
+	    ]
+	```
+
+2. Deploy the Datadog Cluster Agent with `DD_CLUSTER_CHECKS_ENABLED` set to `true`.
```
CharlyF: It also requires
```
        - name: DD_EXTRA_CONFIG_PROVIDERS
          value: "kube_services"
        - name: DD_EXTRA_LISTENERS
          value: "kube_services" 
```

content/en/agent/faq/kubernetes-state-cluster-check.md
```diff
@@ -0,0 +1,102 @@
...
+	    ...
+	  volumes:
+	    - name: empty-dir
+	    emptyDir: {}
+	    ...
+	```
+
+4. Deploy the Agent.
+
+Refer to the [Running Cluster Checks with Autodiscovery][2] documentation for more information. See options that begin with `clusterchecksDeployment` in the Helm chart [README.md][3].
```
CharlyF: Maybe we could be more verbose about the options to use in the agent ?

content/en/agent/faq/kubernetes-state-cluster-check.md
```diff
@@ -0,0 +1,102 @@
...
+Init Config:
+{}
+===
+```
+
+The Agent running on the node returned above should also have a `kubernetes_state` check in its `agent status`
+
+## Background
+
+If you use Autodiscovery from the DaemonSet, one of your Agents (the one running on the same node as `kube-state-metrics`) runs the check and uses a significant amount of memory, while other Agents in the DaemonSet use far less. To prevent the outlier Agent from getting killed, you could increase the memory limit for all Agents, but this wastes resources. The alternative is to use the DaemonSet for lightweight checks to keep the general memory usage low, and use a small (e.g. 2-3 pods) dedicated deployment for heavy checks, where each pod has a large amont of RAM and only runs cluster checks.
```
CharlyF: Good to give background, but we need to document or point to what the full workload looks like, so they can clearly see what the cluster check worker is compared to the regular agent.

content/en/agent/faq/kubernetes-state-cluster-check.md
hkaj: agreed, we can remove this line and the next by default, and maybe add a section to the faq about the check timing out and the need to decrease the frequency and increasing the timeout.

content/en/agent/faq/kubernetes-state-cluster-check.md
hkaj: Yeah we should aim to explain all options, but i think for an FAQ this is ok. We'll see what follow up questions users ask, and go from there.

Simwar:
It is not very clear that the customer needs to use this solution with another daemonset only deployed on a few nodes with higher memory limits.

content/en/agent/faq/kubernetes-state-cluster-check.md
```diff
@@ -0,0 +1,113 @@
...
+4. Deploy the Agent.
+
+Refer to the [Running Cluster Checks with Autodiscovery][2] documentation for more information. See options that begin with `clusterchecksDeployment` in the Helm chart [README.md][3].
+
+### Verify check dispatching
+
+Run:
+
+```
+kubectl exec -it <datadog-cluster-agent-leader> agent clusterchecks
```
kivagant-ba: How to find the leader in the pods list?

content/en/agent/faq/kubernetes-state-cluster-check.md
```diff
@@ -0,0 +1,113 @@
...
+- link: "agent/autodiscovery/clusterchecks"
+  tag: "Documentation"
+  text: "Running Cluster Checks with Autodiscovery"
+---
+
+The `kubernetes_state` check, which monitors `kube-state-metrics`, is enabled automatically by [Autodiscovery][1]. In larger Kubernetes clusters, this check may consume too much RAM. In this case, consider running the `kubernetes_state` check as a [cluster check][2].
+
+## Configuration
+
+1. Deploy `kube-state-metrics` with cluster check annotations on the service:
```
kivagant-ba: Please change to

```
1. Deploy `kube-state-metrics` with cluster check annotations on the Kubernetes service:


```
apiVersion: v1
kind: Service
metadata:
  annotations:
    # ... others
    ad.datadoghq.com/service.check_names: '["kubernetes_state"]'
    ad.datadoghq.com/service.init_configs: '[{}]'
    ad.datadoghq.com/service.instances: |
      [
      {
          "kube_state_url": "http://%%host%%:8080/metrics",
          "prometheus_timeout": 30,
          "min_collection_interval": 30,
          "send_pod_phase_service_checks": false,
          "telemetry": true
      }
      ]
```

File: content/en/agent/faq/kubernetes-state-cluster-check.md [hunk 13]
```diff
@@ -0,0 +1,113 @@
...
+
```
kivagant-ba: Helm chart `kube-state-metrics` 2.0.0 doesn't support custom service annotations, they are hardcoded:

```
apiVersion: v1
kind: Service
metadata:
  name: {{ template "kube-state-metrics.fullname" . }}
  labels:
    app.kubernetes.io/name: {{ template "kube-state-metrics.name" . }}
    helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
    app.kubernetes.io/instance: "{{ .Release.Name }}"
    app.kubernetes.io/managed-by: "{{ .Release.Service }}"
  {{- if .Values.prometheusScrape }}
  annotations:
    prometheus.io/scrape: '{{ .Values.prometheusScrape }}'
  {{- end }}
```

content/en/agent/faq/kubernetes-state-cluster-check.md
```diff
@@ -0,0 +1,113 @@
...
+Instance ID: kubernetes_state:4a95df9b407f14d7
+empty_default_hostname: true
+kube_state_url: http://10.59.249.74:8080/metrics
+min_collection_interval: 30
+prometheus_timeout: 30
+send_pod_phase_service_checks: false
+tags:
+- kube_service:kube-state-metrics
+- kube_namespace:kube-system
+- cluster_name:test-cluster
```
kivagant-ba: Offtopic: is there any way to add a custom tag to the list? All our DD dashboards configured to read `kubernetescluster"`tag instead of `cluster_name`.

kivagant-ba: I tried to use the section: https://github.com/helm/charts/blob/master/stable/datadog/values.yaml#L122
But looks it doesn't work, at least in this case when a separate kube-state-metrics pod is running. Also I found that tags value from the Helm chart is translated into this format that looks wrong according to the documentation:

```
    - name: DD_TAGS
      value: '[map[kubernetescluster:k8s-blue-staging]]'
```

Does this looks like a bug?

kivagant-ba: Found working configuration:

```
clusterAgent:
...
  env:
# This also works but it's better not to break the default value 
# but add another one for backward compatibility
# - name: DD_CLUSTER_CHECKS_CLUSTER_TAG_NAME 
# value: "kubernetescluster" 
    - name: DD_CLUSTER_CHECKS_EXTRA_TAGS
      value: "kubernetescluster:my-lovely-cluster"

```

kivagant-ba:
For people who will be looking for the Helm chart configuration, check the comment:

https://github.com/DataDog/datadog-agent/issues/3923#issuecomment-519962498

cswatt:
updated to webpack, closing this one. see new pr
https://github.com/DataDog/documentation/pull/5232

PR #5050, opened 2019-07-24
update api/v1/logs-queries/list to use GET

pmbauer:
GET and POST both work.
Is POST the correct usage here since this is a read?


pmbauer:
closing as this pull is incorrect; we shouldn't be modifying the response based on request body content, so POST is the more correct verb

PR #5145, opened 2019-08-05
Webpack migration

zbayoff:
### What does this PR do?
The migration from Gulp to Webpack to compile js, scss->css and assets for the Docs build pipeline. 

### Motivation
Webpack is a powerful and customizable bundler that is used to create js and scss->css minified bundles. In this PR, the new webpack config includes [Babel](https://babeljs.io/) to transpile ES6 modern js code into compatible js for older browsers, minifiers for js and css, eslintrc files configured to AirBnb js style guide for better js code linting. This work has already been performed on Corp site and is based off of [Victor Hugo](https://github.com/netlify-templates/victor-hugo), a project merging Hugo and Webpack together. 

### Preview link
https://docs-staging.datadoghq.com/zach/webpack-migration/

### Trello Card
https://trello.com/c/H7pWQrdv/3795-docs-migration-to-webpack-4

### Additional Notes
The current Docs build includes hashing of image names in order to prevent the older cached imaged from being displayed. This work is not so easily replicable with Webpack and Hugo, so for now, we do not include this functionality. If an image is replaced with a newer image of the same name, it must be purged from Imgix, like we do on Corpsite.


.gitlab-ci.yml
```diff
@@ -6,7 +6,7 @@ stages:
 variables:
   PREVIEW_CONFIG: "config/preview.yaml"
   LIVE_CONFIG: "config/live.yaml"
-  ARTIFACT_RESOURCE: "public"
+  ARTIFACT_RESOURCE: "dist"
```
anagio: `corp-ci` references the `public` folder, maybe better to leave all things as `public` in docs so there's less things changing?

zbayoff: I accounted for that in a branch on corp-ci, wanted to aim for consistency across docs and corp. https://github.com/DataDog/corp-ci/pull/44

zbayoff: I was trying to aim for consistency between corp and docs, I've already made a branch for needed changes to corp-ci. I can make the output folder `public` but I'll also have to change webpack configs. Maybe we can change it to `dist` after the merge then?

l0k0ms:
It seems that the Search page is broken now: https://docs-staging.datadoghq.com/zach/webpack-migration/search/?s=agent

davidejones:
not blocking but i noticed we have `RUN_GULP := true` in the example makefile `Makefile.config.example` we should probably remove or change to run webpack?

davidejones:
@zbayoff we talked a bit about the image problem when you encountered it but now I'm thinking it through this might be more of a problem.

The images on a preview site and production both point to the same place `https://datadog-docs.imgix.net/`. I know we won't delete any images as we only `cp` or `sync` without delete command but I'm thinking about overwrites being an issue.

If i create a preview site that doesn't hash images and update an existing image say the Datadog logo will it now update production as well as my preview site because they both share the same location? Where as before them both sharing the same location was fine because it created a unique hashed name when contents changed? 

We might want to test this with a discreet existing image on the site to see if its an issue.

@nsollecito Do you know how long it would take to do a separate imgix/bucket for docs previews/staging assets like we do on corp? I feel like that would probably be the best resolution and move to being more similar between our sites.

PR #5164, opened 2019-08-07
Making index filter required

l0k0ms:
### What does this PR do?
When updating an index, the index filter parameter is required.

### Motivation
Hotjar Feedback


l0k0ms:
This is actually a bug that is going to be fixed.

PR #5202, opened 2019-08-12
adding a complex guide on adding metadata to spans and using it in the UI

omris1:
### What does this PR do?
Adds a new guide to APM about adding metadata to spans and using it in the UI.

### Motivation
The ongoing onboarding efforts on APM.

### Preview link
/tracing/guide/add_span_md_and_graph_it/

https://docs-staging.datadoghq.com/omri/span_md/tracing/guide/add_span_md_and_graph_it/

### Additional Notes
1. Would love a preview link before this goes live and after a formal review to share with the APM engineering group and see about modification to the code blocks cc @palazzem 
2. Trace Analytics will need to be updated to App Analytics @priyanshi-gupta


content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,287 @@
...
+```
+
+[1]: https://github.com/DataDog/dd-trace-java/blob/master/dd-trace-api/src/main/java/datadog/trace/api/DDTags.java
+[2]: https://github.com/opentracing/opentracing-java/blob/master/opentracing-api/src/main/java/io/opentracing/tag/Tags.java
+{{% /tab %}}
+{{% tab "Python" %}}
+
+Add metadata directly to a span by calling `set_tag`. For example, with the following route handler:
+
+```python
```
palazzem: Do we really need that section for most of the languages? Isn't enough showing only the second snippet about "how to add custom tags"?

In Java we have:
> Custom metadata may be set for auto-instrumentation by grabbing the active span out of the global tracer and setting a tag with `setTag`

That I think is cool because it shows:
- you have auto-instrumentation enabled
- you enrich auto-instrumentation without doing anything more
- having two code-blocks could be misleading. It may let people think you need to do "a lot of things" to achieve that, while if we focus only on a) retrieving the span b) adding the tag, it will be straightforward

This is just a thought, in general I prefer to keep at minimum the code we show, unless it's really needed to understand the topic.

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,287 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet
+class ServletImpl extends AbstractHttpServlet {
+  @Override
+  void doGet(HttpServletRequest req, HttpServletResponse resp) {
+    final Span span = GlobalTracer.get().activeSpan();
+    if (span != null) {
+      span.setTag("customer.id", 12345);
```
palazzem: nitpick: later in the screenshot I see the tag is called `customer_id`. I'd say to be consistent and use the same tag name everywhere so the example is "real".

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,287 @@
...
+      span.setTag("http.url", "/login");
```
palazzem: Do we need to show this second tag? I think it's not useful for the scope of this doc page? If we want to show how to add a facet for `customer_id`, let's show only `customer_id`.

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,287 @@
...
+  if current_span:
+    current_span.set_tag('customer.id', customer_id)
+```
+
+{{% /tab %}}
+{{% tab "Ruby" %}}
+
+Add metadata directly to `Datadog::Span` objects by calling `#set_tag`:
+
+```ruby
```
palazzem: If you agree with the previous comment, we can remove this section too.

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,287 @@
...
+end
+```
+
+Access the current active span from any method within your code. **Note**: If the method is called and there is no span currently active, `active_span` is `nil`.
+
+```ruby
+# e.g. adding tag to active span
+
+current_span = Datadog.tracer.active_span
+current_span.set_tag('<TAG_KEY>', '<TAG_VALUE>') unless current_span.nil?
```
palazzem: Let's not use a placeholder. Let's use the real example that is `customer_id` as we wrote for Java.

Also you can copy the previous example such as:
```ruby
# An example of a Sinatra endpoint,
# with Datadog auto-instrumentation enabled.
get '/posts' do
  current_span = Datadog.tracer.active_span
  current_span.set_tag('customer_id', 12345) unless current_span.nil?
end
```

Side note, we can make it parametrized as we did for the Python section to avoid using `12345` that is not really meaningful as example. cc @delner that can help with a better example.

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,287 @@
...
+package main
+
+import (
+    "log"
+    "net/http"
+
+    "gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+func handler(w http.ResponseWriter, r *http.Request) {
```
palazzem: We may not need this section, if you agree to keep the code at minimum.

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,287 @@
...
+    // Retrieve a span for a web request attached to a Go Context.
+    if span, ok := tracer.SpanFromContext(r.Context()); ok {
+        // Set tag
+        span.SetTag("http.url", r.URL.Path)
```
palazzem: If in other examples we use `customer_id`, let's use `customer_id`.

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,287 @@
...
+}
+```
+
+{{% /tab %}}
+{{% tab "Node.js" %}}
+
+
+Add metadata directly to span objects by calling `setTag` or `addTags`:
+
+```javascript
```
palazzem: Same here.

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,287 @@
...
+```
+
+Access the current active span from any method within your code. **Note**: If the method is called and there is no span currently active, `tracer.scope().active()` returns `null`.
+
+```javascript
+// e.g. adding tag to active span
+
+const span = tracer.scope().active()
+
+span.setTag('<TAG_KEY>', '<TAG_VALUE>')
```
palazzem: Let's use a real endpoint with this snippet to add the `customer_id` (same comment as before).

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,287 @@
...
+Add metadata directly to a `Datadog.Trace.Span` object by calling `Span.SetTag()`. For example:
+
+```csharp
+using Datadog.Trace;
+
+// access the active scope through the global tracer (can return null)
+var scope = Tracer.Instance.ActiveScope;
+
+// add a tag to the span
+scope.Span.SetTag("<TAG_KEY>", "<TAG_VALUE>");
```
palazzem: Same here.

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,287 @@
...
+
+
+Add metadata directly to a `DDTrace\Span` object by calling `Span::setTag()`. For example:
+
+```php
+// Get the currently active span (can be null)
+$span = \DDTrace\GlobalTracer::get()->getActiveSpan();
+if (null !== $span) {
+  // Add a tag to the span
+  $span->setTag('<TAG_KEY>', '<TAG_VALUE>');
```
palazzem: Same here.

content/en/tracing/guide/add_span_md_and_graph_it.md
omris1: Agreed, fixing now

content/en/tracing/guide/add_span_md_and_graph_it.md
omris1: Agreed, fixing now.

content/en/tracing/guide/add_span_md_and_graph_it.md
omris1: Agreed. I am also modifying the second Ruby block to call out `client_id` and not <> brackets for consistency.

content/en/tracing/guide/add_span_md_and_graph_it.md
omris1: Sorry, addressed this in the previous comment. I'll leave with `12345` unless @delner has a better example.

content/en/tracing/guide/add_span_md_and_graph_it.md
omris1: Agreed.

content/en/tracing/guide/add_span_md_and_graph_it.md
omris1: Agreed.

content/en/tracing/guide/add_span_md_and_graph_it.md
omris1: Agreed.

content/en/tracing/guide/add_span_md_and_graph_it.md
omris1: Agreed.

content/en/tracing/guide/add_span_md_and_graph_it.md
omris1: I completely agree and think that for the most part we can remove one of the blocks for every language. I'm just not confident I can do this myself. Would you be ok assigning this to folks on your team to just make the modifications needed? 

Purpose would be for each language tab to have the minimal code needed to add "customer_id" to a span and exemplify (as you say) that auto-instrumentation is great and and enriching it is easy.

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,273 @@
...
+  tag: "Documentation"
+  text: "Dive into your resource performance and traces"
+- link: "tracing/visualization/trace"
+  tag: "Documentation"
+  text: "Understand how to read a Datadog Trace"
+---
+
+{{< img src="tracing/guide/add_span_md_and_graph_it/span_md_6.gif" alt="comparaison 2" responsive="true" style="width:90%;">}}
+
+Datadog APM allows you to customize your traces to include any additional information you might need to maintain observability into your business. In this example a customer ID will be added to traces allowing the customers that have the slowest performance to be identified. Customization of traces is based on tags that seamlessly integrate APM with the rest of Datadog and come in the form of key:value pairs of metadata that is added to spans. You can read more about tags and their use in Datadog in the [Tags Overview][1] page.
```
apigirl:
```suggestion
Datadog APM allows you to customize your traces to include any additional information you might need to maintain observability into your business. In this example, a customer ID is added to traces allowing the customers that have the slowest performance to be identified. Customization of traces is based on tags that seamlessly integrate APM with the rest of Datadog and come in the form of `key:value` pairs of metadata added to spans. You can read more about tags and their use in Datadog in the [Tags Overview][1] page.
```

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,273 @@
...
+
+## Instrument your code with custom metadata
+
+1. **Follow the example to get your code instrumented**.
+
+    Depending on the language you are you using, you’ll need to set the metadata to add to your spans differently.
```
apigirl:
```suggestion
    Depending on the programming language you are you using, you’ll need to set the metadata to add to your spans differently.
```
we have to specify now, since we have docs in more than English

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,273 @@
...
+
+    **Note**: take note of the service and [resource names][2] you are working on, these will come in handy later. In this example the service is the Ruby server `web-store` and the resource (endpoint) is `ShoppingCartController#checkout`.
```
apigirl:
```suggestion
    **Note**: take note of the service and [resource names][2] you are working on, these will come in handy later. In this example, the service is the Ruby server `web-store` and the resource (endpoint) is `ShoppingCartController#checkout`.
```

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,273 @@
...
+
+{{< tabs >}}
+{{% tab "Java" %}}
+
+The Datadog UI uses tags to set span level metadata. A full list of these metadata can be found in the [Datadog][1] and [OpenTracing][2] APIs.
```
apigirl:
```suggestion
The Datadog UI uses tags to set span level metadata. A full list of the metadata can be found in the [Datadog][1] and [OpenTracing][2] APIs.
```

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,273 @@
...
+```python
+from ddtrace import tracer
+
+@app.route('/customer/<int:customer_id>')
+def handle_customer(customer_id):
+  with tracer.trace('web.request') as span:
+    span.set_tag('customer.id', customer_id)
+```
+
+The current span can be retrieved from the context in order to set its metadata. This way, if a span was started by the instrumentation, you can retrieve the span and add custom metadata. **Note**: If a span does not exist, `None` is returned:
```
apigirl:
```suggestion
The current span can be retrieved from the context in order to set its metadata. This way, if a span was started by the instrumentation, you can retrieve the span and add custom metadata. 
**Note**: If a span does not exist, `None` is returned:
```

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,273 @@
...
+  # get the active span in the context, put there by tracer.wrap()
+  current_span = tracer.current_span()
+  if current_span:
+    current_span.set_tag('customer.id', customer_id)
+```
+
+{{% /tab %}}
+{{% tab "Ruby" %}}
+
+Access the current active span from any method within your code. **Note**: If the method is called and there is no span currently active, `active_span` is `nil`.
```
apigirl:
```suggestion
Access the current active span from any method within your code. 
**Note**: If the method is called and there is no span currently active, `active_span` is `nil`.
```

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,273 @@
...
+func main() {
+    tracer.Start(tracer.WithServiceName("<SERVICE_NAME>"))
+    defer tracer.Stop()
+    http.HandleFunc("/posts", handler)
+    log.Fatal(http.ListenAndServe(":8080", nil))
+}
+```
+
+Datadog's integrations make use of the `Context` type to propagate the current active span.
+If you want to add span metadata via a tag attached to a `Context`, call the `SpanFromContext` function:
```
apigirl:
```suggestion
If you want to add span metadata using a tag attached to a `Context`, call the `SpanFromContext` function:
```

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,273 @@
...
+app.get('/posts', (req, res) => {
+  const span = tracer.startSpan('web.request')
+
+  span.addTags({
+    'customer_id': customer.id
+  })
+})
+```
+
+Access the current active span from any method within your code. **Note**: If the method is called and there is no span currently active, `tracer.scope().active()` returns `null`.
```
apigirl:
```suggestion
Access the current active span from any method within your code.
**Note**: If the method is called and there is no span currently active, `tracer.scope().active()` returns `null`.
```

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,273 @@
...
+  $span->setTag('customer_id', customer.id);
+}
+```
+
+**Note**: `Tracer::getActiveSpan()` returns `null` if there is no active span.
+
+{{% /tab %}}
+{{< /tabs >}}
+
+<div class="alert alert-info">You might have to wait a few minutes between deploying your updated code and seeing the new metadata in the datadog UI</div>
```
apigirl:
```suggestion
<div class="alert alert-info">You might have to wait a few minutes between deploying your updated code and seeing the new metadata in the Datadog UI</div>
```

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,278 @@
...
+  tag: "Documentation"
+  text: "Dive into your resource performance and traces"
+- link: "tracing/visualization/trace"
+  tag: "Documentation"
+  text: "Understand how to read a Datadog Trace"
+---
+
+{{< img src="tracing/guide/add_span_md_and_graph_it/span_md_6.gif" alt="comparaison 2" responsive="true" style="width:90%;">}}
+
+Datadog APM allows you to customize your traces to include any additional information you might need to maintain observability into your business. In this example, a customer ID is added to traces allowing the customers that have the slowest performance to be identified. Customization of traces is based on tags that seamlessly integrate APM with the rest of Datadog and come in the form of `key:value` pairs of metadata added to spans. You can read more about tags and their use in Datadog in the [Tags Overview][1] page.
```
andrewsouthard1: I don't love the link-out here to `Tags Overview`. That page isn't called Tags Overview and it doesn't provide `more about tags and their use in Datadog`. Ideally this page covers that, or the linked out page provides a better explanation.

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,278 @@
...
+  text: "Discover the list of services reporting to Datadog"
+- link: "tracing/visualization/resource"
+  tag: "Documentation"
+  text: "Dive into your resource performance and traces"
+- link: "tracing/visualization/trace"
+  tag: "Documentation"
+  text: "Understand how to read a Datadog Trace"
+---
+
+{{< img src="tracing/guide/add_span_md_and_graph_it/span_md_6.gif" alt="comparaison 2" responsive="true" style="width:90%;">}}
```
andrewsouthard1: The gif here would be clearer with a top list - the graph as is doesn't necessarily show immediately why I want it. I'd even go further and make it a top list - and export it to a dashboard. The dashboard/monitor use case should definitely be covered here.

content/en/tracing/guide/add_span_md_and_graph_it.md
andrewsouthard1: Agreed this should be as simple as possible and only the minimal code.

Also they should all read customer_id: 22495 to match the screenshot - and we should show that more quickly in this doc to see the immediate impact from code -> UI.

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,278 @@
...
+
+<div class="alert alert-info">You might have to wait a few minutes between deploying your updated code and seeing the new metadata in the Datadog UI</div>
+
+## Leverage the Datadog UI to search for your custom metadata
+
+2) **Go to the Services page** and click on the service that you added metadata to. **Scroll down and click on the specific resource** where the metadata was added in the Resource table.
+
+{{< img src="tracing/guide/add_span_md_and_graph_it/span_md_2.png" alt="comparaison 2" responsive="true" style="width:90%;">}}
+
+The span summary table allows to quickly identify any particular units within the operations that impact the resource. For example, here you can easily find the spans that take the largest portion of average operation time and investigate if they can be optimized.
```
andrewsouthard1: We should scrap the span summary section here. It doesn't have anything to do with metadata - if it showed an aggregate of metadata I'd agree but here let's get straight to the result of adding this otherwise we don't see the benefit. Showing Trace Analytics in a dashboard by metadata is way more powerful/under-utilized.

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,278 @@
...
+
+3) **Scroll down to the Traces table**
```
andrewsouthard1: Similarly - this is turning into a walkthrough of the resource page and not the benefits of adding trace metadata. I think you can scrap this as well in favor of dashboarding/monitoring on metadata

content/en/tracing/guide/add_span_md_and_graph_it.md
delner: Parameter as in a URL parameter from Sinatra? I don't think it'd add much, but if you really think that's better I can figure something out.

content/en/tracing/guide/add_span_md_and_graph_it.md
omris1: I think when possible they should have a parameterized customer ID, not look like we're hard coding a number, as Manu mentioned in the note to @delner below...

content/en/tracing/guide/add_span_md_and_graph_it.md
omris1: Can you give a good example for a dashboard widget I can easily build in the demo app with customer_id and a trace.* metric? I looked for good public about this and couldn't find any (unless you mean export from the App Analytics screen?).

content/en/tracing/guide/add_span_md_and_graph_it.md
omris1: Done.

content/en/tracing/guide/add_span_md_and_graph_it.md
omris1: How's the new gif?

content/en/tracing/guide/add_span_md_and_graph_it.md
omris1: Removed this sentence mostly because that link doesn't include any additional information on top of what's already here.

content/en/tracing/guide/add_span_md_and_graph_it.md
palazzem: We can add a comment before the parameter to say the value they will find in the screenshot later. Let me work on this.

content/en/tracing/guide/add_span_md_and_graph_it.md
omris1: Spoke with Andrew about this IRL, made a last adjustment to resolve.

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,271 @@
...
+1) **Follow the example to get your code instrumented**.
+
+Depending on the programming language you are you using, you’ll need to set the metadata to add to your spans differently.
+
+**Note**: take note of the service and [resource names][1] you are working on, these will come in handy later. In this example, the service is the Ruby server `web-store` and the resource (endpoint) is `ShoppingCartController#checkout`.
+
+{{< tabs >}}
+{{% tab "Java" %}}
+
+The Datadog UI uses tags to set span level metadata. A full list of the metadata can be found in the [Datadog][1] and [OpenTracing][2] APIs.
```
palazzem: @omris1 I'm going to remove this reference. I think it's misleading: "A full list of the metadata can be found in" -> if I read that, it means that this is the list ("full") that I can use to set tags. But there, we find only "special tags" with special meaning that is not the goal of this section.

Here we want to describe that "any trace can be tagged with your custom tag, and we can search and do analytics with infinite cardinality".

Doing the change but let me know if you think it's better I rollback.

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,271 @@
...
+```java
+import io.opentracing.Tracer;
+import io.opentracing.util.GlobalTracer;
+
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet
+class ServletImpl extends AbstractHttpServlet {
```
palazzem: I'm changing all the examples to use `ShoppingCartController`. This makes everything consistent.

content/en/tracing/guide/add_span_md_and_graph_it.md
palazzem: Addressed.

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,251 @@
...
+var scope = Tracer.Instance.ActiveScope;
+
+// add a tag to the span
+scope.Span.SetTag("customer_id", customer.id);
+```
+
+**Note**: `Datadog.Trace.Tracer.Instance.ActiveScope` returns `null` if there is no active span.
+
+{{% /tab %}}
+{{% tab "PHP" %}}
```
palazzem: This is incomplete, adding in a few.

content/en/tracing/guide/add_span_md_and_graph_it.md
```diff
@@ -0,0 +1,251 @@
...
+    // customer_id -> 254889
+    span.setTag('customer.id', req.params.customer_id)
+  }
+
+  // [...]
+})
+```
+
+{{% /tab %}}
+{{% tab ".NET" %}}
```
palazzem: This is incomplete, adding in a few.

PR #5251, opened 2019-08-19
Algolia in staging

l0k0ms:
### What does this PR do?
Allows to test new algolia search config in staging with a manual job to update the indexes.

It uses a new Gitlab script added here: https://github.com/DataDog/corp-ci/pull/51

The idea is now we have two algolia configs one for the prod and one for staging. On normal pace only prod is used as usual, but if an edit is made to the staging one, a dd employee can manually run a job in the CI to update the algolia staging index.

### Motivation

Being able to test search behavior on staging.

### Preview link
* https://docs-staging.datadoghq.com/gus/algolia-in-staging/


ruthnaebeck:
What's the best way to test this?

l0k0ms:
@ruthnaebeck it was tested on staging and it works (since the search works on staging) the best way to test this is to merge it in prod. It will be the last time we have to merge an algolia change in prod. But the config is the same as the current one, only the config path is changing for prod. So i would say *a priori* it's pretty safe.

PR #5265, opened 2019-08-20
fix history state bug

zbayoff:
### What does this PR do?
Fix history issue when using back/forward browser buttons between async and non-async pages. 

https://trello.com/c/402UnG42/3826-docs-history-back-button-issue

### Preview link
https://docs-staging.datadoghq.com/zach/js-fixes/

### DIFF
--- a/src/scripts/datadog-docs.js
+++ b/src/scripts/datadog-docs.js
@@ -28,6 +28,8 @@ gtag('config', gaTag);
 
 $(document).ready(function () {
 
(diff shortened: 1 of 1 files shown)


zbayoff:
@nsollecito @davidejones I know David updated the `corp-ci` repo to not delete `js` or `css` files, so this shouldn't break when merged into master, but we may want to still be aware of how long it takes for the `html` file to reflect the change with the new `js` file's hash change.

PR #5313, opened 2019-08-27
Updated the doc for the env DD_PROCESS_AGENT_ENABLED when set to false

Simwar:
### What does this PR do?
It updates the doc for the env DD_PROCESS_AGENT_ENABLED when set to false

### Motivation
https://github.com/DataDog/datadog-agent/issues/4077

### DIFF
--- a/content/en/agent/docker/_index.md
+++ b/content/en/agent/docker/_index.md
@@ -79,7 +79,7 @@ Optional collection Agents are disabled by default for security or performance r
 |----------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `DD_APM_ENABLED`           | Enable [trace collection][7] with the trace Agent.                                                                                                        |
 | `DD_LOGS_ENABLED`          | Enable [log collection][8] with the logs Agent.                                                                                                            |
-| `DD_PROCESS_AGENT_ENABLED` | Enable [live process collection][9] with the process Agent. The [live container view][10] is already enabled by default if the Docker socket is available. If set to `false`, will disable the live process collection][9] and the [live container view][10]|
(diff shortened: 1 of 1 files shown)


content/en/agent/docker/_index.md
```diff
@@ -79,7 +79,7 @@ Optional collection Agents are disabled by default for security or performance r
 |----------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `DD_APM_ENABLED`           | Enable [trace collection][7] with the trace Agent.                                                                                                        |
 | `DD_LOGS_ENABLED`          | Enable [log collection][8] with the logs Agent.                                                                                                            |
-| `DD_PROCESS_AGENT_ENABLED` | Enable [live process collection][9] with the process Agent. The [live container view][10] is already enabled by default if the Docker socket is available. |
+| `DD_PROCESS_AGENT_ENABLED` | Enable [live process collection][9] with the process Agent. The [live container view][10] is already enabled by default if the Docker socket is available. If set to `false`, will disable the live process collection][9] and the [live container view][10]|
```
l0k0ms:
```suggestion
| `DD_PROCESS_AGENT_ENABLED` | Enable [live process collection][9] with the process Agent. The [live container view][10] is already enabled by default if the Docker socket is available. If set to `false`, the  live process collection][9] and the [live container view][10] are disabled|
```

What's the difference with `disabled` then? 
According to https://github.com/DataDog/datadog-agent/blob/master/pkg/config/config_template.yaml#L478

* `false` would make the Agent collects only containers information.. so the Live container view should still be up ?

Simwar: This is not what's written in the code: https://github.com/DataDog/datadog-agent/blob/master/pkg/process/config/yaml_config.go#L161-L163

l0k0ms:
```suggestion
| `DD_PROCESS_AGENT_ENABLED` | Enable [live process collection][9] with the process Agent. The [live container view][10] is already enabled by default if the Docker socket is available. If set to `false`, the [live process collection][9] and the [live container view][10] are disabled.|
```

PR #5315, opened 2019-08-27
downgrade hugo-bin to 0.43.3 which is hugo version 0.55.3

davidejones:
### What does this PR do?

There seems to be issues changing files with the local hugo server causing a segfault. When downgrading to another hugo version this seemed to resolve itself.

This will likely need some qa effort to make sure it hasn't introduced any new bugs.

### Motivation
Issues running locally

### Preview link
https://docs-staging.datadoghq.com/david.jones/hugo-version/

### Additional Notes
Do not merge until @l0k0ms reviews

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
         "bootstrap": ">=4.1.2",
         "del": "4.1.1",
         "fancy-log": "^1.3.3",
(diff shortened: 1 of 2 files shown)


l0k0ms:
tested and it's working

PR #5341, opened 2019-08-30
Clarify example snippet based upon feedback

burningion:
### Motivation

A user was confused about terminology on a page. They reached out and had suggestions to improve our example.

### Preview link

https://docs-staging.datadoghq.com/kirk.kaiser/clarify-resource-service-example/tracing/faq/resource-trace-doesn-t-show-up-under-correct-service/

### DIFF
--- a/content/en/tracing/faq/resource-trace-doesn-t-show-up-under-correct-service.md
+++ b/content/en/tracing/faq/resource-trace-doesn-t-show-up-under-correct-service.md
@@ -23,7 +23,9 @@ These resources, with a top level name of `web_identification` still appear in t
 An example of modifying the top level name for Python can be found below:
 
 ```
-   @tracer.wrap('tornado.notify', service='tornado-notification', resource='MainHandler.do_something')
+   @tracer.wrap('tornado.notify', 
+                service='tornado-notification', 
+                resource='MainHandler.do_something')
     @tornado.gen.coroutine
     def do_something(self):
         # do something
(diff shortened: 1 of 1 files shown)


content/en/tracing/faq/resource-trace-doesn-t-show-up-under-correct-service.md
```diff
@@ -23,14 +23,20 @@ These resources, with a top level name of `web_identification` still appear in t
...
+                resource='MainHandler.do_something')
     @tornado.gen.coroutine
-    def notify(self):
+    def do_something(self):
         # do something
 ```
 
 This function explicitly sets both the service name and Top Level Name, being `tornado-notification` and `tornado.notify` respectively.
 
+Also note that we set our resource name manually, `MainHandler.do_something`.
```
l0k0ms:
```suggestion
Also note that the resource name is set manually, `MainHandler.do_something`. 
```

content/en/tracing/faq/resource-trace-doesn-t-show-up-under-correct-service.md
```diff
@@ -23,14 +23,20 @@ These resources, with a top level name of `web_identification` still appear in t
...
-    def notify(self):
+    def do_something(self):
         # do something
 ```
 
 This function explicitly sets both the service name and Top Level Name, being `tornado-notification` and `tornado.notify` respectively.
 
+Also note that we set our resource name manually, `MainHandler.do_something`. 
+
+By default our resource name would be set to this as it's the name of the function and the class for which it lives under in Tornado.
```
l0k0ms:
```suggestion
By default the resource name would be set to this as it's the name of the function and the class for which it lives under in Tornado.
```

PR #5361, opened 2019-09-04
force-yes is deprecated

dabcoder:
### What does this PR do?
`--force-yes` is now deprecated in favor of `--allow*` options. 

https://superuser.com/questions/1438031/ubuntu-18-command-apt-get-dist-upgrade-qq-force-yes-deprecated

### Page preview

https://docs-staging.datadoghq.com/davidb/force-yes-deprecated/

### Motivation
Testing


olivielpeau:
we should mention that, if the option `--allow-downgrades` doesn't exist for the `apt` version that the user has, they can use `--force-yes` instead. Old versions of `apt-get` do not support the `--allow-downgrades` option

PR #5413, opened 2019-09-11
add an example of a manually typed attribute

mstbbs:
### What does this PR do?
add an example of non-faceted log attribute.
https://docs-staging.datadoghq.com/michaelw/public-beta-logs2metrics/logs/logs_to_metrics/


l0k0ms:
@mstbbs does it work only with numerical attributes? or all attributes are supported even the strings one?

mstbbs:
Just numerics right now. We plan to support `unique count` in the future (maybe q1?).

PR #5444, opened 2019-09-18
Zach/add pa11y

zbayoff:
### What does this PR do?
Adds [pa11y](https://github.com/pa11y/pa11y) to Docs CI. Pa11y is a website accessibility (a11y) tool that uses [HTML Code Sniffer](https://squizlabs.github.io/HTML_CodeSniffer/) to scrape live urls and output errors and warnings in accordance with [WCAG](https://www.w3.org/TR/WCAG/) guidelines.

By default the pa11y script was written to read all html files from the `public` folder, but this would take the script way too long to complete and give redundant errors.  For example, pa11y doesn't need to run for every language directory (`/fr`, /`ja`) or every `agent/` page as they have similar layouts. For starters, we can check top-level pages, fixing those a11y errors, and add more pages as necessary. This can be adjusted in the pa11y.js script. 

A CSV file is generated that lives in s3 bucket `origin-static-assets` under `pa11y-output-docs` and contains the pa11y errors for specific urls, right down to the HTML element causing the error.

Some errors will be layout specific and need to be handled or approved by Websites. Other issues such as anchors tags with no corresponding ID, or img alt attributes, can be handled by Community.

This job will be run during the nightly master Docs build.

Note: pa11y is live currently on Corpsite and outputs a metric with the number of pa11y errors to DD, shown on the dashboard: https://dd-corpsite.datadoghq.com/dashboard/u3m-7ve-2yv/corpsite-accessibility-pa11y-issues?from_ts=1568212013417&live=true&tile_size=m&to_ts=1568816813417

### Motivation
This tool will help the Docs site to be ADA compliant by fixing accessibility issues sitewide


zbayoff:
> It looks good from here, a couple of things to note
> 
> * we might not need the `package.json` `yarn.lock` changes as pa11y is only run during CI not locally
> * After this releases we probably want to merge the corp-ci branch you made into master

I see, but these changes allow me to test the pa11y script locally.

PR #5484, opened 2019-09-25
Removing unused images

l0k0ms:
### What does this PR do?

This removes all un-used images from the documentation repo representing **17%** of all images.

### Process 

1. Get all images from the static folder:
 * `find ./static/images -name "*.jpg"`
 *  `find ./static/images -name "*.png"`

2. Get all images used within the doc:
 * `grep "{{< img src=" ./content/en/* -R`

3. Clean the two list to get only the file path in both cases

4. Display list of images that are in the static folder but not in the doc:
`awk 'FNR==NR {a[$0]++; next} !a[$0]' ./list_of_images_in_doc.txt ./list_of_images.txt`

5. Remove all those files from the static folder: `rm -f $(<images_to_delete.txt)`


apigirl:
static/images/integrations/fluentd/snapshot-fluentd.png is used in https://github.com/DataDog/documentation/blob/master/content/fr/integrations/fluentd.md

l0k0ms:
@apigirl Those pages should be updated anytime soon since the localised review process is going on.

The script didn't checked localised integrations page since image updates there are quite rare. The fact that there is still an image used on a localised integration page is temporary, i'm going to manually review the new content to fix this in tomorrows PR

PR #5485, opened 2019-09-25
Adding a getting started for sending logs.

l0k0ms:
### What does this PR do?
Adds a first log getting started focused on:

* Sending logs manually through a terminal
* Sending logs manually from a file monitored by the Agent.

### Motivation
More Getting started.

### Preview link

* https://docs-staging.datadoghq.com/gus/log-gs/getting_started/logs/?tab=ussite


content/en/getting_started/logs/_index.md
```diff
@@ -0,0 +1,223 @@
...
+{{< /tabs >}}
+
+#### Validation
+
+Verify the Agent is running with the [status command][8]:
+
+```
+sudo datadog-agent status
+```
+
```
NBParis: Should we show the example of what to expect? (with the line `logs collection not started` for instance)?

content/en/getting_started/logs/_index.md
```diff
@@ -0,0 +1,223 @@
...
+    sudo touch /etc/datadog-agent/conf.d/custom_log_collection.d/conf.yaml
+    ```
+
+3. Copy and Past the following content within this `conf.yaml` file:
+
+      ```
+      logs:
+        - type: file
+          path: /home/ubuntu/log_file_to_monitor.log
+          source: custom
```
NBParis: So here the issue is that we do not really explain what `source`, `service` and `sourcecategory` are for.

1. I would avoid using sourcecategory as this is totally optional and its behaviour might change
2. source I would suggest something like `custom_integration`
3. service something like `application1`

WDYT?

content/en/getting_started/logs/_index.md
l0k0ms: 🤔 we could indeed. i'm going to add it.

content/en/getting_started/logs/_index.md
l0k0ms: i like it, i'm also going to link to log collection page for further reading about those parameters.

l0k0ms: well, now that i'm reading it, we might want to just remove the sourcecategory since in the context of this GS, the source is custom and the service is the user 😸

NBParis: yes, let's remove sourcecategory

content/en/getting_started/logs/_index.md
```diff
@@ -0,0 +1,226 @@
...
+- link: "/logs/log_collection/"
+  tag: "Documentation"
+  text: "Collect logs from your Applications, Containers, and Cloud providers"
+---
+
+## Overview
+
+Datadog Log Management collects logs from your application. This page shows you how get your first logs into Datadog. Follow the sections below:
+
+* [Create a Datadog account](#create-a-datadog-account)
```
cswatt: there's no `#create-a-datadog-account` section on this page, we can probably take this link out

content/en/getting_started/logs/_index.md
```diff
@@ -0,0 +1,226 @@
...
+* [Send logs manually](#sending-logs-manually)
+* [Use the Agent to send logs from a file](#send-logs-from-a-file)
```
cswatt: is there a reason this page has an explicit table of contents here?

content/en/getting_started/logs/_index.md
```diff
@@ -0,0 +1,226 @@
...
+
+Before moving forward:
```
cswatt: if these are prerequisites, put them ABOVE the line that says "follow the sections below"

content/en/getting_started/logs/_index.md
l0k0ms: just wanted to outline the two "getting started" that are on this page.

PR #5523, opened 2019-09-27
Azure Status Metric

ruthnaebeck:
### What does this PR do?
- Update FAQ on Azure status metric
- Rename and alias file

### Motivation
- Trello request

### Preview link
https://docs-staging.datadoghq.com/ruth/azure-status/integrations/faq/azure-status-metric/

### Additional Notes
PM should review before merge


content/en/integrations/faq/azure-status-metric.md
```diff
@@ -0,0 +1,25 @@
+---
+title: Azure Status Metric
+kind: faq
+aliases:
+  - /integrations/faq/azure-vm-status-is-not-reporting
+---
+
+## Overview
+
+Datadog provides an `azure.*.status` metric for each [Azure integration][1]. For example, Azure Virtual Machines monitored by Datadog should report `azure.vm.status`, which identifies if the machine is currently running.
```
steveharrington1: I think it would be better to say:

"Datadog provides an `azure.*.status` metric for each resource monitored via the [Azure integration][1]."

Not sure if that's the right wording exactly- but the idea is that the status metric is provided for the resource, not the integration.

PR #5535, opened 2019-09-30
a11y fixes

l0k0ms:
### What does this PR do?
This PR fixes the last a11y issues we are facing with the Documentation Content.

Two fixes are bigger than the other one:

* Refactor of the Uninstall the Agent page in order to avoid duplication of anchors.
* Refactor of the PHP Log collection page in order to avoid duplication of anchors

The content is the same in both cases, it's the tab logic that changed in order to have unique anchors on the page allowing for a better accessibility.

The content that is removed in the basic_agent_usage page is actually duplicated content from 15lines above.. since it created conflict with the Agent v6 content i just removed it. 

### Preview link

* https://docs-staging.datadoghq.com/gus/a11y-final-fix/agent/faq/how-do-i-uninstall-the-agent/?tab=agentv6#pagetitle

* https://docs-staging.datadoghq.com/gus/a11y-final-fix/logs/log_collection/php/?tab=phpmonolog

### DIFF
--- a/content/en/agent/basic_agent_usage/_index.md
+++ b/content/en/agent/basic_agent_usage/_index.md
@@ -105,24 +105,11 @@ The recommended number of open file descriptors is 1024. You can see this value 
 minfds = 100  # Your hard limit
 ```
 
-### The Collector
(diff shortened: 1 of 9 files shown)


apigirl:
great work! thanks for taking ownership of the a11y fixes!

PR #5558, opened 2019-10-02
delete translated files where the source english file is already deleted

davidejones:
### What does this PR do?

This PR deletes the translated markdown files that should no longer exist because the source English file no longer exists.

During review please take a close look at the files and that they should in fact be deleted.

### Motivation

We have a task to fix the automation aspect of deleting translated files when a source file is deleted. This led me to manually delete any it should have done so far with this PR.

### Preview link
https://docs-staging.datadoghq.com/david.jones/delete-localized-files/


davidejones:
@l0k0ms Instead of delete and pull translations again I've manually done some deleting with this PR

l0k0ms:
Double checked, everything seems fine

PR #5580, opened 2019-10-04
Adding helm example

l0k0ms:
### What does this PR do?
Leverage the Helm `env` param to document how to pass a log global rule within and Agent helm chart.

### Motivation
Customer feedback 

### Preview link

* https://docs-staging.datadoghq.com/gus/helm-log/agent/logs/advanced_log_collection?tab=helm#global-processing-rules


l0k0ms:
@NBParis setting a custom label on a container through Helm is outside of the Datadog documentation i think.

NBParis:
@l0k0ms Actually no. This environment variable is to define GLobal processing rule on the agent that is applied on all the logs collected by the agent. Not from a specific container. Otherwise this is done on the auto discovery label.

PR #5655, opened 2019-10-11
Peter/edit page

anagio:
Adds an Edit This Page link to the footer of all pages
https://docs-staging.datadoghq.com/peter/editPage/

Should maybe use a config to set pages to load on

Needs brand design to style


l0k0ms:
Some integrations documentation have a `dependencies` frontmatter that point to their source file in the corresponding repository (since those files are coming from integrations-core and integrations-extras repo)

Some integrations can't benefit from this feature since their source file is in a private repo. We should maybe redirect to the "create a new issue" from the doc repo: https://github.com/DataDog/documentation/issues/new/choose

The rule can be summarise in:

* If dependencies parameter => this is the link to the source file
* if not
  * if it's in `/integrations/` then redirect to the new issue in github
  * if it's not in `/integrations/` then redirect to the corresponding documentation file (current logic implemented)

PR #5688, opened 2019-10-16
update JS sample code to use required JSON format

ericmschow:
### What does this PR do?

Fixes the JavaScript sample code on the docs page for linking logs to traces to log in the format the Datadog Agent requires.
### Motivation

Gini helped me diagnose a bug in my manual linking of logs to traces, that ultimately stemmed from me following this code. The docs do specify JSON format before the sample code, but the code itself logs the object itself (which does not link the trace).

### Preview link

N/A - working off a fork


l0k0ms:
Hello, Many thanks for this fix :)

PR #5689, opened 2019-10-16
Azure integration tag keys

ruthnaebeck:
### What does this PR do?
- Add Azure integration tag keys

### Motivation
- Updates to included Azure tags

### Preview link
https://docs-staging.datadoghq.com/ruth/azure-tags2/tagging/assigning_tags/#azure

### Additional Notes
Do not merge until PM approves


steveharrington1:
@ruthnaebeck can we do something to mark the new tags as being in beta? Maybe add a * after the name in the Integration column and then a note at the bottom like "_* resource-specific tags are in beta_" ? 

Also, is there any way to clean up the list of tags for app services? I think starting the "plus (as applicable)" on a new line rather than in the list of tags would help.

Open to other ideas for both of these.

steveharrington1:
@ruthnaebeck I'm happy with the beta marking if you are.

The list of tags still looks bad. What if we just get rid of the "Plus (as applicable)" part and list the tags? For example, will anyone will be confused by _not_ having a python_version tag if they're using java? I don't think the qualifier is necessary.

PR #5694, opened 2019-10-17
moving all head script into a partial

l0k0ms:
### What does this PR do?
Follows the logic introduced by: https://github.com/DataDog/documentation/pull/5687/files In order to be consistent on how we manage header script.

* It moves hotjar / DD monitoring / google site tag manager to this logic
* It disables all of them in dev env

### Motivation
Cleaner doc.

### Preview link

* https://docs-staging.datadoghq.com/gus/partial-clean-up/


layouts/partials/head_scripts/datadog_monitoring.html
```diff
@@ -1,19 +1,20 @@
...
-      clientToken: '{{ $.Site.Params.datadog.clientToken }}',
-      applicationId: '{{ $.Site.Params.datadog.applicationId }}',
-    });
-{{ end }}
+  { { if (or(eq $.Site.Params.environment "live")(eq $.Site.Params.environment "preview")) } }
+  DD_RUM.init({
+    clientToken: '{{ $.Site.Params.datadog.clientToken }}',
+    applicationId: '{{ $.Site.Params.datadog.applicationId }}',
+  });
+  { { end } }
```
davidejones: In the page source for the homepage i see references to `$.Site`
I think these extra spaces in the curly braces are causing it to output incorrectly? it might need to be `{{` instead of `{ {`

l0k0ms: indeed, i just removed the logic since now all DD monitoring is disabled on dev env.

layouts/partials/head_scripts/datadog_monitoring.html
```diff
@@ -1,20 +1,18 @@
...
-    DD_RUM.init({
-      clientToken: '{{ $.Site.Params.datadog.clientToken }}',
-      applicationId: '{{ $.Site.Params.datadog.applicationId }}',
-    });
-{{ end }}
+  // RUM init
+  DD_RUM.init({
+    clientToken: '{{ $.Site.Params.datadog.clientToken }}',
+    applicationId: '{{ $.Site.Params.datadog.applicationId }}',
+  });
```
davidejones: Now you removed the environment check this will now run during local development too. Will that skew the data?

layouts/partials/head_scripts/datadog_monitoring.html
davidejones: This seems to be resolved already

layouts/partials/head_scripts/datadog_monitoring.html
l0k0ms: It won't since the script shouldn't be inlined in local build since the default is now to false

PR #5706, opened 2019-10-17
Update metrics_type.md

ijkaylin:
Updated histogram and distribution sections

### Preview link

https://docs-staging.datadoghq.com/kathy.lin/metricstypes/developers/metrics/metrics_type/

### DIFF
--- a/content/en/developers/metrics/metrics_type.md
+++ b/content/en/developers/metrics/metrics_type.md
@@ -144,9 +144,9 @@ Discover how to submit gauge metrics:
 {{% /tab %}}
 {{% tab "HISTOGRAM" %}}
 
-The `HISTOGRAM` metric type allows you to measure the statistical distribution of a set of values. Datadog's `HISTOGRAM` metric type is an extension of the [StatsD timing metric type][1]: it aggregates the values that are sent during a defined time interval (usually defaults to 10 seconds with the Agent) and produces different metrics representing the different aggregations possible for the set of value. Depending on the aggregation, the metric type stored by Datadog is different.
+The `HISTOGRAM` metric type allows you to measure the statistical distribution of a set of values. Datadog's `HISTOGRAM` metric type is an extension of the [StatsD timing metric type][1]: it aggregates the values that are sent during a defined time interval (usually defaults to 10 seconds with the Agent) and produces different timeseries representing the different aggregations possible for the set of values. Depending on the aggregation, the metric type stored by Datadog is different.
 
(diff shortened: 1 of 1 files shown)


content/en/developers/metrics/metrics_type.md
```diff
@@ -36,73 +30,73 @@ But the following metric submission types are accepted:
...
-  * 10 requests for the first 30 seconds, then
-  * 20 requests for the next 30 seconds, then
-  * 0 request for the last 30 seconds.
+To better understand the different metrics types, what they represent, and how to modify them within Datadog, consider the following example:
 
-Then the pattern starts again.
+You have two web servers: `server:web_1` and `server:web_2`. Both web servers continuously receive:
+  * 10 HTTP requests for the first 30 seconds, then
+  * 20 HTTP requests for the next 30 seconds, then
+  * 0 HTTP requests for the next 30 seconds.
```
bradleyjay: For this example, I think we need to not use 0. It’s not clear if that means no value was received, or if a 0 was submitted. 

If that truly was a 0 value, no interpolation is needed, there’s a value - 0. 

If it’s a NaN, we’ll .fill(last) for all of these. I think it starts to muddy the waters a bit.

content/en/developers/metrics/metrics_type.md
```diff
@@ -18,15 +18,9 @@ further_reading:
...
+A metric's type affects how the given metric is aggregated in queries and values displayed in graph visualizations within Datadog. Different metric types can have different default time aggregation functions and different [Metric Type modifiers][1]. A metric's type is displayed on the details sidepanel for the given metric on the [Metrics Summary page][2]. Note: Changing the metric type in this details sidepanel can potentially change metric behavior in all existing visualizations and monitors; potentially rendering historical data as nonsensical.
 
-There are three metric types within the Datadog web application:
-
-* `COUNT`
-* `RATE`
-* `GAUGE`
-
-But the following metric submission types are accepted:
+The following metric submission types are accepted:
```
bradleyjay: I like this distinction, rather than calling them types.

content/en/developers/metrics/metrics_type.md
bradleyjay: The other bit is, I think using a 0 here, if we want to keep these examples, is confusing. If it means "no value submitted", we'll have a NaN, interpolation kicks in, that's one case. Gauge/Count/Rate will all three have fill(last, 5 minutes) applied, and rolllup avg.

If the 0 is an actual 0 submitted by the Agent, all of those types will actually preserve the 0 in-app, no interpolation.

PR #5711, opened 2019-10-18
Correct missing underscores in generated docs

jacastello:
### What does this PR do?
There's a small type-o in some of the docs that strips the underscore in the `last_#x` notation. This aims to correct that. It looks like the following

image

### Motivation
I copy and pasted from these docs in some code for an internal tool and it was wrong!

### Preview link


cswatt:
This is a great catch! I just made some modifications that pertain to our translation process. Thank you @jacastello for the fix!

PR #5818, opened 2019-11-01
remove `DD_CLUSTER_NAME` mention

DylanLovesCoffee:
### What does this PR do?
Removes the section mentioning `DD_CLUSTER_NAME` under "cluster name auto detection" as this is not relevant to how the agent collects the `cluster-name` tag. Cluster name tags will either be auto-detected via a cloud provider's available metadata API or can be set manually via `DD_TAGS`. 

`DD_CLUSTER_NAME` should only be used either to append a unique string to host aliases to prevent hostname overlap from different clusters (host aliasing issues), or to add the [unique tag to clusterchecks](https://github.com/DataDog/documentation/blob/06f1cce7dd7521c33a5ddf7f450016bb788b7b70/content/en/agent/autodiscovery/clusterchecks.md#cluster-agent-setup) as cluster checks cannot resolve to a host.

### Motivation
Customer requests. This causes some confusion as it makes it seem as though the environment variable adds a tag.

### Preview link

https://docs-staging.datadoghq.com/dylan-cluster-name/kubernetes/daemonset_setup/

### DIFF
--- a/content/en/agent/kubernetes/daemonset_setup.md
+++ b/content/en/agent/kubernetes/daemonset_setup.md
@@ -187,9 +187,7 @@ NAME            DESIRED   CURRENT   READY     UP-TO-DATE   AVAILABLE   NODE SELE
 datadog-agent   2         2         2         2            2           <none>          16h
 ```
 
-### Kubernetes cluster name auto detection
-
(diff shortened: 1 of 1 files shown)


hkaj:
Could we make the description more precise rather than removing it? I also think the way it was described would cause confusion, but it's a useful feature for teams who run multiple clusters and face aliasing issues.

PR #5945, opened 2019-11-13
Add DD_AGENT_VERSION option to our EBS sample config

albertvaka:
### What does this PR do?
Allow users to pin a specific version of the Agent.

PR #5944, opened 2019-11-13
Install Agent 7 in our EBS sample config

albertvaka:
### What does this PR do?
Install Agent 7 in our EBS sample config and stop trusting our old RPM signing key. Agent 7 packages are always signed with the new key.

### Motivation
Agent 7 release.

### Additional Notes
To be merged after Agent 7 goes GA.

PR #5966, opened 2019-11-15
Add DD_AGENT_VERSION option to our EBS sample config

albertvaka:
## What does this PR do?
Allow users to pin a specific version of the Agent
## Testing done
Deployed on EBS for with and without specifying DD_AGENT_VERSION.
## Additional notes
Was previously merged and reverted because of a typo in the command: https://github.com/DataDog/documentation/pull/5945

### DIFF
--- a/static/config/99datadog.config
+++ b/static/config/99datadog.config
@@ -3,7 +3,9 @@ option_settings:
     - namespace:  aws:elasticbeanstalk:application:environment
       option_name:  DD_API_KEY
       value:  "<YOUR_DD_API_KEY>"
-
+    - namespace:  aws:elasticbeanstalk:application:environment
+      option_name:  DD_AGENT_VERSION
+      value:  "" # For example, "6.15.1". Leave empty to install the latest version. Only Agent 6 is supported.
 files:
     "/configure_datadog_yaml.sh":
         mode: "000700"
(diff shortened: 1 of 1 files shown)


static/config/99datadog.config
```diff
@@ -3,7 +3,9 @@ option_settings:
     - namespace:  aws:elasticbeanstalk:application:environment
       option_name:  DD_API_KEY
       value:  "<YOUR_DD_API_KEY>"
-
+    - namespace:  aws:elasticbeanstalk:application:environment
+      option_name:  DD_AGENT_VERSION
+      value:  "" # For example, "6.15.1", leave empty to install the latest version.
```
KSerrania: If users set a value that does not correspond to the major version installed by default, they will also need to change the `baseurl` in the [repository file](https://github.com/DataDog/documentation/pull/5966/files#diff-13dacd0da828fde7e9003ec2de5dc109R21-R32). This should be added as a comment here and in the related docs.

albertvaka: We are only doing Agent 6 here. I have another PR to add Agent 7 support [1] and my plan was to stop supporting Agent 6 instead of letting users install either version. The rationale is: this file is something you download when setting up DataDog for the first time and then you keep your own copy of it (so existing users shouldn't use what's in here).

[1] https://github.com/DataDog/documentation/pull/5944/files

KSerrania: I see; the thing I'm worried about is customers seeing this and trying to put an Agent 5 version here, which will make the deployment fail.

albertvaka: We can mention only agent 6 is supported.

albertvaka:
Added note about support for Agent 6 only. @ruthnaebeck I think this is ready to merge together with #41624

PR #6087, opened 2019-11-29
http.version added to standard attributes

gusai-dd:
### What does this PR do?
Adds `http.version` to the list of standard attributes.

### Motivation
The HTTP version is always remapped to `http.version` in the pipelines. This means it's a so-called standard attribute. This information was missing from the public documentation.

### Preview link
https://docs-staging.datadoghq.com/gusai/http-vers-stdAttr/logs/processing/attributes_naming_convention/#http-requests


gusai-dd:
Hello, perfect, thanks a lot!

PR #6156, opened 2019-12-10
Adding lang-redirects

nsollecito:
### What does this PR do?
Moving language-redirect logic client-side in an effort to reduce / eliminate dependencies on Lambda@Edge. 

### Motivation
As the scope of the functionality required at the edge has been reduced, the language-specific redirect functionality can be completed client-side instead. 

### Preview link
https://docs-staging.datadoghq.com/nsollecito/lang-redirects/

Switch to JA:
https://docs-staging.datadoghq.com/nsollecito/lang-redirects/?lang_pref=ja

Back to EN
https://docs-staging.datadoghq.com/nsollecito/lang-redirects/?lang_pref=en

Try a language that's not yet enabled:
https://docs-staging.datadoghq.com/nsollecito/lang-redirects/?lang_pref=es

### Additional Notes
Adds js-cookie as a dependency. Sets `lang_pref` cookie with Default based on navigator.language value

User is redirected based on the following:
1. URL params are checked for `lang_pref` that matches a value in `acceptLanguage`
2. Cookies are checked for a `lang_pref` value, if not found then
3. The browser's navigator.language value is used to determine if a redirect is required. 

The preferred language is then set in a cookie and a redirect is triggered if the requested page. 

Cookie `lang_pref` should update to reflect selected language. Requests to other languages should redirect as expected. The only language currently activated to force a redirect is Japanese. 

Related corp-ci PR: https://github.com/DataDog/corp-ci/pull/64
Related CloudOps PR: https://github.com/DataDog/cloudops/pull/16359


src/scripts/components/lang-redirects.js
```diff
@@ -0,0 +1,100 @@
...
+    const a = document.createElement('a');
+    a.href = data;
+    return a.hostname;
+}
+
+const getPath = (data) => {
+    const a = document.createElement('a');
+    a.href = data;
+    return a.pathname;
+}
```
davidejones: I was thinking you might want to condense these 2 into one function that returns both? I guess its a preference kind of thing if you think its easier to read / maintain together or seperate

```suggestion:-11
const getUrlParts = (data) => {
    const a = document.createElement('a');
    a.href = data;
    return {hostname: a.hostname, pathname: a.pathname};
}
```

then later you could just do

`const {hostname, pathname} = getUrlParts(window.location.href);`

src/scripts/components/lang-redirects.js
```diff
@@ -0,0 +1,100 @@
...
+
+export function handleLanguageBasedRedirects() {
+	const params = getUrlVars();
+
+	const supportedLanguages = navigator.language || navigator.browserLanguage;
```
davidejones: running this gave me `en-US` would it give us `ja-JP` or anything for japan or is it just ja?

src/scripts/components/lang-redirects.js
```diff
@@ -0,0 +1,100 @@
...
+		Cookies.set("lang_pref", acceptLanguage, {path: cookiePath});
+	}
+	else if (Cookies.get('lang_pref') && allowedLanguages.indexOf(Cookies.get('lang_pref')) !== -1 ) {
+		acceptLanguage = Cookies.get('lang_pref');
+		logMsg = `Change acceptLanguage based on lang_pref Cookie: ${ acceptLanguage}`;
+	}
+	else if ( supportedLanguages && subMatch.length ) {
+		logMsg = `Set acceptLanguage based on navigator.language header value: ${  supportedLanguages  } ; DEBUG: ${ supportedLanguages.startsWith('ja') }`;
+
+		if ( supportedLanguages.startsWith('ja') ) {
```
davidejones: should we be checking this against `allowedLanguages` instead of just hardcoding to `ja`?

nsollecito:
Yes should be ja-JP for Japan

https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/languages

src/scripts/components/lang-redirects.js
nsollecito: Initially, we just want to check for JA because that’s the only redirect we’re supporting at the moment

src/scripts/components/lang-redirects.js
nsollecito: Yes should be ja-JP for Japan https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/languages

src/scripts/components/lang-redirects.js
nsollecito: +1 good suggestion, I've updated the function

src/scripts/components/lang-redirects.js
nsollecito: actually, not sure about this.. going to spend some time on it today

nsollecito: So yea, I confirmed.. this is the case because we only want to force redirects for JA at the moment but we want to support the other languages. I'll add some comments so this is clear

davidejones: ok sounds good, when i wrote the comment i was mostly thinking about having this defined at the top of the file or somewhere else where we can just add another language if we support fr or something too.

nsollecito: yea, good call.. working on doing that now

nsollecito: Alright, I've added a separate redirectLanguages where we can set which accept languages should actually trigger a redirect

zbayoff:
@nsollecito When I click the language switching options on a page like https://docs-staging.datadoghq.com/nsollecito/lang-redirects/getting_started/agent/?tab=datadogussite
the language switches, but the cookies doesn't change.

zbayoff:
There's also a console error on the homepage when trying to access a page with a lang pref that's not active yet: https://share.getcloudapp.com/04ug7ALe

nsollecito:
@zbayoff sorry, the preview site had not updated yet, needed to update/rebuild corp-ci... should be good now. Please check again

davidejones:
So i just tried something similar, i was on https://docs-staging.datadoghq.com/nsollecito/lang-redirects/ja/getting_started/agent/?tab=datadogussite and i had a Ja cookie. Then i click french in the footer and it takes me to french but then flashes back to ja.

nsollecito:
@davidejones this is fixed. I needed to update the script to execute handleLanguageBasedRedirects function directly now that it's in the head. 

Otherwise, the `window.history.replaceState` in codetabs script interferes with the `lang_params` URL params before it has a chance to execute.

davidejones:
> @davidejones this is fixed. I needed to update the script to execute handleLanguageBasedRedirects function directly now that it's in the head.
> 
> Otherwise, the `window.history.replaceState` in codetabs script interferes with the `lang_params` URL params before it has a chance to execute.

Ah ok gotcha that makes sense, i was wondering if it was related to the tabs. That seems to have fixed that issue for me.

PR #6203, opened 2019-12-16
update to use node-13

davidejones:
### What does this PR do?

This PR updates the documentation local and CI to use node 13.

### Motivation

We recently updated corp to be using node 13 so for consistency we are doing the same on docs. This also helps in our ability to install security fixed packages.

### Preview link

https://docs-staging.datadoghq.com/david.jones/node-13/


nsollecito:
👍 Tested on local with Node 10.13.0 and 12.10.0

l0k0ms:
Tested on local with node 10.16 and 13.3.0 and it works !

PR #6216, opened 2019-12-17
adjust bullet indent for second or greater lines

davidejones:
### What does this PR do?

This PR updates the bullet point text alignment which is thrown off when bullets are longer than one line.

### Motivation

https://trello.com/c/coDgQBsv/4261-bullet-list-text-is-not-aligned-when-there-is-more-than-one-line

### Preview link

Here is the page i was using as a reference but please check as many places as you can:
https://docs-staging.datadoghq.com/david.jones/bullet-indent/tracing/app_analytics/analytics/#table

### DIFF
--- a/src/styles/pages/_global.scss
+++ b/src/styles/pages/_global.scss
@@ -310,7 +310,7 @@ h5{
         content: "–";
         position: relative;
         margin-right: 10px;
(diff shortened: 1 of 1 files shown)


l0k0ms:
<3

PR #6219, opened 2019-12-17
Update mvc lang

andrewsouthard1:
### What does this PR do?

Updates support versions for ASP.NET MVC

### Motivation

We re-introduced support for ASP.NET MVC 4 in https://github.com/DataDog/dd-trace-dotnet/pull/580 

### Preview link

https://docs-staging.datadoghq.com/update_mvc_lang/tracing/setup/dotnet/?tab=netframeworkonwindows

### DIFF
--- a/content/en/tracing/setup/dotnet.md
+++ b/content/en/tracing/setup/dotnet.md
@@ -282,7 +282,7 @@ The .NET Tracer can instrument the following libraries automatically:
 
 | Framework or library           | NuGet package name                       | Package versions     | Integration Name     |
 |--------------------------------|------------------------------------------|----------------------|----------------------|
-| ASP.NET MVC                    | `Microsoft.AspNet.Mvc`                   | 4.0+               | `AspNetMvc`          |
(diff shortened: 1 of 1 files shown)


content/en/tracing/setup/dotnet.md
```diff
@@ -282,7 +282,7 @@ The .NET Tracer can instrument the following libraries automatically:
 
 | Framework or library           | NuGet package name                       | Package versions     | Integration Name     |
 |--------------------------------|------------------------------------------|----------------------|----------------------|
-| ASP.NET MVC                    | `Microsoft.AspNet.Mvc`                   | 5.1.0+               | `AspNetMvc`          |
+| ASP.NET MVC                    | `Microsoft.AspNet.Mvc`                   | 4.0-5.0+               | `AspNetMvc`          |
```
lucaspimentel: @andrewsouthard1: I think we can just say `4.0+` which means "4.0 and above" and includes the `5.x` versions.

Alternatively, we can be super specific and say `4.0.40804 - 5.2.7` since we haven't really tested `4.0` versions lower than that one.

Your pick!

cswatt: Seconding the above. Would also like this to be `4.0+`

PR #6342, opened 2020-01-07
[Logs] Attribute-based exclusion filter

gusai-dd:
### What does this PR do?
Adds Attribute-based exclusion filter documentation and updates screenshots.

### Motivation
Align doc with deployed feature.

### Preview link
https://docs-staging.datadoghq.com/gusai-dd-patch-1/logs/indexes/#exclusion-filters


content/en/logs/indexes/_index.md
```diff
@@ -78,6 +78,27 @@ Update or remove this quota at any time when editing the Index:
 
 **Note**: Indexes daily quotas reset automatically at 2:00pm UTC (4:00pm CET, 10:00am EDT, 7:00am PDT).
 
+### Attribute based exclusion filter
+
+In addition to the default sampling on the raw log (exclusion percentage set on `all logs`), you can choose to sample on a specific attribute holding an ID, like `UserID`, `SessionID`, `RequestID`.
```
pcarioufr: "raw log" ?
-> "such as User, Session or Request Identifiers" (actual information rather than example path name seems more relevant here)

gusai-dd: Thanks for the review.
Implementing changes.

content/en/logs/indexes/_index.md
```diff
@@ -78,6 +78,27 @@ Update or remove this quota at any time when editing the Index:
...
+
+{{< img src="logs/indexes/index_exclusion_on_attribute.png" alt="index attribute-based exclusion filter"  style="width:70%;">}}
+
+**Note**: 
+
+* for a non faceted attribute, you need to prefix it with the `@` character.
+
+* for low cardinality attributes, we suggest to define an exclusion query and sample on all matching logs.
+
+  For example to exclude 99% of successful Nginx HTTP requests, define the exclusion query `source:nginx @http.status_code:200` on 99% of `all logs`.
```
pcarioufr: This is not the example matching the note up there. 
I think what you want to write here is that better than excluding logs from 50% of status codes, you'd better explicitly exclude 100% logs from status code 201, 204, 403...

But status code is probably not the best example here : might be better to use something you'd actually want to sample on, like internal IPs ... or something like this. And at the end of the day, not sure we want to cast light on this.

gusai-dd: OK, replacing by IP.

gusai-dd:
Hello @pcarioufr ,
changes implemented.
Thanks for your review!

NBParis:
Preview do not work for me, is that expected?

content/en/logs/indexes/_index.md
```diff
@@ -78,6 +78,33 @@ Update or remove this quota at any time when editing the Index:
 
 **Note**: Indexes daily quotas reset automatically at 2:00pm UTC (4:00pm CET, 10:00am EDT, 7:00am PDT).
 
+### Attribute based exclusion filter
```
NBParis: We are already in the exclusion filter section so we could rename this section `Attribute based Sampling` ?

content/en/logs/indexes/_index.md
```diff
@@ -78,6 +78,33 @@ Update or remove this quota at any time when editing the Index:
...
+
+#### Sample all logs of a trace
+
+In order to guarantee the sampling of all logs related to a given trace, define the log sampling on `Trace Id` attribute.
+Refer to [Trace Remapper][10] to get more information on how to connect logs and traces.
+
+For example, setting exclude 99% of `Trace Id`, you end up with:
+- 1% of traces for which we keep all logs
+- 99% of traces for which we keep no logs 
+
```
NBParis: Not sure why we do a specific section out of it. This could be an example of the previous section as it's an attribute like this others even if automatically added by Datadog.

content/en/logs/indexes/_index.md
```diff
@@ -78,6 +78,33 @@ Update or remove this quota at any time when editing the Index:
...
+{{< img src="logs/indexes/index_exclusion_attribute_groups.png" alt="index attribute-based exclusion filter groups"  style="width:70%;">}}
+
+For example, exclude `nginx` logs for 99% of `http.request_id`:
+
+{{< img src="logs/indexes/index_exclusion_on_attribute.png" alt="index attribute-based exclusion filter"  style="width:70%;">}}
+
+**Note**: 
+
+* you need to prefix the attribute name with the `@` character.
```
l0k0ms:
```suggestion
* You must prefix the attribute name with the `@` character.
```

content/en/logs/indexes/_index.md
```diff
@@ -78,6 +78,33 @@ Update or remove this quota at any time when editing the Index:
...
+
+* for low cardinality attributes, we suggest to define a set of distinct Index Exclusion Filters on `all logs`.
```
l0k0ms:
```suggestion
* For low cardinality attributes, you should define a set of distinct Index Exclusion Filters on `all logs`.
```

content/en/logs/indexes/_index.md
```diff
@@ -78,6 +78,33 @@ Update or remove this quota at any time when editing the Index:
...
+
+#### Sample all logs of a trace
+
+In order to guarantee the sampling of all logs related to a given trace, define the log sampling on `Trace Id` attribute.
+Refer to [Trace Remapper][10] to get more information on how to connect logs and traces.
+
+For example, setting exclude 99% of `Trace Id`, you end up with:
```
l0k0ms:
```suggestion
For example, setting exclude 99% of `Trace Id`,  results in:
```

content/en/logs/indexes/_index.md
```diff
@@ -78,6 +78,33 @@ Update or remove this quota at any time when editing the Index:
...
+- 1% of traces for which we keep all logs
```
l0k0ms:
```suggestion
* 1% of traces get all their related logs kept
```

content/en/logs/indexes/_index.md
```diff
@@ -78,6 +78,33 @@ Update or remove this quota at any time when editing the Index:
...
+- 99% of traces for which we keep no logs
```
l0k0ms:
```suggestion
* 99% of traces get their related logs dropped.
```

content/en/logs/indexes/_index.md
```diff
@@ -78,6 +78,33 @@ Update or remove this quota at any time when editing the Index:
 
 **Note**: Indexes daily quotas reset automatically at 2:00pm UTC (4:00pm CET, 10:00am EDT, 7:00am PDT).
 
+### Attribute based exclusion filter
+
+You can sample on a specific attribute holding an ID, such as User, Session, Request Identifiers, in addition to the default log-based sampling (applied when the exclusion percentage is set on `all logs`).
+
+This feature operates on groups rather than the log itself: for example, given an attribute holding 3 distinct values (circles, triangles, squares), defining a 66% exclusion filter on it, keeps all logs holding 1 specific value (circles).
```
l0k0ms: The notion of "groups" is not really detailed through the doc.. could we say `attribute's values` or am i missing the point?

content/en/logs/indexes/_index.md
```diff
@@ -78,6 +78,33 @@ Update or remove this quota at any time when editing the Index:
...
+
+For example, exclude `nginx` logs for 99% of `http.request_id`:
+
+{{< img src="logs/indexes/index_exclusion_on_attribute.png" alt="index attribute-based exclusion filter"  style="width:70%;">}}
+
+**Note**: 
+
+* You must prefix the attribute name with the `@` character.
+
+* For low cardinality attributes, you should define a set of distinct Index Exclusion Filters on `all logs`.
```
NBParis: Let's use passive format. Something like:

* Attribute name must be prefixed by `@` (i.e if the attribute is called `http.request_id`, in the exclusion filter it must be `@http.request_id`)
* The attribute-based sampling is recommended for high cardinality attributes (like session_id, trace_id, ...)

content/en/logs/indexes/_index.md
```diff
@@ -104,3 +131,4 @@ Multiple indexes also provide the ability to define access rules on the data con
 [7]: https://docs.datadoghq.com/logs/archives/
 [8]: https://docs.datadoghq.com/logs/logs_to_metrics/
 [9]: /account_management/rbac
+[10]: /logs/processing/processors/?tab=ui#trace-remapper
```
NBParis: Might be better to link to https://docs.datadoghq.com/tracing/advanced/connect_logs_and_traces/?tab=java#pagetitle, what do you think?

content/en/logs/indexes/_index.md
```diff
@@ -78,6 +78,37 @@ Update or remove this quota at any time when editing the Index:
 
 **Note**: Indexes daily quotas reset automatically at 2:00pm UTC (4:00pm CET, 10:00am EDT, 7:00am PDT).
 
+### Attribute based sampling
```
NBParis: What about moving those two independent section into the main exclusion filter right after the definition?

Even as #### if you want to be able to link directly to it but here it looks a bit far away from the exclusion filter definition.

content/en/logs/indexes/_index.md
```diff
@@ -78,6 +78,37 @@ Update or remove this quota at any time when editing the Index:
...
+* given an attribute `id` holding 3 distinct values (`1`, `2`, `3`),
+* defining a 66% exclusion filter on it,
+* results in indexing all logs holding 1 specific attribute's value (`3`), all the other logs won't be indexed.
+
+{{< img src="logs/indexes/index_exclusion_attribute_groups.png" alt="index attribute-based exclusion filter groups"  style="width:70%;">}}
+
+For example, exclude `nginx` logs for 99% of `http.request_id`:
+
+{{< img src="logs/indexes/index_exclusion_on_attribute.png" alt="index attribute-based exclusion filter"  style="width:70%;">}}
+
```
NBParis: As discussed, let's focus on the http.request_id example and the trace one and put the emphasis on the fact that the kept traces are complete (which is the key functionality of this new sampling mechanism).

content/en/logs/indexes/_index.md
```diff
@@ -78,6 +78,37 @@ Update or remove this quota at any time when editing the Index:
...
+**Note**: 
+
+* You must prefix the attribute name with the `@` character.
+
+* For low cardinality attributes, you should define a set of distinct Index Exclusion Filters on `all logs`.
+
+#### Sample all logs of a trace
+
+In order to guarantee the sampling of all logs related to a given trace, define the log sampling on `Trace Id` attribute.
```
NBParis: We could be clearer here, as I'm not sure it's crystal clear what we are trying to say.

Something like the following:


Once [logs and traces are connected], a `trace_id` attribute is automatically added in the logs. In order to be able to sample those logs but make sure that for the chosen trace ids the full set of logs is kept, an attribute-based exclusion filter should be set on the `trace_id` attribute.

For example, setting exclude 60% of `Trace Id`, results in:

* 40% of all traces get all their related logs indexed.
* 60% of all traces have their logs sampled (still available in livetail and archives)

gusai-dd: Actually the last bullet is not correct: for the 60% of traces we don't get any logs indexed.

content/en/logs/indexes/_index.md
gusai-dd: Yes, `session_id` and `trace_id` examples only.
I prefer session instead of request as the latter has no concept of transaction unless we define an exclusion filter with a `*` query (on every log).

ruthnaebeck:
Replacement PR - https://github.com/DataDog/documentation/pull/6413

PR #6352, opened 2020-01-08
Removes outdated instructions for istio

cgilmour:
### What does this PR do?
Removes outdated parts of the setup instructions for istio.
These were applicable at the time, but now can cause errors when users apply them.

### Motivation
Problems reported to support, resolved by undoing those steps.

### Preview link
https://docs-staging.datadoghq.com/cgilmour/istio-doc-update/tracing/setup/istio/

### Additional Notes
Removed from FR translation also

### DIFF
--- a/content/en/tracing/setup/istio.md
+++ b/content/en/tracing/setup/istio.md
@@ -26,59 +26,6 @@ Datadog APM is available for Istio v1.1.3+ on Kubernetes clusters.
 2. [Make sure APM is enabled for your Agent][2].
 3. Uncomment the `hostPort` setting so that Istio sidecars can connect to the Agent and submit traces.
 
-**Note**: The Agent pods fails to start up if Istio's automatic sidecar injection is enabled on the namespace that Datadog Agent runs in. To avoid this:
-
-- [Disable Sidecar Injection for the Datadog Agent](#disable-sidecar-injection-for-the-datadog-agent)
-- [Create a Headless Service for the Datadog Agent](#create-a-headless-service-for-the-datadog-agent)
-
-#### Disable Sidecar Injection for the Datadog Agent
-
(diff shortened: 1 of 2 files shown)


jaygorrell:
Wait, the agent can run the Istio sidecar and function properly now? Even the new Istio blog series still calls for that: https://www.datadoghq.com/blog/istio-datadog/#set-up-the-datadog-agent

cgilmour:
@jaygorrell it's a little bit complicated to explain, but I'll try. The intention is to update the docs after thoroughly testing all scenarios below.

Istio installations can use a combination of
- mTLS `enabled` or `disabled
- datadog agent installed as `daemonset` or `deployment`
- agent `on mesh` or `off mesh`

Depending on the exact combination, the user may need to create a normal k8s clusterIP service, a headless service, an istio DestinationRule, and have sidecar injection explicitly enabled or disabled.

The original docs worked fine with v1.1.x releases, but it turns out they interfere with things in some releases since then. (Not sure the exact version, sorry.)
So for the most common setups reported by users, it didn't work correctly, and the fix was to _not_ create the mentioned headless service.
Those setups are using datadog installed as a daemonset, agent "off-mesh" and mTLS either enabled or disabled.

The scenario I expect to need a headless service, DestinationRule and sidecar injection enabled is: mTLS enabled, agent installed as a `Deployment` and wanting it "on-mesh". It might also need an istio `VirtualService` and adjustments to the readiness / liveness checks. That'll be discovered during the testing, with updates to the documentation when completed.

jaygorrell:
Thanks for that. Is it not a common piece to call out that sidecar injection for the agent should be disabled though? As far as I know that's always true in default setups that nearly all customers would want. Happy to share anything helpful about my testing as well.

PR #6429, opened 2020-01-17
Adding command to build doc faster

l0k0ms:
### What does this PR do?
Creates a new Makefile command:

`make start-no-pre-build` 

that builds the documentation without any extra content.

This allow to decrease build time from **~90s** to **~40s** when the pulled content is available locally.. it should decrease build time even more if content is not available locally.

Other changes:

* Updates Main Readme.md
* Changes default value for Github Token from `False` to `false`
* If Python 3 is missing => documentation doesn't build 
* Lint bash files

### Motivation

Allowing faster build when not working on pulled content.

### Preview link

* http://docs-staging.datadoghq.com/gus/build-faster

### DIFF
--- a/local/bin/py/build/update_pre_build.py
+++ b/local/bin/py/build/update_pre_build.py
@@ -83,7 +83,7 @@ class Build:
         except:
             if getenv("LOCAL") == 'True':
                 print(
-                    "\x1b[33mWARNING\x1b[0m: Local mode detected: Integration merge failed, documentation is now in degraded mode.")
+                    "\x1b[33mWARNING\x1b[0m:Integration merge failed, documentation is now in degraded mode.")
             else:
                 print(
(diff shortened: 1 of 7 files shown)


local/bin/py/build/update_pre_build.py
```diff
@@ -83,7 +83,7 @@ def build_documentation(self, list_of_contents):
         except:
             if getenv("LOCAL") == 'True':
                 print(
-                    "\x1b[33mWARNING\x1b[0m: Local mode detected: Integration merge failed, documentation is now in degraded mode.")
+                    "\x1b[33mWARNING\x1b[0m:Integration merge failed, documentation is now in degraded mode.")
```
l0k0ms:
```suggestion
                    "\x1b[33mWARNING\x1b[0m: Integration merge failed, documentation is now in degraded mode.")
```

PR #6517, opened 2020-01-28
Update nodejs.md with required parameter

grahamlyons:
Add the `service` parameter to the initialisation example - without the service the trace agent drops the data.

### Preview link

### DIFF
--- a/content/en/tracing/setup/nodejs.md
+++ b/content/en/tracing/setup/nodejs.md
@@ -52,7 +52,9 @@ Finally, import and initialize the tracer:
 
 ```js
 // This line must come before importing any instrumented module.
-const tracer = require('dd-trace').init()
+const tracer = require('dd-trace').init({
+  service: 'service-name',
+})
 ```
 
 ##### TypeScript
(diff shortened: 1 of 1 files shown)


content/en/tracing/setup/nodejs.md
```diff
@@ -52,7 +52,9 @@ Finally, import and initialize the tracer:
 
 ```js
 // This line must come before importing any instrumented module.
-const tracer = require('dd-trace').init()
+const tracer = require('dd-trace').init({
+  service: 'service-name',
```
l0k0ms: let's use the placeholder format here: 

```suggestion
  service: '<SERVICE_NAME>',
```

content/en/tracing/setup/nodejs.md
```diff
@@ -52,7 +52,9 @@ Finally, import and initialize the tracer:
 
 ```js
 // This line must come before importing any instrumented module.
-const tracer = require('dd-trace').init()
+const tracer = require('dd-trace').init({
+  service: '<SERVICE_NAME>',
```
rochdev: This is not a required parameter.

rochdev: From reading the original issue, it looks like there is a bug since the service name should automatically added. I would recommend opening an issue in https://github.com/DataDog/dd-trace-js

l0k0ms:
Closing this PR as it seems to be a bug, not a documentation update.

PR #6563, opened 2020-02-03
layout logic for disabled toc

zbayoff:
### What does this PR do?
1. When TOC is disabled with Front matter: `disable_toc: true`, the mainContent layout will be 9 columns as opposed to the 7 columns when the TOC is visible. 
2. Adds i18n support for TOC heading.

### Motivation
https://trello.com/c/IycJHnm7/4350-docs-layout-updates-phase-2

### Preview link
https://docs-staging.datadoghq.com/zach/layout-updates/

### DIFF
--- a/content/en/getting_started/_index.md
+++ b/content/en/getting_started/_index.md
@@ -1,6 +1,7 @@
 ---
 title: Getting started
 kind: documentation
+disable_toc: true
 aliases:
   - /overview
   - /getting_started/faq/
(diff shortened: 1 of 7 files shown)


l0k0ms:
@zbayoff regarding: https://docs-staging.datadoghq.com/zach/layout-updates/infrastructure/ it seems that the parameter needs to be explicitly set in order to enable the 9 columns layout right?

zbayoff:
> @zbayoff regarding: https://docs-staging.datadoghq.com/zach/layout-updates/infrastructure/ it seems that the parameter needs to be explicitly set in order to enable the 9 columns layout right?

@l0k0ms correct.

PR #6676, opened 2020-02-14
add Pipefy community Rust tracer lib

fhsgoncalves:
### What does this PR do?
It adds `datadog-apm-rust` lib written by Pipefy to the list of tracing libraries.
More info: https://github.com/pipefy/datadog-apm-rust

### Motivation
Give an option to the rust community in use datadog apm with rust.

### Preview link
N.A.

### Additional Notes
No.


phrawzty:
👋 @fhsgoncalves! First off, thanks for letting us know about your project—it's always great to see the community coming up with new and useful tools. 😄 We've taken a look at the library as it stands today, and we all agree that's it's a great start, and with a little more development it could become a great addition to the Datadog ecosystem.

Our suggestion is to first work on some sort of buffering / flow control mechanism; for example, what happens if millions or billions of spans are generated and expect to be sent? That's going to hurt something! Another area for improvement is with the span generation itself—making this more user-friendly would reduce the barrier to entry for usage. Looking further ahead, you'll want to also examine the context / carrier machinery too…

Once the library is closer to production-ready, we'd be more than happy to include it in the official list. Please keep us posted on your progress! And if you have any questions or want some help, don't hesitate to check out `#integrations` on our [public Slack](https://chat.datadoghq.com/), and to participate in our twice-monthly [Office Hours](https://docs.datadoghq.com/developers/office_hours/).

Cheers!

fhsgoncalves:
Hey @phrawzty, thank you for the reply!

The lib currently is a low level client for the datadog apm, and it does not automatically track the trace/spans for the user, but it is in the plan to integrate with libs like [tracing](https://github.com/tokio-rs/tracing)(and [it is documented here](https://docs.rs/datadog-apm/0.1.2/datadog_apm/#features-that-are-not-included-yet-contributions-welcome)).

About handle millions or billions of spans: it is prepared for that, and I perform some load tests against a datadog agent, and I could see that all the traces arrived and were shown on the datadog apm UI. It has builtin buffering using tokio channel (similar to go channel) and can handle a really high amount of spans per second. You can also configure the number of spans to keep in the buffer, the amount of traces sent in a single request [and so on](https://docs.rs/datadog-apm/0.1.2/datadog_apm/struct.Config.html). Also, if for any reason it could not consume the buffer in time, the new traces will start to be discarded until the buffer get capacity again. So we are good on that point :smile:, and [it's also documented](https://docs.rs/datadog-apm/0.1.2/datadog_apm/#overview).

---

In fact, we are using it in production, in a high load system, but I'll notify you guys when the lib offer a more user friendly experience! Thanks :smile:

phrawzty:
Hello @fhsgoncalves! It's been a little while—just checking in to see if you've made any progress or have any updates to share regarding this library.

Stay safe out there!

fhsgoncalves:
@phrawzty hi! 
I didn't made any progress yet, however I'll do and I'll post it here!
Meanwhile, we are using it in production and it is working pretty well for us :)

PR #6728, opened 2020-02-20
Mount the directory containing the docker socket instead of the socket itself

L3n41c:
This is to handle the cases where the docker daemon is restarted.
In this case, the docker daemon will recreate its docker socket.
On k8s context, the containers will not be restarted.
And, if the agent container bind-mounted directly the socket, the container would still have access to the old socket instead of the one of the new docker daemon.
As a consequence, with a bind-mount of the socket directly, a docker restart disconnects the agent from docker forever; until the agent POD is destroyed and recreated.

On the other hand, if we bind-mount the directory containing the socket, when the daemon recreates a new socket following a restart, it will be visible from the agent container. The agent will be disconnected from the old socket, and it will reconnect on the new one.

On non-k8s context link ECS or bare docker, docker is configured by default to restart its containers when the daemon is restarted so that there isn’t the issue described above.

### DIFF
--- a/content/en/agent/kubernetes/daemonset_setup.md
+++ b/content/en/agent/kubernetes/daemonset_setup.md
@@ -118,6 +118,9 @@ spec:
               fieldRef:
                 fieldPath: status.hostIP
 
+          - {name: DD_CRI_SOCKET_PATH, value: /host/var/run/docker.sock}
+          - {name: DOCKER_HOST, value: unix:///host/var/run/docker.sock}
+
         ## Note these are the minimum suggested values for requests and limits.
         ## The amount of resources required by the Agent varies depending on:
         ## * The number of checks
@@ -131,7 +134,7 @@ spec:
             memory: "256Mi"
             cpu: "200m"
         volumeMounts:
-          - {name: dockersocket, mountPath: /var/run/docker.sock}
+          - {name: dockersocketdir, mountPath: /host/var/run}
           - {name: procdir, mountPath: /host/proc, readOnly: true}
           - {name: cgroups, mountPath: /host/sys/fs/cgroup, readOnly: true}
           - {name: s6-run, mountPath: /var/run/s6}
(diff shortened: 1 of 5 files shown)


content/en/agent/kubernetes/daemonset_setup.md
```diff
@@ -276,20 +279,26 @@ Mount `/var/lib/docker/containers` as well, since `/var/log/pods` is symlink to
...
     (...)
     - hostPath:
-        path: /var/run/docker.sock
-      name: dockersocket
+        path: /var/run
+      name: dockersocketdir
   (...)
 ```
 
+**Note**: Mounting only the `docker.sock` socket instead of the whole directory containing it prevents the agent from recovering after a docker daemon restart.
```
ruthnaebeck:
```suggestion
**Note**: Mounting only the `docker.sock` socket instead of the whole directory containing it prevents the Agent from recovering after a Docker daemon restart.
```

content/fr/agent/kubernetes/daemonset_setup.md
```diff
@@ -15,7 +15,7 @@ aliases:
   - /agent/kubernetes/apm
   - /integrations/faq/using-rbac-permission-with-your-kubernetes-integration
 ---
-Tirez profit des DaemonSets pour déployer l'Agent Datadog sur l'ensemble de vos nœuds (ou sur un nœud donné grâce [aux nodeSelectors][1]). 
+Tirez profit des DaemonSets pour déployer l'Agent Datadog sur l'ensemble de vos nœuds (ou sur un nœud donné grâce [aux nodeSelectors][1]).
```
ruthnaebeck: The french documentation should not be updated via PR. We have an translation service for this.

l0k0ms: Indeed @ruthnaebeck is right. We can merge this PR anyway, it's just the the FR changes are going to be overridden by the next translation PR

PR #6735, opened 2020-02-20
Region redirects

zbayoff:
### What does this PR do?
1. Adds Docs region redirect feature. Based on url params, cookies, or referrer, should direct the user to the Docs site with the correct region set (currently only `US` or `EU`). 
The dropdown button is still a WIP and waiting for Design.

2. Creates new shortcode `site-region` to replace `tabs` for region specific markdown. Used like so: 

```
{{< site-region region="us" >}}

//region specific markdown/code

{{< /site-region >}}
```

3. Creates new shortcode `region-param` to use as a placeholder for specific params that will take a value corresponding to the chosen region. This shortcode is used in conjunction to the new `regions.config.js` file which defines some placeholder values for specific regions. For example:

In `regions.config.js`: 
```
export default {
    dd_site: {
        us: 'datadoghq.com',
        eu: 'datadoghq.eu'
    },
    tcp_endpoint: {
        us: 'intake.logs.datadoghq.com',
        eu: 'tcp-intake.logs.datadoghq.eu'
    },
    tcp_endpoint_port: {
        us: '10514',
        eu: '1883'
    }
};
```
one would add a placeholder value in this config file and then use in a markdown like so: 

```
Replace the placeholder value `<DATADOG_SITE>` with {{< region-param key="dd_site" code="true" >}}.
```

The `code="true"` shortcode param will style the output as inline code as if using backticks. 
This page uses this shortcode. See the update by changing the region: 
https://docs-staging.datadoghq.com/zach/region-redirects/integrations/alcide/

If an implementor uses this shortcode, but fails to pass a key, hugo will fail to build and output a message. If an implementor fails to add the key to the region-config.js file, an error is thrown in the console. 

4. Replaces all `app.datadoghq.*` links on load or with new async page with either `app.datadoghq.com` or `app.datadoghq.eu` based on the region. This is implemented only on main content pages (not home, or api pages).

### Motivation
https://docs.google.com/document/d/1a8PG7cgz7exDDMy7Q1TEaw8PykhgPwoblssRGEFINLg/edit#heading=h.qn7fh8626fdq

### Preview link
https://docs-staging.datadoghq.com/zach/region-redirects/getting_started/logs/

This page ^ implements the new `site-region` shortcode. 
Test:
1. load page without any cookie set, should default to `US` region. 
2. cookie will now be set as `US`. 
3. change region to `EU` with dropdown. Markdown inside the shortcode should be updated to `EU` specific. Cookie will also be set as `EU` upon next reload of page. 
4. Load url https://docs-staging.datadoghq.com/zach/region-redirects/getting_started/logs/?site=us
with a query param `?site=eu` or `?site=us`. Dropdown should reflect this region, cookie will be set and appropriate content shown. 
5. Document referrer.
Hard to test a link within app.datadoghq.eu as I don't have access. But hackily, I created a test anchor on the login page to link to the preview site. The preview site should then load the `EU` site. Test video here: https://a.cl.ly/Apurn4ww


src/scripts/region-redirects.js
```diff
@@ -0,0 +1,159 @@
+import Cookies from 'js-cookie';
+import config from '../../regions.config';
+// import datadogLogs from './components/dd-browser-logs-rum';
+
+const allowedRegions = ['us', 'eu'];
```
nsollecito: can we pull this in from the config file as well? This way, when new regions are added, we don't have to change JS and can update everything from config

layouts/partials/sidenav/sidenav.html
```diff
@@ -18,7 +18,17 @@
...
         <div class="col">
-            <a href="{{ "" | absLangURL }}"><h4>{{ i18n "datadog_docs" }}</h4></a>
+            <h4>
+                <a class="text-primary" href="{{ "" | absLangURL }}">{{ i18n "datadog_docs" }}</a>
+                <span>
+                    <label for="region-select" class="sr-only">Region</label>
+                    <select id=region-select class="js-region-selector">
+                        <option disabled selected value>--</option>
+                        <option value="us">US</option>
+                        <option value="eu">EU</option>
```
davidejones: Similarly to Nicks comment I'm wondering if this should come from a config too.

layouts/shortcodes/region-param.html
```diff
@@ -0,0 +1,4 @@
+{{ if  not (.Get "key") }}{{ errorf "region-param shortcode error: Missing value for param 'key': %s" .Position }}{{ end }}
+{{ if eq (.Get "code") "true"  }}<code class="js-region-param" data-region-param="{{ .Get "key" }}"></code>
+{{ else }}<span class="js-region-param" data-region-param="{{ .Get "key" }}"></span>
+{{ end }}
```
davidejones: You are probably aware but thought I'd mention hyphens on on tags can sometimes help with [whitespace](https://gohugo.io/templates/introduction/#whitespace) too

PR #6748, opened 2020-02-24
system-probe: Update install instruction for 7.18

albertvaka:
### What does this PR do?
- Mention that the `start` and `enable` steps are only needed for Agents < 6.18/7.18.
- Change the default instructions to use `sytemctl` and move the alternative instructions using `service` to the note text. Not only `systemctl` (systemd) is more common in modern distros, but also I think that the instructions using `service` were wrong since I don't think that `service enable` exists.
- When on older Agent versions, mention that system-probe needs to be started on boot before the main Agent, but leave up to the user to figure out how to do that on non-systemd systems (since I think that the `service enable` instructions we were giving were wrong).

### Motivation
Since Agent 6.18/7.18 there is no need to start and enable system-probe separately.

### Preview

https://docs-staging.datadoghq.com/albertvaka/system-probe-7.18/network_performance_monitoring/installation?tab=agent#setup

### DIFF
--- a/content/en/network_performance_monitoring/installation.md
+++ b/content/en/network_performance_monitoring/installation.md
@@ -62,7 +62,7 @@ To enable network performance monitoring with the Datadog Agent, use the followi
         enabled: true
     ```
 
-4. If you are running an Agent older than v6.18/7.18, manually start the system-probe and enable it to start on boot (since v6.18/v7.18, system-probe will start automatically when the Agent is started by systemd):
+4. If you are running an Agent older than v6.18/7.18, manually start the system-probe and enable it to start on boot (since v6.18/v7.18, system-probe will start automatically when the Agent is started):
 
     ```shell
     sudo systemctl start datadog-agent-sysprobe 
--- a/content/en/network_performance_monitoring/installation.md
+++ b/content/en/network_performance_monitoring/installation.md
@@ -62,7 +62,7 @@ To enable network performance monitoring with the Datadog Agent, use the followi
         enabled: true
     ```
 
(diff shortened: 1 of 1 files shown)


content/en/network_performance_monitoring/installation.md
```diff
@@ -62,29 +62,22 @@ To enable network performance monitoring with the Datadog Agent, use the followi
         enabled: true
     ```
 
-4. Start the system-probe:
+4. If you are running an Agent older than v6.18/7.18, manually start the system-probe and enable it to start on boot (since v6.18/v7.18, system-probe will start automatically when the Agent is started by systemd):
```
albertvaka:
```suggestion
4. If you are running an Agent older than v6.18/7.18, manually start the system-probe and enable it to start on boot (since v6.18/v7.18, system-probe will start automatically when the Agent is started):
```
Removing mention to systemd since we also added support for upstart.

content/en/network_performance_monitoring/installation.md
```diff
@@ -62,29 +62,22 @@ To enable network performance monitoring with the Datadog Agent, use the followi
         enabled: true
     ```
 
-4. Start the system-probe:
+4. If you are running an Agent older than v6.18/7.18, manually start the system-probe and enable it to start on boot (since v6.18/v7.18, system-probe will start automatically when the Agent is started):
```
l0k0ms:
```suggestion
4. If you are running an Agent older than v6.18 or 7.18, manually start the system-probe and enable it to start on boot (since v6.18/v7.18, system-probe will start automatically when the Agent is started):
```

albertvaka:
We are good to merge this @l0k0ms now that 7.18 is out.

PR #6761, opened 2020-02-25
[PROD] fix links in RBAC instructions for cluster agent install

gml3ff:
Existing RBAC manifests that are linked here are for the _node agent_ not the cluster agent. 

Was following these instructions with a customer recently and it didn't create any of the `dca` RBAC resources that are required for this to work. This should correct things. 

### Preview link


gml3ff:
CC'ing @Simwar as well since he made the recommendation to just swap out the links rather than rewrite this section.

PR #6823, opened 2020-03-03
[PROD] include cluster agent token piece

gml3ff:
### What does this PR do?
The cluster agent is very commonly deployed in Kubernetes environments and requires an auth token to communicate with the node agent. This is mentioned in [these docs](https://docs.datadoghq.com/agent/cluster_agent/setup/?tab=secret#configure-the-datadog-agent), but it will make for a much easier install if the environment is already there and just needs to be commented out (we already do this for a lot of add-ons like APM). 

### Preview link

### DIFF
--- a/content/en/agent/kubernetes/daemonset_setup.md
+++ b/content/en/agent/kubernetes/daemonset_setup.md
@@ -121,6 +121,21 @@ spec:
           - {name: DD_CRI_SOCKET_PATH, value: /host/var/run/docker.sock}
           - {name: DOCKER_HOST, value: unix:///host/var/run/docker.sock}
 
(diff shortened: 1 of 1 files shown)


content/en/agent/kubernetes/daemonset_setup.md
```diff
@@ -121,6 +121,21 @@ spec:
           - {name: DD_CRI_SOCKET_PATH, value: /host/var/run/docker.sock}
           - {name: DOCKER_HOST, value: unix:///host/var/run/docker.sock}
 
+          ## For secure communication with the cluster agent (required to use the cluster agent)
```
cswatt:
```suggestion
          ## For secure communication with the Cluster Agent (required to use the Cluster Agent)
```

PR #7216, opened 2020-04-20
Add API test is retried immediately after failing

lgh2:
### What does this PR do?

Adds that API tests are retried immediately after failing 

### Motivation

Customer request 

### Preview link

https://docs-staging.datadoghq.com/laura.hampton/api-test-retry/getting_started/synthetics/api_test/#test-results


sarina-dd:
To my knowledge, retry has to be set. [Default in the UI is zero](https://a.cl.ly/v1uDrnx1). @MargotLepizzera can you confirm whether or not a test automatically retries upon failure?

MargotLepizzera:
@sarina-dd Retry indeed has to be set, otherwise it defaults to 0, as shown in the UI. 
But if it is set to at least 1, the default interval is 300 ms (ie. we wait for 300ms before retrying a test that failed, so it's re-tried almost immediately). This interval is not configurable in the UI but is through the API. 

So I'd go for something like:

You can also decide after how many failures you want a location to be considered as fail using the option below:

```text
Retry x time before location is marked as failed
```

Note: by default retries, we wait for 300ms before retrying a test that failed. This interval can be configured via the API. 

----

It might be good to mention that in the Browser test getting started + on the regular API & Browser test doc pages.

sarina-dd:
> Note: by default, we wait for 300ms before retrying a test that failed. This interval can be configured via the [API](https://docs.datadoghq.com/api/?lang=bash#create-a-test).

Thanks @MargotLepizzera! Apologies on the delay - didn't get a notification for this until now. I'll update this PR so it includes those notes.

PR #7229, opened 2020-04-21
docs(helm v2): mv rbac.create agents.rbac.create

JordanSussman:
### What does this PR do?

Updates the `rbac.create` option to be `agents.rbac.create` which is documented on https://github.com/helm/charts/blob/master/stable/datadog/docs/Migration_1.x_to_2.x.md.

### Motivation

Confusion when looking at the mix of documentation that references v1 and v2 options for the helm chart.


l0k0ms:
Many thanks for this !

PR #7248, opened 2020-04-23
Update developers/dogstatsd/high_throughput.md to add optimization sections

remeh:
### What does this PR do?

Add sections in the Dogstatsd guides to help optimizing its usage.

### Motivation

Better documentation for DSD high throughput scenarios.

### Preview link

https://docs-staging.datadoghq.com/remeh/dogstatsd-optim-guide/developers/dogstatsd/high_throughput/?tab=go#limit-the-amount-of-packet-drops


content/en/developers/dogstatsd/high_throughput.md
```diff
@@ -218,6 +218,140 @@ Then set the Agent `dogstatsd_so_rcvbuf` configuration option to the same number
...
+```
+
+In the same manner, the `net.core.wmem_max` could be incremented to 25MiB to improve
+client writing performances:
+
+```conf
+net.core.wmem_max = 26214400
+```
+
+## Configuration optimization guide
```
olivielpeau, line 250: Most of this section probably shouldn't live in our public docs:

* I think that the config options mentioned here should not be considered part of the "officially-supported config options", so that we have room to break them in the future as we improve performance and perf options in the daemon. But let me know what you think of course.
* overall this seems a bit too complex for end users, and goes into lots of internal details that are very interesting to us but not really to the end user.

For now, in the public docs here, I think we should stick to general recommendations like:

* recommend to use the latest version of an official dogstatsd client (which should have best performance and by default use the right defaults for UDS and UDP datagram sizes -- if it's not the case now for some clients it will be in a couple of months)
* mention these "optimal" datagram sizes for users using non-official clients.
* if we add a setting that sets at a high level in the Agent the sizes of the buffers, only present here this single setting, and describe what the tradeoff is

content/en/developers/dogstatsd/high_throughput.md
```diff
@@ -218,6 +218,140 @@ Then set the Agent `dogstatsd_so_rcvbuf` configuration option to the same number
...
+sysctl -w net.unix.max_dgram_qlen=512
+```
+
+Add the following configuration to `/etc/sysctl.conf` to make this change permanent:
+
+```conf
+net.unix.max_dgram_qlen = 512
+```
+
+In the same manner, the `net.core.wmem_max` could be incremented to 25MiB to improve
```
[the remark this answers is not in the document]
remeh: In order to have a controlled behavior, let's ask to force the value to `dogstatsd_so_rcvbuf` as well 👍 This way, no default or unmanaged value are in the pipe.

content/en/developers/dogstatsd/high_throughput.md
```diff
@@ -218,6 +218,140 @@ Then set the Agent `dogstatsd_so_rcvbuf` configuration option to the same number
 dogstatsd_so_rcvbuf: 26214400
 ```
 
+### Over UDS (Unix Domain Socket)
+
+#### Linux
+
+Some Linux distributions have a kernel datagram queue size which can be low. This can be confirmed using the following command:
```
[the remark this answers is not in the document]
remeh: I've added detail what it is and slight details on how Linux is using that value / working internally.

remeh:
@olivielpeau 
I've tried to address all your feedback in my last commit:
  - less detailed information on Dogstatsd internals (I'll move some of these either in the agent repo or other docs)
  - here and there fixes with your feedback
  - added a section about using official clients and to keep using latest version available
  - I've decided to stick with `dogstatsd_queue_size` field in the "Limit max memory usage" section, here's why: I think that it is easier to document this "internal" knob for now than to introduce a new one virtually having the same effect, especially when we are not sure how we will maintain this virtual one correctness in the future. If a new knob, better suited for customer-facing, I think that we'll just have to change the documentation at this point and to re-use the value of `dogstatsd_queue_size` if set, not the contrary (the contrary = maintain a "virtual" configuration field with changed internals that could possibly not have this capacity of limiting the buffering). WDYT?

content/en/developers/dogstatsd/high_throughput.md
```diff
@@ -218,6 +224,83 @@ Then set the Agent `dogstatsd_so_rcvbuf` configuration option to the same number
...
+to process them in batches later in the pipeline. The official Dogstatsd clients are
+capable of grouping metrics to have the best ratio of the number of metrics per packet.
+
+If the clients properly send packets having a size close to `dogstatsd_buffer_size`, Dogstatsd
+will perform better. Be careful: the packets must not be larger than this size, otherwise, the
+Dogstatsd server won't be able to load them completely in the buffer and some of metrics will
+be malformed. Please use the corresponding configuration field in your Dogstatsd clients.
+
+Note for UDP: as the UDP packets are most of the time going through the Ethernet and IP layer,
+they inherently have a size limited to 1472.
```
prognant, line 280: 1472 is the "usual limit" (IPv4 header with no option/1500 bytes Ethernet MTU) and it may be larger (bigger MTU) or smaller (due to IPv6 larger header).
```suggestion
Note for UDP: as the UDP packets are most of the time going through the Ethernet and IP layer, fragmentation should be avoided even if supported. Packet size should be limited to fit at most a single Ethernet frame and prevent IP fragmentation.
```
I'm not sure the suggested wording is understandable, but I think we should definitely mentioned that IP fragmentation should be avoided.

remeh: You're right I'll add more precision here

content/en/developers/dogstatsd/high_throughput.md
```diff
@@ -218,6 +224,83 @@ Then set the Agent `dogstatsd_so_rcvbuf` configuration option to the same number
...
+Dogstatsd tries to absorb burst of metrics sent by the clients but in order to do so,
+it needs to use memory. Even if this is for a short amount of time and even if this
+memory is quickly released to the OS, a spike happens and that could be an issue in
+containerized environments where limit on memory usage could evict pods or containers.
+
+This first thing to do to not have Dogstatsd reach its maximum memory usage is
+to avoid sending metrics in bursts in your application.
+
+Another thing to look at to limit the maximum memory usage is to reduce the buffering.
+The main buffer of Dogstatsd is configurable with the `dogstatsd_queue_size` field, its default value of `1024` induce an approximate maximum memory usage of 768MB. Note that reducing this
```
ogaca-dd, line 295: `induce` => `induces`?

content/en/developers/dogstatsd/high_throughput.md
remeh: I've addressed your feedback, please re-open if you think that's not clear enough!

content/en/developers/dogstatsd/high_throughput.md
```diff
@@ -26,6 +26,12 @@ Most of the time the symptoms can be alleviated by tweaking some configuration o
 
 ## General tips
 
+### Use Datadog official clients
+
+Official Dogstatsd clients are provided by Datadog for every major programming language.
+Their features are optimized for usage with Dogstatsd, please use them if one exist
```
olivielpeau, line 32:
```suggestion
Their features are optimized for usage with Dogstatsd, please use them if one exists
```

content/en/developers/dogstatsd/high_throughput.md
```diff
@@ -218,6 +224,85 @@ Then set the Agent `dogstatsd_so_rcvbuf` configuration option to the same number
...
+net.core.wmem_max = 4194304
+```
+
+Then set the Agent `dogstatsd_so_rcvbuf` configuration option to the same number in `datadog.yaml`:
+
+```yaml
+dogstatsd_so_rcvbuf: 4194304
+```
+
+Please note that setting these values will need a reboot of the host.
```
olivielpeau, line 264: is this true for all the options mentioned here? 🤔

Pretty sure at least `net.core.wmem_max` is taken into account at runtime without a reboot?

Can be addressed in a follow-up PR in any case

content/en/developers/dogstatsd/high_throughput.md
```diff
@@ -218,6 +224,85 @@ Then set the Agent `dogstatsd_so_rcvbuf` configuration option to the same number
...
+
+### Ensure proper packet sizes
+
+Sending packets with an adequate size to Dogstatsd will avoid extra work to the server.
```
olivielpeau, line 268: let's keep naming consistent:

```suggestion
Sending packets with an adequate size to Dogstatsd will avoid extra work to the Agent.
```

content/en/developers/dogstatsd/high_throughput.md
```diff
@@ -218,6 +224,85 @@ Then set the Agent `dogstatsd_so_rcvbuf` configuration option to the same number
...
+
+If the packets sent are too small, Dogstatsd will have to pack several together in order
+to process them in batches later in the pipeline. The official Dogstatsd clients are
+capable of grouping metrics to have the best ratio of the number of metrics per packet.
```
olivielpeau, line 272: Since this section is basically only relevant for readers who do not use the official clients, I think it's worth highlighting this first. Something like this as the second sentence of this section could work for example:

```
The latest versions of the official Dogstatsd clients send packets with a size optimized for performance. If you're using them, skip this section, otherwise read on.
```

(not super important, can be addressed separately)

content/en/developers/dogstatsd/high_throughput.md
```diff
@@ -218,6 +224,85 @@ Then set the Agent `dogstatsd_so_rcvbuf` configuration option to the same number
...
+be malformed. Please use the corresponding configuration field in your Dogstatsd clients.
+
+Note for UDP: as the UDP packets are most of the time going through the Ethernet and IP layer,
+it is be better to avoid IP packets fragmentation by limiting the packet size to a value lower
+than a single Ethernet frame on your network. Most of the time, IPv4 networks are configured with
+a MTU of 1500 bytes, in this situation the packet size of sent packets should be limited to 1472.
+
+Note for UDS: for the best performances, UDS packet should have a size of 8192 bytes.
+
+### Limit the maximum memory usage
```
olivielpeau, line 286:
```suggestion
### Limit the maximum memory usage of the Agent
```

content/en/developers/dogstatsd/high_throughput.md
```diff
@@ -218,6 +224,85 @@ Then set the Agent `dogstatsd_so_rcvbuf` configuration option to the same number
...
+
+Dogstatsd tries to absorb burst of metrics sent by the clients but in order to do so,
```
olivielpeau, line 288: AFAIK the terminology we use in our public docs is:

* `Dogstatsd`: the protocol
* the `Agent` (or `Dogstatsd-standalone binary`): the dogstatsd server
* `Dogstatsd clients`: (obviously) the clients :)

So in this section, I think we should replace most (all?) instances of `Dogstatsd` with `Agent`. If we're talking about the memory usage of the Dogstatsd bits of the Agent, I guess we can write `the Dogstatsd server within the Agent`, or something of that kind, WDYT?

content/en/developers/dogstatsd/high_throughput.md
```diff
@@ -218,6 +224,85 @@ Then set the Agent `dogstatsd_so_rcvbuf` configuration option to the same number
...
+it needs to use memory. Even if this is for a short amount of time and even if this
+memory is quickly released to the OS, a spike happens and that could be an issue in
+containerized environments where limit on memory usage could evict pods or containers.
+
+This first thing to do to not have Dogstatsd reach its maximum memory usage is
+to avoid sending metrics in bursts in your application.
+
+Another thing to look at to limit the maximum memory usage is to reduce the buffering.
+The main buffer of Dogstatsd is configurable with the `dogstatsd_queue_size` field, its default value of `1024` induces an approximate maximum memory usage of 768MB. Note that reducing this
```
olivielpeau, line 297: let's mention since which agent version this config option exists

content/en/developers/dogstatsd/high_throughput.md
remeh: No you're right, I've first added this here to be cautious but I read yesterday in kubernertes doc that net sysctls doesn't need a reboot! I'll remove that.

content/en/developers/dogstatsd/high_throughput.md
remeh: 👍

content/en/developers/dogstatsd/high_throughput.md
```diff
@@ -218,12 +222,72 @@ Then set the Agent `dogstatsd_so_rcvbuf` configuration option to the same number
...
+
+Note for UDP: as the UDP packets are most of the time going through the Ethernet and IP layer, it is be better to avoid IP packets fragmentation by limiting the packet size to a value lower than a single Ethernet frame on your network. Most of the time, IPv4 networks are configured with a MTU of 1500 bytes, in this situation the packet size of sent packets should be limited to 1472.
+
+Note for UDS: for the best performances, UDS packet should have a size of 8192 bytes.
+
+### Limit the maximum memory usage of the Agent
+
+The Agent tries to absorb burst of metrics sent by the Dogstatsd clients but in order to do so, it needs to use memory. Even if this is for a short amount of time and even if this memory is quickly released to the OS, a spike happens and that could be an issue in containerized environments where limit on memory usage could evict pods or containers.
+
+This first thing to do to not have the Dogstatsd server within the agent to reach its maximum memory usage is to avoid sending metrics in bursts in your application.
```
apigirl, line 276:
```suggestion
Avoid sending metrics in bursts in your application - this prevents the Datadog Agent from reaching its maximum memory usage.
```

content/en/developers/dogstatsd/high_throughput.md
```diff
@@ -218,12 +222,72 @@ Then set the Agent `dogstatsd_so_rcvbuf` configuration option to the same number
...
+
+Another thing to look at to limit the maximum memory usage is to reduce the buffering. The main buffer of the Dogstatsd server within the Agent is configurable with the `dogstatsd_queue_size` field (since Datadog Agent 6.1.0), its default value of `1024` induces an approximate maximum memory usage of 768MB.
```
apigirl, line 278:
```suggestion
Another thing to look at to limit the maximum memory usage is to reduce the buffering. The main buffer of the DogStatsD server within the Agent is configurable with the `dogstatsd_queue_size` field (since Datadog Agent 6.1.0), its default value of `1024` induces an approximate maximum memory usage of 768MB.
```

content/en/developers/dogstatsd/high_throughput.md
```diff
@@ -218,12 +222,72 @@ Then set the Agent `dogstatsd_so_rcvbuf` configuration option to the same number
...
+
+Note that reducing this buffer could have the caveat of increasing the number of packet drops.
```
apigirl, line 280:
```suggestion
**Note**: reducing this buffer could increase the number of packet drops.
```

content/en/developers/dogstatsd/high_throughput.md
```diff
@@ -218,12 +222,72 @@ Then set the Agent `dogstatsd_so_rcvbuf` configuration option to the same number
...
+
+Find below an example decreasing the max memory usage of Dogstatsd to approximately 384MB:
```
apigirl, line 282:
```suggestion
This example decreases the max memory usage of DogStatsD to approximately 384MB:
```

content/en/developers/dogstatsd/high_throughput.md
```diff
@@ -218,12 +222,72 @@ Then set the Agent `dogstatsd_so_rcvbuf` configuration option to the same number
...
+dogstatsd_queue_size: 512
+```
+
 ## Client side telemetry
 
-Dogstatsd clients send telemetry metrics by default to the Agent. This allows
-you to better troubleshoot where bottleneck exists. Each metric will be
-tagged with the client language and the client version. These metrics will not be
-counted as custom metrics and will not be billed.
+Dogstatsd clients send telemetry metrics by default to the Agent. This allows you to better troubleshoot where bottleneck exists. Each metric will be tagged with the client language and the client version. These metrics will not be counted as custom metrics and will not be billed.
```
apigirl, line 290:
```suggestion
DogStatsD clients send telemetry metrics by default to the Agent. This allows you to better troubleshoot where bottlenecks exist. Each metric is tagged with the client language and the client version. These metrics are not counted as custom metrics.
```

remeh:
Thanks a lot for the detailed review @apigirl 🙇

PR #7249, opened 2020-04-23
[WEB-165] security rules prebuild

davidejones:
### What does this PR do?

Uses the pre build script to combine security rules into the documentation versus the current method. Note this PR doesn't contain the setup for displaying this content only for retrieving it.

### Motivation
https://datadoghq.atlassian.net/browse/WEB-165

### Preview link
https://docs-staging.datadoghq.com/david.jones/prebuild-secrules/


l0k0ms:
I think you need to add this new imported content to the `UNTRACKED_EXTRAS` parameter from the `.gitlab-ci.yml` so that they get translated.

PR #7305, opened 2020-04-30
Api desc update

l0k0ms:
### What does this PR do?

Adds back information about the API rate limits and auth logic.

@zbayoff i tried to put the "overview" section on top but it seems to be sorted alphabetically by nature.. do you know if we could make just this section sticky ?

Another issue is that when selecting an entry in this new section, the api layout changes for some reason :/ 

### Motivation
Iso content with current API doc

### DIFF
--- a/config/_default/menus/menus.en.yaml
+++ b/config/_default/menus/menus.en.yaml
@@ -1351,7 +1351,7 @@ main:
     identifier: dev_tools_metrics_powershell
     weight: 204
   - name: 'Submission - API '
-    url: 'api/#post-timeseries-points'
+    url: api/v1/metrics/#submit-metrics
     parent: dev_tools_metrics
     identifier: dev_tools_metrics_api
     weight: 205
@@ -1386,7 +1386,7 @@ main:
     identifier: dev_tools_events_mail
     weight: 303
   - name: API
-    url: 'api/#post-an-event'
+    url: /api/v1/events/#post-an-event
     parent: dev_tools_events
     identifier: dev_tools_events_api
     weight: 304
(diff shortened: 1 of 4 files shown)


zbayoff:
We have to update the build-api-script.js which updates the menu.yaml file with updated API content. All menu items under api_v1 and api_v2 are automatically generated, and are deleting the hardcoded ones you added in this PR. Will need to fix that first.

config/_default/menus/menus.en.yaml
```diff
@@ -2130,6 +2143,19 @@ api_v1:
...
+  - name: Authentication
+    url: /api/v1/
+    parent: API V1 overview
+    url: authentication
+    identifier: API V1 authetication
+  - name: Rate Limiting
+    url: /api/v1/
+    parent: API V1 overview
+    url: rate-limiting
+    identifier: API V1 rate limiting
```
l0k0ms, line 2158:
```suggestion
```

config/_default/menus/menus.en.yaml
```diff
@@ -1562,6 +1562,19 @@ main:
...
```
l0k0ms, line 1577:
```suggestion
```

config/_default/menus/menus.en.yaml
```diff
@@ -1386,7 +1386,7 @@ main:
     identifier: dev_tools_events_mail
     weight: 303
   - name: API
-    url: 'api/#post-an-event'
+    url: /api/v1/events/#post-an-event
```
l0k0ms, line 1389:
```suggestion
    url: api/v1/events/#post-an-event
```

PR #7317, opened 2020-04-30
Add snmp usage metering API documentation

carlmartensen:
There's a new usage type called SNMP Devices that is available via the public API. This is the documentation for it.


carlmartensen:
hold on merging to PM approves (should be done 5pm est 4/30)

PR #7361, opened 2020-05-05
Allow empty changes after OpenAPI processing

jirikuncar:
### What does this PR do?

Don't fail when there are no changes.

### Preview link

https://docs-staging.datadoghq.com/<BRANCH_NAME>/<PATH>

PR #7362, opened 2020-05-05
turn off openapi job for now

davidejones:
### What does this PR do?

Turns off the openapi github job. It is no longer needed to run on our side and is causing errors with no files to commit.

### Motivation

Is actually causing errors when there are no files to commit.

### Preview link
N/A


jirikuncar:
closes #7361

PR #7432, opened 2020-05-12
Hugo deploy (dryRun)

zbayoff:
### What does this PR do?
1. Uses new `hugo deploy` feature to deploy files to S3. This feature takes care of all diffing and including/excluding files, thus replacing AWS sync and copy commands currently used. 
First we'd like to run the command to deploy with `--dryRun` enabled, before the current AWS commands run, to verify that hugo deploy will upload/delete the correct files.

Upload strategy: 
- Deploy all files (`html` and media) to live Docs bucket: `datadog-docs-live-hugo`.
- Deploy static media file types `jpg,jpeg,png,gif,mp4,svg,json` to shared static bucket for all websites, `origin-static-assets/documentation`.

This is currently how to Docs deploy works, where images and videos are uploaded to two separate buckets. This should probably be changed in the future, with all static files going to one bucket. 

The `--maxDeletes` flag signifies the max number of files to delete. We currently set it to `0` for the statics `origin-static-assets/documentation` bucket, so static file will not be removed. This is just to safeguard that no images and videos will be deleted just yet.
`--maxDeletes` is set to `-1` for the main `datadog-docs-live-hugo` bucket, so files deleted locally will be removed remotely.

Corp-ci PR: https://github.com/DataDog/corp-ci/pull/79

### Preview link
http://docs-staging.datadoghq.com/zach/hugo-deploy-dryrun

### Gitlab Job
See the changes that would be made by Hugo deploy and check that they are made as intended
https://gitlab.ddbuild.io/DataDog/documentation/-/jobs/30281922


nsollecito:
@zbayoff why disable the matchers?

zbayoff:
> @zbayoff why disable the matchers?

Because it changed the file size on too many files so it was hard too see what files had actually changed

PR #7504, opened 2020-05-20
use Goldmark renderer, update code blocks

zbayoff:
### What does this PR do?
Updates hugo to use new `Goldmark` markdown rendering engine.
https://gohugo.io/getting-started/configuration-markup/#goldmark

When upgrading past hugo v. 0.60 we tried using this new engine, but there were issues with how some code blocks were being rendered. Issue documented here: https://datadoghq.atlassian.net/jira/software/projects/WEB/boards/102?selectedIssue=WEB-38

The solution to the issue of using the `code-block` shortcode within the `tabs` and `tab` shortcode is to use the syntax `{{< tab >}}` as opposed to `{{% tab %}}`. This only needs to be done when there is a `code-block` shortcode within a `tab` shortcode. The percent sign syntax for a shortcode will try to render everything within that shortcode as markdown. Usually this is what is desired, but with the new markdown renderer, everything within a `code-block` shortcode within `{{% tab %}}` shortcode would misrender newline characters and add `<p>` tags in the highlighted code. 

There are some cases where we need to have content in the `tab` shortcode rendered as markdown. In this case, I've implemented a new shortcode `markdownify` which uses the hugo function `markdownify` to render any content within the shortcode as markdown.

### Motivation
Gives us new features like markdown rendering hooks.
https://gohugo.io/getting-started/configuration-markup/#markdown-render-hooks

### Preview link
https://docs-staging.datadoghq.com/zach/use-goldmark-renderer/
 Since this PR updates the markdown renderer, we should do a thorough QA, testing code blocks, TOC (desktop and mobile), tabs, async pages, to make sure markdown is being rendered properly.


config/_default/config.yaml
```diff
@@ -26,7 +26,9 @@ enableGitInfo: true
 
 # Allow to render HTML in markdown files
 markup:
-  defaultMarkdownHandler: blackfriday
+  goldmark:
+    renderer:
+      unsafe: true
```
zbayoff, line 31: the `unsafe` parameter lets us render html tags in markdown.

content/en/developers/events/dogstatsd.md
```diff
@@ -171,8 +171,10 @@ $statsd->event('An error occurred.',
   );
 {{< /code-block >}}
 
+{{< markdownify >}}
```
zbayoff, line 174: need to use this new shortcode `markdownify` because we are now using `{{< tab "PHP" >}}` the `<` syntax which will not render inner content as markdown. Links also won't render using [link text][3] syntax with the link at the bottom of the page. Need to add the link inline. The format-links pre-commit hook tries to change this, so will have to commit code with --no-verify.

davidejones:
I think we have a pre commit hook that can format links inside tabs. I'm not sure if you have looked at that but we might need to confirm it isn't breaking this?

zbayoff:
> need to use this new shortcode `markdownify` because we are now using `{{< tab "PHP" >}}` the `<` syntax which will not render inner content as markdown. Link shortcuts then need to be placed within the shortcode. The format-links pre-commit hook tries to change this, so will have to commit code with --no-verify.

Yeah I mentioned that in one my comments. The script should be updated then.

PR #7530, opened 2020-05-22
Debug Log Sample for Python Said Generated by Go

MikeTarkington:
### What does this PR do?
Fixing what appeared to be a typo in the heading before our sample Python debug log since it said the sample was generated by the Go tracer rather than ddtrace for Python.

### Motivation
Minor error

### Preview link
https://docs.datadoghq.com/tracing/troubleshooting/?tab=python#span-generated-by-the-go-tracer

https://a.cl.ly/eDu64Wqk

https://docs-staging.datadoghq.com/MikeTarkington-Py_debug_sample_header/documentation/edit/master/content/en/tracing/troubleshooting/_index.md/

I don't think the preview URL I built works and was a little confused by the instructions.  That said, this change is minor enough and I expect the info I've provided so far is enough to move forward.  Please let me know if more confirmation is needed.

### Additional Notes
n/a


MikeTarkington:
Actually on further review, it looks like we're not providing the key instruction meant to be offered in this section for Python because we're not telling users to run their application with the env var `DATADOG_TRACE_DEBUG` as noted here: 
http://pypi.datadoghq.com/trace/docs/configuration.html?highlight=debug

Instead it's only mentioning using DD_LOGGING_RATE_LIMIT=0, which merely controls the extent to which each span in a trace gets logged.  I'm pretty sure that further revision is necessary to restore these instructions to working order.  I will do a little testing to confirm but I expect to make an additional revision in this PR shortly.

MikeTarkington:
Actually, please disregard, I simply missed a change to the layout of the doc and now I see that all the info is present. 🙇

PR #7536, opened 2020-05-25
Trace service.name fix

jstoja:
In the documentation, it seems that the configuration fields aren't correct.
It should be `dd.service.name` and `DD_SERVICE_NAME` as per https://github.com/DataDog/dd-trace-java/blob/ae9e5a827ada5b94d1bf761897a13abba414b509/dd-trace-api/src/main/java/datadog/trace/api/Config.java#L59


andrewardito:
We've updated our tracers across all languages for `env`, `service` and `version` to make these tags global and take precedence.  Documentation updates are following, which you see here. Everything is backwards compatible, so don't worry if you're using `dd.service.name`

PR #7548, opened 2020-05-26
WEB-271 fix anchors

zbayoff:
### What does this PR do?
Fix anchor issue where linking to a specific hash in the url would not scroll the page to the correct anchor. 

Removing the scroll animation to the header reduces the complexity and makes it easier to link to headers. This is the current behavior on the API pages. 

### Motivation
https://datadoghq.atlassian.net/jira/software/projects/WEB/boards/102?assignee=5d4b49d50fa6d40d14fc7b11&selectedIssue=WEB-271

### Preview link
http://docs-staging.datadoghq.com/zach/anchor-issue

Test clicking Page headers and subheaders, TOC items, and trying loading the page with a hash in the url in a new browser tab, and the page should be scrolled to the correct anchor.

### DIFF
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -59,7 +59,7 @@
                     <div class=" guide-col d-flex flex-column">
                             <div class="inner-content">
                             <div class="guide-no">{{ add $i 1}}.</div>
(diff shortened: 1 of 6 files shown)


zbayoff:
@jiboong This should be fixed.

PR #7576, opened 2020-05-28
Remove [profiling] extra from ddtrace on Python

jd:
Starting with 0.39, profiling is installed directly as part of ddtrace.

### DIFF
--- a/content/en/tracing/profiling/_index.md
+++ b/content/en/tracing/profiling/_index.md
@@ -74,10 +74,10 @@ The Datadog Profiler requires [Java Flight Recorder][1]. The Datadog Profiling l
 
 The Datadog Profiler requires Python 2.7+. Memory profiling only works on Python 3.5+. To begin profiling applications:
 
-1. Install `ddtrace` with the `profile` flavor, which contains both tracing and profiling:
+1. Install `ddtrace`:
 
     ```shell
-    pip install ddtrace[profiling]
+    pip install ddtrace
     ```
 
      **Note**: Profiling is available in the `ddtrace` library for versions 0.36+.


content/en/tracing/profiling/_index.md
```diff
@@ -74,10 +74,10 @@ The Datadog Profiler requires [Java Flight Recorder][1]. The Datadog Profiling l
...
 
-1. Install `ddtrace` with the `profile` flavor, which contains both tracing and profiling:
+1. Install `ddtrace`:
 
     ```shell
-    pip install ddtrace[profiling]
+    pip install ddtrace
     ```
 
      **Note**: Profiling is available in the `ddtrace` library for versions 0.36+.
```
GaurabAryal, line 83: Should we update this as well? Assuming that `install dd-trace` will only work from a newer version than 0.36,

dbenamydd, line 83: Maybe we should add a note along the lines of

> For versions 0.36-0.xx, `pip install ddtrace[profiling]`. After that it was moved into the main distribution.

with whatever terminology is best. Otherwise I can see some support cases where people were already using tracing 0.36, try to enable profiling, and it fails in a confusing way.

jd: `pip install ddtrace` will install 0.38 (or whatever is newer).

jd: Do you really think it's necessary to document older version of the library?

dbenamydd: I think Gaurab and I are basically asking the same thing :-)

dbenamydd: I'm not totally sure, just trying to avoid having people trip on this.

jd: Then I'm going to end up writing:
```
If you want to install an older version (< 0.38) of ddtrace, you need to use:

  pip install ddtrace[profiling]
```

Is that what we want? We're in general pretty eager for people to upgrade.

jd: I expect people reading this being first-timers, no reason they would trip on it.

dbenamydd: They'll be first timers to profiling, but some will be existing apm users who will have older versions of ddtrace installed. @hugokash @GaurabAryal in case you have thoughts. Non-blocking if anyone's really opposed to this.

jd: See https://github.com/DataDog/documentation/pull/7576/files#r432430330

apigirl:
hey @jd - we recently updated our S3 strategy, so in order to get the staging site to work, you'll need to merge in the latest master branch

GaurabAryal:
@jd  I already updated the docs if you want to see them :). Thank you for following through with this! https://docs.datadoghq.com/tracing/profiling/?tab=java

PR #7613, opened 2020-06-01
Encourage upgrading to latest agent 5

dbenamydd:
### What does this PR do?
Tweak wording and doc structure to better encourage upgrading agent 5 rather than removing the pem or upgrading to 6 or 7.

### Motivation
There are rare cases where removing the pem file won't work and 6 and 7 can be more involved upgrades so let's put the easiest and most robust solution first.

### Preview link
https://docs-staging.datadoghq.com/dbenamy-cert-faq-2/agent/faq/certificate_verify_failed-error/


dbenamydd:
Thanks!

PR #7621, opened 2020-06-01
update security info for default logs transport

nicholas-devlin:
https://docs.datadoghq.com/agent/logs/log_transport

https://github.com/DataDog/datadog-agent/blob/master/CHANGELOG.rst#7190--6190

Should I include any notes about how the agent will fall back on TCP if the http check fails, or if the user has configure the agent to us a SOCKS5 proxy server? I figured that would be handled by just linking to that Agent transport doc which has more details about when TCP is used. But am not sure if I should explicitly call it out

### DIFF
--- a/content/en/security/logs.md
+++ b/content/en/security/logs.md
@@ -17,9 +17,9 @@ The Log Management product supports multiple [environments and formats][2], allo
 
 ## Information Security
 
-The Datadog Agent submits logs to Datadog via HTTPS when possible. Logs are sent over port `443` with compression. Datadog uses symmetric encryption at rest (AES-256) for indexed logs. Indexed logs are deleted from the Datadog platform once their retention period, as defined by the customer, expires.
+The Datadog Agent submits logs to datadog either through HTTPS or through TLS-encrypted TCP connection on port 10516, requiring outbound communication (see [Agent Transport for logs][8]).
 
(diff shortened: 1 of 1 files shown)


content/en/security/logs.md
```diff
@@ -17,7 +17,9 @@ The Log Management product supports multiple [environments and formats][2], allo
 
 ## Information Security
 
-The Datadog Agent submits logs to Datadog over a TLS-encrypted TCP connection, requiring an outbound communication over port `10516`. The Datadog Agent can also be configured to send logs over secured HTTPS requests, requiring an outbound communication over port `443`. Datadog uses symmetric encryption at rest (AES-256) for indexed logs. Indexed logs are deleted from the Datadog platform once their retention period, as defined by the customer, expires.
+The Datadog Agent submits logs to Datadog via HTTPS when possible. Logs are sent over port `443` with compression. Datadog uses symmetric encryption at rest (AES-256) for indexed logs. Indexed logs are deleted from the Datadog platform once their retention period, as defined by the customer, expires.
```
PierreVanRoy, line 20: I would go for a shorter version which don't specify the behavior of old version. 
"The Datadog Agent submits logs to data through HTTPS or through TLS-encrypted TCP connection on port `10516`, requiring outbound communication (see [Agent Transport for logs][8]).

nicholas-devlin: changed the wording to match the suggestion of this comment

PR #7679, opened 2020-06-08
Add note for journald logs source

NBParis:
### What does this PR do?
Add details about setting the source for journald in container mode

### Motivation
Clarifying the behaviour

### Preview link

https://docs-staging.datadoghq.com/nils/journald-log-source/integrations/journald/#log-collection

### DIFF
--- a/content/en/integrations/journald.md
+++ b/content/en/integrations/journald.md
@@ -56,6 +56,7 @@ logs:
 ```
 
 **Note**: With Agent 7.17+ if `container_mode` is set to `true`, the `source` attribute of your logs is automatically set to the corresponding short image name of the container instead of simply `docker`.
+Indeed by default, the source and service are set to the unit name in journald which by default is `docker` for all logs coming from containers.
 
 Finally, [restart the agent][2].
 


content/en/integrations/journald.md
```diff
@@ -56,6 +56,7 @@ logs:
 ```
 
 **Note**: With Agent 7.17+ if `container_mode` is set to `true`, the `source` attribute of your logs is automatically set to the corresponding short image name of the container instead of simply `docker`.
+Indeed by default, the source and service are set to the unit name in journald which by default is `docker` for all logs coming from containers.
```
PierreVanRoy, line 59: Let's try to be more specific about how source and service are collected by default.
By default, the agent try to collect `SYSLOG_IDENTIFIER `, `_SYSTEMD_UNIT ` and `_COMM` and then set the source and service to the first non empty value. 
Since Agent 7.17+ this behavior is overwrite when `container_mode` is set to `true`. in that case, that case `short_image` tag overwrites the source.

PierreVanRoy:
Let's abandon this PR to push this one :D https://github.com/DataDog/documentation/pull/7683

PR #7723, opened 2020-06-11
Andrew/custom cpp

andrewardito:
### What does this PR do?
Update C++ Custom Instrumentation section with more content & context

### Preview link

https://docs-staging.datadoghq.com/andrew/CustomCPP/tracing/manual_instrumentation/cpp/

### DIFF
--- a/content/en/tracing/manual_instrumentation/cpp.md
+++ b/content/en/tracing/manual_instrumentation/cpp.md
@@ -15,6 +15,14 @@ further_reading:
 
 Add custom [span tags][1] to your [spans][2] to customize your observability within Datadog.  The span tags are applied to your incoming traces, allowing you to correlate observed behavior with code-level information such as merchant tier, checkout amount, or user ID.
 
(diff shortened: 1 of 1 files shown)


content/en/tracing/manual_instrumentation/cpp.md
```diff
@@ -1,23 +1,68 @@
...
+```cpp
+auto tracer = ...
+auto span = tracer->StartSpan("operation_name");
+span->SetTag("key must be string", "Values are variable types");
+span->SetTag("key must be string", 1234);
+```
+
+Values are of [variable type][6] and can be complex objects. Values are serialized as JSON, with the exception of a string value being serialized bare (without extra quotation marks).
+
+### Adding tags globally to all spans
```
cgilmour, line 40: This section should probably reference the `DD_TAGS` environment variable instead.

andrewardito: Thanks - updated!

PR #7760, opened 2020-06-17
Fix `check_names` annotation in JMX autodisovery doc

DocX:
### What does this PR do?
This PR changes it to `check_names` which should be correct according to the https://docs.datadoghq.com/agent/kubernetes/integrations/?tab=kubernetes

### Motivation
There is a typo in the annotation key with dot in `check.name` instead of `check_names`

### Preview link


DocX:
Let me know if I'm missing something and the `check.names` with dot is actually correcy? But it looks like it is typo and it does not work for us. We spent great deal of time debugging it :)

sarina-dd:
Thanks for the contribution, it appears you are correct. Merging these changes. Thanks again!

PR #7775, opened 2020-06-18
Describing injection of trace_id and span_id to json logs in PHP

obukhov:
### What does this PR do?
This PR adds a description of an alternative way to inject trace_id and span_id if the application uses json logs format.

### Motivation

We were looking for this way of connecting logs for our app, didn't find it in the logs, then just replicated the way node.js logs are connected

### Preview link

https://docs.datadoghq.com/tracing/connect_logs_and_traces/php/


apigirl:
thanks for this PR @obukhov! We are looking into this and we will get back to you soon

PR #7793, opened 2020-06-19
Ruby Custom Instrumentation

andrewardito:
### What does this PR do?
Adds content to existing Ruby Custom Instrumentation Page

### Motivation
Consolidating usecases for Custom Instrumentation

### Preview link
https://docs-staging.datadoghq.com/andrew.ardito/rubyinstrumentation/tracing/manual_instrumentation/ruby


content/en/tracing/manual_instrumentation/ruby.md
```diff
@@ -1,23 +1,130 @@
 ---
-title: Ruby Manual Instrumentation
+title: Ruby Custom Instrumentation
 kind: documentation
-decription: 'Manually instrument your Ruby application to send custom traces to Datadog.'
+aliases:
+    - /tracing/opentracing/ruby
+description: 'Implement the OpenTracing standard with the Datadog Java APM tracer.'
```
delner, line 6: I think you meant Ruby, not Java. Also this isn't OpenTracing?

content/en/tracing/manual_instrumentation/ruby.md
```diff
@@ -1,23 +1,130 @@
...
+{{% /tab %}}
+{{< /tabs >}}
+
+### Adding tags globally to all spans
+
+Add [tags][1] to all [spans][2] by configuring the tracer with the `tags` option:
+
+```ruby
+Datadog.configure do |c|
+  c.tracer tags: { 'env' => 'dev' }
```
delner, line 84: This works but is deprecated; keep in mind this will be changing in the future to:

```
Datadog.configure do |c|
  c.tags = { 'team' => 'qa' }
end
```

content/en/tracing/manual_instrumentation/ruby.md
```diff
@@ -1,23 +1,130 @@
...
+end
+```
+
+You can also use the `DD_TAGS` environment variable to set tags on all spans for an application. For more information on Ruby environment variables, refer to the [setup documentation][3].
+
+### Set errors on a span
+
+TODO
```
delner, line 92: TODO? Did you have a code snippet for this or did you need one from me?

content/en/tracing/manual_instrumentation/ruby.md
andrewardito: Ah yes I've been lax about changing these. Will update

content/en/tracing/manual_instrumentation/ruby.md
andrewardito: Need from you

content/en/tracing/manual_instrumentation/ruby.md
andrewardito: Can we change the snippet now to the snippet you provided? Don't want to be recommending a deprecated way of doing this.

content/en/tracing/manual_instrumentation/ruby.md
ericmustin: @andrewardito @delner something like this ok?

-----

### Setting Errors on a Span

There are two ways to set an error on a span

- The first is to simply call `span.set_error` and pass in the Exception Object.
- This will automatically extract the error type, message, and backtrace.

```ruby
require 'ddtrace'
require 'timeout'

def example_method
  span = Datadog.tracer.trace('example.trace')
  puts 'some work'
  sleep(1)
  raise StandardError.new "This is an exception"
rescue StandardError => error
  span = Datadog.tracer.active_span
  span.set_error(error) unless span.nil?
  raise error
ensure
  span.finish
end

example_method()
```

- The second is to use `tracer.trace` which will by default set the error type, message, and backtrace. 
- To configure this behavior you can use the `on_error` option, which is the Handler invoked when a block is provided to `trace`, and the block raises an error. 
- The Proc is provided `span` and `error` as arguments. 
- By default, `on_error` Sets error on the span.

Default Behavior: `on_error`

```ruby
require 'ddtrace'
require 'timeout'

def example_method
  puts 'some work'
  sleep(1)
  raise StandardError.new "This is a exception"
end

Datadog.tracer.trace('example.trace', on_error: custom_error_handler) do |span|
  example_method()
end
```

Custom Behavior: `on_error`

```ruby
require 'ddtrace'
require 'timeout'

def example_method
  puts 'some work'
  sleep(1)
  raise StandardError.new "This is a special exception"
end

custom_error_handler = proc do |span, error|
  span.set_tag('custom_tag', 'custom_value')
  span.set_error(error) unless error.message.include?("a special exception")
end

Datadog.tracer.trace('example.trace', on_error: custom_error_handler) do |span|
  example_method()
end  
```

andrewardito: 🙇 Thank you! @delner Can you give a 👍  and then we're good to merge.

PR #7846, opened 2020-06-25
Integrations HTTP check default

sarina-dd:
### What does this PR do?
Adding note that integrations that make HTTP(S) requests now default to proxy settings defined in `datadog.yaml` if none are specified at the integration level.

### Motivation
Jira request

### Preview link
https://docs-staging.datadoghq.com/sarina/agent-proxy/agent/proxy


content/en/agent/proxy.md
```diff
@@ -431,7 +432,7 @@ apm_config:
 
 process_config:
     process_dd_url: http://haproxy.example.com:3836
-    
+
```
sarina-dd, line 435:
```suggestion
```

PR #7896, opened 2020-07-01
enable process agent to see containers

lgh2:
If the process agent is not running, containers in the cluster will not show up in the Live Containers view.

### What does this PR do?

Add information that the process agent must be enabled for containers to show up on Live Containers view 

### Motivation

customer request 

### Preview link

### DIFF
--- a/content/en/agent/kubernetes/_index.md
+++ b/content/en/agent/kubernetes/_index.md
@@ -252,6 +252,8 @@ Optional collection Agents are disabled by default for security or performance r
 | `DD_PROCESS_AGENT_ENABLED` | Enable [live process collection][8] with the Process Agent. The [live container view][9] is already enabled by default if the Docker socket is available. If set to `false`, the [live process collection][8] and the [live container view][9] are disabled. |
 | `DD_COLLECT_KUBERNETES_EVENTS ` | Enable event collection with the Agent. If you are running multiple Agent in your cluster, set `DD_LEADER_ELECTION` to `true` as well. |
 
+To enable the Live Container view, make sure you are running the process agent in addition to setting DD_PROCESS_AGENT_ENABLED to `true`. 
+
 ### DogStatsD (custom metrics)
 
 Send custom metrics with [the StatsD protocol][10]:


content/en/agent/kubernetes/_index.md
```diff
@@ -252,6 +252,8 @@ Optional collection Agents are disabled by default for security or performance r
 | `DD_PROCESS_AGENT_ENABLED` | Enable [live process collection][8] with the Process Agent. The [live container view][9] is already enabled by default if the Docker socket is available. If set to `false`, the [live process collection][8] and the [live container view][9] are disabled. |
 | `DD_COLLECT_KUBERNETES_EVENTS ` | Enable event collection with the Agent. If you are running multiple Agent in your cluster, set `DD_LEADER_ELECTION` to `true` as well. |
 
+To enable the Live Container view, make sure you are running the process agent in addition to setting DD_PROCESS_AGENT_ENABLED to `true`.
```
chriszacny, line 255: You might want to say something related to the fact that the process agent runs as a separate container from the datadog agent and needs to be included as part of the deployment manifest. 

This is where I got hung up - I was assuming they ran together in the same binary. I was also using this manifest (the vanilla one) - https://docs.datadoghq.com/resources/yaml/datadog-agent-vanilla.yaml, and I was trying to make the containers show up in the live view using that... I assumed that if I just set the environment variable, it would work.

Maybe also mention: "To verify the process agent is running, `kubectl exec` into the datadog-agent pod and run `ps -ef` to confirm it is running, and if is not, to make sure the container specification is made in the YAML or Helm chart." or something to that effect.

Perhaps this isn't a big deal and most users would recognize this, but just figured I'd offer a bit of clarity (hopefully) as to where I got confused.

sarina-dd: +1 to these ideas, we should make a note of these instances in the doc.

sarina-dd: I've made note in a Jira issue for our team to add these additional notes to the docs. Going to approve and merge this change, and then we can create a separate PR for these additional changes. Thanks again for the feedback 👍

PR #7938, opened 2020-07-07
[agent] Give more specific Agent service commands

KSerrania:
### What does this PR do?

Give more precise Agent service commands for given platforms / versions (`service` does not work everywhere, and the `initctl` workaround for upstart doesn't work in all cases).
Update the general Agent commands page with a redirection to the main Agent page for Linux.

### Motivation

Prevent some cases where the commands wouldn't work because they're too generic.

### Preview link

[Amazon Linux](https://docs-staging.datadoghq.com/kylian.serrania/clearer-agent-commands/agent/basic_agent_usage/amazonlinux/?tab=agentv6v7)
[CentOS](https://docs-staging.datadoghq.com/kylian.serrania/clearer-agent-commands/agent/basic_agent_usage/centos/?tab=agentv6v7)
[Fedora](https://docs-staging.datadoghq.com/kylian.serrania/clearer-agent-commands/agent/basic_agent_usage/fedora/?tab=agentv6v7)
[Red Hat](https://docs-staging.datadoghq.com/kylian.serrania/clearer-agent-commands/agent/basic_agent_usage/redhat/?tab=agentv6v7)
[SUSE](https://docs-staging.datadoghq.com/kylian.serrania/clearer-agent-commands/agent/basic_agent_usage/suse/?tab=agentv6v7)
[Agent commands page](https://docs-staging.datadoghq.com/kylian.serrania/clearer-agent-commands/agent/guide/agent-commands/?tab=agentv6v7)

### Additional Notes

- For now, only Agent 6/7 commands have been updated.
- I'm currently testing that the commands do work on all supported systems.
  - Update: I tested the commands on CentOS 6, CentOS 7, Fedora 30, Amazon Linux 1 and Amazon Linux 2, and they work.
- I've updated the general Agent page with a redirection to the main Agent page (which has a list of platforms), because I couldn't find a better generic place that would send them to the service commands. There might be a better place to redirect to, and/or we maybe should split "Linux" into separate OSes.
- Should Ubuntu and Debian use the `systemctl` / `upstart`-specific commands, even though we know `service` is correctly supported there by default?

### DIFF
--- a/content/en/agent/basic_agent_usage/amazonlinux.md
+++ b/content/en/agent/basic_agent_usage/amazonlinux.md
@@ -71,7 +71,7 @@ In Agent v6 and v7, the service manager provided by the operating system is resp
 
 **Note**: If the `service` wrapper is not available on your system, use:
 
-* On `upstart`-based systems: `sudo initctl start/stop/restart/status datadog-agent`
+* On `upstart`-based systems: `sudo start/stop/restart/status datadog-agent`
 * On `systemd`-based systems: `sudo systemctl start/stop/restart/status datadog-agent`
 
 [Learn more about Service lifecycle commands][2]
--- a/content/en/agent/basic_agent_usage/amazonlinux.md
+++ b/content/en/agent/basic_agent_usage/amazonlinux.md
@@ -29,12 +29,27 @@ In Agent v6 and v7, the service manager provided by the operating system is resp
 {{< tabs >}}
 {{% tab "Agent v6 & v7" %}}
 
+### Amazon Linux 2
+
 | Description                        | Command                                                |
 |------------------------------------|--------------------------------------------------------|
-| Start Agent as a service           | `sudo service datadog-agent start`                     |
-| Stop Agent running as a service    | `sudo service datadog-agent stop`                      |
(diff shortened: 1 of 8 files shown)


albertvaka:
What should we do with the non-distro-specific https://docs.datadoghq.com/agent/guide/agent-commands/?tab=agentv6v7 ?

KSerrania:
> What should we do with the non-distro-specific https://docs.datadoghq.com/agent/guide/agent-commands/?tab=agentv6v7 ?

This probably needs to be discussed with the doc team; I'd either remove the tables, or redirect to the specific pages for each OS (what we already do for Windows and Docker).

KSerrania:
I went with the option of redirecting the user to the `/agent/` page in the non-distro-specific page. It's not great (that's two clicks to get to your desired platform's service commands), but it's more accurate, and I couldn't find a better place to redirect to. Another option would be to split the "Linux" row into one row per OS on this page, but that makes it redundant information, and will take a lot of space.

I also updated all reference to upstart commands to match what we use elsewhere (ie. not use `initctl start/stop` but `start/stop` directly).

PR #7939, opened 2020-07-07
updating the phrasing of the private locations access

apigirl:
### What does this PR do?
Clarifies that you should only request access if you don't already have it. 

### Motivation
PM request


MargotLepizzera:
LGTM - thanks Kaylyn!

PR #7941, opened 2020-07-07
adding documentation for host endpoint

az0977776:
### What does this PR do?
Adds documentation to `v1/hosts` for two new query parameters:
* `include_muted_hosts_data`
* `include_hosts_metadata`

### Motivation
To improve documentation

### Preview link
Affected page:
https://docs.datadoghq.com/api/v1/hosts/#get-all-hosts-for-your-organization

After changes:
https://docs-staging.datadoghq.com/aaron.wang/host-flags/api/v1/hosts/#get-all-hosts-for-your-organization

### DIFF
--- a/data/api/v1/full_spec.yaml
+++ b/data/api/v1/full_spec.yaml
@@ -10803,6 +10803,20 @@ paths:
         schema:
           format: int64
           type: integer
+      - description: Include whether or not hosts are muted and when they will be unmuted
+          your hosts.
+        in: query
+        name: include_muted_hosts_data
+        required: false
+        schema:
+          type: boolean
+      - description: Include host metadata
+          your hosts.
+        in: query
+        name: include_hosts_metadata
+        required: false
+        schema:
+          type: boolean
       responses:
         '200':
           content:


data/api/v1/full_spec.yaml
```diff
@@ -10803,6 +10803,20 @@ paths:
         schema:
           format: int64
           type: integer
+      - description: Include whether or not hosts are muted and when they will be unmuted
```
jamesbibby, line 10806: I would update to something like: "Include information about the muted status of hosts and when the mute expires"

data/api/v1/full_spec.yaml
```diff
@@ -10803,6 +10803,20 @@ paths:
...
+          your hosts.
+        in: query
+        name: include_muted_hosts_data
+        required: false
+        schema:
+          type: boolean
+      - description: Include host metadata
```
jamesbibby, line 10813: "Include additional metadata about the hosts (agent_version, processor, ...)". We should include a full list of the metadata fields that are available in the parenthesis.

PR #7980, opened 2020-07-13
Remove need for HIPAA specific log endpoints

PierreVanRoy:
### What does this PR do?
Now that all our log endpoint are HIPAA compliant, remove the the Hipaa specific endpoint from the requirement. 

### Preview link

https://docs-staging.datadoghq.com/pvr/hipaa-log-compliance/security/logs/

### DIFF
--- a/content/en/security/logs.md
+++ b/content/en/security/logs.md
@@ -46,7 +46,7 @@ If you have any questions about how the Log Management Service satisfies the app
 
 **Note:**
 
-HIPAA-enabled custommer used to require the use of specific endpoints to submit logs in order to enforce some specific encryptions. This is not needed anymore as that those encryptions are now enabled on all log submission endpoints.
+Previously, HIPAA-enabled customers needed to use specific endpoints to submit logs in order to enforce specific encryptions. This is no longer necessary. The encryptions are enabled on all log submission endpoints.
 
 These are the legacy endpoints that are not supported after version {edit suggestion to add version}:
 
--- a/content/en/security/logs.md
+++ b/content/en/security/logs.md
@@ -48,7 +48,7 @@ If you have any questions about how the Log Management Service satisfies the app
 
 Previously, HIPAA-enabled customers needed to use specific endpoints to submit logs in order to enforce specific encryptions. This is no longer necessary. The encryptions are enabled on all log submission endpoints.
 
-These are the legacy endpoints that are not supported after version {edit suggestion to add version}:
+These legacy endpoints are still supported:
 
 * `tcp-encrypted-intake.logs.datadoghq.com`
 * `lambda-tcp-encrypted-intake.logs.datadoghq.com`
--- a/content/en/security/logs.md
+++ b/content/en/security/logs.md
@@ -33,7 +33,7 @@ For customers using release 6, the Agent can be configured to obfuscate specific
 
 Datadog will sign a Business Associate Agreement (BAA) with customers that transmit protected health information (ePHI) via Datadog’s Log Management Service.
 
-Some features are not available at the moment to customers who have signed Datadog's BAA, notably:
+These features are not available to customers who have signed Datadog's BAA:
 
 * Users cannot request support via chat
 * Rehydration from Log Archives is disabled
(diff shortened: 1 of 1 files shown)


content/en/security/logs.md
```diff
@@ -62,6 +44,17 @@ Additionally, certain features are not available at the moment to customers who
 
 If you have any questions about how the Log Management Service satisfies the applicable requirements under HIPAA, please contact your account manager.
 
+**Note:**
+
+HIPAA-enabled custommer used to require the use of specific endpoints to submit logs in order to enforce some specific encryptions. This is not needed anymore as that those encryptions are now enabled on all log submission endpoints.
+
+Here are the historical endpoints which are now deprecated :
```
NBParis, line 51: Deprecated is not the right word here as we are still maintaining them. Maybe `legacy` would be better? @atspangler what do you think?

content/en/security/logs.md
```diff
@@ -62,6 +44,17 @@ Additionally, certain features are not available at the moment to customers who
...
+**Note:**
+
+HIPAA-enabled custommer used to require the use of specific endpoints to submit logs in order to enforce some specific encryptions. This is not needed anymore as that those encryptions are now enabled on all log submission endpoints.
+
+Here are the historical endpoints which are now deprecated : 
+
+* tcp-encrypted-intake.logs.datadoghq.com
+* lambda-tcp-encrypted-intake.logs.datadoghq.com
+* gcp-encrypted-intake.logs.datadoghq.com
+* http-encrypted-intake.logs.datadoghq.com
```
NBParis, line 56:
```suggestion
* `tcp-encrypted-intake.logs.datadoghq.com`
* `lambda-tcp-encrypted-intake.logs.datadoghq.com`
* `gcp-encrypted-intake.logs.datadoghq.com`
* `http-encrypted-intake.logs.datadoghq.com`
```

content/en/security/logs.md
```diff
@@ -62,6 +44,17 @@ Additionally, certain features are not available at the moment to customers who
 
 If you have any questions about how the Log Management Service satisfies the applicable requirements under HIPAA, please contact your account manager.
 
+**Note:**
+
+HIPAA-enabled custommer used to require the use of specific endpoints to submit logs in order to enforce some specific encryptions. This is not needed anymore as that those encryptions are now enabled on all log submission endpoints.
```
apigirl, line 49:
```suggestion
Before version {edit suggestion to add version}, HIPAA-enabled customers were required to use specific endpoints to submit logs. These endpoints enforced HIPPA required encryptions.  For version {edit suggestion to add version} +, these HIPAA required encryptions are enabled on all log submission endpoints.

File: content/en/security/logs.md (hunk 1 above)
Comment (apigirl) [inline content/en/security/logs.md:51]:
```suggestion
These are the legacy endpoints that are not supported after version {edit suggestion to add version}:
```

content/en/security/logs.md
NBParis: It is not a version change. More a date.

The reason we did not simply remove those endpoints from the page is because we don't want users to think those endpoints do not exist anymore and that their setup is incorrect and they need to change something.

Maybe we can find a better phrasing for this but the idea is to kind of get rid of those legacy endpoints.

ruthnaebeck, line 49:
```suggestion
Previously, HIPAA-enabled customers needed to use specific endpoints to submit logs in order to enforce specific encryptions. This is no longer necessary. The encryptions are enabled on all log submission endpoints.
```

content/en/security/logs.md
```diff
@@ -62,6 +44,17 @@ Additionally, certain features are not available at the moment to customers who
...
+
+These are the legacy endpoints that are not supported after version {edit suggestion to add version}:
```
ruthnaebeck, line 51: Version?

mstbbs: @NBParis Should we instead just phrase it as:

"These are legacy endpoints that may not be supported in the future."

NBParis: Yeah or we can even not say anything as I don't think we have a plan to deprecate them at some point. This might happen but we would likely do a DNS routing rather than removing them.

mstbbs: ok - let's say "These legacy endpoints are still supported:"

PR #8003, opened 2020-07-15
adds service as an attribute in sending a log example

le-he-hoo:
### What does this PR do?

Changes the request body example for sending a log to include `service` as an attribute.

### Motivation

Sending `service` as a tag does not result in the log having an official "service" set properly, but sending it as an attribute at the root level does.

### Preview link

https://docs-staging.datadoghq.com/tj/update_request_body_example_for_sending_a_log/api/v1/logs/#send-logs


le-he-hoo:
As per: https://github.com/DataDog/datadog-api-spec/wiki/Documentation

The PR needs to be here instead: https://github.com/DataDog/datadog-api-spec/blob/master/spec/v1/not-generated/logs_intake.yaml

PR #8011, opened 2020-07-15
Add a glog format example

case:
`glog` is the format that Kubernetes components use, but "glog" isn't discoverable in the Pipeline Library. These examples are from the Kube Scheduler item in the Pipeline Library.

### What does this PR do?

Adds an example to the Logs Parsing documentation, showing how to parse the `glog` format that Kubernetes components use for logging. 

### Motivation

Ideally this would be discoverable in the Pipeline Library for "glog", but it wasn't. Nor did it turn up on StackOverflow, etc.

### Preview link

### DIFF
--- a/content/en/logs/processing/parsing.md
+++ b/content/en/logs/processing/parsing.md
@@ -50,33 +50,6 @@ After processing, the following structured log is generated:
 
 {{< img src="logs/processing/processors/_parser.png" alt="Parsing example 1"  style="width:80%;">}}
 
-Another example, parsing the `glog` format that Kubernetes uses (via the Kube Scheduler item in the Pipeline Library):
-
-```text
-W0424 11:47:41.605188       1 authorization.go:47] Authorization is disabled
-```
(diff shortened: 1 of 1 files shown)


NBParis:
Hello @case 

Thanks for submitting this and helping us improving our documentation.

This looks great but it would be likely be better as a specific examples near the others ones (date, json, xml, ...) as you can see here: https://docs.datadoghq.com/logs/processing/parsing/?tab=matcher#examples

So it's just about adding a new `glolg` section there with the exact same content you provided.

Let me know your thoughts on this and happy to help if needed.

case:
Sure thing! I updated the text just now.

NBParis:
Thanks looks good to me.

I'll let @sarina-dd do the final review and taking care of merging.

Once again, thanks for submitting this PR.

apigirl:
Thanks for the update @case!

case:
🎉

PR #8024, opened 2020-07-17
Remove duplicate content from the troubleshooting section

GaurabAryal:
### What does this PR do?

Removes duplicate content from the troubleshooting section. :)

### Motivation

Existing doc has repeated paragraphs!


kayayarai:
Nice catch. Thank you!

PR #8030, opened 2020-07-18
update anchor to parsing-xml

bradfair:
### What does this PR do?
Fixes a broken link from the filter tab's xml info to a list of examples parsing xml.

### Motivation
I had to scroll, but wanted to click. ¯\_(ツ)_/¯ 

### Preview link
I'm external, so this isn't a thing I can do.

### Additional Notes

I also searched the rest of the content for links to parsing/#xml to see if there were other links using the invalid anchor, and found none. There were several references to #parsing-dates though, so the #parsing-xml one seems to fit that pattern.


cswatt:
great catch, thanks @bradfair !

PR #8049, opened 2020-07-21
Notifications only fire for automated tests

paulclarkin:
### What does this PR do?
Added a note that Synthetic Browser Tests only trigger notifications on automated runs, not via "Run Test Now"

### Motivation
I spent quite a bit of time wondering why the notifications weren't sending, re-reading this bit of the docs to see what I missed  and then opened a support ticket for help! They told me that it's working as it's supposed to!

### Preview link

### Additional Notes

Hope I've helped!

### DIFF
--- a/content/en/synthetics/browser_tests/_index.md
+++ b/content/en/synthetics/browser_tests/_index.md
@@ -62,7 +62,7 @@ You can customize alert conditions to define the circumstances under which you w
 
 ### Notify your team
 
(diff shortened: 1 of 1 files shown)


paulclarkin:
Support have been back in touch and suggested that their advice was wrong - Notifications should fire for Run Test Now.

It appears that it doesn't fire if paused, and then click "Run Now". 

I'll close this, and create a new ticket once support have understood if its a bug, or undocumented!

PR #7900, opened 2020-07-01
[anomaly detection] explain how anomaly monitors are configured via the API

StephenKappel:
### Preview link

https://docs-staging.datadoghq.com/<BRANCH_NAME>/<PATH>

Check preview base path using the URL in details in `preview` status check.

PR #8183, opened 2020-08-03
[forecast] Add an explanation of the format of a forecast query string

StephenKappel:
### What does this PR do?

Like #7900 , this PR adds more details about how to configure a data science monitor via the API. These monitors are more complicated than the "standard" monitors and when customers try to use tf or similar to manage their monitors in code, a lack of good documentation about how to configure these monitors leads to confusion.

### Preview link

https://docs-staging.datadoghq.com/StephenKappel-patch-1/monitors/monitor_types/anomaly/#api
https://docs-staging.datadoghq.com/StephenKappel-patch-1/monitors/monitor_types/forecasts/#api

### DIFF
--- a/content/en/monitors/monitor_types/forecasts.md
+++ b/content/en/monitors/monitor_types/forecasts.md
@@ -110,17 +110,17 @@ The `query` property in the request body should contain a query string in the fo
 <aggregator>(<query_window>):forecast(<metric_query>, '<algorithm>', <deviations>, interval=<interval>[, history='<history>'][, model='<model>'][, seasonality='<seasonality>']) <comparator> <threshold>
 ```
 
-* `aggregator`: `min` if the alert should trigger when the forecast goes below the threshold, and `max` if the alert should trigger when the forecast goes above the threshold 
-* `query_window`: a timeframe like `last_4h` or `last_7d`; the time window displayed in graphs in notifications; must be at least as large as the `alert_window` and is recommended to be around 5 times the `alert_window`
-* `metric_query`: a standard Datadog metric query, for example: `min:system.disk.free{service:database,device:/data}by{host}`
+* `aggregator`: Use `min` if the alert should trigger when the forecast goes below the threshold. Use `max` if the alert should trigger when the forecast goes above the threshold.
+* `query_window`: A timeframe, for example: `last_4h` or `last_7d`. The timeframe is recommended to be around five times the `alert_window`, but it must be at least as large as `alert_window`. This parameter controls the time range displayed in graphs included in notifications. 
+* `metric_query`: A standard Datadog metric query, for example: `min:system.disk.free{service:database,device:/data}by{host}`.
 * `algorithm`: `linear` or `seasonal`
-* `deviations`: a number greater than or equal to 1; controls the size of the confidence bounds, allowing a monitor to be made more or less sensitive
-* `interval`: a positive integer representing the number of seconds in the rollup interval
-* `history`: a string representing the amount of past data that should be used for making the forecast, for example: `1w`, `3d`. This parameter is only used with the `linear` algorithm.
+* `deviations`: A number greater than or equal to one. This parameter controls the size of the confidence bounds, allowing a monitor to be made more or less sensitive.
+* `interval`: A positive integer representing the number of seconds in the rollup interval.
+* `history`: A string representing the amount of past data that should be used for making the forecast, for example: `1w`, `3d`. This parameter is only used with the `linear` algorithm.
 * `model`: The type of model to use: `default`, `simple`, or `reactive`. This parameter is only used with the `linear` algorithm.
 * `seasonality`: The seasonality to use: `hourly`, `daily`, or `weekly`. This parameter is only used with the `seasonal` algorithm
 * `comparator`: Use `<=` to alert when the forecast goes below the threshold. Use `>=` to alert when the forecast goes above the threshold.
-* `threshold`: a number; a critical alert will trigger when the confidence bounds around the forecast reach this threshold
+* `threshold`: A critical alert will trigger when the forecast's confidence bounds reach this threshold.
 
 ## Troubleshooting
 
--- a/content/en/monitors/monitor_types/forecasts.md
+++ b/content/en/monitors/monitor_types/forecasts.md
@@ -112,14 +112,14 @@ The `query` property in the request body should contain a query string in the fo
 
 * `aggregator`: `min` if the alert should trigger when the forecast goes below the threshold, and `max` if the alert should trigger when the forecast goes above the threshold 
 * `query_window`: a timeframe like `last_4h` or `last_7d`; the time window displayed in graphs in notifications; must be at least as large as the `alert_window` and is recommended to be around 5 times the `alert_window`
-* `metric_query`: a standard Datadog metric query (e.g., `min:system.disk.free{service:database,device:/data}by{host}`)
+* `metric_query`: a standard Datadog metric query, for example: `min:system.disk.free{service:database,device:/data}by{host}`
 * `algorithm`: `linear` or `seasonal`
 * `deviations`: a number greater than or equal to 1; controls the size of the confidence bounds, allowing a monitor to be made more or less sensitive
 * `interval`: a positive integer representing the number of seconds in the rollup interval
-* `history`: a string representing the amount past data that should be used when making the forecast (e.g., `1w`, `3d`); this parameter is only for use with the `linear` algorithm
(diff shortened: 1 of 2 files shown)


content/en/monitors/monitor_types/forecasts.md
```diff
@@ -100,7 +100,27 @@ For detailed instructions on the **Say what's happening** and **Notify your team
...
+
+```text
+<aggregator>(<query_window>):forecast(<metric_query>, '<algorithm>', <deviations>, interval=<interval>[, history='<history>'][, model='<model>'][, seasonality='<seasonality>']) <comparator> <threshold>
+```
+
+* `aggregator`: `min` if the alert should trigger when the forecast goes below the threshold, and `max` if the alert should trigger when the forecast goes above the threshold 
+* `query_window`: a timeframe like `last_4h` or `last_7d`; the time window displayed in graphs in notifications; must be at least as large as the `alert_window` and is recommended to be around 5 times the `alert_window`
+* `metric_query`: a standard Datadog metric query (e.g., `min:system.disk.free{service:database,device:/data}by{host}`)
+* `algorithm`: `linear` or `seasonal`
+* `deviations`: a number greater than or equal to 1; controls the size of the confidence bounds, allowing a monitor to be made more or less sensitive
```
byassine93, line 117: should we keep the same definition for deviations for both anomalies and forecasts for consistency?

content/en/monitors/monitor_types/forecasts.md
```diff
@@ -100,7 +100,27 @@ For detailed instructions on the **Say what's happening** and **Notify your team
...
+* `interval`: a positive integer representing the number of seconds in the rollup interval
+* `history`: a string representing the amount past data that should be used when making the forecast (e.g., `1w`, `3d`); this parameter is only for use with the `linear` algorithm
+* `model`: `default`, `simple`, or `reactive`; this parameter is only for use with the `linear` algorithm
+* `seasonality`: `hourly`, `daily`, or `weekly`; this parameter is only for use with the `seasonal` algorithm
+* `comparator`: `<=` to alert when the forecast goes below the threshold; `>=` to alert when the forecast goes above the threshold
```
byassine93, line 122: It feels a bit redundant with the `aggregator` param. Can the user send a `min` aggregator with a `>=` (or do we block it), if so should we specify it ?

StephenKappel: You're right; they need to be synchronized. Good question about validation on the backend... I'll have to check if we actually enforce it, but if a monitor was created with these two things mismatched, it'd act very strangely...

content/en/monitors/monitor_types/forecasts.md
StephenKappel: The range of expected values is actually a little different. For anomalies, a bounds arg of 0.5 is fine, but for forecast it doesn't really make sense. I used the "confidence bounds" terminology here because this is how we describe it in the forecast monitor UI (which is different than the description in the anomalies UI).

content/en/monitors/monitor_types/forecasts.md
StephenKappel: It doesn't look like we explicitly block conflicting aggregators and comparators, and the monitor should evaluate, but it would give undesirable results.

content/en/monitors/monitor_types/forecasts.md
```diff
@@ -100,7 +100,27 @@ For detailed instructions on the **Say what's happening** and **Notify your team
...
+Forecast monitors are managed using the [same API][9] as other monitors, but the contents of the `query` property deserves further explanation.
+
+The `query` property in the request body should contain a query string in the following format:
+
+```text
+<aggregator>(<query_window>):forecast(<metric_query>, '<algorithm>', <deviations>, interval=<interval>[, history='<history>'][, model='<model>'][, seasonality='<seasonality>']) <comparator> <threshold>
+```
+
+* `aggregator`: `min` if the alert should trigger when the forecast goes below the threshold, and `max` if the alert should trigger when the forecast goes above the threshold 
+* `query_window`: a timeframe like `last_4h` or `last_7d`; the time window displayed in graphs in notifications; must be at least as large as the `alert_window` and is recommended to be around 5 times the `alert_window`
```
ruthnaebeck, line 114:
```suggestion
* `query_window`: a timeframe, for example: `last_4h` or `last_7d`. It is recommended that the time window displayed in graphs and notifications be five times the `alert_window`, but must be at least the size of the `alert_window`.
```

content/en/monitors/monitor_types/forecasts.md
```diff
@@ -100,7 +100,27 @@ For detailed instructions on the **Say what's happening** and **Notify your team
...
+* `metric_query`: a standard Datadog metric query (e.g., `min:system.disk.free{service:database,device:/data}by{host}`)
```
ruthnaebeck, line 115:
```suggestion
* `metric_query`: a standard Datadog metric query, for example: `min:system.disk.free{service:database,device:/data}by{host}`
```

content/en/monitors/monitor_types/forecasts.md
ruthnaebeck, line 117:
```suggestion
* `deviations`: a number greater than or equal to one, which controls the size of the confidence bounds, allowing a monitor to be made more or less sensitive.
```

content/en/monitors/monitor_types/forecasts.md
```diff
@@ -100,7 +100,27 @@ For detailed instructions on the **Say what's happening** and **Notify your team
...
+* `algorithm`: `linear` or `seasonal`
+* `deviations`: a number greater than or equal to 1; controls the size of the confidence bounds, allowing a monitor to be made more or less sensitive
+* `interval`: a positive integer representing the number of seconds in the rollup interval
+* `history`: a string representing the amount past data that should be used when making the forecast (e.g., `1w`, `3d`); this parameter is only for use with the `linear` algorithm
```
ruthnaebeck, line 119:
```suggestion
* `history`: a string representing the amount of past data that should be used for making the forecast, for example: `1w`, `3d`. This parameter is only used with the `linear` algorithm.
```

content/en/monitors/monitor_types/forecasts.md
```diff
@@ -100,7 +100,27 @@ For detailed instructions on the **Say what's happening** and **Notify your team
...
+* `model`: `default`, `simple`, or `reactive`; this parameter is only for use with the `linear` algorithm
```
ruthnaebeck, line 120:
```suggestion
* `model`: The type of model to use: `default`, `simple`, or `reactive`. This parameter is only used with the `linear` algorithm.
```

content/en/monitors/monitor_types/forecasts.md
```diff
@@ -100,7 +100,27 @@ For detailed instructions on the **Say what's happening** and **Notify your team
...
+* `seasonality`: `hourly`, `daily`, or `weekly`; this parameter is only for use with the `seasonal` algorithm
```
ruthnaebeck, line 121:
```suggestion
* `seasonality`: The seasonality to use: `hourly`, `daily`, or `weekly`. This parameter is only used with the `seasonal` algorithm
```

content/en/monitors/monitor_types/forecasts.md
ruthnaebeck, line 122:
```suggestion
* `comparator`: Use `<=` to alert when the forecast goes below the threshold. Use `>=` to alert when the forecast goes above the threshold.
```

content/en/monitors/monitor_types/forecasts.md
```diff
@@ -100,7 +100,27 @@ For detailed instructions on the **Say what's happening** and **Notify your team
...
+* `comparator`: `<=` to alert when the forecast goes below the threshold; `>=` to alert when the forecast goes above the threshold
+* `threshold`: a number; a critical alert will trigger when the confidence bounds around the forecast reach this threshold
```
ruthnaebeck, line 123:
```suggestion
* `threshold`: A number used to trigger a critical alert when the confidence bounds around the forecast reach the threshold.
```

content/en/monitors/monitor_types/forecasts.md
byassine93: Do you think we should create a card to prevent this behavior ?

StephenKappel: 👍 validation for this has been implemented and will be rolled out shortly.

StephenKappel:
@ruthnaebeck does this updated version look better?

StephenKappel:
Thanks! All set to merge.

PR #8208, opened 2020-08-06
Add Custom Link to see logs by index

NoisomePossum:
### What does this PR do?
Adds a Custom Link to the "logs by index" widgets in the estimated usage dashboard JSON definition.

### Motivation
The "view related logs" links for these two dashboards currently doesn't show any logs because it generates a query with the wrong parameter (it uses the tag `datadog_index` instead of inserting `index=<value>`). This is a quick workaround so customers have a useable link as a fallback.

### Preview link

N/A

### DIFF
--- a/static/resources/json/estimated_log_usage_dashboard_configuration.json
+++ b/static/resources/json/estimated_log_usage_dashboard_configuration.json
@@ -241,6 +241,7 @@
                         "q": "top(sum:datadog.estimated_usage.logs.ingested_events{datadog_is_excluded:false,$service,$index} by {datadog_index}.as_count(), 50, 'sum', 'desc')"
                     }
                 ],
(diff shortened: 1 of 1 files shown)


static/resources/json/estimated_log_usage_dashboard_configuration.json
```diff
@@ -241,6 +241,7 @@
                         "q": "top(sum:datadog.estimated_usage.logs.ingested_events{datadog_is_excluded:false,$service,$index} by {datadog_index}.as_count(), 50, 'sum', 'desc')"
                     }
                 ],
+                "custom_links":[{"link":"https://dd.datad0g.com/logs?index={{datadog_index.value}}","label":"See Logs in Explorer"}],
```
kayayarai, line 244: Hi @NoisomePossum Should these be `app.datadoghq.com` links instead of `dd.datadog.com`?

PR #7732, opened 2020-06-12
DOCS-853 - container env vars

cswatt:
### What does this PR do?
adds new agent 7.2 container include/exclude env vars

### Preview link

https://docs-staging.datadoghq.com/cswatt/container-env-vars/infrastructure/livecontainers
https://docs-staging.datadoghq.com/cswatt/container-env-vars/agent/kubernetes/#ignore-containers
https://docs-staging.datadoghq.com/cswatt/container-env-vars/agent/guide/environment-variables
https://docs-staging.datadoghq.com/cswatt/container-env-vars/agent/guide/autodiscovery-management
https://docs-staging.datadoghq.com/cswatt/container-env-vars/agent/docker

Check preview base path using the URL in details in `preview` status check.

PR #8224, opened 2020-08-07
Adding the version of the Agent

FatineBts:
Adding some useful info about the Environment Variables useable depending on the version of the agent.

### What does this PR do?
Adds some useful info. 

### Motivation
Being more explicit and thus helping customers.

### DIFF
--- a/content/en/agent/guide/autodiscovery-management.md
+++ b/content/en/agent/guide/autodiscovery-management.md
@@ -25,13 +25,13 @@ Exclude containers from the Agent Autodiscovery perimeter with an exclude rule b
 {{< tabs >}}
 {{% tab "Containerized Agent" %}}
 
-To remove a given Docker container with the image `<IMAGE_NAME>` from Autodiscovery, add the following environment variable to the Datadog Agent:
+In **Agent v7.20+**, to remove a given Docker container with the image `<IMAGE_NAME>` from Autodiscovery, and thus exclude the **logs and metrics**, add the following environment variable to the Datadog Agent:
 
 ```shell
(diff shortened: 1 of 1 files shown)


content/en/agent/guide/autodiscovery-management.md
```diff
@@ -25,13 +25,13 @@ Exclude containers from the Agent Autodiscovery perimeter with an exclude rule b
 {{< tabs >}}
 {{% tab "Containerized Agent" %}}
 
-To remove a given Docker container with the image `<IMAGE_NAME>` from Autodiscovery, add the following environment variable to the Datadog Agent:
+In **Agent v7.20+**, to remove a given Docker container with the image `<IMAGE_NAME>` from Autodiscovery, and thus exclude the **logs and metrics**, add the following environment variable to the Datadog Agent:
```
ahmed-mez, line 28: should we add a note to say that `AC_INCLUDE` and `AC_EXCLUDE` can be used for prior agent versions? I believe there is a note about its deprecation we can add it there.
Let's see what the docs team say about that as well

FatineBts: Yes I agree, let's wait for their reply! Thanks Ahmed!

cswatt: @ahmed-mez yes! let's do that :)

ahmed-mez: here is the original PR https://github.com/DataDog/documentation/pull/7732/files
seems like there are other places we should update

PR #8292, opened 2020-08-13
Initialize dd-trace to enable tracing

hghotra:
### What does this PR do?

`dd-trace` needs to be initialized to enable tracing. 

### Preview link

https://docs-staging.datadoghq.com/hghotra-patch-1/serverless/installation/nodejs/?tab=serverlessframework

### DIFF
--- a/content/en/serverless/installation/nodejs.md
+++ b/content/en/serverless/installation/nodejs.md
@@ -250,7 +250,7 @@ If you would like to submit a custom metric or manually instrument a function, s
 
 ```javascript
 const { sendDistributionMetric } = require("datadog-lambda-js");
-const tracer = require("dd-trace");
+const tracer = require("dd-trace").init();
 
 // Submit a custom APM span named "sleep"
 const sleep = tracer.wrap("sleep", (ms) => {


content/en/serverless/installation/nodejs.md
```diff
@@ -250,7 +250,7 @@ If you would like to submit a custom metric or manually instrument a function, s
 
 ```javascript
 const { sendDistributionMetric } = require("datadog-lambda-js");
-const tracer = require("dd-trace");
+const tracer = require("dd-trace").init();
```
tianchu, line 253: This is unnecessary, when `DD_TRACE_ENABLED` set to `true`, we initialize the tracer on users' behalf https://github.com/DataDog/datadog-lambda-js/blob/3b0f2a8b7c2736d30a07b557caae536b2b138d8f/src/handler.ts#L8

hghotra: No you're right, if `DD_TRACE_ENABLED` is set to `true`, this is not required - I missed step 4 here: https://docs.datadoghq.com/serverless/installation/nodejs/?tab=custom#configure-the-function

PR #8303, opened 2020-08-15
Update dogstatsd.md with date_happened 

piyushjain18:
When calling the statsd.event we cannot just pass the timestamp or timestamp=<event_time> we need to pass `date_happened=<event_time>`
We need to give the clear example on how to use the timestamp parameter. Because this is confusing so better to provide example as below for clarity  `dogStatsdService.Event("An error occurred", "Error message", alertType: "error", date_happened='TIMESTAMP', tags: new[] { "env:dev" });`

### Preview link


ogaca-dd (approved):
LGTM

PR #8378, opened 2020-08-27
split js into exports, lint files

zbayoff:
### What does this PR do?
1. makes js functions exportable for future use with code lang selection. Easier to read and manage.
2. eslint files. 
3. remove old `codenav` files and deps leftover from old API page.

### Motivation
cleanup 

### Preview link
https://docs-staging.datadoghq.com/zach/js-exportable

Test async functionality, gTag is populating correct value.


nsollecito:
Separate from this, I noticed the Async nav isn't setup to work on the API pages

zbayoff:
> Separate from this, I noticed the Async nav isn't setup to work on the API pages

Yeah, it was never ported over after the new API design. It has a different layout, so the async loading would have to be adjusted for it.

PR #8438, opened 2020-09-01
Admission Controller manual setup (without Helm)

FOSSforlife:
### What does this PR do?
Adds instructions for orchestrating the Cluster Agent Admission Controller without a Helm chart, just like on some of the other Kubernetes pages.

I copied configs from these files, which is the same thing I did at my company do get it working:

- https://github.com/helm/charts/blob/d6da827531/stable/datadog/templates/agent-rbac.yaml
- https://github.com/helm/charts/blob/d6da827531/stable/datadog/templates/agent-services.yaml 
- https://github.com/helm/charts/blob/d6da827531/stable/datadog/templates/cluster-agent-deployment.yaml 

In addition, I've fixed a couple small bugs/typos:
- The `agent-rbac.yaml` link was broken and pointed to `rbac.yaml`
- The `datadog-agent-security` config map file was named `system-probe-configmap.yaml` instead of `security-configmap.yaml`

**One additional thing:** I didn't know where to add this in, but somewhere during the Cluster Agent setup (before applying the `deployment` file), the user needs to run `kubectl apply -f https://raw.githubusercontent.com/DataDog/datadog-agent/master/Dockerfiles/manifests/agent-only/install_info-configmap.yaml`. If someone could point me to a good spot to add this in, I will gladly do so.

### Motivation
I spent a few days scratching my head at how to get this working at my company, since we're not using Helm. APM and the automatic setup of new pods using the Admission Controller were must-have features for us.

### Preview link (images)

Screenshot from 2020-09-01 16-40-55
Screenshot from 2020-09-01 16-41-05
Screenshot from 2020-09-01 16-41-12

### Additional Notes

I want to make sure my comment above about the additional `kubectl apply` that needs to be added isn't missed. 

I hope this pull request is helpful. 🙂 Please let me know if there is anything else I need to do.

### DIFF
--- a/content/en/agent/cluster_agent/admission_controller.md
+++ b/content/en/agent/cluster_agent/admission_controller.md
@@ -63,12 +63,70 @@ To enable the admission controller for the Datadog operator, set the parameter `
 [...]
 ```
 
+### Manual setup
+
+To enable the admission controller without using Helm or the Datadog operator, you'll need to add a few things to your configuration:
+
+1. Download the [Cluster Agent RBAC permissions][3] manifest, and add the following under `rules`:
+
+{{< code-block lang="yaml" filename="cluster-agent-rbac.yaml" disable_copy="true" >}}
+- apiGroups:
+  - admissionregistration.k8s.io
+  resources:
+  - mutatingwebhookconfigurations
+  verbs: ["get", "list", "watch", "update", "create"]
+- apiGroups: [""]
+  resources: ["secrets"]
+  verbs: ["get", "list", "watch", "update", "create"]
+- apiGroups: ["batch"]
+  resources: ["jobs", "cronjobs"]
+  verbs: ["get"]
+- apiGroups: ["apps"]
+  resources: ["statefulsets", "replicasets", "deployments"]
+  verbs: ["get"]
+{{< /code-block >}}
+
+2. Add the following to the bottom of `agent-services.yaml`:
+
+{{< code-block lang="yaml" filename="agent-services.yaml" disable_copy="true" >}}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: datadog-cluster-agent-admission-controller
+  labels:
+    app: "datadog"
+    app.kubernetes.io/name: "datadog"
+spec:
+  selector:
+    app: datadog-cluster-agent
+  ports:
+  - port: 443
+    targetPort: 8000
+
+{{< /code-block >}}
+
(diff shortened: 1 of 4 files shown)


content/en/agent/cluster_agent/admission_controller.md
```diff
@@ -63,12 +63,72 @@ To enable the admission controller for the Datadog operator, set the parameter `
...
+
+{{< /code-block >}}
+
+Add environment variables to the Cluster Agent deployment which enable the Admission Controller:
+
+{{< code-block lang="yaml" filename="cluster-agent-deployment.yaml" disable_copy="true" >}}
+- name: DD_ADMISSION_CONTROLLER_ENABLED
+  value: "true"
+- name: DD_ADMISSION_CONTROLLER_MUTATE_UNLABELLED
+  value: "false"
```
ahmed-mez, line 115: this is a an optional config, and it's `false` by default - should we add it as a note?

content/en/agent/cluster_agent/admission_controller.md
```diff
@@ -63,12 +63,72 @@ To enable the admission controller for the Datadog operator, set the parameter `
 [...]
 ```
 
+### Manual setup
+
+To enable the admission controller without using Helm or the Datadog operator, you'll need to add a few things to your configuration:
+
+First, Download the [Cluster Agent RBAC permissions][3] manifest, and add the following under `rules`:
```
cswatt, line 70:
```suggestion
First, download the [Cluster Agent RBAC permissions][3] manifest, and add the following under `rules`:
```

cswatt (changes requested):
just one styleguide comment. otherwise looks good! thanks!

content/en/agent/cluster_agent/admission_controller.md
FOSSforlife: Ok, how does this look:

```yaml
- name: DD_ADMISSION_CONTROLLER_ENABLED
  value: "true"
- name: DD_ADMISSION_CONTROLLER_SERVICE_NAME
  value: "datadog-cluster-agent-admission-controller"

# Uncomment this to configure APM tracers automatically (see below)
# - name: DD_ADMISSION_CONTROLLER_MUTATE_UNLABELLED
#   value: "true"
```

ahmed-mez: looks good thanks - to be validated by docs cc @cswatt

cswatt (approved):

PR #8443, opened 2020-09-02
Zach/gitlab cache

zbayoff:
### What does this PR do?
use Gitlab cache to install node_modules instead of installing in the Docker image (already implemented on corpsite)

Corp-ci PR: https://github.com/DataDog/corp-ci/pull/93

### Motivation
remove need to update two package.json files, easier to manage dependencies.

### Preview link
http://docs-staging.datadoghq.com/zach/gitlab-cache


.gitlab-ci.yml
```diff
@@ -284,6 +327,7 @@ evaluate_missing_metrics:
 manage_translations:on-schedule:
   <<: *base_template
   stage: post-deploy
+  cache: {}
```
davidejones, line 330: i think the cache might be needed to run translate node package

package copy.json
```diff
@@ -0,0 +1,119 @@
...
+    "terser-webpack-plugin": "^2.2.3",
+    "webpack-assets-manifest": "^3.1.1",
+    "webpack-bundle-analyzer": "^3.3.2",
+    "webpack-cli": "^3.3.6",
+    "webpack-merge": "^4.2.1"
+  },
+  "resolutions": {
+    "browserslist": "4.11.1"
+  }
+}
```
davidejones, line 119: do we need this?

.gitlab-ci.yml
```diff
@@ -10,27 +11,49 @@ variables:
   LIVE_DOMAIN: "https://docs.datadoghq.com/"
   PREVIEW_DOMAIN: "https://docs-staging.datadoghq.com/"
   FORCE_COLOR: '1'
+  GIT_DEPTH: 50
```
davidejones, line 14: might need to be careful using git_depth on all jobs and check them. it tends to cause finding tags or comparing file changes with git commands to not always work.

package.json
```diff
@@ -1,8 +1,8 @@
 {
     "name": "docs",
     "version": "1.0.0",
-    "description": "Corpweb built on HUGO",
-    "private": true,
+    "description": "Datadog Docs built on HUGO",
+    "private": false,
```
davidejones, line 5: Might want to keep this as true to avoid accidently publishing to the npm public repo. When its private you need to have a user login

package copy.json
zbayoff: whoops. no.

davidejones (approved):

sarina-dd (approved):

PR #8518, opened 2020-09-11
Fix small typos

serjaum:
### What does this PR do?

Just fixing some mistakes in the docs.

1. The `metric` spec doesn't exists. it is `metrics`

```error: error validating "hpa.yaml": error validating data: ValidationError(HorizontalPodAutoscaler.spec): unknown field "metric" in io.k8s.api.autoscaling.v2beta1.HorizontalPodAutoscalerSpec; if you choose to ignore these errors, turn validation off with --validate=false```

2. The number of maximun replicas in the example is wrong

### Motivation

I'm trying to enable the external metrics on my hpa and when reading the docs, I found the errors

### DIFF
--- a/content/en/agent/cluster_agent/external_metrics.md
+++ b/content/en/agent/cluster_agent/external_metrics.md
@@ -77,7 +77,7 @@ Once you have the Datadog Cluster Agent running and the service registered, crea
 
 ```yaml
 spec:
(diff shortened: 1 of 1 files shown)


kayayarai (approved):
Great fix! Thanks for the PR.

PR #8580, opened 2020-09-17
Fix error logging in Makefile

adw1n:
Try running `make help` while there is no `Make.config` file. Result:

```
Makefile:17: *** recipe commences before first target.  Stop.
```

With my changes:

```
Makefile:17: *** Makefile.config not found. See Makefile.config.example..  Stop
```

Other possible fix is to not use indentation or to replace the tab with a single space / multiple spaces but we risk that this change gets reverted by someone's editor.


davidejones (approved):

cswatt (approved):

cswatt:
Thanks, @adw1n !

PR #8674, opened 2020-09-29
Minor fixes for recent additions to pull config

davidejones:
### What does this PR do?

This PR fixes the issue with

```
Error: Error building site: "/go/src/github.com/DataDog/documentation/content/en/serverless/serverless_integrations/cli.md:1:1": plain HTML documents not supported
```

Currently we are pulling a `.md` file and the config expects to replace a h1 tag and inject frontmatter.
In some situations the `.md` we are pulling doesn't have a h1 so we never inject the front matter and the build seems to fail on not allowing html in md files.

The fix updates the intentions of the script so we replace h1 only if it exists and we update frontmatter if we have it.

### Motivation

Seeing build failures

### Preview

Need to test integrations pages look good as well as other sections pulling content
https://docs-staging.datadoghq.com/david.jones/fixfm/

### Additional Notes

---

### Reviewer checklist

### DIFF
--- a/local/bin/py/build/actions/pull_and_push_file.py
+++ b/local/bin/py/build/actions/pull_and_push_file.py
@@ -5,6 +5,13 @@ import yaml
 
 from os.path import basename
 
+TEMPLATE = """\
+---
+{front_matter}
+---
+{content}
+"""
+
 
 def pull_and_push_file(content, content_dir):
     """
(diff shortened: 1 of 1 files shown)


zbayoff (approved):

apigirl (approved):
looks great! Including the one I pulled in that caused this issue to begin with 😬🤦‍♀️

PR #8717, opened 2020-10-02
Replace - with NOT operator for Search String exclusion

nirmaljpatel:
the "-" Operator would not work for me in the search strings. "NOT" works for exclusion.

### What does this PR do?
Document use of "NOT" operator for search string exclusion

### Motivation
Update documentation

---

### Reviewer checklist


le-he-hoo:
Hey @nirmaljpatel - Thanks for making a PR.  
In this case, "-" is the expected way to exclude in the log search syntax.  If you're experiencing different behavior, can you reach out to support@datadoghq.com so that the team can take a closer look at the situation?

PR #8750, opened 2020-10-07
use partialCached

zbayoff:
### What does this PR do?
Use hugo's [partialCached](https://gohugo.io/functions/partialcached/) to speed up build time. 
To know when to use `partialCached`, I used `--templateMetrics` and `--templateMetricsHints` while developing locally, which logs the partials that can benefit from the `partialCached` method. Specifcally, for the partials I've modified,  they logged `cache potential` at `100`, which means they can use `partialCached`. 
 
### Motivation
speed up hugo build

### Preview
http://docs-staging.datadoghq.com/zach/speed-hugo-build


nsollecito (approved):

ruthnaebeck (approved):
👍  for docs

PR #8775, opened 2020-10-09
[WEB-840] - update frontmatter formatting on integrations

davidejones:
### What does this PR do?

When building integration markdown files we opted to use `default_style='"'` to wrap entries with quotes in order to avoid issues with localized text breaking yaml format. A consequence of doing this caused boolean values to be output as `!!bool "false"` or `!!bool "true"`. See related PR #6059. While this works fine Hugo doesn't respect the values in this format and we now want to make use of some of these booleans.

This PR:

- Will cause integrations defailed pages (`/integrations/some-integration/`) to not build if `is_public: false` 
- Will cause generated integration frontmatter booleans to change formatting from `!!bool "false"` to `false` etc. 

### Motivation
https://datadoghq.atlassian.net/browse/WEB-840

### Preview
https://docs-staging.datadoghq.com/david.jones/draftintegrations/integrations/

### Additional Notes

---

### Reviewer checklist

### DIFF
--- a/config/preview/config.yaml
+++ b/config/preview/config.yaml
@@ -1,6 +1,6 @@
 baseURL: https://docs-staging.datadoghq.com/$CI_COMMIT_REF_NAME/
 buildfuture: true
-builddrafts: true
+builddrafts: false
 enableGitInfo: true
 
 deployment:
--- a/config/preview/config.yaml
+++ b/config/preview/config.yaml
@@ -1,6 +1,6 @@
 baseURL: https://docs-staging.datadoghq.com/$CI_COMMIT_REF_NAME/
 buildfuture: true
-builddrafts: false
+builddrafts: true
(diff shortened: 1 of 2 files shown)


config/preview/config.yaml
```diff
@@ -1,6 +1,6 @@
 baseURL: https://docs-staging.datadoghq.com/$CI_COMMIT_REF_NAME/
 buildfuture: true
-builddrafts: true
+builddrafts: false
```
davidejones, line 3:
```suggestion
builddrafts: true
```

Apply this suggestion before merging. I temporarily disabled builddrafts on preview so we could test the page is respecting the `draft: true` frontmatter. We want it back the way it was on release.

nsollecito (approved):

sarina-dd (approved):

PR #8786, opened 2020-10-13
Update example to better reflect the mathematical differences between as-count and as-rate in monitor

john-bodley:
### What does this PR do?

I found the [as-count-in-monitor-evaluations](https://docs.datadoghq.com/monitors/guide/as-count-in-monitor-evaluations) document to be helpful but per the recommendations, 

> In general, avg time aggregation with .as_rate() is reasonable, but sum aggregation with .as_count() is recommended for error rates. Aggregation methods other than sum (shown as in total in-app) do not make sense to use with .as_count().

It seems like in the example the `classic_eval_path` should be using the `avg(last_5m)` time aggregation as opposed to `sum(last_5m)` which also provides a better apples-to-apples comparison with the `as_count_eval_path` metric. 

Further I tweaked the numbers (and added a comment) to better illustrate how these calculations differ, i.e., the `classic_eval_path` weights each interval (timestamp) equally whereas the `as_count_eval_path` weights each interval by the number of requests in said interval, i.e., 

```
(1 + 1 + 0 + 1 + 0) / (1 + 2 + 3 + 4 + 5) = (1/1 * 1/15 + 1/2 * 2/15 + 0 * 3/15 + 1/4 * 4/15 + 0 * 5/15)
```

### Motivation

I felt reading through the documentation a slightly different formulation of the problem would aid readers with understanding the nuances between these approaches.

### Preview

### Additional Notes

---

### Reviewer checklist

### DIFF
--- a/content/en/monitors/guide/as-count-in-monitor-evaluations.md
+++ b/content/en/monitors/guide/as-count-in-monitor-evaluations.md
@@ -58,7 +58,7 @@ Compare the result of the evaluation depending on the path:
 | **`classic_eval_path`**  | Aggregation function applied _after_ division  | **(1/1 + 1/2 + ... + 0/5)/5**           | **0.35** |
 | **`as_count_eval_path`** | Aggregation function applied _before_ division | **(1 + 1 + ... + 0)/(1 + 2 + ... + 5)** | **0.20** |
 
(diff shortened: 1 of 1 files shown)


apigirl:
hey @john-bodley - thanks for this! we've decided not to accept this change at this time because of other updates that are in our pipeline currently that will impact this. We really appreciate you taking the time to make this PR though!

PR #8890, opened 2020-10-22
content/en/tracing/setup/go.md: remove unnecessary instruction

knusbaum:
### What does this PR do?
This removes an unnecessary step in using the Go Tracer.
The `go build` command will resolve dependencies anyway, so we do not need to tell users to download it manually with `go get`. 

Telling users to do this can even cause confusion and inconvenience, since the command listed *also* downloads all of the *potential* dependencies for dd-trace-go (the libraries we provide integrations for), including multiple web frameworks, kafka libraries, grpc, etc. Some of these throw errors since specific versions need to be selected, which the command does not do.

The user can just put the import in the code and the toolchain will resolve it correctly.

### Motivation
Scanning the go documentation for consistency.

### Preview

https://docs-staging.datadoghq.com/knusbaum/update-go-doc/tracing/setup/go/#installation

### Reviewer checklist

### DIFF
--- a/content/en/tracing/setup/go.md
+++ b/content/en/tracing/setup/go.md
@@ -44,12 +44,6 @@ Follow the [Quickstart instructions][6] within the Datadog app for the best expe
 
 Otherwise, [install and configure the Datadog Agent][7]. See the additional documentation for [tracing Docker applications][8] or [Kubernetes applications][9].
 
-Next, install the Go tracer from its canonical import path:
-
-```go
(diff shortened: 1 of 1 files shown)


andrewardito (approved):
As discussed, looks good to me!

gbbr (approved):
The nice side effect of this is that there are no more installation "errors" displayed now.

apigirl (approved):

PR #8987, opened 2020-11-04
Kubernetes resources for live containers

sarina-dd:
### What does this PR do?
Adds a section to the Kubernetes docs focusing on live Kubernetes resources for live containers

### Motivation
Jira request

### Preview
https://docs-staging.datadoghq.com/sarina/k8s-resource-collections/agent/kubernetes/

---

### Reviewer checklist


content/en/agent/kubernetes/_index.md
```diff
@@ -255,6 +254,100 @@ For further details on setting up Operator, including information about using to
       fsGroup: <DOCKER_GROUP_ID>
 ```
 
+## Kubernetes resources for Live Containers
+
+The Agent and [Cluster Agent][3] can be configured to retrieve Kubernetes resources for [Live Containers][4]. This feature allows you to monitor the state of pods or deployments in a specific namespace or availability zone, view resource specifications for failed pods within a deployment, correlate node activity with related logs, and more.
```
cohenyair, line 259: "pods or deployments" --> "pods, deployments and other Kubernetes concepts"

cohenyair (approved):

content/en/agent/kubernetes/_index.md
```diff
@@ -255,6 +254,100 @@ For further details on setting up Operator, including information about using to
...
+    valueFrom:
+    configMapKeyRef:
+      name: datadog-cluster-id
+      key: id
+  {{< /code-block >}}
+
+  It must also be configured to communicate with the Cluster Agent, see the [Cluster Agent Setup documentation][5] for more information about this.
+
+- Cluster Agent version >= 1.9.0
+- Cluster Agent is running, and the Agent must be able to communicate with it
```
hkaj, line 279: All this block should be part of the DaemonSet tab, as the helm chart will take care of all of this automatically if it's used.

content/en/agent/kubernetes/_index.md
```diff
@@ -255,6 +254,100 @@ For further details on setting up Operator, including information about using to
...
+  - get
+  - watch
+{{< /code-block >}}
+
+These permissions are needed in order to create a datadog-cluster-id ConfigMap in the same Namespace as the Agent DaemonSet and the Cluster Agent Deployment, as well as to collect Deployments and ReplicaSets.
+
+If the cluster-id ConfigMap doesn't get created by the Cluster Agent, the Agent pod will not start, and fall in CreateContainerConfigError status. If the Agent pod is stuck because this ConfigMap doesn't exist, update the Cluster Agent permissions and restart its pods to let it create the ConfigMap and the Agent pod will recover automatically.
+
+{{% /tab %}}
+{{% tab "Operator" %}}
```
hkaj, line 344: We probably won't have this ready by the time we're going GA, so we should remove it for now and release without it imho. We can add it when the change in the operator is merged and the next operator version is released.

content/en/agent/kubernetes/_index.md
```diff
@@ -255,6 +254,100 @@ For further details on setting up Operator, including information about using to
...
+
+
+[1]: https://github.com/DataDog/helm-charts
+[2]: https://github.com/DataDog/helm-charts/blob/master/charts/datadog/values.yaml
+{{% /tab %}}
+{{% tab "DaemonSet" %}}
+
+Update the Agent DaemonSet, the Cluster Agent Deployment, and the Cluster Agent ClusterRole with the following settings:
+
+## Cluster Agent Settings
```
hkaj, line 299: It may look like a nitpick, but can we move the cluster agent settings above the agent settings (after we bring the agent settings in this DaemonSet tab) please? I'm worried that some users do the agent change and try to deploy without reading the cluster agent changes, and this will block their process-agent from starting, which i'd rather avoid.

hkaj (changes requested):
this looks great! Left a couple of comments mostly about the order in which we should present things.
One more point: can we make this part https://docs.datadoghq.com/infrastructure/livecontainers/#kubernetes-resources-1 point to these new instructions directly instead of giving some instructions specific to helm? Now that we have this page that includes all install methods it makes more sense i think.

sarina-dd:
@hkaj @cohenyair Thanks for the great feedback! @hkaj after your link feedback I realized it probably makes more sense to shift this information onto the lives containers page and to update the Kubernetes landing page with a link back to the live containers page setup instructions. This will allows users to easily configure resources and read through how to engage with them in-app, all in one place. Let me know your thoughts.

content/en/infrastructure/livecontainers.md
```diff
@@ -13,73 +13,100 @@ further_reading:
...
-   clusterName: <PLACEHOLDER>
-   ...
-{{< /code-block >}}
-
-**Note**: The cluster name must be 80-characters or less.
-
-On Google's GKE, AWS EKS, and Azure AKS, this is unnecessary, unless the Agent and the cluster Agent don't have access to the cloud metadata APIs, or the cluster name is longer than 80 characters.
+The Agent and [Cluster Agent][5] can be configured to retrieve Kubernetes resources for [Live Containers][6]. This feature allows you to monitor the state of pods, deployments and other Kubernetes concepts in a specific namespace or availability zone, view resource specifications for failed pods within a deployment, correlate node activity with related logs, and more.
+
+Kubernetes resources for Live Containers requires [Agent version >= 7.21.1][7] prior to the configurations below.
```
hkaj, line 32: we should also mention `Cluster Agent version >= 1.9.0` to make sure people who know about the cluster agent but don't run it on purpose (for various reasons) understand the implications

content/en/infrastructure/livecontainers.md
```diff
@@ -13,73 +13,100 @@ further_reading:
...
+
+{{< tabs >}}
+{{% tab "Helm" %}}
+
+If you are using the official [Datadog Helm Chart][1]:
+
+- Use chart version 2.4.5 or above
```
hkaj, line 39: we should also advise users to ensure that if they're hardcoding agent and cluster-agent versions they need to update them to the minimum versions required or above. Otherwise some of them will set the option but run old agents and it won't work.

content/en/infrastructure/livecontainers.md
```diff
@@ -13,73 +13,100 @@ further_reading:
...
+- Set `datadog.orchestratorExplorer.enabled` to true in [values.yaml][2]
+- Deploy a new release
+
+
+[1]: https://github.com/DataDog/helm-charts
+[2]: https://github.com/DataDog/helm-charts/blob/master/charts/datadog/values.yaml
+{{% /tab %}}
+{{% tab "DaemonSet" %}}
```
hkaj, line 47: not sure we should call this DaemonSet, because it requires the agent daemonset and the cluster agent deployment. If we want to distinguish this method from the Helm install I'd go with "raw manifests" or something similar

content/en/infrastructure/livecontainers.md
```diff
@@ -13,73 +13,100 @@ further_reading:
...
+
+To configure a DaemonSet, [Cluster Agent][1] version >= 1.9.0 is required. The Cluster Agent must be running, and the Agent must be able to communicate with it. See the [Cluster Agent Setup documentation][2] for configuration.
```
hkaj, line 49: `To configure a DaemonSet` --> `Before configuring the DaemonSet` maybe? Basically people need to configure and run the Cluster Agent deployment, and then the DaemonSet will work. Practically they can apply both at the same time or close to one another, and it will stabilize.

content/en/infrastructure/livecontainers.md
```diff
@@ -13,73 +13,100 @@ further_reading:
...
+
+1. Set the Cluster Agent container with the following environment variable:
+
+      ```
+      - name: DD_ORCHESTRATOR_EXPLORER_ENABLED
+        value: "true"
+      ```
+
+2. Set the Cluster Agent ClusterRole with the following RBAC permissions:
```
hkaj, line 58: I don't know if it's clear enough that the clusterrole needs to be updated with these additional permissions, but they might also need more for other features. It's not an extensive list. If you think it's clear let's keep it that way, otherwise maybe update with "Make sure these permissions are included in the Cluster Agent ClusterRole"

content/en/infrastructure/livecontainers.md
```diff
@@ -13,73 +13,100 @@ further_reading:
...
+        - list
+        - get
+        - watch
+      ```
+
+    These permissions are needed in order to create a `datadog-cluster-id` ConfigMap in the same Namespace as the Agent DaemonSet and the Cluster Agent Deployment, as well as to collect Deployments and ReplicaSets.
+
+    If the `cluster-id` ConfigMap doesn't get created by the Cluster Agent, the Agent pod will not start, and fall in `CreateContainerConfigError` status. If the Agent pod is stuck because this ConfigMap doesn't exist, update the Cluster Agent permissions and restart its pods to let it create the ConfigMap and the Agent pod will recover automatically.
+
+3. The Process-agent must be running (it doesn't need to run the process collection), and configured with the following options:
```
hkaj, line 93: This step, unlike the previous two, is to be applied to the DaemonSet. That's because the process-agent runs in the agent DaemonSet, not in the Cluster Agent Deployment. We might also want to point this out.

apigirl (commented):
lgtm other than one tiny nit and Haïssams suggestions

content/en/infrastructure/livecontainers.md
sarina-dd: Ah, good catch: do they need to update the version in the agent config, or within the helm chart?

content/en/infrastructure/livecontainers.md
sarina-dd: @cohenyair Thoughts on this? I think DaemonSet makes sense for tabbing based off of the initial beta docs / the Jira card request for use to flesh out into Helm, DaemonSet and Operator tabs going forward.

content/en/infrastructure/livecontainers.md
sarina-dd: I think this is descriptive enough as-is. 👍

content/en/infrastructure/livecontainers.md
cohenyair: I agree with @hkaj that using DaemonSet can be misleading, but since we're using it in many places in the docs and the product, we should stick to it in the short term here as well (for consistency)

content/en/infrastructure/livecontainers.md
```diff
@@ -13,73 +13,100 @@ further_reading:
...
+1. [Cluster Agent][1] version >= 1.9.0 is required before configuring the DaemonSet. The Cluster Agent must be running, and the Agent must be able to communicate with it. See the [Cluster Agent Setup documentation][2] for configuration.
+
+- Set the Cluster Agent container with the following environment variable:
+
+      ```
+      - name: DD_ORCHESTRATOR_EXPLORER_ENABLED
+        value: "true"
+      ```
+
+- Set the Cluster Agent ClusterRole with the following RBAC permissions:
```
nammn, line 59: Do we want and can link to an example RBAC? The only one I found was the one from helm: https://github.com/DataDog/helm-charts/blob/master/charts/datadog/templates/cluster-agent-rbac.yaml#L93-L121 This contains the up to date Helm RBAC "code", so not sure whether this helps or confuses people more.

content/en/infrastructure/livecontainers.md
```diff
@@ -274,20 +301,17 @@ You can see logs that you have chosen to index and persist by selecting a corres
...
+[7]: /agent/cluster_agent/setup/
+[8]: /tagging/assigning_tags?tab=agentv6v7#host-tags
+[9]: /getting_started/tagging/unified_service_tagging
+[10]: /dashboards/widgets/timeseries/
+[11]: /logs
+[12]: /metrics
+[13]: /tracing
+[14]: /events
+[15]: /logs/live_tail/
+[16]: https://gist.github.com/hkaj/404385619e5908f16ea3134218648237
```
nammn, line 316: @hkaj this points to nothing. Do you remember what we had there?

sarina-dd: +1, forgot to mention this yesterday: we had an RBAC doc we used to link here that was in the private beta doc for live containers, but looks like a 404 now.

content/en/infrastructure/livecontainers.md
hkaj: in the helm chart's values.yaml

content/en/infrastructure/livecontainers.md
hkaj: oh good catch, probably some RBAC example. Let's use this one instead: https://github.com/DataDog/datadog-agent/blob/7.23.1/Dockerfiles/manifests/cluster-agent/rbac.yaml

content/en/infrastructure/livecontainers.md
hkaj: i'd advise using https://github.com/DataDog/datadog-agent/blob/7.23.1/Dockerfiles/manifests/cluster-agent/cluster-agent-rbac.yaml
it's not helm specific, and the link is permanent so it won't 404

apigirl (approved):

hkaj (approved):

PR #9098, opened 2020-11-19
[DOCS-1587] Adding some info about dedicated containers

FatineBts:
### What does this PR do?
Adds an info about dedicated containers explaining that it's not possible to set the dd_log_level at runtime.

### Motivation
An escalation.


apigirl (approved):

apigirl:
verified in slack with @knusbaum

PR #9150, opened 2020-11-26
Correction of the Japanese translation.( CentOS/RHEL 7.6+)

taku333:
The Japanese translation was incorrect and has been corrected.
Before the modification, [CentOS/RHEL 7.6+] is listed as unavailable.

### What does this PR do?
Corrections in the Japanese language documentation.

### Motivation
After checking with support, we found out that the Japanese translation is incorrect, so we decided to use.

### DIFF
--- a/content/ja/network_performance_monitoring/installation.md
+++ b/content/ja/network_performance_monitoring/installation.md
@@ -24,7 +24,7 @@ further_reading:
 - Amazon AMI 2016.03 以降
 - Amazon Linux 2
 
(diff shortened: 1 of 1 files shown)


kayayarai:
Thank you for this improvement on our translation. The source of our Japanese docs is in our translation management system, so I cannot just accept this change here (it would be overwritten with the next build). But I have entered the suggested translation into our TMS to be evaluated and published by our translators.

Thank you so much for contributing!

taku333:
Thank you for your respons!!

PR #9156, opened 2020-11-27
Update livecontainers.md

nammn:
### Preview

https://docs-staging.datadoghq.com/nammn-patch-1/infrastructure/livecontainers

### Additional Notes

---

### Reviewer checklist

### DIFF
--- a/content/en/infrastructure/livecontainers.md
+++ b/content/en/infrastructure/livecontainers.md
@@ -87,7 +87,22 @@ In some setups, the Process Agent and Cluster Agent are unable to automatically 
             - get
             - watch
         ```
-
+Live Containers need permissions to collect common kubernetes resources (pods, services, nodes ..), which should be already in the RBAC if following [Cluster Agent Setup documentation][2].
+If not, the missing resources should be added:
+```
+          - apiGroups:  # to collect new resource types
+            - "apps"
+            resources:
+            - deployments
+            - replicasets
+            - pods
+            - nodes
+            - services
+            verbs:
+            - list
+            - get
+            - watch
+```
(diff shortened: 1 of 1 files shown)


content/en/infrastructure/livecontainers.md
```diff
@@ -87,7 +87,22 @@ In some setups, the Process Agent and Cluster Agent are unable to automatically
...
+Live Containers need permissions to collect common kubernetes resources (pods, services, nodes ..), which should be already in the RBAC if following [Cluster Agent Setup documentation][2].
+If not, the missing resources should be added:
+```
+          - apiGroups:  # to collect new resource types
+            - "apps"
+            resources:
+            - deployments
+            - replicasets
+            - pods
+            - nodes
```
mfpierre, line 99: shouldn't it be added to the above block?

nammn: hmm, yes true. I just wanted to make clear that they only need to add this part, if one did not follow the typical DCA setup. 🤔

kayayarai: Maybe we should add this instruction in line 60, something like: 
```
    - Set the Cluster Agent ClusterRole with the following RBAC permissions. Note particularly that for the `apps` apiGroups, Live Containers need permissions to collect common kubernetes resources (`pods`, `services`, `nodes`, etc.), which should be already in the RBAC if you followed [Cluster Agent Setup documentation][2]. But if they are missing, ensure they are added (after `deployments`, `replicasets`, etc.):
```
... and then add `pods` `nodes` etc after line 84. 

What do you think?

nammn: This sounds great! I added this + a comment to point that pods etc. only needs to be added if missing

apigirl (approved):

PR #9329, opened 2020-12-21
Update manual span example

mackjmr:
### What does this PR do?
Add example of creating span with startSpan as a child of a parent span

### Motivation
Specify that startSpan doesn't activate the span on the current scope, and provide an example of creating span with startSpan as a child of a parent span

### Reviewer checklist


rochdev (changes requested):
We're planning to remove all references to `tracer.startSpan` since it's almost always misused and `tracer.trace` should be used instead.

andrewardito:
Addressing as part of this: https://github.com/DataDog/documentation/pull/9333

PR #9348, opened 2020-12-22
fix yaml spacing for config examples

nicholas-devlin:
i tried copy pasting these but the yaml formatting was off


ruthnaebeck (approved):
👍

PR #9358, opened 2020-12-23
Require syslog-ng to accept only trusted certificates

fdr:
There's no purpose in downloading a root CA bundle if the instructions
go on to suggest turning off CA validation altogether.


apigirl (approved):

apigirl:
thanks for this update @fdr!

fdr:
No problem. Might want to watch out that your certificate bundle is actually seen on popular platforms. I found CentOS 8 did not have the whole bundle. I wound up having to fish through sectigo support and found the correct certificates.

PR #9862, opened 2021-02-24
Use AddAction instead of AddUserAction as it's deprecated

ThibautGeriz:
### What does this PR do?

Replace deprecated function name with new function

### Preview

https://docs-staging.datadoghq.com/thibaut.gery/replace-deprecated-function-name

### Additional Notes

---

### Reviewer checklist


bcaudan (approved):
Great, thx!

ThibautGeriz:
Maybe make sense from a browser point of view but from an mobile point of view, does not seems to be correct as I can only reference to `addUserAction` in ios & android SDK. 
As far as I can tell, there is only a mobile doc in French & Japonese not english.

kayayarai (approved):
Hi! The changes you've made to the English files so far are fine, approved. The English source for iOS and Android are here, and you can make the same changes there:
- https://github.com/DataDog/dd-sdk-ios/blob/master/docs/rum_collection.md
- https://github.com/DataDog/dd-sdk-android/blob/master/docs/rum_collection.md
The changes you made to JA and FR, while appreciated, might get overwritten by the current state of the JA and FR translation project, until they catch up from these latest English changes. It won't take long for them to find their way back through the translation pipeline though.

PR #9868, opened 2021-02-24
error status config example

nicholas-devlin:
The examples for `DD_HTTP_CLIENT_ERROR_STATUSES` and `DD_HTTP_SERVER_ERROR_STATUSES` were the exact same as the default, which is confusing, and it didn't show off the comma separated list syntax that is required to make the field useful.

Example of the config from the code 

```
"Invalid config for {}: '{}'. Must be formatted like '400-403,405,410-499'."
```
https://github.com/DataDog/dd-trace-java/blob/0567633c0561825c9c26951e3449af69f94f6489/internal-api/src/main/java/datadog/trace/bootstrap/config/provider/ConfigConverter.java#L117-L126


jtappa (approved):
Wow, thanks so much for the example upgrade this is great!

PR #9936, opened 2021-03-04
[agent] Update k8s manifests based on helm chart 2.9.11

ahmed-mez:
### What does this PR do?

Update the agent k8s manifests to be aligned with the latest helm chart

### Motivation

Follow-up on https://github.com/DataDog/datadog-agent/pull/7461

### Preview

### Additional Notes

---

### Reviewer checklist

### DIFF
--- a/static/resources/yaml/datadog-agent-npm.yaml
+++ b/static/resources/yaml/datadog-agent-npm.yaml
@@ -1,5 +1,4 @@
-# Source: datadog/templates/secrets.yaml
-# API Key
+# Source: datadog/templates/secret-api-key.yaml
 apiVersion: v1
 kind: Secret
 metadata:
@@ -8,8 +7,6 @@ metadata:
 type: Opaque
 data:
   api-key: PUT_YOUR_BASE64_ENCODED_API_KEY_HERE
-
-# APP Key
 ---
 # Source: datadog/templates/install_info-configmap.yaml
 apiVersion: v1
@@ -18,7 +15,7 @@ metadata:
   name: datadog-agent-installinfo
   labels: {}
   annotations:
-    checksum/install_info: 9f58f4fe71f1b79dfabae5311eb8f5373bd03072d4636e248cd3f581f8752627
+    checksum/install_info: 092f34522792579934b4afae83bc34fc9fa2ea9d89ab29459dd8594579a3c8fa
 data:
   install_info: |
     ---
@@ -44,7 +41,11 @@ data:
       bpf_debug: false
       enable_tcp_queue_length: false
       enable_oom_kill: false
-      collect_dns_stats: true
+      collect_dns_stats: false
+      max_tracked_connections: 65536
+      conntrack_max_state_size: 131072
+    network_config:
+      enabled: true
     runtime_security_config:
       enabled: false
       debug: false
@@ -264,7 +265,7 @@ spec:
     spec:
       containers:
         - name: agent
-          image: "gcr.io/datadoghq/agent:7.23.1"
+          image: "gcr.io/datadoghq/agent:7.26.0"
           imagePullPolicy: IfNotPresent
           command: ["agent", "run"]
           resources: {}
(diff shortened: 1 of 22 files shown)


L3n41c (changes requested):
It seems that the default values of a lot of parameters in the Helm chart have changed so that now, the DCA, the `security-agent`, the RBAC, etc. are enabled by default.

As a consequence, the DCA for ex. is enabled in some sample manifests where it used to not be.

I think that those manifests are used to illustrate specific features from the documentation.
For ex., the `datadog-agent-apm.yaml` file is used as a reference from the doc to illustrate what are the specific parameters to use to enable APM.
In that context, it might make sense to keep this `datadog-agent-apm.yaml` as simple as possible with only the minimum number of options required to enable APM.

So, I would rather explicitly disable the settings in the `values-*.yaml` files that are now enabled by default in the latest version of the chart to not add new features in the sample manifests.

clamoriniere (approved):
Ok, so maybe for the  `datadog-agent-apm.yaml ` we can disable the cluster-agent

L3n41c:
I took `datadog-agent-apm.yaml` as an example, but I think that most of those manifests are used to illustrate a specific part of the documentation. Hence my recommendation to keep them as simple as possible to highlight only the parameters for a specific feature.
We can review what are all the samples and where they are referenced from, but I think that what I described could be generalised to the others.

ahmed-mez:
I agree with the suggestions, thank you both! PR updated accordingly

static/resources/yaml/datadog-agent-npm.yaml
```diff
@@ -498,12 +525,16 @@ spec:
...
+        - name: tmpdir
+          emptyDir: {}
         - hostPath:
             path: /proc
           name: procdir
         - hostPath:
             path: /sys/fs/cgroup
           name: cgroups
+        - name: s6-run
+          emptyDir: {}
```
L3n41c, line 537: This looks like a bug in the Helm chart rather than in this PR, but a new volume is declared here whereas it is mounted in no container.
So, it’s useless.
And it’s expected to be useless since there’s no S6 in the multi-container setup.

static/resources/yaml/datadog-agent-npm.yaml
```diff
@@ -44,7 +41,11 @@ data:
       bpf_debug: false
       enable_tcp_queue_length: false
       enable_oom_kill: false
-      collect_dns_stats: true
+      collect_dns_stats: false
```
L3n41c, line 44: This change of `collect_dns_stats` from `true` to `false` is surprising, but I’m OK to stick with the default value here.

static/resources/yaml/datadog-agent-vanilla.yaml
```diff
@@ -140,30 +146,43 @@ spec:
...
-            - name: DD_ORCHESTRATOR_EXPLORER_ENABLED
+            - name: DD_SYSTEM_PROBE_ENABLED
               value: "false"
+            - name: DD_ORCHESTRATOR_EXPLORER_ENABLED
+              value: "true"
+            - name: DD_ORCHESTRATOR_CLUSTER_ID
+              valueFrom:
+                configMapKeyRef:
+                  name: datadog-cluster-id
+                  key: id
```
L3n41c, line 161: This manifest now references a new `ConfigMap` but this `ConfigMap` hasn’t been added to the manifest.
So, the instantiation will most probably fail because the `CM` is missing.

L3n41c (commented):
We let the DCA being switched from `disabled` to `enabled` in the `datadog-agent-all-features.yaml`.
Given the name, it makes sense even i feel that this name used to mean “all the features of the **node** agent”.

I’m more concerned by the reference to the `ConfigMap` which isn’t part of the manifest. This would make the manifest instantiation fail.

L3n41c (approved):

ruthnaebeck (approved):

PR #10142, opened 2021-03-29
Update otel java trace and logs

mackjmr:
- getTraceIdAsHexString() was changed to getTraceId() in v.0.16 of the tracer
- add example for span


ericmustin (approved):
👍 lgtm from a technical standpoint

kayayarai (approved):

PR #10145, opened 2021-03-29
[WEB-1250] fixing multi code lang navigation bug

davidejones:
### What does this PR do?

Fixes a bug with persisting the users selected code lang and navigating the site

### Motivation

https://datadoghq.atlassian.net/browse/WEB-1250

### Preview

If, for example, they're on /tracing/setup_overview/setup/python, and they click "Compatibility" via the sidebar TOC or a link, it should set the code-lang cookie to Python and take them to /tracing/setup_overview/compatibility_requirements/python (not java).

https://docs-staging.datadoghq.com/david.jones/code-lang-tabfix/tracing/setup_overview/setup/python

### Additional Notes

---

### Reviewer checklist

### DIFF
--- a/src/scripts/datadog-docs.js
+++ b/src/scripts/datadog-docs.js
@@ -1,4 +1,3 @@
-import Cookies from 'js-cookie';
 import Stickyfill from 'stickyfilljs';
 import algoliasearch from 'algoliasearch';
 import Choices from 'choices.js';
--- a/src/scripts/datadog-docs.js
+++ b/src/scripts/datadog-docs.js
@@ -840,16 +840,6 @@ function navClickEventHandler(event) {
 
     newUrl = event.target.closest('li').querySelector('a').href;
 
-    /*
-    If this is a multi code lang page we are clicking on
-    lets check if we have a lang already and load that url
-    */
-    const type = event.target.closest('li').querySelector('a').dataset.type;
-    if(type === 'multi-code-lang') {
-      const codeLang = Cookies.get('code-lang') || document.documentElement.dataset.pageCodeLang || 'python';
-      newUrl = newUrl.replace(/\/[^\/]*$/, `/${codeLang}`)
-    }
-
     const domain = window.location.origin;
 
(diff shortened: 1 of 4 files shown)


src/scripts/datadog-docs.js
```diff
@@ -839,6 +840,16 @@ function navClickEventHandler(event) {
 
     newUrl = event.target.closest('li').querySelector('a').href;
 
+    /*
+    If this is a multi code lang page we are clicking on
+    lets check if we have a lang already and load that url
+    */
+    const type = event.target.closest('li').querySelector('a').dataset.type;
+    if(type === 'multi-code-lang') {
+      const codeLang = Cookies.get('code-lang') || document.documentElement.dataset.pageCodeLang || 'python';
```
bgdeutsch, line 849: this might never happen, but if for some reason there is no `code-lang` cookie and no `pageCodeLang` data attribute, can we be certain there will always be a page for python?

davidejones: Yea that is a good point and it doesn't feel right to do it here. Now i've tried a few different options I'm thinking the best approach might be to just make sure the nav that the user clicks on is already showing the language they are on if possible. This seems better than trying to catch this during async loading or some other area.

davidejones: @bgdeutsch i've committed a different approach, if you get time to take a look at that instead and see what you think.

bgdeutsch (approved):
looks good from what I can see!   would be good to have some extra 👀  as well i think.

kayayarai (approved):
Yaaaas! Works for me

src/scripts/datadog-docs.js
```diff
@@ -1,3 +1,4 @@
+import Cookies from 'js-cookie';
```
stanleegoodspeed, line 1: Is this import statement needed? I'm not seeing it referenced anywhere else in this file

davidejones: Yea you are right it was from previous code changes, removed now

PR #10563, opened 2021-05-12
[WEB-1393] fail build pipeline if code examples cant be pulled

bgdeutsch:
### What does this PR do?
cause the build pipeline to fail if we cannot pull code examples from external repos.

### Motivation
https://datadoghq.atlassian.net/browse/WEB-1393

### Preview
n/a

testing:

- pull down this branch locally
- manually change something inside the examples so it will fail.  for example here: https://github.com/DataDog/documentation/blob/brian.deutsch/fail-pipeline-without-code-examples/Makefile#L248 change `extract-code-blocks.sh` to have a fake name or so that its within a directory that does not exist.
- run `make start`, the build should fail.
- here's an example of the build pipeline failing: https://gitlab.ddbuild.io/DataDog/documentation/-/jobs/66192938#L40

### Additional Notes

---

### Reviewer checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -245,7 +245,7 @@ EXAMPLES_DIR = $(shell pwd)/examples/content/en/api
 
 examples/go: examples/datadog-api-client-go clean-go-examples
 	echo $(EXAMPLES_DIR)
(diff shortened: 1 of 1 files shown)


davidejones (approved):
looks good, tested by breaking an awk script and seeing the build stop

jtappa (approved):

PR #10753, opened 2021-06-04
rate limit: number notation change

sguyon:
### What does this PR do?

Change 500'000 to 500,000


sarina-dd:
@sguyon Pushed a change for this yesterday that is now live 😄 Thanks for fixing, though! Closing this PR as this is already resolved.

PR #10773, opened 2021-06-08
Fix the URN in ja/account_management/saml

umireon:
### What does this PR do?
Fixes the URN error.

### Motivation
Wrong URN in a ja document.

### Preview

### Additional Notes

---

### Reviewer checklist


jtappa:
Hi @umireon, thanks for your contribution! I'm going to add a card on our Jira to track this change and confirm that the URN needs updating. If that's the case, it will get updated in the /en/ version of the docs and work its way into our localization automation. If we committed it to the /ja file it would just end up being overwritten by that automation. :) Thank you!

umireon:
@jtappa This URN in the en document is correct.
Wrong only in ja document.
Renewing the translated content will fix this error.

PR #10918, opened 2021-06-23
Adding a new library

apigirl:
Adding a new library to the developer libraries page:

https://docs-staging.datadoghq.com/kaylyn/adding-finagle/developers/libraries/

---

### Reviewer checklist

### DIFF
--- a/data/libraries.yaml
+++ b/data/libraries.yaml
@@ -315,6 +315,12 @@ Tracing:
     - name: scala-opentracing
       link: https://github.com/Colisweb/scala-opentracing
       authors: Colisweb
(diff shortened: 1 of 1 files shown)


jtappa (approved):

jtappa (changes requested):
actually, it looks like this table is alphabetized, maybe we should add Finagle in alphabetic order of the table?

apigirl:
hey, this is a good catch! for this library I would prefer to have it at the bottom just because it is not going to be used nearly as much as the other ones.

jtappa (approved):

PR #10970, opened 2021-06-28
Fixes for Serverless Docs

nhinsch:
### What does this PR do?
Update docs to fix a few issues related to Serverless installation:
- The Datadog Lambda Extension can now be used in all AWS regions, including GovCloud. (The exceptions are the mainland China regions, but I don't think our docs really apply there at all).
- Include an API key when setting up the CloudFormation Macro. This is required when adding the Lambda Extension.
- Correct the parameter that sets the API key when using the CDK Construct Library in Python from `dd_api_key` to `api_key`.

---

### Reviewer checklist
- [ ] Check images for PII


kayayarai (approved):

Hesperide (approved):
Thank you Nick :)

PR #10993, opened 2021-06-30
Include latest Lambda Layer versions in documentation

nhinsch:
### What does this PR do?
Include the latest Lambda Layer versions directly in our documentation. Add a shortcode called `latest-lambda-layer-version` which prints out the version in order to make this maintainable going forward.

### Motivation
Currently, we frequently ask our users to click a link to GitHub or Amazon ECR Public to determine the latest version of a particular Lambda Layer. This is not ideal because:
- The user needs to leave our documentation and come back
- It can be hard to determine exactly what the latest layer version is. For example, the latest release of the dd-lambda-python is called `v3.56.0` in GitHub, but the layer version (what we care about) is just `56`.
It would be much simpler to just state the latest version number in our documentation. Because the version numbers appear in many places, we need to store the version number in a single place and then interpolate it into the documentation using a shortcode so that we can easily go back in and update the version numbers in the future.

### Preview
https://docs-staging.datadoghq.com/layer-versions/content/en/serverless/guide/datadog_forwarder_node.md
https://docs-staging.datadoghq.com/layer-versions/content/en/serverless/guide/datadog_forwarder_python.md
https://docs-staging.datadoghq.com/layer-versions/content/en/serverless/installation/dotnet.md
https://docs-staging.datadoghq.com/layer-versions/content/en/serverless/installation/java.md
https://docs-staging.datadoghq.com/layer-versions/content/en/serverless/installation/nodejs.md
https://docs-staging.datadoghq.com/layer-versions/content/en/serverless/installation/python.md
https://docs-staging.datadoghq.com/layer-versions/content/en/serverless/installation/ruby.md
https://docs-staging.datadoghq.com/layer-versions/content/en/serverless/libraries_integrations/extension.md

### Additional Notes
I explored and rejected an alternative solution using [shields.io](https://shields.io/). Shields.io is a service for creating "badges" with version numbers. I rejected the solution for two reasons:
- To use shields.io, we would need to expose an API endpoint in our production AWS account to tell shields.io what the latest layer versions are. This would likely take the form of a Lambda function and an API Gateway endpoint. I would rather not deploy and indefinitely maintain production infrastructure, even if it is relatively simple, in order to support this change.
- Shields.io creates badges, which are images. The images wouldn't fit into our documentation as seamlessly as text.

---

### Reviewer checklist
- [ ] Check images for PII


apigirl (commented):
name your branch with a `/`

nhinsch:
Re-opening with a different branch name.

PR #10999, opened 2021-07-01
Forwarder/Extension Migration Guide

cswatt:
### Motivation

https://docs-staging.datadoghq.com/cswatt/extension-migration/serverless/guide/forwarder_extension_migration

### Preview

https://docs-staging.datadoghq.com/cswatt/extension-migration/serverless/guide/extension_motivation/

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII


content/en/serverless/guide/forwarder_extension_migration.md
```diff
@@ -0,0 +1,123 @@
...
+If this is a significant deterrent to migration, but you still wish to benefit from the advantages of the Datadog Lambda Extension, you may consider a [hybrid approach](#hybrid-approach).
+
+## How to migrate from the Forwarder to the Extension
+
+Datadog recommends that you set up the Extension **before** removing the Forwarder. This means that for the period during which both the Extension and Forwarder are active, all telemetry is sent twice. The alternative (removing the Forwarder before setting up the Extension) results in a period of zero visibility. Depending on your organization's particular needs, you may prefer one outcome over the other.
+
+### Setting up the Datadog Lambda Extension {#extension-setup}
+
+{{< tabs >}}
+{{% tab "Lambda Layer" %}}
```
tianchu, line 32: These instructions won't work. Many customers use serverless framework and out plugin for instance, which is not covered in this guide.

I think we need to cover all the installation methods for Node.js and Python -- Serverless Framework, AWS SAM, etc...

Hesperide: Can we roll with this for now (better than having no migration docs), and consider covering all the installation methods as part of the more detailed installation instructions?

content/en/serverless/guide/forwarder_extension_migration.md
```diff
@@ -0,0 +1,123 @@
...
+
+For more information, see the [Datadog Lambda Extension documentation][1].
+
+### Removing the Datadog Forwarder
+
+Remove subscriptions (TK)
+
+#### Prevent the Forwarder from attaching to new functions {#prevent-forwarder}
+
+When you deploy a new function, the Datadog Forwarder detects an unsubscribed function and automatically creates a new subscription. Depending on whether the Forwarder was configured automatically or manually, complete the following steps:
```
Hesperide, line 93: I think the instructions here are the same regardless of whether the Forwarder was configured automatically or manually. The main difference with automatic is that the Forwarder will not automatically re-subscribe to your function's log groups if the Extension is installed on that function.

We should instead look to split this up based on the deployment method you are using (Serverless Framework, AWS SAM, AWS CDK, Custom)

content/en/serverless/guide/forwarder_extension_migration.md
```diff
@@ -0,0 +1,123 @@
...
+
+{{< tabs >}}
+{{% tab "Automatic" %}}
+
+tk
+
+{{% /tab %}}
+{{% tab "Manual" %}}
+
+1. Remove the resource name of the Forwarder from your `serverless.yaml` file.
```
Hesperide, line 103: Related to the above comment, you only have a `serverless.yaml` if you are deploying your function with the Serverless Framework. For example, if you are using SAM, you have a `template.yml` and in CDK you need to modify your CDK stack.

Hesperide (changes requested):
Looks really good so far, left some feedback on tweaks most notably in the last section on removing the Forwarder.

content/en/serverless/guide/forwarder_extension_migration.md
```diff
@@ -0,0 +1,123 @@
...
+2. Redeploy your cloud permissions stack.
```
Hesperide, line 104: You don't necessarily deploy Lambda functions via CloudFormation stacks. It's a popular way of managing infrastructure as code, but typically customers don't directly work with CloudFormation. For example, the Serverless Framework, AWS SAM and AWS CDK are tools built on top of CloudFormation.

When you modify your functions directly in the console, you aren't using CloudFormation there either.

content/en/serverless/guide/forwarder_extension_migration.md
```diff
@@ -0,0 +1,123 @@
...
+
+{{% /tab %}}
+{{< /tabs >}}
+
+For more information, see the [Datadog Forwarder documentation][2].
```
Hesperide, line 109: We'll probably need to add a note at the end here outlining how to remove the Forwarder once you complete the migration.

content/en/serverless/guide/extension_motivation.md
```diff
@@ -0,0 +1,54 @@
...
+| ------- | ----------------- | ----------------- |
+| Python  | Yes               | Yes               |
+| Node.js | Yes               | Yes               |
+| Go      |                   | Yes               |
+| Java    |                   | Yes               |
+| .NET    |                   |                   |
+| Ruby    |                   |                   |
+| PHP     |                   |                   |
+
+For Python and Node.js runtimes, the Lambda extension has full feature parity with the Forwarder. The Lambda extension can submit Lambda logs, Lambda traces (generated by `dd-trace`, not X-Ray), real-time enhanced Lambda metrics, and custom metrics generated by Lambda functions.
```
pinkerton, line 30: And Go now

content/en/serverless/guide/extension_motivation.md
```diff
@@ -0,0 +1,54 @@
...
+- **Less infrastructure to manage**: The Lambda extension's simpler configuration enables you to save time on infrastructure management. **Note**: You still require the Forwarder for non-Lambda AWS integrations.
+- **Skip CloudWatch Logs**: The Forwarder converts logs to metrics and traces, which are then sent to Datadog. The Datadog Lambda Extension sends traces, metrics, and logs directly to Datadog, enabling you to reduce the cost associated with CloudWatch Logs.
+
+**Note**: If you are reporting to Datadog sites in the same regions as the ones in which you are deployed (for example: if you are deployed in APAC regions and are reporting to APAC Datadog sites), your use-case is well suited for the extension. However, if you are reporting to a Datadog site in a different region than the one in which you are deployed (for example: if you are deployed in APAC regions and are reporting to US Datadog sites) the Forwarder may be a cheaper option.
```
pinkerton, line 40: We can now confidently recommended the extension for these use cases. Tian addressed in this PR: https://github.com/DataDog/documentation/pull/11264

content/en/serverless/guide/extension_motivation.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Trade-offs
+
+Migration from the Forwarder to the Lambda extension can be an involved process. There is one Forwarder setup per region and account, but there is one Lambda extension setup **per function**. That is, if you are shipping data from *N* functions using the Forwarder, you must set up the Lambda extensio *N* times before the migration is complete. Since you must reconfigure and redeploy **each** function, the full process of migration may take some time.
```
pinkerton, line 44: We should note that our plugins for the Serverless Framework, AWS CDK, AWS SAM, and the Datadog CLI can automate this by letting you add the extension to an application composed of many Lambda functions all at once.

content/en/serverless/guide/extension_motivation.md
```diff
@@ -0,0 +1,54 @@
...
+
+If this is a significant deterrent to migration, but you still wish to benefit from the advantages of the Datadog Lambda extension, you may consider a [hybrid approach][4].
+
+The extension adds overhead to your Lambda functions, compared to functions that have zero instrumentation. The added overhead affects your AWS bill and Lambda concurrency, and could cause worse cold starts. The majority of the added duration **does not** affect your function's performance.
```
pinkerton, line 48: We share the ballpark ranges for that overhead here: https://github.com/DataDog/documentation/pull/11264

We should either link to that or put the same information here

tianchu: We should also mention that we are actively improving the overhead.

content/en/serverless/guide/extension_motivation.md
tianchu: Actually don't think this trade-off holds true. Because the functions emitting data via forwarder already have the Datadog Lambda library layer installed, which also requires set up per function. Installing one more layer doesn't seem to be too much extra work.

tianchu: This holds true only if the customer has only been collecting Lambda logs using the Forwarder, aka, no lambda library installed to each lambda function -- I would recommend them continue doing so.

If the customer has installed Lambda library to their individual Lambda functions, and been collecting metrics and traces in addition to logs, I would recommend them migrate to using Extension.

content/en/serverless/guide/extension_motivation.md
tianchu: Actually based on the latest benchmarking results, the cost overhead will always be lower (or about the same when reporting data from remote regions) between using extension vs forwarder.

This is indeed a penalty on the cold start time, but we have identified a few quick wins that can bring the overhead down significantly (even lower than using the forwarder today), so soon it will no longer be a concern.

content/en/serverless/guide/forwarder_extension_migration.md
tianchu: I'm not sure we want to recommend that, since the forwarder is likely used for sending aws logs from other services. At least it seems a good idea to me to keep it, in case next week you want to start sending api gateway logs for example. Having the Datadog Forwarder Lambda existing in your account doesn't cost a penny. This is the same idea behind the idea of installing the forwarder as part of the aws integration, no matter you use it or not.

tianchu (approved):

content/en/serverless/guide/forwarder_extension_migration.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Hybrid approach
+
+You may also wish to keep using the Datadog Forwarder for existing functions while using the Datadog Lambda Extension for all new functions. 
+
+1. Ensure that the Forwarder [does not attach automatically](#prevent-forwarder) to new functions.
+2. For each new function you deploy, [set up the Extension](#extension-setup)
```
alai97, line 72: Can we explain the user workflow with more detail?

Does this belong at the bottom of the doc or are these part of some setup instructions in Datadog Lambda Extension?

alai97 (changes requested):
Docs review completed!

content/en/serverless/guide/extension_motivation.md
```diff
@@ -0,0 +1,47 @@
+---
+title: Deciding to migrate to the Datadog Lambda extension
+kind: guide
+---
+
+## Should I migrate to the Datadog Lambda extension?
+
+AWS Lambda extensions run within the Lambda execution environment, alongside your Lambda function code. Datadog partnered with AWS to create the [Datadog Lambda extension][1], a lightweight version of the Datadog Agent that submits custom metrics, enhanced metrics, traces, and logs.
```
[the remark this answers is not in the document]
cswatt: feeling like "within" and "alongside" are more accurate adverbs here

content/en/serverless/guide/extension_motivation.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Advantages
+
+The Datadog Lambda Extension offers the following advantages over the Datadog Forwarder:
+
+- **Simple to set up**: Triggers need to be set up on the Forwarder for every new Lambda function. The Datadog Lambda extension can be easily added as a Lambda layer. Unlike with the Forwarder, configuring the Lambda extension does not require permissions to install third-party AWS CloudFormation stacks. Additionally, since the Lambda extension sends telemetry directly to Datadog, you do not need to manage CloudWatch Log group subscriptions for your Lambda functions.
+- **Less infrastructure to manage**: The Lambda extension's simpler configuration enables you to save time on infrastructure management. **Note**: You still require the Forwarder for non-Lambda AWS integrations.
+- **Skip CloudWatch Logs**: The Forwarder converts logs to metrics and traces, which are then sent to Datadog. The Datadog Lambda Extension sends traces, metrics, and logs directly to Datadog, enabling you to reduce the cost associated with CloudWatch Logs.
```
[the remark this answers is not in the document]
cswatt: not necessarily a reduction, only an enablement of reduction

content/en/serverless/guide/forwarder_extension_migration.md
```diff
@@ -0,0 +1,75 @@
+---
+title: Migrating from the Datadog Forwarder to the Datadog Lambda Extension
+kind: guide
+---
+
+The [Datadog Lambda Extension][1] is a lightweight version of the Datadog Agent built to run alongside your code with minimal performance overhead. 
+
+This guide explains how to migrate from the [Datadog Forwarder][2] to the Datadog Lambda Extension
+
+Datadog recommends that you set up the Extension **before** removing the Forwarder. This means that for the period during which both the Extension and Forwarder are active, all telemetry is sent twice. The alternative (removing the Forwarder before setting up the Extension) results in a period of zero visibility. Depending on your organization's particular needs, you may prefer one outcome over the other.
```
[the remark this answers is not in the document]
cswatt: "before" is the key word, as the alternative is "after." the last sentence is also still necessary

content/en/serverless/guide/forwarder_extension_migration.md
cswatt: each thing links to a detailed workflow so i don't think that's necessary

and yep this is the hybrid approach so it's like the alternative process where you just do things in a different order

content/en/serverless/guide/forwarder_extension_migration.md
```diff
@@ -0,0 +1,75 @@
...
+
+Manually remove the subscription from the AWS console.
+
+{{% /tab %}}
+{{% tab "Custom" %}}
+
+Manually remove the subscription from the AWS console.
+
+{{% /tab %}}
+{{< /tabs >}}
```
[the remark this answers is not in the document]
cswatt: super depends on people's setup so unfortunately this is about as specific as it gets

content/en/serverless/guide/extension_motivation.md
```diff
@@ -0,0 +1,47 @@
...
+
+| Runtime | Extension support | Forwarder support |
+| ------- | ----------------- | ----------------- |
+| Python  | Yes               | Yes               |
+| Node.js | Yes               | Yes               |
+| Go      | Yes               | Yes               |
+| Java    |                   | Yes               |
+| .NET    |                   |                   |
+| Ruby    |                   |                   |
+| PHP     |                   |                   |
```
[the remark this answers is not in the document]
cswatt: updating to checkmarks to align with other tables

Hesperide (approved):

apigirl (approved):

PR #11002, opened 2021-07-01
[CIAPP-2213] Update testing and CI provider instructions

fermayo:
### What does this PR do?

Improves the instructions for CI/testing instrumentation:
* Removes the CI provider compatibility section (we are really compatible with all providers, we would just miss build metadata in case the CI provider is not supported, so the list is confusing)
* Updates interpreter/framework compatibility section on all languages
* Bring service name and environment configuration to the "instrument your tests" section as something required (putting service name as something defined "in code" and environment defined via environment variables where possible).
* Homogenize the "additional configuration settings" section across languages
* Add `enabled` setting to  tracers where applicable
* Refactors the JUnit XML upload section
* Updates GitLab instructions with latest version
* Adds tabs where applicable

### Motivation

Current instructions are too different between languages, which makes it hard to translate them into an in-app onboarding experience as-is.

### Preview

https://docs-staging.datadoghq.com/fermayo/ciapp-bump-env-service-name/continuous_integration/setup_pipelines/gitlab/
https://docs-staging.datadoghq.com/fermayo/ciapp-bump-env-service-name/continuous_integration/setup_pipelines/jenkins/
https://docs-staging.datadoghq.com/fermayo/ciapp-bump-env-service-name/continuous_integration/setup_tests/dotnet/
https://docs-staging.datadoghq.com/fermayo/ciapp-bump-env-service-name/continuous_integration/setup_tests/java/
https://docs-staging.datadoghq.com/fermayo/ciapp-bump-env-service-name/continuous_integration/setup_tests/javascript/
https://docs-staging.datadoghq.com/fermayo/ciapp-bump-env-service-name/continuous_integration/setup_tests/python/
https://docs-staging.datadoghq.com/fermayo/ciapp-bump-env-service-name/continuous_integration/setup_tests/ruby/
https://docs-staging.datadoghq.com/fermayo/ciapp-bump-env-service-name/continuous_integration/setup_tests/swift/
https://docs-staging.datadoghq.com/fermayo/ciapp-bump-env-service-name/continuous_integration/setup_tests/junit_upload/

### Additional Notes

Ruby docs are still not accessible, which is expected, until release 0.51.0 is out. This will be re-added in a separate PR.

---

### Reviewer checklist
- [ ] Check images for PII


content/en/continuous_integration/setup_tests/junit_upload.md
```diff
@@ -10,87 +10,89 @@ further_reading:
...
 
-For example:
+You will need to specify a valid [Datadog API key][3] in the `DATADOG_API_KEY` environment variable, and the environment where tests were run (e.g. `local` when uploading results from a developer workstation, or `ci` when uploading them from a CI provider) in the `DD_ENV` environment variable. For example:
 
 {{< code-block lang="bash" >}}
-datadog-ci junit upload --service my-service \
-  --tags key1:value1 --tags key2:value2 \
-  unit-tests/junit-reports acceptance-tests/junit-reports e2e-tests/single-report.xml
+export DATADOG_API_KEY=a5f...ea170229
+export DD_ENV=ci
```
AdrianLC, line 39: isn't this optional? at least I never set it on my tests

content/en/continuous_integration/setup_tests/javascript.md
```diff
@@ -70,56 +59,69 @@ module.exports = {
...
 
-{{< code-block lang="javascript" >}}
-require('dd-trace').init({
-  enabled: !!process.env.CI // the environment variable to use depends on the CI provider
-})
-{{< /code-block >}}
+`service`
+: Name of the service or library under test.<br/>
+**Environment variable**: `DD_SERVICE`<br/>
+**Default**: `unnamed-java-app`<br/>
```
juan-fernandez, line 99: this isn't right

content/en/continuous_integration/setup_tests/javascript.md
```diff
@@ -70,56 +59,69 @@ module.exports = {
...
-{{< code-block lang="javascript" >}}
-// package.json
+{{< code-block lang="javascript" filename="package.json" >}}
 'scripts': {
-  'test': 'mocha --require dd-trace/init'
+  'test': 'DD_SERVICE=my-ui-app mocha --require dd-trace/init'
 },
 {{< /code-block >}}
 
-## Disabling instrumentation in local development
```
juan-fernandez, line 98: we might want to keep this section. If the instrumentation is added without taking this into account, running local tests will fail, which is annoying (it has been when dogfooding)

content/en/continuous_integration/setup_tests/javascript.md
```diff
@@ -70,56 +59,69 @@ module.exports = {
...
+
+`hostname`
+: The Datadog Agent hostname.<br/>
+**Environment variable**: `DD_TRACE_AGENT_HOSTNAME`<br/>
+**Default**: `localhost`
+
+`port`
+: The Datadog Agent trace collection port.<br/>
+**Environment variable**: `DD_TRACE_AGENT_PORT`<br/>
+**Default**: `8126`
```
juan-fernandez, line 121: do we want to single out these parameters?

content/en/continuous_integration/setup_tests/junit_upload.md
```diff
@@ -10,87 +10,89 @@ further_reading:
...
+**Note**: Tags specified using `--tags` and via the `DD_TAGS` environment variable are merged. If the same key appears in both `--tags` and `DD_TAGS`, the value in the environment variable `DD_TAGS` takes precedence.
 
 `--max-concurrency`
-: The number of concurrent uploads to the API.<br>
+: The number of concurrent uploads to the API.<br/>
 **Default**: `20`
 
 `--dry-run`
-: Runs the command without the final upload step. All other checks are performed.<br>
+: Runs the command without the final upload step. All other checks are performed.<br/>
```
AdrianLC, line 65: @juan-fernandez does this actually do anything for junit upload? What does it print

content/en/continuous_integration/setup_tests/junit_upload.md
```diff
@@ -10,87 +10,89 @@ further_reading:
...
 {{< /code-block >}}
 
+## Additional configuration settings
+
+This is the full list of options available when using the `datadog-ci junit upload` command:
+
 `--service`
-: The name of the service you're uploading JUnit XML reports for.<br>
-**Default**: the `DD_SERVICE` environment variable value.
+: Name of the service or library under test.<br/>
```
juan-fernandez, line 48: can we consider `--service` "additional"? It is mandatory and explained earlier

juan-fernandez (changes requested):
the default service name for javascript is wrong. Otherwise, only minor comments

content/en/continuous_integration/setup_tests/junit_upload.md
juan-fernandez: the files to be uploaded. It's something from sourcemaps. We can remove it if we don't think it adds value

content/en/continuous_integration/setup_tests/junit_upload.md
fermayo: It is optional but to avoid confusion and mixing local and CI test results that are not easily fixable later (i.e. requires deleting test services), we should put it on the "quickstart" instructions

content/en/continuous_integration/setup_tests/dotnet.md
```diff
@@ -71,38 +53,28 @@ dd-trace --help
...
+: Datadog Agent URL for traces.<br/>
 **Environment variable**: `DD_TRACE_AGENT_URL`<br/>
 **Default**: `http://localhost:8126`
 
-Additionally, all [Datadog Tracer configuration][2] options can be used during test phase. 
-
-For example, to run a test suite with a custom agent URL and a custom service name:
+`--set-ci`
+: Sets up the clr profiler environment variables for all the CI pipeline. <br/>
+**Default**: `false`
```
fermayo, line 75: @tonyredondo do we need this here?

content/en/continuous_integration/setup_tests/javascript.md
fermayo: what should we put there?

juan-fernandez: `mocha` for mocha test suites and `jest` for jest test suites. 
```
**Default**: depends on the testing framework
```
maybe?

fermayo: Is it the name of the package or something like that?

content/en/continuous_integration/setup_tests/dotnet.md
```diff
@@ -13,55 +13,37 @@ further_reading:
...
-{{< /code-block >}}
-
 ## Instrumenting your tests
 
-To instrument your test suite, prefix your test command with `dd-trace`. For example:
+To instrument your test suite, prefix your test command with `dd-trace`, providing the name of the service or library under test as the `--dd-service` parameter, and the environment where tests are being run (i.e. `local` when running tests on a developer workstation, or `ci` when running them on a CI provider) as the `--dd-env` parameter. For example:
 
 {{< code-block lang="bash" >}}
-dd-trace dotnet test
+dd-trace --dd-service my-dotnet-app --dd-env ci dotnet test
```
tonyredondo, line 41:
```suggestion
dd-trace --dd-service=my-dotnet-app --dd-env=ci dotnet test
```

content/en/continuous_integration/setup_tests/dotnet.md
tonyredondo: Not for now, or at least not in this way, could mislead the user. 

I'll work to increase the CI support for that option (only Azure Pipeline is supported) first.

content/en/continuous_integration/setup_tests/dotnet.md
```diff
@@ -111,7 +83,7 @@ If the application expects command line arguments, use a `--` separator before t
 The following example shows how to instrument the command `dotnet test --framework netcoreapp3.1` with `ci` as environment:
 
 {{< code-block lang="bash" >}}
-dd-trace --dd-env=ci -- dotnet test --framework netcoreapp3.1
+dd-trace --dd-service my-dotnet-app --dd-env=ci -- dotnet test --framework netcoreapp3.1
```
tonyredondo, line 86:
```suggestion
dd-trace --dd-service=my-dotnet-app --dd-env=ci -- dotnet test --framework netcoreapp3.1
```

tonyredondo (approved):
LGTM just a few nits

content/en/continuous_integration/setup_tests/java.md
```diff
@@ -79,22 +67,24 @@ dependencies {
...
+Configure the [Maven Surefire Plugin][3] and/or the [Maven Failsafe Plugin][4] to use Datadog Java agent, specifiying the name of the service or library under test with the `-Ddd.service` property:
 
-{{< code-block lang="xml" >}}
+{{< code-block lang="xml" filename="pom.xml" >}}
 <plugin>
   <groupId>org.apache.maven.plugins</groupId>
   <artifactId>maven-surefire-plugin</artifactId>
   <configuration>
-    <argLine>${dd.java.agent.arg} -Ddd.prioritization.type=ENSURE_TRACE -Ddd.jmxfetch.enabled=false -Ddd.integrations.enabled=false -Ddd.integration.junit.enabled=true</argLine>
+    <argLine>${dd.java.agent.arg} -Ddd.service=my-java-app Ddd.prioritization.type=ENSURE_TRACE -Ddd.jmxfetch.enabled=false -Ddd.integrations.enabled=false -Ddd.integration.junit.enabled=true -Ddd.integration.testng.enabled=true</argLine>
```
drodriguezhdez, line 77:
```suggestion
    <argLine>${dd.java.agent.arg} -Ddd.service=my-java-app -Ddd.prioritization.type=ENSURE_TRACE -Ddd.jmxfetch.enabled=false -Ddd.integrations.enabled=false -Ddd.integration.junit.enabled=true -Ddd.integration.testng.enabled=true</argLine>
```

content/en/continuous_integration/setup_pipelines/jenkins.md
```diff
@@ -10,6 +10,11 @@ further_reading:
       text: "Troubleshooting CI"
 ---
 
+## Compatibility
+
+Supported Jenkins versions:
+* Jenkins >= 2.164.1
+
```
drodriguezhdez, line 17: Probably, we want to update Jenkins instructions with this PR (https://github.com/DataDog/documentation/pull/11007)

drodriguezhdez (commented):
Small typo in Java docs. Finally, you can have merge conflicts with Jenkins instructions as we're updating it upfront in this PR (https://github.com/DataDog/documentation/pull/11007)

content/en/continuous_integration/setup_tests/javascript.md
fermayo: Yes, as in the case of running this in a SaaS provider, they would have to change them to point to the agent running in a sidecar container instead of the default `localhost`

content/en/continuous_integration/setup_tests/javascript.md
```diff
@@ -70,56 +59,69 @@ module.exports = {
...
-{{< code-block lang="javascript" >}}
-require('dd-trace').init({
-  enabled: !!process.env.CI // the environment variable to use depends on the CI provider
-})
-{{< /code-block >}}
+`service`
+: Name of the service or library under test.<br/>
+**Environment variable**: `DD_SERVICE`<br/>
+**Default**: (test framework name)<br/>
+**Example**: `my-java-app`
```
juan-fernandez, line 100:
```suggestion
**Example**: `my-ui`
```

juan-fernandez (approved):

content/en/continuous_integration/setup_tests/junit_upload.md
fermayo: Maybe "additional" is not the best title for the section, but I think it's good to include service and environment in this list even if they are explained earlier, just for completeness. Maybe we can use "Advanced configuration" or something like that.

content/en/continuous_integration/setup_tests/dotnet.md
fermayo: Ok, I'll remove it for now and we can readd it later if needed

content/en/continuous_integration/setup_pipelines/jenkins.md
fermayo: Yes, once that's merged I'll rebase this PR

content/en/continuous_integration/setup_tests/ruby.md
```diff
@@ -44,76 +41,92 @@ To install the Ruby tracer:
...
+
+
+## Additional configuration settings
+
+The following is a list of the most important configuration settings that can be used with the tracer, either in code by using a `Datadog.configure` block, or using environment variables:
+
+`service`
+: Name of the service or library under test.<br/>
+**Environment variable**: `DD_SERVICE`<br/>
+**Default**: `"pytest"`<br/>
```
Yun-Kim, line 100: Does ruby also have pytest? Looks like this may be a wrong default service for ruby.

fermayo: Good catch, my bad

content/en/continuous_integration/setup_tests/java.md
```diff
@@ -107,91 +97,62 @@ Configure the [Maven Surefire Plugin][3] and/or the [Maven Failsafe Plugin][4] t
...
+    jvmArgs = ["-javaagent:${configurations.ddTracerAgent.asPath}", "-Ddd.service=my-java-app", "-Ddd.prioritization.type=ENSURE_TRACE", "-Ddd.jmxfetch.enabled=false", "-Ddd.integrations.enabled=false", "-Ddd.integration.junit.enabled=true", "-Ddd.integration.testng.enabled=true"]
 }
 {{< /code-block >}}
 
-Run your tests as you normally do, for example:
+Run your tests as you normally do, specifying the environment where test are being run (e.g. `local` when running tests on a developer workstation, or `ci` when running them on a CI provider) in the `DD_ENV` environment variable. For example:
 
 {{< code-block lang="bash" >}}
-./gradlew cleanTest test --rerun-tasks
+DD_ENV=ci ./gradlew cleanTest test --rerun-tasks
```
drodriguezhdez, line 119: Probably we should add here some property discriminant to be used to activate CI App. Now, we're activating it by default for Gradle. For Maven we're using profiles, though.

content/en/continuous_integration/setup_tests/python.md
```diff
@@ -11,25 +11,17 @@ further_reading:
...
-* CircleCI
-* GitHub Actions
-* GitLab
-* Jenkins
-* TravisCI
+Supported Python interpreters:
+* Python >= 2.7 and >= 3.5
+
+Supported test frameworks:
+* pytest >= 4.0.0
```
Yun-Kim, line 18: Our pytest integration is compatible with:
- Python 2: pytest<=4.6 (4.6 is last to support python 2)
- Python 3: pytest >= 3.0.0

content/en/continuous_integration/setup_tests/python.md
```diff
@@ -39,36 +31,53 @@ pip install "ddtrace>=0.50.0rc4"
...
 
-{{< code-block lang="ini" >}}
-[pytest]
-ddtrace = 1
-{{< /code-block >}}
+The following is a list of the most important configuration settings that can be used with the tracer, either in code or using environment variables:
+
+`ddtrace.config.service`
+: Name of the service or library under test.<br/>
+**Environment variable**: `DD_SERVICE`<br/>
```
Yun-Kim, line 48: It seems like the Python tracer is unique among other tracers to include not only a global configuration setting `DD_SERVICE` but also integration specific configurations including `DD_PYTEST_SERVICE`. 
Our preferred way to set the service name is through the integration specific `DD_PYTEST_SERVICE`, but `DD_SERVICE` should technically be fine. I'll get back to you with more feedback once I discuss with the Python team tomorrow.

content/en/continuous_integration/setup_tests/python.md
```diff
@@ -39,36 +31,53 @@ pip install "ddtrace>=0.50.0rc4"
...
-**Default**: `"pytest"`
+`port`
+: The Datadog Agent trace collection port.<br/>
+**Environment variable**: `DD_TRACE_AGENT_PORT`<br/>
+**Default**: `8126`
 
-`ddtrace.config.pytest["operation_name"]`
-: The operation name reported by default for pytest traces.<br/>
-**Environment variable**: `DD_PYTEST_OPERATION_NAME`<br/>
-**Default**: `"pytest.test"`
```
Yun-Kim, line 67: Can I ask why this was removed?

content/en/continuous_integration/setup_tests/python.md
fermayo: Ok - I changed it to be consistent with the other tracers, as consistency will mean an easier in-app onboarding flow where we could reuse more between languages. If we need to do something different here is fine, but we should try to have a common way to configure any language.

content/en/continuous_integration/setup_tests/python.md
fermayo: No other language exposes this (at least so prominently), and for testing I cannot think of a reason to change it. We could still link to the tracer's docs where this is explained, as an advanced setup. Do you think this is something users will change often?

Yun-Kim: I don't see this being crucial to put on the CI docs, so no problem leaving this out. I don't see this as a setting users can change often, but I do agree it would be nice to link to the tracer docs as part of advanced setup.

content/en/continuous_integration/setup_tests/ruby.md
```diff
@@ -12,29 +12,26 @@ further_reading:
...
 
 [Install the Datadog Agent to collect tests data][1].
 
-## Installing tracing
+## Installing the Ruby tracer
 
 To install the Ruby tracer:
 
-1. Add the `ddtrace` gem (version 0.50.0 or above) to your `Gemfile`:
+1. Add the `ddtrace` gem to your `Gemfile`:
```
marcotc, line 31: Given the changes are available in the very latest version of `ddtrace`, wouldn't it be prudent to keep the minimum version here?

Unless we are having the pattern of not having the version number for any of our APM libraries, as long as the latest one supports the feature, then I think adding the minimum version is helpful.

fermayo: I've removed it from the step description but it's still on the `Gemfile` code below. I can readd it to the description though.

content/en/continuous_integration/setup_pipelines/gitlab.md
```diff
@@ -10,6 +10,17 @@ further_reading:
...
+* GitLab >= 14.1
+
+We also support the following versions with additional configuration:
+* GitLab >= 13.7.0, by enabling the `datadog_ci_integration` feature flag
+
+GitLab versions prior to 13.7.0 are supported using the [webhooks integration](#integrating-through-webhooks).
+
+## Enabling the Datadog integration
 
 To enable traces for your GitLab CI pipelines a [native integration][1] is available and hidden under a [feature flag][2] since release `13.7.0`.
```
edrevo, line 25: Any reason the "native integration" link doesn't take you to the Datadog specific page (https://docs.gitlab.com/ee/integration/datadog.html) vs the general one (https://docs.gitlab.com/ee/integration/)?

edrevo: More feedback, which doesn't relate to this PR specifically but I think is valuable either way:

This page mentions that the integration needs to be enabled per project. The GitLab page for the Datadog integration says it can be enabled per project or globally, but there aren't any instruction anywhere on how to enable it globally:

> You can also activate the integration for the entire GitLab instance.

Maybe it is obvious from the GitLab UI (I haven't used it), but it might be worth mentioning how.

content/en/continuous_integration/setup_tests/java.md
```diff
@@ -37,9 +25,9 @@ Supported CI providers:
 
 ### Maven
```
edrevo, line 26: Would it be possible to add Maven Central to the docs or would that not be allowed because it uses an external service? It allows users to know what the latest version is without navigating away from the docs page, which is nice.

content/en/continuous_integration/setup_tests/javascript.md
```diff
@@ -13,22 +13,11 @@ further_reading:
...
-* Azure Pipelines
-* BitBucket
-* BuildKite
-* CircleCI
-* GitHub Actions
-* GitLab
-* Jenkins
-* TravisCI
+* Mocha >= 5.2.0
+  * Mocha >= 9.0.0 has [partial support](#known-limitations).
```
edrevo, line 20: nit: Does this mean mean that mocha 8 is not supported? If mocha 8 is supported, then it seems like different pages use different conventions to talk about versioning. For example, the .NET page states that .NET Core 2.1 and .NET Core 3 are supported.

content/en/continuous_integration/setup_tests/python.md
```diff
@@ -11,25 +11,18 @@ further_reading:
...
+  * pytest < 5 when using Python 2
 
 ## Prerequisites
 
 [Install the Datadog Agent to collect tests data][1].
 
-## Installing tracing
+## Installing the Python tracer
 
 Install the Python tracer by running:
```
edrevo, line 27: ℹ️  rc5 exists now. Not sure if we want people using that version instead

content/en/continuous_integration/setup_tests/python.md
```diff
@@ -39,36 +32,53 @@ pip install "ddtrace>=0.50.0rc4"
...
+**Environment variable**: `DD_ENV`<br/>
+**Default**: `(empty)`<br/>
+**Examples**: `local`, `ci`
+
+The following configuration settings can be passed in as parameters to `tracer.configure()`, or using environment variables:
+
+`enabled`
+: Setting this to `false` completely disables the instrumentation.<br/>
+**Environment variable**: `DD_TRACE_ENABLED`<br/>
+**Default**: `true`
```
edrevo, line 64: This seems like a weird option. Does removing `--ddtrace` do the same thing?

content/en/continuous_integration/setup_tests/junit_upload.md
```diff
@@ -10,87 +10,87 @@ further_reading:
...
 {{< /code-block >}}
 
-For example:
+You will need to specify a valid [Datadog API key][3] in the `DATADOG_API_KEY` environment variable, and the environment where tests were run (e.g. `local` when uploading results from a developer workstation, or `ci` when uploading them from a CI provider) in the `DD_ENV` environment variable. For example:
 
 {{< code-block lang="bash" >}}
-datadog-ci junit upload --service my-service \
-  --tags key1:value1 --tags key2:value2 \
-  unit-tests/junit-reports acceptance-tests/junit-reports e2e-tests/single-report.xml
+DD_ENV=ci DATADOG_API_KEY=<api_key> datadog-ci junit upload --service my-api-service unit-tests/junit-reports e2e-tests/single-report.xml
```
edrevo, line 38: This command line is too long for a 1080p screen. Maybe it would be better to split into different lines?

content/en/continuous_integration/setup_tests/junit_upload.md
```diff
@@ -10,87 +10,87 @@ further_reading:
...
+## Additional configuration settings
+
+This is the full list of options available when using the `datadog-ci junit upload` command:
+
 `--service`
-: The name of the service you're uploading JUnit XML reports for.<br>
-**Default**: the `DD_SERVICE` environment variable value.
+: Name of the service or library under test.<br/>
+**Environment variable**: `DD_SERVICE`<br/>
+**Default**: (none)<br/>
```
edrevo, line 48: Not sure if the default value is the literal `"(none)"` or if it is the literal `""`. If it really has no value, that surprises me because every other tracer does provide a value (e.g. `unnamed-java-app` in the Java tracer).

I was assuming that all the tracers were providing non-empty values because the UX would be bad with an empty value for the service. Is that the case? If that's not the case, maybe it is better to have an empty value everywhere?

content/en/continuous_integration/setup_pipelines/gitlab.md
```diff
@@ -10,6 +10,17 @@ further_reading:
       text: "Troubleshooting CI"
 ---
 
+## Compatibility
```
edrevo, line 13: There's a note in this page that says:

> Note: Only datadoghq.com is currently supported.

2 things:

1. Maybe it would be better to move that note to the Compatibility section?
2. Is that really a limitation? The JUnit uploader doc says everything is supported (`Possible values: datadoghq.com, datadoghq.eu or us3.datadoghq.com`)

edrevo (commented):
Here's a comment I couldn't fit in the page because it was unchanged:

The Jenkins page says:

> Global job tags (globalJobTags): A comma-separated list of regexes to match a job and a list of tags to apply to that job. Tags can include environment variables that are defined in the Jenkins controller instance. Tags can reference match groups in the regex using the $ symbol: (.*?)_job_(*?)_release:someValue

I don't really understand this as a user. What syntax is the rexexp using? It says it is a "omma-separated list of regexes to match a job and a list of tags to apply to that job", but that sentence doesn't quite parse in my brain. Maybe explaining the effect of the example would help clarify.

content/en/continuous_integration/setup_tests/javascript.md
juan-fernandez: mocha 8 is supported. We support mocha >= 5.2.0 *but* from 9.0.0 on we "lose" _some_ support: we no longer instrument _within_ the test (when using .mjs test files), for example, http requests wouldn't appear in the trace view. 
Makes sense? Do you think there'd be a better way to express this?

content/en/continuous_integration/setup_tests/junit_upload.md
juan-fernandez: yeah good catch. The service name (whether via `--service` or `DD_SERVICE`) is mandatory when uploading junit results via `datadog-ci`. There shouldn't be a default

content/en/continuous_integration/setup_tests/python.md
Yun-Kim: Update from the Python team discussion: We agree it's more important to be consistent with the other tracers, and `DD_SERVICE` should be fine for CIApp customers.

content/en/continuous_integration/setup_tests/javascript.md
edrevo: I'm not sure what the best way to express this is, but it seems the doc isn't consistent between different pages. The other pages define all of the major versions they support whereas this one has the same syntax as the other ones but it ranges over major versions.

Maybe it would be best to adopt a standard syntax like the npm one? That would mean this page is correct, and the other ones would need to either state the minimum major version they support (and support anything above that) or switch to the `^3.0.0` syntax to indicate that they support everything in the 3.x.y range but not necessarily the 4.x.y range. Just an idea.

marcotc (approved):

content/en/continuous_integration/setup_pipelines/gitlab.md
fermayo: Because we are not deployed anywhere other than us1, we should really remove all mentions to this until we do support something else, and then only document the ones that we support

content/en/continuous_integration/setup_pipelines/gitlab.md
fermayo: I refactored a bit the instructions, because they were out of date. I added how to enable it at the instance level as well. Please have another look.

content/en/continuous_integration/setup_tests/python.md
fermayo: It doesn't come with anything that we need, so we can wait until the stable version is released to update this.

content/en/continuous_integration/setup_tests/python.md
fermayo: Yes, in the case of Python is weird, and I don't think it's worth having it for consistency. I'll remove it.

content/en/continuous_integration/setup_tests/junit_upload.md
fermayo: Fixed

content/en/continuous_integration/setup_tests/java.md
fermayo: Any suggestions to do that @drodriguezhdez ? It's the same problem as with JS and Ruby

content/en/continuous_integration/setup_tests/python.md
fermayo: Great, thanks for the update @Yun-Kim !

content/en/continuous_integration/setup_tests/java.md
```diff
@@ -107,91 +97,62 @@ Configure the [Maven Surefire Plugin][3] and/or the [Maven Failsafe Plugin][4] t
...
 ### Gradle
 
-Configure the `test` Gradle task by adding to the `jvmArgs` attribute the `-javaagent` argument targeting the Datadog Java tracer based on the `configurations.ddTracerAgent` property (use `-Ddd.integration.testng.enabled=true` if your testing framework is TestNG rather than JUnit):
+Configure the `test` Gradle task by adding to the `jvmArgs` attribute the `-javaagent` argument targeting the Datadog Java tracer based on the `configurations.ddTracerAgent` property, specifiying the name of the service or library under test with the `-Ddd.service` property:
 
-{{< code-block lang="groovy" >}}
+{{< code-block lang="groovy" filename="build.gradle" >}}
 test {
-    jvmArgs = ["-javaagent:${configurations.ddTracerAgent.asPath}", "-Ddd.prioritization.type=ENSURE_TRACE", "-Ddd.jmxfetch.enabled=false", "-Ddd.integrations.enabled=false", "-Ddd.integration.junit.enabled=true"]
+    jvmArgs = ["-javaagent:${configurations.ddTracerAgent.asPath}", "-Ddd.service=my-java-app", "-Ddd.prioritization.type=ENSURE_TRACE", "-Ddd.jmxfetch.enabled=false", "-Ddd.integrations.enabled=false", "-Ddd.integration.junit.enabled=true", "-Ddd.integration.testng.enabled=true"]
```
drodriguezhdez, line 112:
```suggestion
    if(project.hasProperty("ci-app")) {
        jvmArgs = ["-javaagent:${configurations.ddTracerAgent.asPath}", "-Ddd.service=my-java-app", "-Ddd.prioritization.type=ENSURE_TRACE", "-Ddd.jmxfetch.enabled=false", "-Ddd.integrations.enabled=false", "-Ddd.integration.junit.enabled=true", "-Ddd.integration.testng.enabled=true"]
        }
```

content/en/continuous_integration/setup_tests/java.md
drodriguezhdez:
```suggestion
DD_ENV=ci ./gradlew cleanTest test --rerun-tasks -Pci-app
```

Using this approach, the command would be pretty similar to the Maven. 
Notice the suggestion in the code from above.

content/en/continuous_integration/setup_tests/java.md
```diff
@@ -107,91 +97,62 @@ Configure the [Maven Surefire Plugin][3] and/or the [Maven Failsafe Plugin][4] t
 </plugin>
 {{< /code-block >}}
 
-Run your tests using the `ci-app` profile, for example:
+Run your tests as you normally do, specifying the environment where tests are being run (e.g. `local` when running tests on a developer workstation, or `ci` when running them on a CI provider) in the `DD_ENV` environment variable. For example:
 
 {{< code-block lang="bash" >}}
-mvn clean verify -P ci-app
+DD_ENV=ci mvn clean verify -P ci-app
```
drodriguezhdez, line 103: Nit: At this point, I don't know if we should use `ci-visibility` instead of `ci-app`.

content/en/continuous_integration/setup_pipelines/gitlab.md
```diff
@@ -26,24 +35,32 @@ sudo gitlab-rails runner "Feature.enable(:datadog_ci_integration)"
...
+* **Active** enables the integration.
+* **Datadog site** specifies which [Datadog site][11] to send data to. Please note that only `datadoghq.com` is currently supported.
+* **API URL** (optional) allows overriding the API URL used for sending data directly, only used in advanced scenarios.
+* **API key** specifies which API key to use when sending data. You can generate one in the [APIs tab][6] of the Integrations section on Datadog.
+* **Service** (optional) specifies which service name to attach to each span generated by the integration. Use this to differentiate between GitLab instances.
+* **Env** (optional) specifies which environment (`env` tag) to attach to each span generated by the integration. Use this to differentiate between groups of GitLab instances (i.e. staging vs production).
 
-Test the integration with the *Test settings* button. After it's successful, click *Save changes* to finish the integration set up.
+You can test the integration with the **Test settings** button. After it’s successful, click **Save changes** to finish the integration set up.
```
AdrianLC, line 63: The Test button is only available at project-level config. Maybe we should reflect that

content/en/continuous_integration/setup_pipelines/gitlab.md
```diff
@@ -26,24 +35,32 @@ sudo gitlab-rails runner "Feature.enable(:datadog_ci_integration)"
...
+Fill in the integration configuration settings:
 
-There are a couple of optional parameters:
-- _service_: Sets the service name for all traces for this GitLab instance. Defaults to `gitlab-ci`. Useful if you're using more than one GitLab instance.
-- _env_: Sets the environment tag for all traces for this GitLab instance. Useful in advanced setups where grouping GitLab instances per environment is required.
+* **Active** enables the integration.
+* **Datadog site** specifies which [Datadog site][11] to send data to. Please note that only `datadoghq.com` is currently supported.
+* **API URL** (optional) allows overriding the API URL used for sending data directly, only used in advanced scenarios.
+* **API key** specifies which API key to use when sending data. You can generate one in the [APIs tab][6] of the Integrations section on Datadog.
+* **Service** (optional) specifies which service name to attach to each span generated by the integration. Use this to differentiate between GitLab instances.
```
AdrianLC, line 60: Do we remove "Defaults to `gitlab-ci`"?

AdrianLC (approved):

content/en/continuous_integration/setup_pipelines/gitlab.md
edrevo:
```suggestion
You can test the integration with the **Test settings** button. After it's successful, click **Save changes** to finish the integration set up.
```

content/en/continuous_integration/setup_tests/javascript.md
juan-fernandez: nvm, it does not fail. It logs an error, but only if you've added the debug flag.

content/en/continuous_integration/setup_tests/java.md
fermayo: Changed to `dd-civisibility`

content/en/continuous_integration/setup_tests/java.md
fermayo: I'll add it and see what the docs team says on the review

content/en/continuous_integration/setup_tests/junit_upload.md
fermayo: I have rephrased this a bit.

content/en/continuous_integration/setup_tests/dotnet.md
```diff
@@ -71,56 +71,24 @@ dd-trace --help
...
-: Environment name for unified service tagging.<br/>
-**Environment variable**: `DD_ENV`<br/>
-**Default**: `(empty)`
-
 `--dd-service`
-: Service name for unified service tagging. <br/>
+: Name of the service or library under test.<br/>
 **Environment variable**: `DD_SERVICE`<br/>
-**Default**: `(ProcessName)` 
+**Default**: `(ProcessName)`<br/>
```
tonyredondo, line 77: In .NET the default is the repo name (https://github.com/DataDog/dd-trace-dotnet/blob/master/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/Common.cs#L43)

tonyredondo (changes requested):

fermayo:
> Here's a comment I couldn't fit in the page because it was unchanged:
> 
> The Jenkins page says:
> 
> > Global job tags (globalJobTags): A comma-separated list of regexes to match a job and a list of tags to apply to that job. Tags can include environment variables that are defined in the Jenkins controller instance. Tags can reference match groups in the regex using the $ symbol: (._?)_job_(_?)_release:someValue
> 
> I don't really understand this as a user. What syntax is the rexexp using? It says it is a "omma-separated list of regexes to match a job and a list of tags to apply to that job", but that sentence doesn't quite parse in my brain. Maybe explaining the effect of the example would help clarify.

@edrevo I checked with @drodriguezhdez and the example was wrong - I updated it. I think the example now explains the description properly.

content/en/continuous_integration/setup_tests/dotnet.md
```diff
@@ -13,55 +13,55 @@ further_reading:
...
-To instrument your test suite, prefix your test command with `dd-trace`. For example:
+You will also need to configure the tracer to use Call Target instrumentation by setting `DD_TRACE_CALLTARGET_ENABLED=true`, which is not the default value.
+
+For example:
 
 {{< code-block lang="bash" >}}
-dd-trace dotnet test
+DD_TRACE_CALLTARGET_ENABLED=true dd-trace \
+  --dd-service=my-dotnet-app --dd-env=ci -- \
+  dotnet test
```
tonyredondo, line 59: This is a common thing in bash but not supported in the default Windows command line (unless customer is using WSL2). Due the majority of users of .NET uses Windows, maybe we should add a windows compatible version of this script. 
See: 
- https://superuser.com/questions/424001/launch-windows-program-with-custom-environment-variable
- https://stackoverflow.com/a/34769146

content/en/continuous_integration/setup_tests/java.md
```diff
@@ -107,91 +97,64 @@ Configure the [Maven Surefire Plugin][3] and/or the [Maven Failsafe Plugin][4] t
 </plugin>
 {{< /code-block >}}
 
-Run your tests using the `ci-app` profile, for example:
+Run your tests as you normally do, specifying the environment where tests are being run (for example, `local` when running tests on a developer workstation, or `ci` when running them on a CI provider) in the `DD_ENV` environment variable. For example:
 
 {{< code-block lang="bash" >}}
-mvn clean verify -P ci-app
+DD_ENV=ci mvn clean verify -P dd-civisibility
```
drodriguezhdez, line 103:
```suggestion
DD_ENV=ci mvn clean verify -Pdd-civisibility
```
nit

drodriguezhdez (approved):

content/en/continuous_integration/setup_tests/dotnet.md
tonyredondo: For **CMD**:
```Batchfile
SET DD_TRACE_CALLTARGET_ENABLED=true && dd-trace --dd-service=my-dotnet-app --dd-env=ci -- dotnet test
```
For **PowerShell**:
```PowerShell
$env:DD_TRACE_CALLTARGET_ENABLED="true"; dd-trace --dd-service=my-dotnet-app --dd-env=ci -- dotnet test
```

content/en/continuous_integration/setup_tests/javascript.md
juan-fernandez: that makes sense. I've created a task to track this as this PR is already huge

juan-fernandez (approved):

content/en/continuous_integration/setup_pipelines/jenkins.md
```diff
@@ -145,9 +150,15 @@ You can configure the Jenkins Plugin to send custom tags in all pipeline traces:
...
+**Global tags**
+: A comma-separated list of tags to apply to all metrics, traces, events, and service checks. Tags can include environment variables that are defined in the Jenkins controller instance<br/>
+**Environment variable**: `DATADOG_JENKINS_PLUGIN_GLOBAL_TAGS`
+**Example**: `key1:value1,key2:${SOME_ENVVAR},${OTHER_ENVVAR}:value3`
 
-**Global job tags** (`globalJobTags`): A comma-separated list of regexes to match a job and a list of tags to apply to that job. Tags can include environment variables that are defined in the Jenkins controller instance. Tags can reference match groups in the regex using the `$` symbol: `(.*?)_job_(*?)_release:someValue`
+**Global job tags**
+: A comma-separated list of regexes to match a job and a list of tags to apply to that job. Tags can include environment variables that are defined in the Jenkins controller instance. Tags can reference match groups in the regex using the `$` symbol.
+**Environment variable**: `DATADOG_JENKINS_PLUGIN_GLOBAL_JOB_TAGS`
+**Example**: `(.*?)_job_(*?)_release, owner:$1, release_env:$2, optional:Tag3`
```
edrevo, line 161: I still think there's room for improvement. How would I go about having multiple regex values or is there only one? If there's only one, the text is a bit misleading.

Also, is `(*?)` valid regex syntax or should it be `(.*?)`? It might be good to specify what the regex syntax is specifically (Java probably? Or maybe it's PCRE...).

Anyway, this is a low-pri feedback.

edrevo (approved):

content/en/continuous_integration/setup_pipelines/jenkins.md
fermayo: @drodriguezhdez do you know if it's possible to add more than one regex?

jtappa:
Since this is a pretty big PR, I added a card on our board (DOCS-2209) so that we have someone take a look for editorial review. It should be assigned during our next planning session. :)

tonyredondo (approved):

content/en/continuous_integration/setup_tests/java.md
```diff
@@ -12,36 +12,25 @@ further_reading:
...
 
-### Maven
+{{< tabs >}}
+{{% tab "Maven" %}}
 
-Add a new Maven profile in your root `pom.xml` configuring the Datadog Java tracer dependency and the `javaagent` arg property, replacing `$VERSION` with the latest version of the tracer accessible from the [Maven Repository][2]: 
+Add a new Maven profile in your root `pom.xml` configuring the Datadog Java tracer dependency and the `javaagent` arg property, replacing `$VERSION` with the latest version of the tracer accessible from the [Maven Repository][2] (without the preceding `v`): ![Maven Central][7]
 
-{{< code-block lang="xml" >}}
+{{< code-block lang="xml" filename="pom.xml" >}}
```
edrevo, line 31: ⚠️  this is not rendering correctly

content/en/continuous_integration/setup_tests/java.md
```diff
@@ -12,36 +12,25 @@ further_reading:
...
 [Install the Datadog Agent to collect tests data][1].
 
 ## Installing the Java tracer
 
-### Maven
+{{< tabs >}}
+{{% tab "Maven" %}}
 
-Add a new Maven profile in your root `pom.xml` configuring the Datadog Java tracer dependency and the `javaagent` arg property, replacing `$VERSION` with the latest version of the tracer accessible from the [Maven Repository][2]: 
+Add a new Maven profile in your root `pom.xml` configuring the Datadog Java tracer dependency and the `javaagent` arg property, replacing `$VERSION` with the latest version of the tracer accessible from the [Maven Repository][2] (without the preceding `v`): ![Maven Central][7]
```
edrevo, line 29: ⚠️  this is not rendering correctly

edrevo (dismissed):

content/en/continuous_integration/setup_tests/java.md
edrevo: Link rendering seems to be broken in general for this page

fermayo: Good catch - it might have been when adding the tabs, fixing.

fermayo: @edrevo should be fixed now

content/en/continuous_integration/setup_pipelines/jenkins.md
drodriguezhdez: > @drodriguezhdez do you know if it's possible to add more than one regex?
Yes, it's possible using `\r\n` as the separator. This is oriented to be configured using the Jenkins UI.

edrevo (approved):

content/en/continuous_integration/setup_tests/junit_upload.md
```diff
@@ -10,87 +10,90 @@ further_reading:
...
-* CircleCI
-* GitHub Actions
-* GitLab
-* Jenkins
-* TravisCI
+JUnit test report files are XML files that contain test execution information, such as test and suite names, pass/fail status, duration, and sometimes error logs. Although it was introduced by the [JUnit][4] testing framework, lots of other popular frameworks are able to output results using this format.
+
+As an alternative to instrumenting your tests natively using Datadog tracers, which is the recommended option as it provides the most comprehensive test results, you can also upload JUnit XML test reports.
+
+Test results imported via JUnit XML reports will appear alongside tests reported via tracers. However, there are some limitations when using this method, such as the lack of distributed traces on integration tests or structured stacktraces. For this reason, only use this method if there is no native support for the language or testing framework being used.
```
kayayarai, line 17:
```suggestion
Test results imported via JUnit XML reports appear alongside test data reported by tracers. However, there are some limitations when using this method, such as the lack of distributed traces on integration tests or structured stacktraces. For this reason, only use this method if there is no native support for the language or testing framework being used.
```

content/en/continuous_integration/setup_tests/junit_upload.md
```diff
@@ -10,87 +10,90 @@ further_reading:
...
-To upload your JUnit XML test reports to Datadog:
+To upload your JUnit XML test reports to Datadog, run the following command, specifying the name of the service or library that was tested using the `--service` parameter, and one or more file paths to either the XML report files directly or directories containing them:
 
 {{< code-block lang="bash" >}}
-datadog-ci junit upload [--service] [--tags] [--max-concurrency] [--dry-run] <paths>
+datadog-ci junit upload --service <service_name> <path> [<path> ...]
 {{< /code-block >}}
 
-For example:
+You will need to specify a valid [Datadog API key][3] in the `DATADOG_API_KEY` environment variable, and the environment where tests were run for example, `local` when uploading results from a developer workstation, or `ci` when uploading them from a CI provider) in the `DD_ENV` environment variable. For example:
```
kayayarai, line 35:
```suggestion
Specify a valid [Datadog API key][3] in the `DATADOG_API_KEY` environment variable, and the environment where tests were run for example, `local` when uploading results from a developer workstation, or `ci` when uploading them from a CI provider) in the `DD_ENV` environment variable. For example:
```

content/en/continuous_integration/setup_tests/junit_upload.md
```diff
@@ -10,87 +10,90 @@ further_reading:
...
+: The Datadog site to upload results to.<br/>
+**Default**: `datadoghq.com`<br/>
+**Possible values**: `datadoghq.com`, `datadoghq.eu` or `us3.datadoghq.com`
+-->
 
-## Optional dependencies
+## Collecting repository and commit metadata
 
-- [`git`][1] is used for extracting repository metadata.
+The Datadog CI CLI will try to extract git repository and commit metadata from CI provider environment variables and from the local `.git` directory and attach it to test executions. In order to read this directory, the [`git`][1] binary is required.
```
kayayarai, line 90:
```suggestion
The Datadog CI CLI tries to extract git repository and commit metadata from CI provider environment variables and from the local `.git` directory and attach it to test executions. In order to read this directory, the [`git`][1] binary is required.
```

content/en/continuous_integration/setup_tests/swift.md
```diff
@@ -13,369 +13,407 @@ further_reading:
...
+{{% /tab %}}
+{{< /tabs >}}
+
 
 ## Instrumenting your tests
 
 ### Configuring Datadog
 
-To enable testing instrumentation, add the following environment variables to your test target. You must also select your main target in the `Expand variables based on:`  combo or `Target for Variable Expansion` if using test plans.
+To enable testing instrumentation, add the following environment variables to your test target. You must also select your main target in the `Expand variables based on:`  combo or `Target for Variable Expansion` if using test plans:
```
kayayarai, line 65:
```suggestion
To enable testing instrumentation, add the following environment variables to your test target. You must also select your main target in `Expand variables based on` or `Target for Variable Expansion` if using test plans:
```

content/en/continuous_integration/setup_tests/swift.md
```diff
@@ -13,369 +13,407 @@ further_reading:
...
 
 For UITests, environment variables need to be set only in the test target, because the framework automatically injects these values to the application.
 
-| Environment variable     | Value                       |
-|--------------------------|-----------------------------|
-| `DD_TEST_RUNNER`           | `true`                         |
-| `DATADOG_CLIENT_TOKEN`     | Your current Datadog client token  |
-| `DD_ENDPOINT`            | The server endpoint of your Datadog Account: `us` (default), `us3`, `eu`, `gov`             |
-| `SRCROOT`                  | `$(SRCROOT)`                  |
+You should set all these variables in your test target:
```
kayayarai, line 71:
```suggestion
Set all these variables in your test target:
```

content/en/continuous_integration/setup_tests/swift.md
```diff
@@ -13,369 +13,407 @@ further_reading:
...
 
+`DD_TEST_RUNNER`
+: Enables or disables the instrumentation of tests. You should set this value to `$(DD_TEST_RUNNER)` so you can enable and disable test instrumentation with a environment variable defined outside of the test process (for example, in the CI build).<br/>
```
kayayarai, line 74:
```suggestion
: Enables or disables the instrumentation of tests. Set this value to `$(DD_TEST_RUNNER)` so you can enable and disable test instrumentation with a environment variable defined outside of the test process (for example, in the CI build).<br/>
```

content/en/continuous_integration/setup_tests/swift.md
```diff
@@ -13,369 +13,407 @@ further_reading:
...
+**Default**: `(empty)`<br/>
+**Example**: `pub0zxxxyyyxxxyyxxxzzxxyyxxxyyy`
 
-| Environment variable     | Value                       |
-|--------------------------|-----------------------------|
-|`DD_ENV `                   | The environment you want to report              |
-|`DD_SERVICE`                | The name of the service you want to report      |
+`DD_SERVICE`
+: Name of the service or library under test.<br/>
+**Default**: `(Repository name)`<br/>
```
kayayarai, line 86:
```suggestion
**Default**: The repository name<br/>
```

content/en/continuous_integration/setup_tests/swift.md
```diff
@@ -13,369 +13,407 @@ further_reading:
...
+**Example**: `my-ios-app`
 
+`DD_ENV`
+: Name of the environment where tests are being run. You should set this value to `$(DD_ENV)` so you can use an environment variable at runtime for setting it.<br/>
```
kayayarai, line 90:
```suggestion
: Name of the environment where tests are being run. Set this value to `$(DD_ENV)` so you can use an environment variable at runtime for setting it.<br/>
```

content/en/continuous_integration/setup_tests/swift.md
```diff
@@ -13,369 +13,407 @@ further_reading:
...
+**Default**: `none`<br/>
+**Recommended**: `$(DD_ENV)`<br/>
+**Examples**: `ci`, `local`
 
-### Configuring your CI service
+`SRCROOT`
+: This is the path to the project SRCROOT environment variable. The recommended approach is to use `$(SRCROOT)` for the value, as this is automatically set by Xcode.<br/>
```
kayayarai, line 96:
```suggestion
: The path to the project SRCROOT environment variable. Use `$(SRCROOT)` for the value, because it is automatically set by Xcode.<br/>
```

content/en/continuous_integration/setup_tests/swift.md
```diff
@@ -13,369 +13,407 @@ further_reading:
...
 
-| Environment variable     | Value                       |
-|--------------------------|-----------------------------|
-| `GIT_COMMIT`             | `$(GIT_COMMIT)`             |
-| `GIT_URL`                | `$(GIT_URL)`                |
-| `GIT_BRANCH`             | `$(GIT_BRANCH)`             |
+### Running tests
 
-#### CircleCI
+After installation, you can run your tests as you normally do, for example using the `xcodebuild test` command. Tests, network requests, and application logs will be instrumented automatically. You can pass your environment variables when running your tests in the CI, e.g:
```
kayayarai, line 116:
```suggestion
After installation, run your tests as you normally do, for example using the `xcodebuild test` command. Tests, network requests, and application logs are instrumented automatically. Pass your environment variables when running your tests in the CI, for example:
```

kayayarai (changes requested):
Just nits, I'll commit these myself and merge

kayayarai (approved):

PR #11064, opened 2021-07-09
update docs to include ddtrace-run --status command

ZStriker19:
**DO NOT MERGE,** as this feature is not yet released.

The ddtrace-run --status command output helpful information for customers so they can see if their settings are working and that there are not connection issues between the tracer and trace-agent.

### What does this PR do?
Let's customer's know about a new flag that will inform them about their setup and configuration, as well as make suggestions in case certain settings have not yet been set, e.g service, env, version

### Motivation
Customer's often have issues figuring out how to setup tracing in the beginning. the --status flag should help them with this.


apigirl (approved):

apigirl:
merging this since the PR adding the functionality has been merged in!

PR #11104, opened 2021-07-14
update gitignore for rum ios changes

kayayarai:
splat was not working


kayayarai:
Disregard, yes it was.

PR #11059, opened 2021-07-09
Link documentation on enabling allocation profiler in enabling the profiler

luhenry:
### What does this PR do?

As we are disabling the allocation profiler by default in Java, we want
to make it as clear as possible for users on how to enable it back.

### Preview

https://docs-staging.datadoghq.com/<BRANCH_NAME>/<PATH>

### Additional Notes

---

### Reviewer checklist
- [ ] Review the changed files.
- [ ] Review the URLs listed in the [Preview](#preview) section.
- [ ] Check images for PII
- [ ] Review any mentions of "Contact Datadog support" for internal support documentation.

PR #11144, opened 2021-07-19
Merge Java profiler setup steps

luhenry:
### What does this PR do?

It merges the optional Java Profiler setup steps to avoid a "do this, do that, ok you're all done, just kidding! go back and change the thing you did before!" effect.

### Motivation

See https://github.com/DataDog/documentation/pull/11059/files/82c1af76fa307dce9e13a1b059042a1d7a57e659#r668125247

### Preview

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/tracing/profiler/enabling.md
+++ b/content/en/tracing/profiler/enabling.md
@@ -44,7 +44,7 @@ The Datadog Profiler requires [JDK Flight Recorder][1]. The Datadog Profiler lib
     java -javaagent:dd-java-agent.jar -Ddd.profiling.enabled=true -XX:FlightRecorderOptions=stackdepth=256 -jar <YOUR_SERVICE>.jar <YOUR_SERVICE_FLAGS>
     ```
 
-4. **Optional** - Datadog strongly recommends that you specify the `service` and `version` as this gives you the ability to slice and dice your profiles across these dimensions. [Use environment variables](#environment-variables) to set the parameters.
+4. **Recommendation**: Specify the `service` and `version` so you can to slice and dice your profiles across these dimensions. [Use environment variables](#environment-variables) to set the parameters.
 
 5. After a minute or two, you can visualize your profiles on the [Datadog APM > Profiling page][5].
 
--- a/content/en/tracing/profiler/enabling.md
+++ b/content/en/tracing/profiler/enabling.md
@@ -44,7 +44,7 @@ The Datadog Profiler requires [JDK Flight Recorder][1]. The Datadog Profiler lib
     java -javaagent:dd-java-agent.jar -Ddd.profiling.enabled=true -XX:FlightRecorderOptions=stackdepth=256 -jar <YOUR_SERVICE>.jar <YOUR_SERVICE_FLAGS>
     ```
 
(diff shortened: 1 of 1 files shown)


jbachorik (approved):

content/en/tracing/profiler/enabling.md
```diff
@@ -44,8 +44,9 @@ The Datadog Profiler requires [JDK Flight Recorder][1]. The Datadog Profiler lib
     java -javaagent:dd-java-agent.jar -Ddd.profiling.enabled=true -XX:FlightRecorderOptions=stackdepth=256 -jar <YOUR_SERVICE>.jar <YOUR_SERVICE_FLAGS>
     ```
 
-4. After a minute or two, you can visualize your profiles on the [Datadog APM > Profiling page][5].
+4. **Optional** - Datadog strongly recommends that you specify the `service` and `version` as this gives you the ability to slice and dice your profiles across these dimensions. [Use environment variables](#environment-variables) to set the parameters.
```
kayayarai, line 47:
```suggestion
4. **Recommendation**: Specify the `service` and `version` so you can to slice and dice your profiles across these dimensions. [Use environment variables](#environment-variables) to set the parameters.
```

kayayarai (changes requested):
What do you think of this edit? It's less wordy, more to the point.

kayayarai (approved):

content/en/tracing/profiler/enabling.md
```diff
@@ -44,8 +44,9 @@ The Datadog Profiler requires [JDK Flight Recorder][1]. The Datadog Profiler lib
     java -javaagent:dd-java-agent.jar -Ddd.profiling.enabled=true -XX:FlightRecorderOptions=stackdepth=256 -jar <YOUR_SERVICE>.jar <YOUR_SERVICE_FLAGS>
     ```
 
-4. After a minute or two, you can visualize your profiles on the [Datadog APM > Profiling page][5].
+4. **Recommendation**: Specify the `service` and `version` so you can slice and dice your profiles across these dimensions. [Use environment variables](#environment-variables) to set the parameters.
```
dbenamydd, line 47: Thanks for improving this!

We should take one more step, which is to merge this with step 3 or otherwise structure it so when we show an example snippet of code, it matches our recommendations. Otherwise, again, it's sort of like:

User: "Ok this page has a bunch of words, skimming... Oh great, a code snippet, let me copy/paste that and fill in my values."
Us: "Gotcha! You didn't include service or version (or env) so your experience won't be as cool!"
User: "But, but, I used the code snippet you showed me!"
Lurker: <reacts with :itsatrap:>

:-)

Or maybe they notice step 4 in time but rather than seeing exactly how to use these tags, they have to jump elsewhere to learn.

You can see how we get this right on the other language tabs- https://docs.datadoghq.com/tracing/profiler/enabling/

PS We should also show env.
PPS You can set these with system properties or env vars. You can see in Python and Ruby how we show snippets for both versions, and in Go where we prefer one up top and show the alternate below.

luhenry: > PPS You can set these with system properties or env vars. You can see in Python and Ruby how we show snippets for both versions, and in Go where we prefer one up top and show the alternate below.
There is no way to do that in Java. The Profiler can only be enabled when launching the JVM, not once the JVM has already started.

PR #11239, opened 2021-07-28
HashMap should be of type object rather than string for tags.

surajtikoo:
The current hashmap which is defined in the code is a type <string, string> which is not defined in the class of DDLambda.
The custom metric method is breaking due to the type of map as <string, object> in DDLambda
 
  Map<String,Object> myTags = new HashMap<String, Object>();
        myTags.put("product", "coffee");
        myTags.put("product", "coffee");
        ddl.metric(
                "coffee_house.order_value", // Metric name
                12.45,                      // Metric value
                myTags);

### Preview

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII


agocs (approved):
Thanks again for making this change!

jtappa (approved):

PR #11245, opened 2021-07-28
Add known caveats for Python/Ruby DogStatsD clients in detailed info docs

sgnn7:
### What does this PR do?

- Adds important caveat note about multi-processing to Python and Ruby client detailed/high-throughput usage docs
- Improves a few notes/warnings on those pages to real note/warning stylings

### Motivation

Since Python and Ruby have somewhat big caveats for multi-process use,
those needed to be added to the places where detailed and
high-performance usage was described to prevent issues for users.

Incidental to this change a few notes/warnings had their styling
improved in those two pages.

### Preview

_Note: I don't know why but the links don't show the previews_

- https://docs-staging.datadoghq.com/sgnn7/dogstatsd-client-caveats/developers/dogstatsd
- https://docs-staging.datadoghq.com/sgnn7/dogstatsd-client-caveats/developers/dogstatsd/high_throughput

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/developers/dogstatsd/high_throughput.md
+++ b/content/en/developers/dogstatsd/high_throughput.md
@@ -296,7 +296,7 @@ If the packets sent are too small, the Datadog Agent packs several together to p
 The Datadog Agent performs most optimally if the DogStatsD clients send packets the size of the `dogstatsd_buffer_size`. The packets must not be larger than the buffer size, otherwise, the Agent won't be able to load them completely in the buffer and some of metrics will be malformed. Use the corresponding configuration field in your DogStatsD clients.
 
 <div class="alert alert-info">
-  <strong>Note for UDP</strong>: since UDP packets usually go through the Ethernet and IP layer, avoid IP packets fragmentation by limiting the packet size to a value lower than a single Ethernet frame on your network. Most of the time, IPv4 networks are configured with a MTU of 1500 bytes, so in this situation the packet size of sent packets should be limited to 1472.
+  <strong>Note for UDP</strong>: Because UDP packets usually go through the Ethernet and IP layer, you can avoid IP packets fragmentation by limiting the packet size to a value lower than a single Ethernet frame on your network. Most of the time, IPv4 networks are configured with a MTU of 1500 bytes, so in this situation the packet size of sent packets should be limited to 1472.
 </div>
 
 <div class="alert alert-info">
--- a/content/en/developers/dogstatsd/high_throughput.md
+++ b/content/en/developers/dogstatsd/high_throughput.md
@@ -300,7 +300,7 @@ The Datadog Agent performs most optimally if the DogStatsD clients send packets 
 </div>
 
 <div class="alert alert-info">
-  <strong>Note for UDS</strong>: for the best performances, UDS packet should have a size of 8192 bytes.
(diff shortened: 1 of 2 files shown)


content/en/developers/dogstatsd/high_throughput.md
```diff
@@ -287,9 +295,13 @@ If the packets sent are too small, the Datadog Agent packs several together to p
 
 The Datadog Agent performs most optimally if the DogStatsD clients send packets the size of the `dogstatsd_buffer_size`. The packets must not be larger than the buffer size, otherwise, the Agent won't be able to load them completely in the buffer and some of metrics will be malformed. Use the corresponding configuration field in your DogStatsD clients.
 
-Note for UDP: since UDP packets usually go through the Ethernet and IP layer, avoid IP packets fragmentation by limiting the packet size to a value lower than a single Ethernet frame on your network. Most of the time, IPv4 networks are configured with a MTU of 1500 bytes, so in this situation the packet size of sent packets should be limited to 1472.
+<div class="alert alert-info">
+  <strong>Note for UDP</strong>: since UDP packets usually go through the Ethernet and IP layer, avoid IP packets fragmentation by limiting the packet size to a value lower than a single Ethernet frame on your network. Most of the time, IPv4 networks are configured with a MTU of 1500 bytes, so in this situation the packet size of sent packets should be limited to 1472.
```
jtappa, line 299:
```suggestion
  <strong>Note for UDP</strong>: Because UDP packets usually go through the Ethernet and IP layer, you can avoid IP packets fragmentation by limiting the packet size to a value lower than a single Ethernet frame on your network. Most of the time, IPv4 networks are configured with a MTU of 1500 bytes, so in this situation the packet size of sent packets should be limited to 1472.
```

content/en/developers/dogstatsd/high_throughput.md
```diff
@@ -287,9 +295,13 @@ If the packets sent are too small, the Datadog Agent packs several together to p
...
+</div>
 
-Note for UDS: for the best performances, UDS packet should have a size of 8192 bytes.
+<div class="alert alert-info">
+  <strong>Note for UDS</strong>: for the best performances, UDS packet should have a size of 8192 bytes.
```
jtappa, line 303: Is this saying the UDS packet size should be exactly 8192 bytes, or is this a value that it should be less than?

jtappa (changes requested):

content/en/developers/dogstatsd/high_throughput.md
sgnn7: @jtappa It should be exactly 8192 bytes I think as it is the general max size that the server will accept per packet over UDS. While smaller values are allowed, they won't be as performant since it would take more request to send the same amount of data.

jtappa:
```suggestion
  <strong>Note for UDS</strong>: for the best performances, the UDS packet size should be 8192 bytes.
```

jtappa: So in that case, I'd suggest just small modification to this sentence ^^

jtappa (approved):

PR #11356, opened 2021-08-06
Remove performance_schema_max_sql_text_length from Mysql 5.6 Flags

alexandre-normand:
CloudSQL's UI doesn't have the `performance_schema_max_sql_text_length` flag because it doesn't exist in Mysql 5.6 so this adds a tab for the mysql flags and omits this one for 5.6. 

That said, we also have this config in other hosting options (like [self-hosted](https://docs-staging.datadoghq.com/kari/dbm-docs/database_monitoring/setup_mysql/selfhosted?tab=mysql80#configure-mysql-settings)) where it might just get ignored. Should we also have version tabs for those?


alexbarksdale (approved):
Confirmed `performance_schema_max_sql_text_length` does not exist on MySQL RDS v5.6

PR #11383, opened 2021-08-11
Update serverless installation docs for chalice (EN,FR)

jeffWelling:
Current docs describe using the environment variable to set the Datadog API key
but this goes against best practices because it often results in the API key
being added to code repositories unencrypted.
The datadog lambda python layer documentation itself calls out not to use
the environment variable to set the API key and to instead use the Secret Manager
ARN.
https://github.com/DataDog/datadog-lambda-python

This change updates the English and French documentation to reflect those best practices, but I must ask for help updating the Japanese instructions to match.

### What does this PR do?
This PR adjusts the English installation docs for Serverless (Python Chalice) to use the environment variable to set the ARN for where to retrieve the Datadog API key instead of recommending people use the less secure DATADOG_API_KEY env var.

### Motivation
The documentation as it is currently written recommends an insecure installation setup by default for python chalice serverless apps which we cannot deploy in production environments, so we went looking for a work-around and found one. I would like to update the documentation to reflect Datadog's own Best Practices so that fewer customers have to toil arriving at their own secure implementation. 

### Preview
https://docs.datadoghq.com/serverless/installation/python/?tab=chalice

### Additional Notes
I've adjusted the English notes and have made an attempt (which needs to be reviewed and likely corrected) for the French but I apologize, I am not able to recommend complete updates to the JP installation docs and must ask for assistance there.

---

### Reviewer checklist
- [ ] Check images for PII


content/en/serverless/installation/python.md
```diff
@@ -309,7 +309,7 @@ For example:
     {{< /site-region >}}
 2. Replace the following placeholders with appropriate values: 
 
-- Replace `<DATADOG_API_KEY>` with your Datadog API key on the [API Management page][2].
```
tianchu, line 312: My only hesitation is that `DD_API_KEY_SECRET_ARN` requires actual step to store the key in Secrets Manager and when resolving the key on cold start, it adds extra latency. So it's a trade-off between easy to setup, performanc and security.

I think we can keep `DD_API_KEY` as the default in the example, while mentioning the `SECRET_ARN_DATADOG_API_KEY` as a more secure alternative?

jeffWelling: I'm not sure if this was meant for me or for internal discussion but incase it's meant for me I'll respond :)

Those are valid concerns. I didn't note adding the secret because I expected it to be implied but that's a poor assumption on my part, you're right. That could be detailed somewhere, perhaps even with a note towards storing the secret in the same region as the lambda that will be using it. 

What if it was re-framed as "use the DD_API_KEY for proof of concepts but we recommend the more secure SECRET_ARN_DATADOG_API_KEY approach for production deployments, described in more detail..."?

tianchu: Yeah, I think we would probably mention both options and the trade-off, while leave the decision to customers. @ceciliawatt Could you help make some edits on the language here?

@jeffWelling thanks for your PR!

roimor:
@jeffWelling @tianchu 
if I export DD_API_KEY_SSM_NAME or DD_API_KEY_SECRET_ARN to the lambda environment, what should be the value of the 
```
    Parameters:
      apiKey: <DATADOG_API_KEY>
```
should i just leave it empty ?

cswatt (approved):

PR #11437, opened 2021-08-17
[WEB-1485] Adds agent-config shortcode

bgdeutsch:
### What does this PR do?
Adds a new `agent-config` shortcode that outputs a code block containing specific agent configuration based on [this agent config template file](https://github.com/DataDog/datadog-agent/blob/main/pkg/config/config_template.yaml).

### Motivation
https://datadoghq.atlassian.net/browse/WEB-1485

### Preview
https://docs-staging.datadoghq.com/brian.deutsch/dd-agent-config/
- Test examples have been removed; pull down this branch and experiment with the shortcode to see different config outputs.

Usage (will add this as a wiki as well upon merge/approvals):
shortcode: `{{< agent-config  >}}`

options:
 - `type`:  which agent configuration to output.   there is an auto-generated list of types available in the root of the repo, `agent_config_types_list.txt` for referencing.   (required).
 - `filename`: file name where users would typically find this configuration; could also be used for a title.
 - `collapsible`: include the ability to collapse the code block.
 - `disable_copy`: include the ability to disable the copy code functionality.

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII


content/en/agent/basic_agent_usage/aix.md
```diff
@@ -97,6 +97,8 @@ DogStatsD relies on the same configuration file defined for the Agent, where a D
 
 To enable DogStatsD, edit `/etc/datadog-agent/datadog.yaml` and set the relevant configuration options.
 
+{{< agent-config type="dogstatsd configuration" collapsible="true" >}}
```
bgdeutsch, line 100: This is for testing purposes; will revert after approvals

kayayarai:
There seems to be an extra newline at the top and bottom of the code block -- that is, it has two newlines instead of just one. Could we get rid of the extra?

local/bin/py/build/process_agent_config.py
```diff
@@ -0,0 +1,83 @@
...
+
+    # Documenting what config types are available for Docs team to use in agent config shortcode.
+    with open('agent_config_types_list.txt', 'w+') as agent_config_types_list_outfile:
+      agent_config_types_list_outfile.write(config_types_string)
+  except Exception as err:
+    print('An error occurred building agent config data:')
+    print(err)
+
+    if getenv("LOCAL") != 'True':
+      sys.exit(1)
```
davidejones, line 83: Can you pep8 format this?

local/bin/py/build/actions/pull_and_push_file.py
```diff
@@ -30,13 +31,17 @@ def pull_and_push_file(content, content_dir):
...
-    ) as f:
-        f.write(file_content)
+        elif "datadog-agent" in content["repo_name"] and "config_template.yaml" in "".join(content["globs"]):
+            process_agent_config(file_content)
+
+    if "output_content" in content["options"] and content["options"]["output_content"] != False:
+        with open(
+            "{}{}{}".format(
+                content_dir,
+                content["options"]["dest_path"][1:],
```
davidejones, line 41: instead of slicing this string can we do something to normalize this?
if someone changes the dest_path to not start with `/` this won't work

local/bin/py/build/actions/pull_and_push_file.py
```diff
@@ -30,13 +31,17 @@ def pull_and_push_file(content, content_dir):
...
-            content_dir,
-            content["options"]["dest_path"][1:],
-            basename(content["options"]["file_name"]),
-        ),
-        mode="w+",
-        encoding="utf-8",
-    ) as f:
-        f.write(file_content)
+        elif "datadog-agent" in content["repo_name"] and "config_template.yaml" in "".join(content["globs"]):
+            process_agent_config(file_content)
```
davidejones, line 35: I was originally wondering if we should make this a separate script instead of adding this logic into push_pull but i see why as its very similar behavior. We can make bigger architecture plans to improve this kind of thing instead.

davidejones: Also just noticed in the config we have a repo called `puppet-datadog-agent`  is that intended to run this logic?

local/bin/py/build/process_agent_config.py
```diff
@@ -0,0 +1,83 @@
...
+  config_type_header_delimiter = '#######'
+  agent_config_array = dd_agent_config_string.splitlines()
+  agent_config_dict = {}
+  current_config_type = ''
+
+  for index, line in enumerate(agent_config_array):
+    if config_type_header_delimiter in line:
+      config_type = agent_config_array[index + 1].replace('#', '').strip().lower()
+
+      if config_type != '':
```
davidejones, line 38: you can probably just do `if config_type` as empty string will evaluate to false

local/bin/py/build/process_agent_config.py
```diff
@@ -0,0 +1,83 @@
...
+        agent_config_dict.setdefault(config_type, '')
+        current_config_type = config_type
+    else:
+      # Skip any Go template strings or comment boxes used to delineate config types.
+      if '{{' in line or line.startswith('##') and line.endswith('##'):
+        continue
+      else:
+        formatted_string = format_agent_config_string(line)
+
+      agent_config_dict[current_config_type] += formatted_string
```
davidejones, line 48: should this be indented? formatted_string only changes in this else

davidejones (commented):
Looks good, just did a first pass and added some notes

local/bin/py/build/actions/pull_and_push_file.py
bgdeutsch: @davidejones no `puppet-datadog-agent` isn't intended to run this logic, i noticed that as well so to address i've added the second condition there to make sure `config_template.yaml` is present in the globs.

local/bin/py/build/process_agent_config.py
bgdeutsch: I enabled `PyLint` on vscode and fixed the one complaint about importing `sys`.

local/bin/py/build/process_agent_config.py
```diff
@@ -0,0 +1,83 @@
...
+  return agent_config_dict
+
+def document_available_config_types(agent_config_dict):
+  """
+  Returns a single string containing all available config types, for Docs team to reference.
+  """
+  config_types_string = 'Available config types: \n\n'
+
+  for config_type in agent_config_dict.keys():
+    config_types_string += config_type + '\n'
```
davidejones, line 59: might be able to make use of join instead to be a bit shorter

```python
'\n'.join(agent_config_dict.keys())
```

davidejones (approved):
I added a few ideas/nitpicks but other than that its looking good here.

sarina-dd:
@bgdeutsch Hey there, this LGTM! 🎉  One ask: is it possible to add back the default title, I believe we had `Code Example` before. for a code block if it doesn't have a title? I think it'd be beneficial so a user can reference where on a page the examples exist.

kayayarai:
> @bgdeutsch Hey there, this LGTM! 🎉 One ask: is it possible to add back the default title, I believe we had `Code Example` before. for a code block if it doesn't have a title? I think it'd be beneficial so a user can reference where on a page the examples exist.

`Code example` (lowercase e) though, please.

cswatt (approved):

PR #11493, opened 2021-08-23
[CIAPP-2577] Fix API key link for GitLab CI integration

cg14823:
### What does this PR do?
Put the correct links to the API page for the GitLab CI integration documentation page.

### Preview

https://docs-staging.datadoghq.com/carlos.gonzalez/fix-api-key-links-in-ci-docs/continuous_integration/setup_pipelines/gitlab/?tab=gitlabcom

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII


AdrianLC:
I think it's very recent that they moved the API keys page. Perhaps this only applies to org2 and is not changed yet for customers? We also have links to this from Gitlab so if they really removed this without any redirect we'll need more changes.

edrevo (approved):

cg14823:
> I think it's very recent that they moved the API keys page. Perhaps this only applies to org2 and is not changed yet for customers?

It is a recent change, not sure how to check if it only applies to org2 though, do you know who the correct person to ask would be?

> We also have links to this from Gitlab so if they really removed this without any redirect we'll need more changes.

Currently it does not redirect to the new place but it does show a page telling you to go to the correct place so it is not world ending at the moment.

Screenshot 2021-08-23 at 11 33 01

fermayo (approved):

AdrianLC:
> > I think it's very recent that they moved the API keys page. Perhaps this only applies to org2 and is not changed yet for customers?
> 
> It is a recent change, not sure how to check if it only applies to org2 though, do you know who the correct person to ask would be?
> 

I checked with our testing org `CI Integration Team (443497)` and the new page is not up yet so I'd say it's too soon to change the link.

> > We also have links to this from Gitlab so if they really removed this without any redirect we'll need more changes.
> 
> Currently it does not redirect to the new place but it does show a page telling you to go to the correct place so it is not world ending at the moment.
> 
Great then I was afraid of a 404

fermayo:
Let's be safe and use the old link. We can replace all of them later when necessary. We can still fix the EU1 instructions in this PR.

PR #11740, opened 2021-09-20
Note on metrics monitor threshold value

DaniCalifornia:
### What does this PR do?
Adds a note on the metric monitor's threshold value form validation. If there is no leading zero for values between 0 and 1, the monitor cannot be saved.

### Motivation
https://datadog.zendesk.com/agent/tickets/559532
This customer pointed out that they could not save their metric monitor when they inputted a threshold value of .9 instead of 0.9. I have also reproduced and tested and got the same results.

### Preview

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/monitors/create/types/metric.md
+++ b/content/en/monitors/create/types/metric.md
@@ -106,7 +106,7 @@ The alert conditions vary slightly based on the chosen detection method.
 {{< tabs >}}
 {{% tab "Threshold" %}}
 
-* Trigger when the metric is `above`, `above or equal to`, `below`, or `below or equal to` (if value is between 0 and 1 a leading 0 is required, eg. 0.3)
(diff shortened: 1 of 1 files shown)


content/en/monitors/create/types/metric.md
```diff
@@ -106,7 +106,7 @@ The alert conditions vary slightly based on the chosen detection method.
 {{< tabs >}}
 {{% tab "Threshold" %}}
 
-* Trigger when the metric is `above`, `above or equal to`, `below`, or `below or equal to`
+* Trigger when the metric is `above`, `above or equal to`, `below`, or `below or equal to` (if value is between 0 and 1 a leading 0 is required, eg. 0.3)
```
kayayarai, line 109:
```suggestion
* Trigger when the metric is `above`, `above or equal to`, `below`, or `below or equal to`. If the value is between zero and one, a leading zero is required. For example, `0.3`.
```

kayayarai (approved):

PR #11767, opened 2021-09-22
[WEB-1613] Migrate JavaScript compilation from Webpack to Hugo.

bgdeutsch:
### What does this PR do?
Migrates our JavaScript compilation/bundling from Webpack to Hugo.

### Why?
Hugo uses `es-build` to compile JavaScript which is significantly more performant than Webpack.   Additionally this change helps to reduce complexity by removing webpack for a simpler option.

### Motivation
https://datadoghq.atlassian.net/browse/WEB-1613

### Preview
https://docs-staging.datadoghq.com/brian.deutsch/hugo-js-pipes/

General:
- [ ] `jQuery` and `bootstrap` still work.
- [ ] JS is minified and fingerprinted.
- [ ] No console errors (originating from our JS) should be present around the site.
- [ ] Algolia/DocSearch still works.
- [ ] Gitlab Jobs/Node scripts still work.

`lang-redirects` & `region-redirects` used to be loaded in the document head, now are loaded with the bundle at the foot of the document.

- [ ] `lang-redirects` still works: https://docs-staging.datadoghq.com/brian.deutsch/hugo-js-pipes/?lang_pref=ja (should redirect to JA site, also should be a `lang_pref=ja` cookie stored).    Using the language dropdown around the site should also work.
- [ ] `region-redirects` still works: https://docs-staging.datadoghq.com/brian.deutsch/hugo-js-pipes/agent/proxy/?tab=agentv6v7#haproxy (changing the `site` dropdown should update the content according to selected option).

DD Logs and Rum still work:

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII


.gitlab-ci.yml
```diff
@@ -235,7 +235,6 @@ sourcemaps_preview:
     policy: pull
   environment: "preview"
   script:
-    - yarn run build:webpack:preview -- --devtool source-map
     - DATADOG_API_KEY="$(get_secret 'dd_synthetic_api_key_prod')" ./node_modules/.bin/datadog-ci sourcemaps upload ./public/static --service docs --minified-path-prefix "https://docs-staging.datadoghq.com/${CI_COMMIT_REF_NAME}/static/" --release-version "${CI_COMMIT_SHORT_SHA}"
```
davidejones, line 238: Can we build the source map with esbuild standalone if there isn't an option with the one bundled with hugo? https://esbuild.github.io/api/#sourcemap

data/integrations/docker_daemon.yaml
```diff
@@ -1085,3 +1085,12 @@ docker_daemon:
...
   unit_name: thread
+- description: Time since the container was started
+  integration: docker
+  interval: ''
+  metric_name: docker.uptime
+  metric_type: gauge
+  orientation: '0'
+  per_unit_name: ''
+  short_name: uptime
+  unit_name: second
```
davidejones, line 1096: i don't think these changes should be here

local/bin/sh/build-vendor-js.sh
```diff
@@ -0,0 +1,11 @@
...
+
+echo "Bundling vendor JS dependencies..."
+
+ASSETS_VENDOR_DIR=assets/scripts/vendor/
+mkdir -p $ASSETS_VENDOR_DIR
+cp node_modules/jquery/dist/jquery.min.js $ASSETS_VENDOR_DIR
+cp node_modules/popper.js/dist/umd/popper.min.js $ASSETS_VENDOR_DIR
+cp node_modules/bootstrap/dist/js/bootstrap.min.js $ASSETS_VENDOR_DIR
+cp node_modules/bootstrap/dist/js/bootstrap.min.js.map $ASSETS_VENDOR_DIR
+
```
davidejones, line 11: do we need this script?

davidejones: Can't we just reference the files from node_modules instead? https://gohugo.io/hugo-pipes/js/#include-dependencies-in-packagejson--node_modules

local/bin/sh/run-site-no-pre-build.sh
```diff
@@ -1,20 +1,21 @@
...
+  #   printf "checking that node modules are installed and up-to-date"
+  #   npm --global install yarn && \
+  #   npm cache clean --force && yarn install --frozen-lockfile
+  #   printf "starting webpack and hugo build"
+	#   yarn run start
 
-    sleep 5
-	fi
+  #   sleep 5
+	# fi
```
davidejones, line 18: I think we may still need the `yarn install --frozen-lockfile`, you can test by deleting your node modules dir and running make start

local/bin/sh/run-site.sh
```diff
@@ -37,11 +36,12 @@ if [ ${RUN_SERVER} = true ]; then
...
-		echo "Checking that node modules are installed and up-to-date."
-    npm --global install yarn && \
-    npm cache clean --force && yarn install --frozen-lockfile
-    echo "Starting webpack and hugo build."
+	# if [ ${RUN_WEBPACK} = true ]; then
+	# 	echo "Checking that node modules are installed and up-to-date."
+  #   npm --global install yarn && \
+  #   npm cache clean --force && yarn install --frozen-lockfile
+  #   echo "Starting webpack and hugo build."
+
```
davidejones, line 44: I think we may still need the `yarn install --frozen-lockfile`, you can test by deleting your node modules dir and running make start

layouts/partials/head.html
```diff
@@ -0,0 +1,36 @@
...
+
+  <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+  <link rel="icon" type="image/png" href="https://docs.datadoghq.com/favicon.ico">
+
+  {{- partialCached "css.html" . -}}
+
+  {{- if ne $.Params.disable_opengraph_meta_tags true -}}
+    {{- partial "meta.html" . -}}
+  {{- end -}}
+</head>
```
davidejones, line 36: I believe we are going to need to bring back `<script src="{{ (index $.Site.Data.manifest "dd-browser-logs-rum.js") | relURL }}"></script>` somehow in here. It needs to be separate or we risk redeploying every html file every build.

local/bin/sh/build-vendor-js.sh
bgdeutsch: @davidejones i've tried the `hugo mod npm pack` as described there but wasn't having much luck (it is marked as experimental so perhaps there are more updates to come).  i will have another look at that however.

another thought i had was the `vendor.js` JS on corp/docs are bloated and a source of much unused code that is being downloaded.   the idea here was that we copy over these dependencies as needed.

davidejones: I feel like you should just be able to do `import mixitup from 'mixitup';` type syntax in the js file for node_modules libraries we are using and it get bundled in automatically. I could be wrong though or it could be a newer hugo version

local/bin/sh/run-site-no-pre-build.sh
bgdeutsch: @davidejones thanks!  i have commented back in the previous script that was used here and `run-site.sh`.   it works when deleting `node_modules` and re-starting server.

layouts/partials/head.html
bgdeutsch: @davidejones the `dd-browser-logs-rum.js` JS file is now being loaded in separately from the main bundle, without fingerprinting:  https://github.com/DataDog/documentation/blob/brian.deutsch/hugo-js-pipes/layouts/partials/javascript.html#L21

local/bin/sh/run-site.sh
bgdeutsch: @davidejones above comment should apply here as well and resolve this

layouts/partials/head.html
davidejones: Thanks! this works at least until we can figure out a better solution

nsollecito:
Seems like on API the Subnav items aren't staying open/focused as they are on live: https://docs-staging.datadoghq.com/brian.deutsch/hugo-js-pipes/api/latest/using-the-api/

bgdeutsch:
this work is captured in #14018

PR #11803, opened 2021-09-27
2022 Agent Key Rotation docs

albertvaka:
### What does this PR do?
Adds documentation for the upcoming Agent signing key rotation for Linux (deb/rpm) packages.

### Motivation
Will be linked from an email that customers that potentially need to trust the new key will receive.

### Preview
https://docs-staging.datadoghq.com/albertvaka/agent-key-rotation-2022/agent/faq/linux-agent-2022-key-rotation

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/agent/faq/linux-agent-2022-key-rotation.md
+++ b/content/en/agent/faq/linux-agent-2022-key-rotation.md
@@ -44,8 +44,8 @@ For hosts running older versions of the install methods listed above or older ve
 Run the following commands on the host:
 
 ```bash
-$ curl -o /tmp/DATADOG_APT_KEY_F14F620E.public https://keys.datadoghq.com/DATADOG_APT_KEY_F14F620E.public
-$ sudo apt-key add /tmp/DATADOG_APT_KEY_F14F620E.public
+$ curl -o /tmp/DATADOG_APT_KEY_F14F620E https://keys.datadoghq.com/DATADOG_APT_KEY_F14F620E.public
+$ sudo apt-key add /tmp/DATADOG_APT_KEY_F14F620E
 ```
 
 {{% /tab %}}
@@ -54,8 +54,8 @@ $ sudo apt-key add /tmp/DATADOG_APT_KEY_F14F620E.public
 Run the following commands on the host:
 
 ```
-$ curl -o /tmp/DATADOG_RPM_KEY_FD4BF915.public https://keys.datadoghq.com/DATADOG_RPM_KEY_FD4BF915.public
-$ sudo rpm --import /tmp/DATADOG_RPM_KEY_FD4BF915.public
+$ curl -o /tmp/DATADOG_RPM_KEY_FD4BF915 https://keys.datadoghq.com/DATADOG_RPM_KEY_FD4BF915.public
+$ sudo rpm --import /tmp/DATADOG_RPM_KEY_FD4BF915
 ```
 
 {{% /tab %}}
--- a/content/en/agent/faq/linux-agent-2022-key-rotation.md
+++ b/content/en/agent/faq/linux-agent-2022-key-rotation.md
@@ -44,7 +44,8 @@ For hosts running older versions of the install methods listed above or older ve
 Run the following commands on the host:
 
 ```bash
-$ wget -qO - https://keys.datadoghq.com/DATADOG_APT_KEY_F14F620E.public | sudo apt-key add -
+$ curl -o /tmp/DATADOG_APT_KEY_F14F620E.public https://keys.datadoghq.com/DATADOG_APT_KEY_F14F620E.public
+$ sudo apt-key add /tmp/DATADOG_APT_KEY_F14F620E.public
 ```
 
 {{% /tab %}}
--- a/content/en/agent/faq/linux-agent-2022-key-rotation.md
+++ b/content/en/agent/faq/linux-agent-2022-key-rotation.md
@@ -3,7 +3,7 @@ title: 2022 Linux Agent Key Rotation
 kind: faq
 ---
 
-As a common best practice, Datadog periodically rotates the keys and certificates used to sign our Agent packages. The following GPG keys, used to sign the Agent RPM and DEB packages, reach their end-of-life in June 2022 and will be rotated in April 2022:
+As a common best practice, Datadog periodically rotates the keys and certificates used to sign Datadog's Agent packages. The following GPG keys, used to sign the Agent RPM and DEB packages, reach their end-of-life in June 2022 and will be rotated in April 2022:
 
 - The RPM signing key with hash [`A4C0B90D7443CF6E4E8AA341F1068E14E09422B3`][1] will be rotated in April 4th 2022 and will be replaced by the key with hash [`C6559B690CA882F023BDF3F63F4D1729FD4BF915`][2]
 - The DEB signing key with hash [`A2923DFF56EDA6E76E55E492D3A80E30382E94DE`][3] will be rotated in April 18th 2022 and will be replaced by the key with hash `D75CEA17048B9ACBF186794B32637D44F14F620E`[4]
@@ -11,7 +11,7 @@ As a common best practice, Datadog periodically rotates the keys and certificate
 Customers using Datadog's RPM or DEB packages might require a manual action to import the new key on their systems in order to install or upgrade the Agent after the rotation takes place.
 
 <div class="alert alert-info">
-Note that this DOES NOT affect the functionality of already running Agents, and only limits the ability to install or upgrade to a newer version of the Agent. This also doesn't affect Windows or MacOS Agents, nor Dockerized Linux Agents.
+**Note**: This DOES NOT affect the functionality of already running Agents, and only limits the ability to install or upgrade to a newer version of the Agent. This also doesn't affect Windows or MacOS Agents, nor Dockerized Linux Agents.
 </div>
 
 If you're using one of the following install methods, your hosts trust the key automatically and no further action is needed:
@@ -24,13 +24,13 @@ If you're using one of the following install methods, your hosts trust the key a
 * [Elastic Beanstalk][10] config templates updated as of Mar 29, 2021 (should contain `DATADOG_RPM_KEY_FD4BF915.public` under `gpgkey`)
 * Containerized Agents (Docker/Kubernetes): No action needed on any version
 
-Additionally, if you are using the DEB version of the Agent version 7.31.0 or greater, your hosts should also trust the new key autmatically.
+Additionally, if you are using the DEB version of the Agent version 7.31.0 or greater, your hosts should have installed the `datadog-signing-keys` package that automatically trusts the new key.
 
-If unsure, refer to the steps in [Check if a host trusts the new GPG key](#check-if-a-host-trusts-the-new-gpg-key).
+If unsure, read the steps in [Check if a host trusts the new GPG key](#check-if-a-host-trusts-the-new-gpg-key).
 
-Trying to install or upgrade the Agent package without trusting the new key will result in `NOKEY` errors when installing the RPM package and `NO_PUBKEY` errors when installing the DEB package.
+Trying to install or upgrade the Agent package without trusting the new key results in `NOKEY` errors when installing the RPM package and `NO_PUBKEY` errors when installing the DEB package.
 
-For hosts running older versions of the install methods listed above or older versions of the DEB package, we recommend you update your install method to the last available version. Alternatively, for Debian/Ubuntu users, update the Agent to version 7.31.0 or greater. Otherwise, you can also add the new keys manually following these instructions:
+For hosts running older versions of the install methods listed above or older versions of the DEB package, Datadog recommends updating the install method to the latest version. Alternatively, Debian and Ubuntu users can update the Agent to version 7.31.0 or greater. Otherwise, the keys can be manually updated on host by following these instructions:
 
 ### Manual update
 
@@ -46,7 +46,7 @@ $ wget -qO - https://keys.datadoghq.com/DATADOG_APT_KEY_F14F620E.public | sudo a
 {{% /tab %}}
 {{% tab "RedHat/CentOS" %}}
 
-Edit the `/etc/yum.repos.d/datadog.repo` and make sure it includes both the old and new keys in the `gpgkey` section:
+Edit the `/etc/yum.repos.d/datadog.repo` and make sure it includes the following list of keys in the `gpgkey` section:
 
 ```
 gpgkey=https://keys.datadoghq.com/DATADOG_RPM_KEY_CURRENT.public
@@ -58,7 +58,7 @@ gpgkey=https://keys.datadoghq.com/DATADOG_RPM_KEY_CURRENT.public
 {{% /tab %}}
 {{% tab "SUSE" %}}
 
-Edit the `/etc/zypp/repos.d/datadog.repo` and make sure it includes both the old and new keys in the `gpgkey` section:
+Edit the `/etc/zypp/repos.d/datadog.repo` and make sure it includes the following list of keys in the `gpgkey` section:
 
 ```
 gpgkey=https://keys.datadoghq.com/DATADOG_RPM_KEY_CURRENT.public
--- a/content/en/agent/faq/linux-agent-2022-key-rotation.md
+++ b/content/en/agent/faq/linux-agent-2022-key-rotation.md
@@ -1,0 +1,122 @@
+---
+title: 2022 Linux Agent Key Rotation
+kind: faq
+---
+
+As a common best practice, Datadog periodically rotates the keys and certificates used to sign our Agent packages. The following GPG keys, used to sign the Agent RPM and DEB packages, reach their end-of-life in June 2022 and will be rotated in April 2022:
+
+- The RPM signing key with hash [`A4C0B90D7443CF6E4E8AA341F1068E14E09422B3`][1] will be rotated in April 4th 2022 and will be replaced by the key with hash [`C6559B690CA882F023BDF3F63F4D1729FD4BF915`][2]
+- The DEB signing key with hash [`A2923DFF56EDA6E76E55E492D3A80E30382E94DE`][3] will be rotated in April 18th 2022 and will be replaced by the key with hash `D75CEA17048B9ACBF186794B32637D44F14F620E`[4]
+
+Customers using Datadog's RPM or DEB packages might require a manual action to import the new key on their systems in order to install or upgrade the Agent after the rotation takes place.
+
+<div class="alert alert-info">
+Note that this DOES NOT affect the functionality of already running Agents, and only limits the ability to install or upgrade to a newer version of the Agent. This also doesn't affect Windows or MacOS Agents, nor Dockerized Linux Agents.
+</div>
+
+If you're using one of the following install methods, your hosts trust the key automatically and no further action is needed:
+
+* [Agent install script][5] v1.5.0 (released Jun 29, 2021)
+* [Chef cookbook][6] v4.11.0 (released Aug 10, 2021)
+* [Ansible role][7] v4.9.0 (released May 6, 2021)
+* [Puppet module][8] v3.13 (released Aug 11, 2021)
+* [SaltStack formula][9] v3.4 (released Aug 12, 2021)
+* [Elastic Beanstalk][10] config templates updated as of Mar 29, 2021 (should contain `DATADOG_RPM_KEY_FD4BF915.public` under `gpgkey`)
+* Containerized Agents (Docker/Kubernetes): No action needed on any version
+
+Additionally, if you are using the DEB version of the Agent version 7.31.0 or greater, your hosts should also trust the new key autmatically.
+
+If unsure, refer to the steps in [Check if a host trusts the new GPG key](#check-if-a-host-trusts-the-new-gpg-key).
+
+Trying to install or upgrade the Agent package without trusting the new key will result in `NOKEY` errors when installing the RPM package and `NO_PUBKEY` errors when installing the DEB package.
+
+For hosts running older versions of the install methods listed above or older versions of the DEB package, we recommend you update your install method to the last available version. Alternatively, for Debian/Ubuntu users, update the Agent to version 7.31.0 or greater. Otherwise, you can also add the new keys manually following these instructions:
+
+### Manual update
+
+{{< tabs >}}
+{{% tab "Debian/Ubuntu" %}}
+
+Run the following commands on the host:
+
+```bash
+$ wget -qO - https://keys.datadoghq.com/DATADOG_APT_KEY_F14F620E.public | sudo apt-key add -
+```
+
+{{% /tab %}}
+{{% tab "RedHat/CentOS" %}}
+
+Edit the `/etc/yum.repos.d/datadog.repo` and make sure it includes both the old and new keys in the `gpgkey` section:
+
+```
+gpgkey=https://keys.datadoghq.com/DATADOG_RPM_KEY_CURRENT.public
+       https://keys.datadoghq.com/DATADOG_RPM_KEY_FD4BF915.public
+       https://keys.datadoghq.com/DATADOG_RPM_KEY_E09422B3.public
+       https://keys.datadoghq.com/DATADOG_RPM_KEY.public
+```
+
+{{% /tab %}}
+{{% tab "SUSE" %}}
+
+Edit the `/etc/zypp/repos.d/datadog.repo` and make sure it includes both the old and new keys in the `gpgkey` section:
+
+```
+gpgkey=https://keys.datadoghq.com/DATADOG_RPM_KEY_CURRENT.public
+       https://keys.datadoghq.com/DATADOG_RPM_KEY_FD4BF915.public
+       https://keys.datadoghq.com/DATADOG_RPM_KEY_E09422B3.public
+       https://keys.datadoghq.com/DATADOG_RPM_KEY.public
+```
+
+{{% /tab %}}
+{{< /tabs >}}
+
+## Check if a host trusts the new GPG key
+
+
+{{< tabs >}}
+{{% tab "Debian/Ubuntu" %}}
+
+Run the following commands on the host:
+
+```bash
+$ apt-key list
+```
+
+And check the output for Datadog's key fingerprint `F14F620E` (sometimes a space is included: `F14F 620E`).
+
+{{% /tab %}}
+{{% tab "RedHat/CentOS" %}}
+
+Run the following commands on the host:
+
+```bash
+$ grep DATADOG_RPM_KEY_FD4BF915.public /etc/yum.repos.d/*datadog*.repo
+```
+
+And check the output is not empty.
+
+{{% /tab %}}
+{{% tab "SUSE" %}}
+
+Run the following commands on the host:
+
+```bash
+$ grep DATADOG_RPM_KEY_FD4BF915.public /etc/zypp/repos.d/datadog.repo
+```
+
+And check the output is not empty.
+
+{{% /tab %}}
+{{< /tabs >}}
+
+
+[1]: https://keys.datadoghq.com/DATADOG_RPM_KEY_E09422B3.public
+[2]: https://keys.datadoghq.com/DATADOG_RPM_KEY_FD4BF915.public
+[3]: https://keys.datadoghq.com/DATADOG_APT_KEY_382E94DE.public
+[4]: https://keys.datadoghq.com/DATADOG_APT_KEY_F14F620E.public
+[5]: https://app.datadoghq.com/account/settings#agent
+[6]: https://github.com/DataDog/chef-datadog
+[7]: https://github.com/DataDog/ansible-datadog
+[8]: https://github.com/DataDog/puppet-datadog-agent
+[9]: https://github.com/DataDog/datadog-formula
+[10]: https://docs.datadoghq.com/integrations/amazon_elasticbeanstalk
--- a/content/en/agent/faq/linux-agent-2022-key-rotation.md
+++ b/content/en/agent/faq/linux-agent-2022-key-rotation.md
@@ -5,8 +5,8 @@ kind: faq
 
 As a common best practice, Datadog periodically rotates the keys and certificates used to sign Datadog's Agent packages. The following GPG keys, used to sign the Agent RPM and DEB packages, reach their end-of-life in June 2022 and will be rotated in April 2022:
 
-- The RPM signing key with hash [`A4C0B90D7443CF6E4E8AA341F1068E14E09422B3`][1] will be rotated in April 4th 2022 and will be replaced by the key with hash [`C6559B690CA882F023BDF3F63F4D1729FD4BF915`][2]
-- The DEB signing key with hash [`A2923DFF56EDA6E76E55E492D3A80E30382E94DE`][3] will be rotated in April 18th 2022 and will be replaced by the key with hash [`D75CEA17048B9ACBF186794B32637D44F14F620E`][4]
+- The RPM signing key with hash [`A4C0B90D7443CF6E4E8AA341F1068E14E09422B3`][1] will be rotated in April 11th 2022 and will be replaced by the key with hash [`C6559B690CA882F023BDF3F63F4D1729FD4BF915`][2]
+- The DEB signing key with hash [`A2923DFF56EDA6E76E55E492D3A80E30382E94DE`][3] will be rotated in May 2nd 2022 and will be replaced by the key with hash [`D75CEA17048B9ACBF186794B32637D44F14F620E`][4]
 
 Customers using Datadog's RPM or DEB packages might require a manual action to import the new key on their systems in order to install or upgrade the Agent after the rotation takes place.
 
@@ -34,7 +34,7 @@ If you're unsure if a host trusts the new signing key, you can [check](#check-if
 
 For hosts running older versions of the install methods listed above or older versions of the DEB package, Datadog recommends updating the install method to the latest version. Alternatively Debian and Ubuntu users can update the Agent to version 7.31.0 or greater. Otherwise, the key can be [manually updated](#manual-update).
 
-## What happens if I don't trust the new key by April 2022?
+## What happens if I don't trust the new key before it is rotated?
 
 {{< tabs >}}
 {{% tab "Debian/Ubuntu" %}}
--- a/content/en/agent/faq/linux-agent-2022-key-rotation.md
+++ b/content/en/agent/faq/linux-agent-2022-key-rotation.md
@@ -32,9 +32,31 @@ Additionally, if you are using the DEB version of the Agent version 7.31.0 or gr
 
 If you're unsure if a host trusts the new signing key, you can [check](#check-if-a-host-trusts-the-new-gpg-key).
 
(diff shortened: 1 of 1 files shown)


content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,124 @@
+---
+title: 2022 Linux Agent Key Rotation
+kind: faq
+---
+
+As a common best practice, Datadog periodically rotates the keys and certificates used to sign Datadog's Agent packages. The following GPG keys, used to sign the Agent RPM and DEB packages, reach their end-of-life in June 2022 and will be rotated in April 2022:
+
+- The RPM signing key with hash [`A4C0B90D7443CF6E4E8AA341F1068E14E09422B3`][1] will be rotated in April 4th 2022 and will be replaced by the key with hash [`C6559B690CA882F023BDF3F63F4D1729FD4BF915`][2]
+- The DEB signing key with hash [`A2923DFF56EDA6E76E55E492D3A80E30382E94DE`][3] will be rotated in April 18th 2022 and will be replaced by the key with hash `D75CEA17048B9ACBF186794B32637D44F14F620E`[4]
```
mx-psi, line 9:
```suggestion
- The DEB signing key with hash [`A2923DFF56EDA6E76E55E492D3A80E30382E94DE`][3] will be rotated in April 18th 2022 and will be replaced by the key with hash [`D75CEA17048B9ACBF186794B32637D44F14F620E`][4]
```

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,124 @@
...
+**Note**: This DOES NOT affect the functionality of already running Agents, and only limits the ability to install or upgrade to a newer version of the Agent. This also doesn't affect Windows or MacOS Agents, nor Dockerized Linux Agents.
+</div>
+
+If you're using one of the following install methods, your hosts trust the key automatically and no further action is needed:
+
+* [Agent install script][5] v1.5.0 (released Jun 29, 2021)
+* [Chef cookbook][6] v4.11.0 (released Aug 10, 2021)
+* [Ansible role][7] v4.9.0 (released May 6, 2021)
+* [Puppet module][8] v3.13 (released Aug 11, 2021)
+* [SaltStack formula][9] v3.4 (released Aug 12, 2021)
```
mx-psi, line 23: We should say "vX.Y.Z or later" here

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,124 @@
...
+* [Elastic Beanstalk][10] config templates updated as of Mar 29, 2021 (should contain `DATADOG_RPM_KEY_FD4BF915.public` under `gpgkey`)
+* Containerized Agents (Docker/Kubernetes): No action needed on any version
+* Heroku: No action needed on any version.
+* Windows/MacOS Agents: No action needed on any version.
+
+Additionally, if you are using the DEB version of the Agent version 7.31.0 or greater, your hosts should have installed the `datadog-signing-keys` package that automatically trusts the new key.
```
mx-psi, line 29: Why not add this as a bullet point to the list?

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Manual update
+
+{{< tabs >}}
+{{% tab "Debian/Ubuntu" %}}
+
+Run the following commands on the host:
+
+```bash
+$ wget -qO - https://keys.datadoghq.com/DATADOG_APT_KEY_F14F620E.public | sudo apt-key add -
```
mx-psi, line 45: Maybe split into two steps and check the key hash before adding?

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,124 @@
...
+
+As a common best practice, Datadog periodically rotates the keys and certificates used to sign Datadog's Agent packages. The following GPG keys, used to sign the Agent RPM and DEB packages, reach their end-of-life in June 2022 and will be rotated in April 2022:
+
+- The RPM signing key with hash [`A4C0B90D7443CF6E4E8AA341F1068E14E09422B3`][1] will be rotated in April 4th 2022 and will be replaced by the key with hash [`C6559B690CA882F023BDF3F63F4D1729FD4BF915`][2]
+- The DEB signing key with hash [`A2923DFF56EDA6E76E55E492D3A80E30382E94DE`][3] will be rotated in April 18th 2022 and will be replaced by the key with hash `D75CEA17048B9ACBF186794B32637D44F14F620E`[4]
+
+Customers using Datadog's RPM or DEB packages might require a manual action to import the new key on their systems in order to install or upgrade the Agent after the rotation takes place.
+
+<div class="alert alert-info">
+**Note**: This DOES NOT affect the functionality of already running Agents, and only limits the ability to install or upgrade to a newer version of the Agent. This also doesn't affect Windows or MacOS Agents, nor Dockerized Linux Agents.
```
julien-lebot, line 14: Notes are usually not inside `alert-info` boxes (see other `faq` in this folder). I suggest we chose between keeping `**note**` and the `alert-info`.

content/en/agent/faq/linux-agent-2022-key-rotation.md
julien-lebot: Wouldn't that be implied though ? Since those packages include the new signing key starting with the release stated, it would be weird if they later removed it.
Maybe we could modify the sentence to make it clearer (e.g. "starting from version X") but it seemed clear to me.

julien-lebot (approved):
LGTM, a few minor comments

content/en/agent/faq/linux-agent-2022-key-rotation.md
mx-psi: >  Maybe we could modify the sentence to make it clearer (e.g. "starting from version X")

That also works for me. In general I think one needs to be as explicit and precise as possible in docs

content/en/agent/faq/linux-agent-2022-key-rotation.md
albertvaka: Because it overlaps: you can be using an older chef but a newer Agent, or the other way around, and still be fine. How would you include it in the list above in a way that the overlap is clear?

content/en/agent/faq/linux-agent-2022-key-rotation.md
albertvaka:
```suggestion
* [Agent install script][5] v1.5.0 or later (released Jun 29, 2021)
* [Chef cookbook][6] v4.11.0 or later (released Aug 10, 2021)
* [Ansible role][7] v4.9.0 or later (released May 6, 2021)
* [Puppet module][8] v3.13 or later (released Aug 11, 2021)
* [SaltStack formula][9] v3.4 or later (released Aug 12, 2021)
```

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Manual update
+
+{{< tabs >}}
+{{% tab "Debian/Ubuntu" %}}
+
+Run the following commands on the host:
+
+```bash
+$ wget -qO - https://keys.datadoghq.com/DATADOG_APT_KEY_F14F620E.public | sudo apt-key add -
```
bkabrda, line 45: So while this would work, technically it's not ideal because the usage of `apt-key` for 3rd part repos is discouraged per https://wiki.debian.org/DebianRepository/UseThirdParty#OpenPGP_Key_distribution

I understand that we're doing this to accomodate for people who haven't yet adopted the approach of using `signed-by` option for the Datadog DEB repository, so I guess it's ok to keep this, but we should at least provide a clear note about this not being a good way to go. Alternatively, we could provide link to the in-app installation instructions at https://app.datadoghq.com/account/settings#agent/debian which do the whole thing with keyring and `signed-by`.

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,124 @@
...
+gpgkey=https://keys.datadoghq.com/DATADOG_RPM_KEY_CURRENT.public
+       https://keys.datadoghq.com/DATADOG_RPM_KEY_FD4BF915.public
+       https://keys.datadoghq.com/DATADOG_RPM_KEY_E09422B3.public
+       https://keys.datadoghq.com/DATADOG_RPM_KEY.public
+```
+
+{{% /tab %}}
+{{% tab "SUSE" %}}
+
+Edit the `/etc/zypp/repos.d/datadog.repo` and make sure it includes the following list of keys in the `gpgkey` section:
```
bkabrda, line 63: I think we should provide a comment about the fact that for SUSE 11 this entry can only contain one key (the `CURRENT` one).

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,124 @@
...
+## Check if a host trusts the new GPG key
+
+
+{{< tabs >}}
+{{% tab "Debian/Ubuntu" %}}
+
+Run the following commands on the host:
+
+```bash
+$ apt-key list
```
bkabrda, line 84: If we keep doing `apt-key add` above, then this is true; otherwise if we advise to use the separate keyring, this will need to be `gpg --no-default-keyring --keyring /usr/share/keyrings/datadog-archive-keyring.gpg --list-keys`.

content/en/agent/faq/linux-agent-2022-key-rotation.md
mx-psi: That's a fair point, we can leave it as a separate sentence. I didn't get/think about the overlap when reading it like this, but probably that is just me

bkabrda (changes requested):
Mostly looks good, but I left couple comments inline to get some places improved.

content/en/agent/faq/linux-agent-2022-key-rotation.md
albertvaka: The problem with the better way is that the `datadog.list` contents change depending on the Agent major version, and that would make the docs quite more complex... futhermore, if you have use a config management tool, it will rewrite the file on the next run.

And now that I think about it, this is a problem for RPM as well :/

content/en/agent/faq/linux-agent-2022-key-rotation.md
albertvaka: This should actually ask you to run both commands, because you could have the key added either way.

content/en/agent/faq/linux-agent-2022-key-rotation.md
bkabrda: Hmm, that's a good point. I think this should be highlighted in the docs - if you're using an older version of a cfg management tool, your hand changes will be overwritten.

In terms of RPM platforms, importing the key to the `rpm` database will work for package-level signatures. For repodata signatures, it will work on SUSE, because zypper respects rpm-imported GPG keys when validating repodata signatures. However, for `yum`/`dnf` this is not true - people who modify repofile manually on a RH platform, but then rerun a cfg management tool and clear yum/dnf cache will not have the new key for repodata verification.

content/en/agent/faq/linux-agent-2022-key-rotation.md
bkabrda: Thinking about this further, I *think* that all cfg management tools versions that create the separate keyring also import the new key into it, but you're right that if people want to verify, they need to be provided both versions of what to do based on how they set the system up.

content/en/agent/faq/linux-agent-2022-key-rotation.md
albertvaka: On the other hand, repodata validation is only turned on by install methods that already add the `CURRENT` key to the list of keys, so we don't need to worry about that. I think we can just `rpm --import` (but we would need to test all distros and different versions of config management tools...).

bkabrda: Oh, that's right! Well, at least that makes it easier for us :)

content/en/agent/faq/linux-agent-2022-key-rotation.md
albertvaka: I've left it just as `rpm -q` on rpm-based and `apt-key list` on debian-based. Once I test it I'll check if it works on all distros (it will only work if we always add the key to the "generic" keyring instead of adding it per-repo).

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -84,29 +72,28 @@ Run the following commands on the host:
...
+$ rpm -q gpg-pubkey-fd4bf915
 ```
 
-And check the output is not empty.
+If the key is trusted, the command has a 0 exit code and outputs:
 
-{{% /tab %}}
-{{% tab "SUSE" %}}
+```
+gpg-pubkey-e09422b3-57744e9e
```
bkabrda, line 89: Isn't this the old key? For me, the operation actually outputs `gpg-pubkey-fd4bf915-5f573efe`.

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -84,29 +72,28 @@ Run the following commands on the host:
...
 
-Run the following commands on the host:
+Otherwise, the command returns a non-0 exit code and the following output:
 
-```bash
-$ grep DATADOG_RPM_KEY_FD4BF915.public /etc/zypp/repos.d/datadog.repo
 ```
-
-And check the output is not empty.
+package gpg-pubkey-e09422b3 is not installed
```
bkabrda, line 95: Should this be `gpg-pubkey-fd4bf915`?

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -46,27 +46,15 @@ $ wget -qO - https://keys.datadoghq.com/DATADOG_APT_KEY_F14F620E.public | sudo a
...
-       https://keys.datadoghq.com/DATADOG_RPM_KEY_FD4BF915.public
-       https://keys.datadoghq.com/DATADOG_RPM_KEY_E09422B3.public
-       https://keys.datadoghq.com/DATADOG_RPM_KEY.public
-```
-
-{{% /tab %}}
-{{% tab "SUSE" %}}
-
-Edit the `/etc/zypp/repos.d/datadog.repo` and make sure it includes the following list of keys in the `gpgkey` section:
+{{% tab "" %}}
```
bkabrda, line 53: Should this tab have empty title?

bkabrda (changes requested):

content/en/agent/faq/linux-agent-2022-key-rotation.md
albertvaka: Indeed.

content/en/agent/faq/linux-agent-2022-key-rotation.md
albertvaka: Oops

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,110 @@
...
+
+Additionally, if you are using the DEB version of the Agent version 7.31.0 or greater, your hosts should have the `datadog-signing-keys` package installed, which will make APT recognize the signature by the new key on the Datadog APT repository.
+
+If unsure, read the steps in [Check if a host trusts the new GPG key](#check-if-a-host-trusts-the-new-gpg-key).
+
+Trying to install or upgrade the Agent package without trusting the new key results in `NOKEY` errors when installing the RPM package and `NO_PUBKEY` errors when installing the DEB package.
+
+For hosts running older versions of the install methods listed above or older versions of the DEB package, Datadog recommends updating the install method to the latest version. Alternatively Debian and Ubuntu users can update the Agent to version 7.31.0 or greater. Otherwise, the keys can be manually updated on host by following these instructions:
+
+### Manual update
```
AriDDOG, line 37: I think this should be a ## Manual Update 
That way the heading of this and the next paragraph are the same

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,110 @@
...
+* Containerized Agents (Docker/Kubernetes): No action needed on any version
+* Windows/MacOS Agents: No action needed on any version.
+
+Additionally, if you are using the DEB version of the Agent version 7.31.0 or greater, your hosts should have the `datadog-signing-keys` package installed, which will make APT recognize the signature by the new key on the Datadog APT repository.
+
+If unsure, read the steps in [Check if a host trusts the new GPG key](#check-if-a-host-trusts-the-new-gpg-key).
+
+Trying to install or upgrade the Agent package without trusting the new key results in `NOKEY` errors when installing the RPM package and `NO_PUBKEY` errors when installing the DEB package.
+
+For hosts running older versions of the install methods listed above or older versions of the DEB package, Datadog recommends updating the install method to the latest version. Alternatively Debian and Ubuntu users can update the Agent to version 7.31.0 or greater. Otherwise, the keys can be manually updated on host by following these instructions:
```
AriDDOG, line 35: "... can be manually updated."

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,110 @@
...
+- The RPM signing key with hash [`A4C0B90D7443CF6E4E8AA341F1068E14E09422B3`][1] will be rotated in April 4th 2022 and will be replaced by the key with hash [`C6559B690CA882F023BDF3F63F4D1729FD4BF915`][2]
+- The DEB signing key with hash [`A2923DFF56EDA6E76E55E492D3A80E30382E94DE`][3] will be rotated in April 18th 2022 and will be replaced by the key with hash [`D75CEA17048B9ACBF186794B32637D44F14F620E`][4]
+
+Customers using Datadog's RPM or DEB packages might require a manual action to import the new key on their systems in order to install or upgrade the Agent after the rotation takes place.
+
+<div class="alert alert-info">
+<strong>Note</strong>: This DOES NOT affect the functionality of already running Agents, and only limits the ability to install or upgrade to a newer version of the Agent. This also doesn't affect Windows or MacOS Agents, nor Dockerized Linux Agents.
+</div>
+
+If you're using one of the following install methods, your hosts trust the key automatically and no further action is needed:
```
AriDDOG, line 17: Maybe add a title here such as - "## Install methods that automatically trust the new PGP key"

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,110 @@
...
+* [Puppet module][8] v3.13.0 or later (released Aug 11, 2021)
+* [SaltStack formula][9] v3.4 or later (released Aug 12, 2021)
+* [Heroku buildpack][10] v1.26  or later (released May 26, 2021)
+* [Elastic Beanstalk][11] config templates updated as of Mar 29, 2021 or later (should contain `DATADOG_RPM_KEY_FD4BF915.public` under `gpgkey`)
+* Containerized Agents (Docker/Kubernetes): No action needed on any version
+* Windows/MacOS Agents: No action needed on any version.
+
+Additionally, if you are using the DEB version of the Agent version 7.31.0 or greater, your hosts should have the `datadog-signing-keys` package installed, which will make APT recognize the signature by the new key on the Datadog APT repository.
+
+If unsure, read the steps in [Check if a host trusts the new GPG key](#check-if-a-host-trusts-the-new-gpg-key).
```
AriDDOG, line 31: If you're unsure if a host trusts the new PGP key, you can [check](#check-if-a-host-trusts-the-new-gpg-key).

bkabrda (approved):
LGTM from my perspective now, assuming testing on all the mentioned platforms confirms the provided commands work as expected.

content/en/agent/faq/linux-agent-2022-key-rotation.md
mx-psi: No strong feelings about this so I am approving

mx-psi (approved):

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,136 @@
+---
+title: 2022 Linux Agent Key Rotation
+kind: faq
+---
+
+As a common best practice, Datadog periodically rotates the keys and certificates used to sign Datadog's Agent packages. The following GPG keys, used to sign the Agent RPM and DEB packages, reach their end-of-life in June 2022 and will be rotated in April 2022:
+
+- The RPM signing key with hash [`A4C0B90D7443CF6E4E8AA341F1068E14E09422B3`][1] will be rotated in April 11th 2022 and will be replaced by the key with hash [`C6559B690CA882F023BDF3F63F4D1729FD4BF915`][2]
+- The DEB signing key with hash [`A2923DFF56EDA6E76E55E492D3A80E30382E94DE`][3] will be rotated in May 2nd 2022 and will be replaced by the key with hash [`D75CEA17048B9ACBF186794B32637D44F14F620E`][4]
```
ruthnaebeck, line 9:
```suggestion
- The DEB signing key with hash [`A2923DFF56EDA6E76E55E492D3A80E30382E94DE`][3] will be rotated on May 2, 2022 and replaced by the key with hash [`D75CEA17048B9ACBF186794B32637D44F14F620E`][4]
```

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,136 @@
...
+
+Customers using Datadog's RPM or DEB packages might require a manual action to import the new key on their systems in order to install or upgrade the Agent after the rotation takes place.
```
ruthnaebeck, line 11:
```suggestion
Customers using Datadog's RPM or DEB packages might require a manual action to import the new key on their systems to install or upgrade the Agent after the rotation takes place.
```

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,136 @@
...
+
+<div class="alert alert-info">
+<strong>Note</strong>: This DOES NOT affect the functionality of already running Agents, and only limits the ability to install the Agent or upgrade to a newer version of the Agent. This also doesn't affect Dockerized Linux Agents nor Windows or MacOS Agents.
```
ruthnaebeck, line 14:
```suggestion
<strong>Note</strong>: This DOES NOT affect the functionality of already running Agents, and only limits the ability to install or upgrade to a newer version of the Agent. Also, this doesn't affect Dockerized Linux Agents, Windows, or macOS Agents.
```

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,136 @@
...
+</div>
+
+## Install methods that automatically trust the new GPG key
+
+If you're using one of the following install methods, your hosts trust the key automatically and no further action is needed:
```
ruthnaebeck, line 19:
```suggestion
Your host automatically trusts the new key (no further action is required) if you're using one of the following install methods:
```

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,136 @@
...
+
+* [Agent install script][5] v1.6.0 or later (released Jul 26, 2021)
```
ruthnaebeck, line 21:
```suggestion
- [Agent install script][5] v1.6.0+ (released Jul 26, 2021)
```

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,136 @@
...
+* [Chef cookbook][6] v4.11.0 or later (released Aug 10, 2021)
+* [Ansible role][7] v4.10.0 or later (released May 25, 2021)
+* [Puppet module][8] v3.13.0 or later (released Aug 11, 2021)
+* [SaltStack formula][9] v3.4 or later (released Aug 12, 2021)
+* [Heroku buildpack][10] v1.26 or later (released May 26, 2021)
+* [Elastic Beanstalk][11] config templates updated as of Mar 29, 2021 or later (should contain `DATADOG_RPM_KEY_FD4BF915.public` under `gpgkey`)
+* Containerized Agents (Docker/Kubernetes): No action needed on any version
```
ruthnaebeck, line 28:
```suggestion
- Containerized Agents (Docker/Kubernetes) for any version
```

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,136 @@
...
+* Windows/MacOS Agents: No action needed on any version.
```
ruthnaebeck, line 29:
```suggestion
- Windows/MacOS Agents for any version
```

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,136 @@
...
+
+Additionally, if you are using the DEB version of the Agent version 7.31.0 or greater, your hosts should have the `datadog-signing-keys` package installed, which will make APT recognize the signature by the new key on the Datadog APT repository.
```
ruthnaebeck, line 31:
```suggestion
Additionally, if you are using DEB Agent v7.31.0+, your hosts should have the `datadog-signing-keys` package installed, which makes APT recognize the signature by the new key on the Datadog APT repository.
```

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,136 @@
...
+<div class="alert alert-info">
+<strong>Note</strong>: This DOES NOT affect the functionality of already running Agents, and only limits the ability to install the Agent or upgrade to a newer version of the Agent. This also doesn't affect Dockerized Linux Agents nor Windows or MacOS Agents.
+</div>
+
+## Install methods that automatically trust the new GPG key
+
+If you're using one of the following install methods, your hosts trust the key automatically and no further action is needed:
+
+* [Agent install script][5] v1.6.0 or later (released Jul 26, 2021)
+* [Chef cookbook][6] v4.11.0 or later (released Aug 10, 2021)
```
ruthnaebeck, line 22:
```suggestion
- [Chef cookbook][6] v4.11.0+ (released Aug 10, 2021)
```

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,136 @@
...
+* [Ansible role][7] v4.10.0 or later (released May 25, 2021)
```
ruthnaebeck, line 23:
```suggestion
- [Ansible role][7] v4.10.0+ (released May 25, 2021)
```

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,136 @@
...
+* [Puppet module][8] v3.13.0 or later (released Aug 11, 2021)
```
ruthnaebeck, line 24:
```suggestion
- [Puppet module][8] v3.13.0+ (released Aug 11, 2021)
```

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,136 @@
...
+* [SaltStack formula][9] v3.4 or later (released Aug 12, 2021)
```
ruthnaebeck, line 25:
```suggestion
- [SaltStack formula][9] v3.4+ (released Aug 12, 2021)
```

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,136 @@
...
+* [Heroku buildpack][10] v1.26 or later (released May 26, 2021)
```
ruthnaebeck, line 26:
```suggestion
- [Heroku buildpack][10] v1.26+ (released May 26, 2021)
```

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,136 @@
...
+* [Elastic Beanstalk][11] config templates updated as of Mar 29, 2021 or later (should contain `DATADOG_RPM_KEY_FD4BF915.public` under `gpgkey`)
```
ruthnaebeck, line 27:
```suggestion
- [Elastic Beanstalk][11] config templates updated as of Mar 29, 2021 or later (should contain `DATADOG_RPM_KEY_FD4BF915.public` under `gpgkey`)
```

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,136 @@
...
+* Containerized Agents (Docker/Kubernetes): No action needed on any version
+* Windows/MacOS Agents: No action needed on any version.
+
+Additionally, if you are using the DEB version of the Agent version 7.31.0 or greater, your hosts should have the `datadog-signing-keys` package installed, which will make APT recognize the signature by the new key on the Datadog APT repository.
+
+If you're unsure if a host trusts the new signing key, you can [check](#check-if-a-host-trusts-the-new-gpg-key).
+
+For hosts running older versions of the install methods listed above or older versions of the DEB package, Datadog recommends updating the install method to the latest version. Alternatively Debian and Ubuntu users can update the Agent to version 7.31.0 or greater. Otherwise, the key can be [manually updated](#manual-update).
```
ruthnaebeck, line 35:
```suggestion
For hosts running older versions of the install methods listed above or older versions of the DEB package, Datadog recommends updating the install method to the latest version. Alternatively Debian and Ubuntu users can update the Agent to version 7.31.0+. Otherwise, the key can be [manually updated](#manual-update).
```

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,136 @@
...
+
+{{< tabs >}}
+{{% tab "Debian/Ubuntu" %}}
+
+If the file `/usr/share/keyrings/datadog-archive-keyring.gpg` exists, the new key is trusted and no further action is needed.
+
+{{% /tab %}}
+{{% tab "RedHat/CentOS/SUSE" %}}
+
+Run the following commands on the host:
```
ruthnaebeck, line 98:
```suggestion
Run the following command on the host:
```

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,136 @@
...
+Otherwise, the command returns a non-0 exit code and the following output:
+
+```
+package gpg-pubkey-fd4bf915 is not installed
+```
+
+{{% /tab %}}
+{{< /tabs >}}
+
+## Impact for Agent 5 users
```
ruthnaebeck, line 119:
```suggestion
## Impact for Agent v5 users
```

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,136 @@
...
+
+Agent 5 users on DEB-based systems (Debian/Ubuntu) are also required to trust the new signing key in order to install or upgrade the Agent after the rotation date. Agent 5 users on RPM-based systems (RedHat/CentOS/SUSE) are not affected by this rotation.
```
ruthnaebeck, line 121:
```suggestion
Agent v5 users on DEB-based systems (Debian/Ubuntu) are also required to trust the new signing key to install or upgrade the Agent after the rotation date. Agent v5 users on RPM-based systems (RedHat/CentOS/SUSE) are not affected by this rotation.
```

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,136 @@
...
+
+Note that Agent 5 uses Python 2 which reached end-of-life on January 1st 2021. We recommend you [upgrade to Agent 7][12].
```
ruthnaebeck, line 123:
```suggestion
**Note**: Agent v5 uses Python 2 which reached end-of-life on January 1, 2021. Datadog recommends [upgrading to Agent v7][12].
```

ruthnaebeck (changes requested):

ruthnaebeck (approved):
👍 for docs

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,136 @@
...
+
+{{< tabs >}}
+{{% tab "Debian/Ubuntu" %}}
+
+Run the following commands on the host:
+
+```bash
+$ curl -o /tmp/DATADOG_APT_KEY_F14F620E https://keys.datadoghq.com/DATADOG_APT_KEY_F14F620E.public
+$ sudo apt-key add /tmp/DATADOG_APT_KEY_F14F620E
+$ cat /tmp/DATADOG_APT_KEY_F14F620E | sudo gpg --import --batch --no-default-keyring --keyring /usr/share/keyrings/datadog-archive-keyring.gpg
```
bkabrda, line 71: This is the scenario which assumes that the keyring file doesn't exist yet. I just tested it and running the command fails when the keyring doesn't exist. We should add `touch /usr/share/keyrings/datadog-archive-keyring.gpg` before this command.

content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -0,0 +1,137 @@
...
+## Manual update
+
+{{< tabs >}}
+{{% tab "Debian/Ubuntu" %}}
+
+Run the following commands on the host:
+
+```bash
+$ curl -o /tmp/DATADOG_APT_KEY_F14F620E https://keys.datadoghq.com/DATADOG_APT_KEY_F14F620E.public
+$ sudo apt-key add /tmp/DATADOG_APT_KEY_F14F620E
```
bkabrda, line 70: We might add a comment saying that this is only necessary on Debian < 9 and Ubuntu < 16, but it's not strictly necessary.

albertvaka: I think it's better to keep it simple and not add more conditionals.

bkabrda: I agree with not adding conditionals. I just think users might wonder why the key has to be added twice, so a comment explaining it could be helpful.

albertvaka: It's a bit hard to explain because it's a bit more confusing than "only necessary on Debian < 9 and Ubuntu < 16". 

Actually, the only part that's needed on all distros is the `apt-key add` because the keyring file is not going to be used by the `sources.list.d` file: if it is used it means it was created with an install method that also created the keyring, so this is not needed.

The only reason I'm adding the keyring file is because the way we are telling people to check if a system is in a good/bad state is to check for the file. The keyring file could even be empty and everything would still hold.

bkabrda: Oh you're right, we would need to modify the sources.list.d file, which would be even more complicated, so it's probably best to use apt-key add 👍

PR #11866, opened 2021-10-04
Monitor template variables for all attributes & tags

dussault-antoine:
### What does this PR do?

Update the monitor variables documentation page to add attribute and tag variables syntax guidelines to render all attributes and tags associated with events (for log, rum and trace analytics monitors)

### Motivation

Documentation for this [new feature](https://datadoghq.atlassian.net/browse/MNOTIF-989)

### Preview

https://docs-staging.datadoghq.com/antoine.dussault/monitor_template_variables_attributes_tags/monitors/notify/variables

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/monitors/notify/variables.md
+++ b/content/en/monitors/notify/variables.md
@@ -108,6 +108,8 @@ The message will render the `error.message` attribute of a chosen log matching t
 
 <div class="alert alert-info"><strong>Note</strong>: If the picked event does not contain the attribute or the tag key, the variable will render empty in the notification message. To avoid missing notifications, using these variables for routing notification with {{#is_match}} handles is not recommended.</div>
 
+##### Top level attributes
+
 Logs, spans and RUM events also have some first level attributes, that you can use in variables following the following syntax :
 
 | Monitor type    | Variable syntax                       | First level attributes |
@@ -115,6 +117,8 @@ Logs, spans and RUM events also have some first level attributes, that you can u
 | RUM             | `{{rum.key}}`                         | `service`, `status`, `timestamp` |
 | Trace Analytics | `{{span.key}}`                        | `env`, `operation_name`, `resource_name`, `service`, `status`, `span_id`, `timestamp`, `trace_id`, `type` |
 | Log             | `{{log.key}}`                         | `message`, `service`, `status`, `source`, `span_id`, `timestamp`, `trace_id` |
+
+##### Explorer link
 
 ### Check monitor variables
 
--- a/content/en/monitors/notify/variables.md
+++ b/content/en/monitors/notify/variables.md
@@ -84,9 +84,9 @@ If your facet has periods, use brackets around the facet, for example:
 {{ [@network.client.ip].name }}
 ```
 
-### Attributes and Tag variables
+### Attribute and Tag variables
 
-_Available for Log monitor, Trace Analytics monitor (APM) and RUM monitor_
+_Available for [Log monitor][3], [Trace Analytics monitor][4] (APM) and [RUM monitor][5]_
 
 To include **any** attribute or tag from an event matching the monitor query, use the following variables :
 
@@ -96,9 +96,9 @@ To include **any** attribute or tag from an event matching the monitor query, us
 | Trace Analytics |  `{{span.attributes.key}}` / `{{span.tags.key}}`        |
 | Log             |  `{{log.attributes.key}}` / `{{log.tags.key}}`          |
 
-For any `key:value` pair, the variable `{{log.attributes.key}}` renders `value` in the alert message.
+For any `key:value` pair, the variable `{{log.tags.key}}` renders `value` in the alert message.
 
-**Example**: if a log monitor is grouped by `@error_code`, to include the error message in the notification message, use the variable :
+**Example**: if a log monitor is grouped by `@error.code`, to include the error message in the notification message, use the variable :
 
 ```text
 {{ log.attributes.[error.message] }}
@@ -108,9 +108,9 @@ The message will render the `error.message` attribute of a chosen log matching t
 
 <div class="alert alert-info"><strong>Note</strong>: If the picked event does not contain the attribute or the tag key, the variable will render empty in the notification message. To avoid missing notifications, using these variables for routing notification with {{#is_match}} handles is not recommended.</div>
 
-##### Top level attributes
+#### First level attributes
 
-Logs, spans and RUM events also have some first level attributes, that you can use in variables following the following syntax :
+Logs, spans and RUM events have generic first level attributes, that you can also use in variables following the following syntax :
 
 | Monitor type    | Variable syntax                       | First level attributes |
 |-----------------|---------------------------------------|------------------------|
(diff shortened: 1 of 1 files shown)


content/en/monitors/notify/variables.md
```diff
@@ -303,6 +202,153 @@ This is the escalation message @dev-team@company.com
...
+If your monitor triggers an alert for each `host`, then the tag variables `{{host.name}}` and `{{host.ip}}` are available as well as any host tag that is available on this host.
+To see a list of tag variables based on your tag selection, click **Use message template variables** in the **Say what's happening** section.
+Some specific host metadata are available as well:
+
+- Agent Version : {{host.metadata_agent_version}}
+- Machine : {{host.metadata_machine}}
+- Platform : {{host.metadata_platform}}
+- Processor : {{host.metadata_processor}}
+
+##### Query group by a facet
```
NBParis, line 230: Not sure that this one should be its own section. It's definitely part of the group by element and is a common case that should show up before the group by host which is a specific one.

content/en/monitors/notify/variables.md
```diff
@@ -303,6 +202,153 @@ This is the escalation message @dev-team@company.com
 {{% /tab %}}
 {{< /tabs >}}
 
+
+## Notification message enrichment
+
+### Attribute and tag variables
+
+#### Multi-alert tag variables
```
NBParis, line 210: We should likely add a screenshot to illustrate this

content/en/monitors/notify/variables.md
NBParis: We could potentially use tabs to get tags and attribute template variable separated instead of two sections. WDYT?

content/en/monitors/notify/variables.md
```diff
@@ -303,6 +202,153 @@ This is the escalation message @dev-team@company.com
...
+
+```text
+{{ [@network.client.ip].name }}
+```
+
+#### Matching attribute/tag variables
+
+_Available for [Log monitor][1], [Trace Analytics monitor][2] (APM) and [RUM monitor][3]_
+
+To include **any** attribute or tag from an event matching the monitor query, use the following variables :
```
NBParis:
```suggestion
To include **any** attribute or tag from **one** event matching the monitor query, use the following variables :
```

content/en/monitors/notify/variables.md
```diff
@@ -303,6 +202,153 @@ This is the escalation message @dev-team@company.com
...
+
+| Monitor type    | Variable syntax                                         |
+|-----------------|---------------------------------------------------------|
+| RUM             |  `{{rum.attributes.key}}` / `{{rum.tags.key}}`          |
+| Trace Analytics |  `{{span.attributes.key}}` / `{{span.tags.key}}`        |
+| Log             |  `{{log.attributes.key}}` / `{{log.tags.key}}`          |
+
+For any `key:value` pair, the variable `{{log.tags.key}}` renders `value` in the alert message.
+
+**Example**: if a log monitor is grouped by `@error.code`, to include the error message in the notification message, use the variable :
```
NBParis, line 274: Let's do an example section and add screenshot to illustrate the example

content/en/monitors/notify/variables.md
```diff
@@ -303,6 +202,163 @@ This is the escalation message @dev-team@company.com
...
 {{< /tabs >}}
 
+
+## Notification message enrichment
+
+### Attribute and tag variables
+
+#### Multi-alert variables
+
+Multi-alert variables can be used in [multi-alert monitors][1] based on the dimension selected in the multi-alert group box.
```
NBParis, line 212:
```suggestion
Use multi-alert variables in [multi-alert monitors][1] to enrich your notification message based on the dimension selected in the multi-alert group box.
```

NBParis: Crusade against `can`

content/en/monitors/notify/variables.md
```diff
@@ -303,6 +202,163 @@ This is the escalation message @dev-team@company.com
...
+Enrich notification to dynamically include the value associated with the group by dimension in each alert.
```
NBParis, line 213: Doesn't this sentence repeat the first one? We should likely merge the two.

content/en/monitors/notify/variables.md
```diff
@@ -303,6 +202,163 @@ This is the escalation message @dev-team@company.com
...
+
+##### Query group by tag
```
NBParis, line 215: Let's do two tabs: tags and attributes instead of two sections. 

As the two is the exact same behaviour, the only difference is that for attribute a @ is required. However the @ is included in the query path so in a way it's the same thing but is worth calling out. 

WDYT about using tabs instead?

content/en/monitors/notify/variables.md
```diff
@@ -303,6 +202,163 @@ This is the escalation message @dev-team@company.com
 {{% /tab %}}
 {{< /tabs >}}
 
+
+## Notification message enrichment
+
+### Attribute and tag variables
+
+#### Multi-alert variables
```
NBParis, line 210: I'm a big fan of Screenshots to illustrate something, we should likely find a way to show the group by in the query and the notification message using its value.

content/en/monitors/notify/variables.md
```diff
@@ -303,6 +202,163 @@ This is the escalation message @dev-team@company.com
...
+
+```text
+{{ log.attributes.[error.message] }}
+```
+
+The message renders the `error.message` attribute of a chosen log matching the query, **if the attribute exists**.
+
+<div class="alert alert-info"><strong>Note</strong>: If the picked event does not contain the attribute or the tag key, the variable renders empty in the notification message. To avoid missing notifications, using these variables for routing notification with {{#is_match}} handles is not recommended.</div>
+
+##### First level attributes
```
NBParis, line 286: Should we call this as Special attributes or Reserved attributes? First level sounds strange.

content/en/monitors/notify/variables.md
```diff
@@ -303,6 +202,163 @@ This is the escalation message @dev-team@company.com
...
+
+Logs, spans and RUM events have generic first level attributes, that you can also use in variables following the following syntax :
+
+| Monitor type    | Variable syntax                       | First level attributes |
+|-----------------|---------------------------------------|------------------------|
+| RUM             | `{{rum.key}}`                         | `service`, `status`, `timestamp` |
```
NBParis, line 292: We likley don't want to highlight timestamp there though. Because it will be a epoch ms value which is barely usable.

content/en/monitors/notify/variables.md
```diff
@@ -303,6 +202,163 @@ This is the escalation message @dev-team@company.com
...
+| Trace Analytics | `{{span.key}}`                        | `env`, `operation_name`, `resource_name`, `service`, `status`, `span_id`, `timestamp`, `trace_id`, `type` |
+| Log             | `{{log.key}}`                         | `message`, `service`, `status`, `source`, `span_id`, `timestamp`, `trace_id` |
+
+If the matching event does not contain the attribute in its definition, the variable is rendered empty.
+
+##### Explorer link
+
+Use `{{log.link}}`, `{{rum.link}}` and `{{span.link}}` to enrich the notification with a link to the log, rum or trace explorer, scoped on the events matching the query.
+
+#### Check monitor variables
```
NBParis, line 302: Sounds like we should use Tabs instead of sections to put them on the same level.

content/en/monitors/notify/variables.md
```diff
@@ -303,6 +202,163 @@ This is the escalation message @dev-team@company.com
...
+
+For check monitor variables (custom check and integration check), the variable `{{check_message}}` is available and renders the message specified in the custom check or the integration check.
+
+#### Composite monitor variables
```
NBParis, line 306: Let's use tabs for this as well no?

content/en/monitors/notify/variables.md
dussault-antoine: changing this one 👍

content/en/monitors/notify/variables.md
dussault-antoine: do we want to remove it from the available top level attributes ?
I'd say it's up to RUM team to decide this

PR #12139, opened 2021-10-29
Remove overhead sentence for Java tracing

hugokash:
### Motivation
This sentence has been there for a long time. It is wrong, as we haven't evaluated the overhead, and often source of confusion for our customers and support.
It happens quite often that our library's overhead is bigger than 3%, and it's often expected

### Preview
https://docs-staging.datadoghq.com/hugokash-patch-1/tracing/setup_overview/setup/java/?tab=containers#performance


cswatt (approved):

dougqh (commented):
I see, so we'd already changed this from a hard guarantee.

I'm still fine with removing it.  If it is too high, we're going to hear about it regardless of what our docs say.

PR #12140, opened 2021-10-29
Updating Configure Logback Section

john-reiner:
The first step of the Configure Logback section is we are giving an example implementation for the “LogstashTcpSocketAppender” using `<remoteHost>` and `<port>`. This still works, but it has been deprecated and the project documentation suggests using `<destination>` instead: https://github.com/logstash/logstash-logback-encoder#tcp-appenders

### What does this PR do?

Updates the Configure Logback section to replace `<remoteHost>` and `<port>` to `<destination>` as per https://github.com/logstash/logstash-logback-encoder#tcp-appenders

### Motivation

Internal Zendesk ticket: https://datadog.zendesk.com/agent/tickets/574676

### Preview

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/logs/log_collection/java.md
+++ b/content/en/logs/log_collection/java.md
@@ -379,7 +379,7 @@ Use the [logstash-logback-encoder][4] logging library along with Logback to stre
       <encoder class="net.logstash.logback.encoder.LogstashEncoder" />
     </appender>
     <appender name="JSON_TCP" class="net.logstash.logback.appender.LogstashTcpSocketAppender">
-      <destination>tcp-intake.logs.datadoghq.eu:1883</destination>
+      <destination>intake.logs.datadoghq.com:10514</destination>
       <keepAliveDuration>20 seconds</keepAliveDuration>
       <encoder class="net.logstash.logback.encoder.LogstashEncoder">
           <prefix class="ch.qos.logback.core.encoder.LayoutWrappingEncoder">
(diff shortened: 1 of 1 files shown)


cswatt:
@john-reiner I don't have a ton of context on this, but it looks like this is an EU endpoint. Is this site-agnostic?

john-reiner:
@cswatt yes, this is for the eu logs. I am not sure if thats what you meant by site-agnostic? It was pointed out by a customer that if you check out the [TCP Appenders example](https://github.com/logstash/logstash-logback-encoder#tcp-appenders), you can see that the example is not using <remoteHost> and <port>  but using <destination> instead.

kayayarai:
@john-reiner I think what @cswatt was pointing out is that the you updated the US example (Lines 275-397) with a EU URL, when maybe that should remain a US endpoint URL (`intake.logs.datadoghq.com`) and the EU example (lines 404-428) should also be updated (lines 411-412) should be updated with the `<destination>` code as well. Can you confirm that the `destination` syntax should be used in both examples?

john-reiner:
@kayayarai ah yes thank you for the clarification. Yes that URL should be for eu only. Should I do another pull request?

kayayarai:
@john-reiner No, you can commit more changes to this branch.

apigirl (dismissed):

kayayarai:
Hi @john-reiner I updated this to reflect the change I _think_ you meant to make. Please confirm it's correct now.

kayayarai:
In the meantime, https://github.com/DataDog/documentation/pull/12218 happened, where the `<destination>` tag was added, but so was `<ssl/>` which uses different port numbers. I'm going to close this one, as already updated. Thanks for the contribution!

PR #12169, opened 2021-11-02
Run Yarn in Documentation Repo

alai97:
### What does this PR do?

Add an instruction to run `yarn` in your documentation repo to the public documentation README. 

### Motivation

Slack

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@ Built with [hugo][1], a static website generation tool.
 
 6. Download the documentation repo `git clone https://github.com/DataDog/documentation.git`
 
+7. Run `yarn` in the documentation repo
+
 ### Run the server
 
 Inside `documentation/` folder, create a `Makefile.config` file from the [Makefile.config.example][5]


HugoPct (approved):

README.md
```diff
@@ -18,6 +18,8 @@ Built with [hugo][1], a static website generation tool.
 
 6. Download the documentation repo `git clone https://github.com/DataDog/documentation.git`
 
+7. Run `yarn` in the documentation repo
+
```
davidejones, line 22: I don't think we should add this step as typically the make file will do this for you and will use `yarn install --frozen-lockfile` to ensure its using the frozen versions.

davidejones: Maybe we can strongly recommend pyenv further up? instead of just the callout. Also we could add or link to some hugo module troubleshooting. What do you think?

HugoPct: yes I agree with you.
Emphasising a bit more on pyenv would be a good solution. 
Link to hugo module troubleshooting too

alai97: @davidejones the thing is the corp-hugo repo is private whereas the Documentation repo is public, so adding the link to the troubleshooting wiki does not work 🧐

PR #12449, opened 2021-12-01
Update which Agent services run as root/non-root

albertvaka:
### What does this PR do?
Update the "Running as an unprivileged user" about which Agent services run as root/non-root.

### Motivation
The previous docs where incorrect.

### Preview
https://docs-staging.datadoghq.com/albertvaka/process-agent-runs-as-root/security/agent/#running-as-an-unprivileged-user

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/security/agent.md
+++ b/content/en/security/agent.md
@@ -64,7 +64,7 @@ If you believe you've discovered a bug in Datadog's security, get in touch at [s
 
 ## Running as an unprivileged user
 
-By default, the Agent runs as the `dd-agent` user on Linux and as the `ddagentuser` account on [Windows][18]. Note the `system-probe` and `security-agent` services are an exception to this, and still need to run as `root` on Linux and `LOCAL_SYSTEM` on Windows.
+By default, the Agent runs as the `dd-agent` user on Linux and as the `ddagentuser` account on [Windows][18]. Note there are two exceptions to this: the `system-probe` which runs as `root` on Linux and as `LOCAL_SYSTEM` on Windows and the `process-agent` which runs as `LOCAL_SYSTEM` on Windows.
 
 ## Secrets management
 
--- a/content/en/security/agent.md
+++ b/content/en/security/agent.md
@@ -64,7 +64,7 @@ If you believe you've discovered a bug in Datadog's security, get in touch at [s
 
 ## Running as an unprivileged user
 
-By default, the Agent runs as the `dd-agent` user on Linux and as the `ddagentuser` account on [Windows][18]. Note there are two exceptions to this: the `system-probe` which runs as `root` on Linux and as `LOCAL_SYSTEM` on Windows and the `process-agent` which runs as `LOCAL_SYSTEM` on Windows.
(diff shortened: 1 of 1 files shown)


KSerrania (approved):

content/en/security/agent.md
```diff
@@ -64,7 +64,9 @@ If you believe you've discovered a bug in Datadog's security, get in touch at [s
 
 ## Running as an unprivileged user
 
-By default, the Agent runs as the `dd-agent` user on Linux and as the `ddagentuser` account on [Windows][18]. Note the `system-probe` and `security-agent` services are an exception to this, and still need to run as `root` on Linux and `LOCAL_SYSTEM` on Windows.
+By default, the Agent runs as the `dd-agent` user on Linux and as the `ddagentuser` account on [Windows][18]. Note there are some exceptions to this: the `system-probe` which runs as `root` on Linux and as `LOCAL_SYSTEM` on Windows, the `process-agent` which runs as `LOCAL_SYSTEM` on Windows, and the `security-agent` which runs as `root` on Linux.
+
+The `process-agent` can be configured to run as the unprivileged `ddagentuser` on Windows, but some functionality will be missing. In specific, the command line of each process won't be available in Datadog.
```
derekwbrown, line 69: I don't know that this is true.
I mean, we could in theory do that, but AFAIK there's nothing in the install to do this, and the result will be very continuous errors (there's nothing in the code to detect we're _not_ running as root so the various api calls would throw errors all over) .  Again, AFAIK.

jtappa: Sounds like we need to confirm or correct this before adding it to the docs.

robertjli: Is there a reason we want to state this particular exception? TBH I don't _quite_ remember what the effect is, but if the command line is not available then the process can't be reasonably ingested by our pipeline, so it's either dropped at by the agent or very early in our pipeline. Why not just leave it be and "force" the customer to run with privileges?

albertvaka: I removed that line.

robertjli (approved):

content/en/security/agent.md
```diff
@@ -64,7 +64,7 @@ If you believe you've discovered a bug in Datadog's security, get in touch at [s
 
 ## Running as an unprivileged user
 
-By default, the Agent runs as the `dd-agent` user on Linux and as the `ddagentuser` account on [Windows][18]. Note the `system-probe` and `security-agent` services are an exception to this, and still need to run as `root` on Linux and `LOCAL_SYSTEM` on Windows.
+By default, the Agent runs as the `dd-agent` user on Linux and as the `ddagentuser` account on [Windows][18]. Note there are some exceptions to this: the `system-probe` which runs as `root` on Linux and as `LOCAL_SYSTEM` on Windows, the `process-agent` which runs as `LOCAL_SYSTEM` on Windows, and the `security-agent` which runs as `root` on Linux.
```
ruthnaebeck, line 67:
```suggestion
By default, the Agent runs as the `dd-agent` user on Linux and as the `ddagentuser` account on [Windows][18]. The exceptions are as follows:

- The `system-probe` runs as `root` on Linux and as `LOCAL_SYSTEM` on Windows.
- The `process-agent` runs as `LOCAL_SYSTEM` on Windows.
- The `security-agent` runs as `root` on Linux.
```

ruthnaebeck (changes requested):

derekwbrown (approved):
Approve (with ruth's suggestions)

ruthnaebeck (approved):

PR #12559, opened 2021-12-13
[2022 key rotation docs] Recommend updating the install method

albertvaka:
### What does this PR do?
Recommend updating the install method instead of trusting the
new key manually, since this will continue to work in the future.

### Preview
https://docs-staging.datadoghq.com/albertvaka/signing-keys-recommend-updating-install-method/agent/faq/linux-agent-2022-key-rotation

### DIFF
--- a/content/en/agent/faq/linux-agent-2022-key-rotation.md
+++ b/content/en/agent/faq/linux-agent-2022-key-rotation.md
@@ -60,6 +60,8 @@ Check that the correct key URLs are configured for this repository.
 
 ## Manual update
 
+We encourage users to use one of the install methods above which automatically trust the new GPG key, as they will also automatically fetch future keys as well. If this is not an option, these instructions will download and trust the new key.
+
 {{< tabs >}}
 {{% tab "Debian/Ubuntu" %}}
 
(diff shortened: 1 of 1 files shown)


content/en/agent/faq/linux-agent-2022-key-rotation.md
```diff
@@ -60,6 +60,8 @@ Check that the correct key URLs are configured for this repository.
 
 ## Manual update
 
+We encourage users to use one of the install methods above which automatically trust the new GPG key, as they will also automatically fetch future keys as well. If this is not an option, these instructions will download and trust the new key.
```
sarina-dd, line 63:
```suggestion
Datadog encourages you to use one of the install methods above, which automatically trusts the new GPG key as it also automatically fetches future keys. If this is not an option, the following instructions also downloads and trusts the new key.
```

sarina-dd (changes requested):

content/en/agent/faq/linux-agent-2022-key-rotation.md
albertvaka:
```suggestion
Datadog encourages you to use one of the install methods above, which trust the new GPG key as well as all future keys automatically. If this is not an option, the following instructions manually download and trust the new key.
```

KSerrania (approved):

sarina-dd (approved):

PR #12628, opened 2021-12-18
APM > .NET > Update trace logs correlation docs for 2.0

zacharycmontoya:
### What does this PR do?
Updates the setup instructions for logs correlation with the upcoming .NET Tracer 2.0.1 release

### Motivation
We are releasing a new .NET Tracer 2.0.1 release which requires automatic instrumentation for the logs injection to work

### Preview
https://docs-staging.datadoghq.com/zach.montoya/apm-dotnet-logs-injection-2.0/tracing/connect_logs_and_traces/dotnet/?tab=serilog

### Additional Notes
N/A

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/tracing/connect_logs_and_traces/dotnet.md
+++ b/content/en/tracing/connect_logs_and_traces/dotnet.md
@@ -25,7 +25,7 @@ You can set up your logging library and .NET tracing configurations so that trac
 Configure the .NET Tracer with [Unified Service Tagging][1] for the best experience and helpful context when correlating application traces and logs.
 
 The .NET Tracer supports the following logging libraries:
-- [Serilog][2] (v2.0+)
+- [Serilog][2] (v1.4+)
 - [log4net][3]
 - [NLog][4]
 - [Microsoft.Extensions.Logging][5] (added in v1.28.6) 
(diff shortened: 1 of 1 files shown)


jaycdave88:
@zacharycmontoya could we add a table to the "The .NET Tracer supports the following logging libraries:" 

(Or including that text) 

To display: 
- lib log dll 
- tracer version for support with 1.x & 2.x (similar to what we did for comp corp doc) 
- version of lib log dll we support
- & what ever else you think would be good

pierotibou (approved):

zacharycmontoya:
> @zacharycmontoya could we add a table to the "The .NET Tracer supports the following logging libraries:"
> 
> (Or including that text)
> 
> To display:
> 
> * lib log dll
> * tracer version for support with 1.x & 2.x (similar to what we did for comp corp doc)
> * version of lib log dll we support
> * & what ever else you think would be good

We no longer use LibLog so the only information we'd need to convey is support between 1.x and 2.x. The support is identical among versions (except for Tracer 2.x which improves support slightly), so I don't think we really need a two-column table. If you feel strongly otherwise, maybe we can revisit the table idea after the new year so we can push out this updated info ASAP.

sarina-dd (approved):

PR #12682, opened 2021-12-28
add error tracking to collection/retention table

vandonr:
On a suggestion from @nicholas-devlin 

Error tracking pulls data from RUM and APM. I wasn't too sure how to formulate it. Decided to list error tracking twice, as a sub-product of RUM and APM. And listed APM  spans as the data source, but we could also say that it's the agent like for other APM entries.

60s precision because errors are aggregated in 60 seconds buckets, and 2 minutes collection interval because we keep the 1 minute bucket open for 2 minutes before we flush it.


vandonr:
https://docs-staging.datadoghq.com/raphael.vandon/et/developers/guide/data-collection-resolution-retention/

maycmlee (approved):
👍  from docs

PR #12782, opened 2022-01-11
use sandwich analogy in c++ custom instrumentation example

dgoffredo:
In addition to Wan's proposed update for setting-resource-name, use this opportunity to make some sense of the example values for operation name and resource name.


wantsui (approved):
LGTM 👍

PR #12886, opened 2022-01-22
Add note for aiohttp

ZStriker19:
Since additional configuration is required for the aiohttp integration to work, we want to warn/inform users so they're not confused if they don't see the expected spans for the application out of the box: https://ddtrace.readthedocs.io/en/stable/integrations.html#aiohttp

### What does this PR do?
Adds a note for customers that additional setup is required for the aiohttp integration to work

### Motivation
Customer confusion about expectations and how to setup aiohttp integration

---

### Reviewer checklist
- [ ] Check images for PII


kayayarai (approved):
Yes, but I think if we are going to bother with this, we should put it where people will see it :) I moved it and pushed. Will let it build, and then merge it if you are cool with that.

kayayarai:
> 

Oh, no build because of branch name. I built locally and it looks fine. Will merge.

PR #12972, opened 2022-02-01
Update python.md

someboredkiddo:
### What does this PR do?
Adds installation instruction for installing the datadog-lambda-extension in a Container Image an architecture that is compatible with a user's dockerfile.

### Motivation
Avoid other engineers from giving up on upgrading the DataDog extension past Tag 11 because of unfamiliarity with docker architecture hints.

### Preview

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/serverless/installation/python.md
+++ b/content/en/serverless/installation/python.md
@@ -437,6 +437,12 @@ COPY --from=public.ecr.aws/datadog/lambda-extension:<TAG> /opt/extensions/ /opt/
 
 Replace `<TAG>` with either a specific version number (for example, `{{< latest-lambda-layer-version layer="extension" >}}`) or with `latest`. You can see a complete list of possible tags in the [Amazon ECR repository][1].
 
+Beginning with lambda-extension TAG=11, both the linux/amd64 & linux/arm platforms are included in the image's manifest. Make sure to set the proper platform when building your Dockerfile.
+
+```
+docker build --platform "linux/amd64" .
+```
+
 ### Configure the function
 
 1. Set your image's `CMD` value to `datadog_lambda.handler.handler`. You can set this in AWS or directly in your Dockerfile. Note that the value set in AWS overrides the value in the Dockerfile if you set both.


someboredkiddo:
Probably want to do this for all Lambda runtimes - just didn't want to go through that before this copy is approved

nhinsch:
Hi @someboredkiddo, I was not able to replicate a build error when building a Dockerfile including the extension.

Can you elaborate a bit on what happened when you ran `docker build` without the additional argument, versus what you expected to happen? Also, would you mind sharing what base image, if any, you were using to build your container image?

Thanks!

someboredkiddo:
@nhinsch The image will build fine without the `--platform` argument - however when the image is deploy to a Lambda, the Lambda will fail to execute with the below error. 
`Error: fork/exec /opt/extensions/datadog-agent: exec format error Extension.LaunchError`
Guessing this is because the generated will be linux/amd64 but will contain a arm version of the lambda-extension

Just to emphasize - this is **only** an  issue when the Lambda's Docker image is amd64/x86 based

Steps to reproduce
1. Build Docker image that is linux/amd64 based using the public.ecr.aws/datadog/lambda-extension without explicitly specifying the platform arg.  (Sample Dockerfile below) 
ex. `docker build .`
2. Upload that image to ECR
3. Deploy that image to a Lambda selecting the x86_64 architecture option
4. Execute Lambda, the below error will be generated
`Error: fork/exec /opt/extensions/datadog-agent: exec format error Extension.LaunchError`

If step 1. was built with `docker build --platform "linux/amd64" .` , the lambda will execute successfully (posted some more details here https://github.com/DataDog/datadog-lambda-extension/issues/49)

Dockerfile
```dockerfile
FROM amazon/aws-lambda-python:3.8

COPY --from=public.ecr.aws/datadog/lambda-extension:18 /opt/extensions/ /opt/extensions

COPY requirements.txt ./
RUN pip install --upgrade pip && \
    pip install -r requirements.txt

COPY app.py ./

ENV DD_LAMBDA_HANDLER=app.lambda_handler
CMD [ "datadog_lambda.handler.handler" ]
```

app.py
```python
import os
import platform
import sys

def machine():
    """Return type of machine."""
    if os.name == 'nt' and sys.version_info[:2] < (2,7):
        return os.environ.get("PROCESSOR_ARCHITEW6432", 
               os.environ.get('PROCESSOR_ARCHITECTURE', ''))
    else:
        return platform.machine()

def os_bits(machine=machine()):
    """Return bitness of operating system, or None if unknown."""
    machine2bits = {'AMD64': 64, 'x86_64': 64, 'i386': 32, 'x86': 32}
    return machine2bits.get(machine, None)

def lambda_handler(event, context):
  return {
    'statusCode': 200,
    'body': 'Hello world extension 18 - python 3.8',
    'machine': machine(),
    'os_bits': os_bits()
  }
```

requirements.txt
```
datadog-lambda==3.50.0
```

nhinsch:
Hi @someboredkiddo, thanks for the detailed reproduction steps. Before I investigate the issue further, I have one more question: what was the architecture of the machine you used to build the image? If you are building the image locally, are you using a Mac with an Apple silicon chip by any chance?

someboredkiddo:
ah yep - you guessed correctly, I am on a M1. Makes sense Docker looks to the current machine it's building on to determine default architecture (guess I shouldn't have asked for a new laptop ha). I've verified that generating the image with these steps on a x86 machine creates a Docker image that successfully executes on Lambda.

Not sure what the best way to document this particular gotcha is - I understand we don't want to overcomplicate the setup instructions / turn the docs into a Docker tutorial, but I imagine more folks will fall into this build mistake for local development as M1 macs get more adoption. I guess this is why [aws is so tag happy](https://gallery.ecr.aws/lambda/provided) with explicit tags for `latest-x86_64` & `latest-arm64` images in addition to a `latest` image that includes both of this platforms 😵.

If we want something in the docs warning about this gotcha I could re-phase as
```
Beginning with lambda-extension TAG=11, both the linux/amd64 & linux/arm platforms are included in the image's manifest. If building a image on a machine with a different architecture than what your lambda will generate, make sure to set the proper platform when building your Dockerfile.
```

Or we could just trust google to index this convo and close this. I'm fine with any of these approaches & will punt to you @nhinsch to decide.

nhinsch:
Let's close the PR for now -- I think you are right that it is more of a Docker gotcha rather than an issue pertaining to Datadog per se. But this conversation was still illuminating, we will be on the lookout for similar issues in the future and if necessary we can add something to our documentation.

someboredkiddo:
sounds good - thanks for the help @nhinsch !

PR #13022, opened 2022-02-04
Document trigger latency in log-based monitor

sveniu:
### What does this PR do?
Ref Datadog support ticket 657964.

This change documents the latency between a log event being ingested, and a log-based monitor being able to evaluate/trigger on the ingested event. In my experience, the latency is a bit above 180 seconds. Datadog support confirms the latency in the referenced support ticket, and recommends using the combination of log-derived metric and a metric-based monitor instead.

### Motivation
I'm unable to find any mention of latency between log indexing and log-based monitor triggering. With Datadog support confirming the latency issue, it would make sense that it is documented. Then users can quickly determine whether a log-based metric would suit their use case, or whether to immediately go for a log-derived metric + metric monitor instead.

It would of course be better if this inherent latency was fixed, rather than documenting it. Maybe this PR can help push a fix forward.

### Additional Notes
I put the info box prominently at the top, which is maybe a bit too much. At least it would save users time if their monitor is timing sensitive.

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/monitors/create/types/log.md
+++ b/content/en/monitors/create/types/log.md
@@ -19,6 +19,8 @@ further_reading:
 
 Once [log management is enabled][1] for your organization, you can create a logs monitor to alert you when a specified type of log exceeds a user-defined threshold over a given period of time.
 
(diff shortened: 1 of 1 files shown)


txominpelu (approved):

maycmlee (commented):
Hello! We have had a chance to look at this internally. Currently, this is an internal metric and for most customers it is within a few seconds. For those reasons, we don’t document this today. We’ll re-evaluate exposing this metric in the future. Thanks for the suggestion!

PR #13026, opened 2022-02-05
Bowen/update node startup debug docs

bojbrook:
### What does this PR do?
Updates the APM troubleshooting debug and startup log guides for the NodeJS tracer

### Motivation
Breaking changes to the node tracer in the release of the 2.X version  - https://github.com/DataDog/dd-trace-js/blob/master/MIGRATING.md 

### Preview

Debug logs: https://docs-staging.datadoghq.com/bowen/update_node_startup_debug_docs/tracing/troubleshooting/tracer_debug_logs

Startup logs: https://docs-staging.datadoghq.com/bowen/update_node_startup_debug_docs/tracing/troubleshooting/tracer_startup_logs

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/tracing/troubleshooting/tracer_debug_logs.md
+++ b/content/en/tracing/troubleshooting/tracer_debug_logs.md
@@ -88,6 +88,7 @@ const tracer = require('dd-trace').init({
   debug: true
 })
 ```
+**Note:** For version 2.X debug mode can only be configured by the environment variable `DD_TRACE_DEBUG`
 
 **Application Logs**
 
--- a/content/en/tracing/troubleshooting/tracer_startup_logs.md
+++ b/content/en/tracing/troubleshooting/tracer_startup_logs.md
@@ -165,6 +165,8 @@ The Go Tracer prints one of two possible diagnostic lines, one for when the Agen
 {{< /programming-lang >}}
 {{< programming-lang lang="nodejs" >}}
 
+**Note:** Startup logs are disabled by default in version 2.X+. They can enabled using the environment variable `DD_TRACE_STARTUP_LOGS=true`.
+
 **Configuration:**
 
 ```text


content/en/tracing/troubleshooting/tracer_debug_logs.md
```diff
@@ -88,6 +88,7 @@ const tracer = require('dd-trace').init({
   debug: true
```
rochdev, line 88: This should also be removed since this option no longer exists.

content/en/tracing/troubleshooting/tracer_debug_logs.md
```diff
@@ -88,6 +88,7 @@ const tracer = require('dd-trace').init({
...
 })
 ```
+**Note:** For version 2.X debug mode can only be configured by the environment variable `DD_TRACE_DEBUG=true`.
```
rochdev, line 91: Can we do the opposite and make the doc all for v2 and then the note would simply note the deprecation of the option that was available in v1?

content/en/tracing/troubleshooting/tracer_startup_logs.md
```diff
@@ -165,6 +165,8 @@ The Go Tracer prints one of two possible diagnostic lines, one for when the Agen
 {{< /programming-lang >}}
 {{< programming-lang lang="nodejs" >}}
 
+**Note:** Startup logs are disabled by default in version 2.X+. They can enabled using the environment variable `DD_TRACE_STARTUP_LOGS=true`.
```
rochdev, line 168: Similar to the above, I would word this as a deprecation instead of a note specific to v2.

content/en/tracing/troubleshooting/tracer_debug_logs.md
bojbrook: @rochdev Is this more what you had in mind?

To enable debug mode for the Datadog Node.js Tracer, use the environment variable `DD_TRACE_DEBUG=true`. 

**Note:** For versions below 2.X, debug mode can be enabled programmatically inside the tracer initialization.
```javascript

const tracer = require('dd-trace').init({
  debug: true
})
```

rochdev: Probably more along the line of "For versions below 2.X, debug mode could be enabled programmatically inside the tracer initialization but this is no longer available.". And I would remove the code example. The reasoning is that when code exists in any documentation, some users end up copying that code even it's explicitly stated to not use it, so it's better to avoid code completely and just have a note for users who are upgrading.

bojbrook: Just updated the language to reflect this

content/en/tracing/troubleshooting/tracer_debug_logs.md
rochdev: There are still other occurrences of this in the file.

bojbrook: Removed the other occurrences of the `debug:true `,

content/en/tracing/troubleshooting/tracer_debug_logs.md
```diff
@@ -108,8 +104,7 @@ const tracer = require('dd-trace').init({
   logger: {
     debug: message => logger.trace(message),
     error: err => logger.error(err)
-  },
-  debug: true
+  }
```
rochdev, line 107: The above paragraph still has instructions using the old `debug` configuration.

content/en/tracing/troubleshooting/tracer_startup_logs.md
```diff
@@ -165,6 +165,9 @@ The Go Tracer prints one of two possible diagnostic lines, one for when the Agen
 {{< /programming-lang >}}
 {{< programming-lang lang="nodejs" >}}
 
+Startup logs are disabled by default starting in verson 2.X+ of the tracer. They can enabled using the environment variable `DD_TRACE_STARTUP_LOGS=true`.
```
rochdev, line 168: Was there any mention of the startup logs previously? It seems to me like this sentence is not super useful with its current location.

bojbrook: I added this sentence so as not to cause confusion on why Startup logs aren't present as they used to be enabled by default. We have this [paragraph at the bottom](https://docs.datadoghq.com/tracing/troubleshooting/tracer_startup_logs/#disabling-startup-logs) that on how to disable startup logs but this section is present for all languages so would need a bigger document overhaul to change this.

content/en/tracing/troubleshooting/tracer_debug_logs.md
bojbrook: I have removed this reference

rochdev (approved):

sarina-dd (approved):

PR #13217, opened 2022-02-24
[CIAPP] explicitly exclude http mode for jenkins

alxch-:
### What does this PR do?

This PR adds a clarification about the fact the HTTP mode is not supported for CI App for the Jenkins plugin.

### Motivation

We had some customers wondering if they could use the HTTP mode, so we clarify it here.

### Preview

https://docs-staging.datadoghq.com/alexc/CIAPP-jenkins-http/continuous_integration/setup_pipelines/jenkins/?tab=usingui

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/continuous_integration/setup_pipelines/jenkins.md
+++ b/content/en/continuous_integration/setup_pipelines/jenkins.md
@@ -44,7 +44,7 @@ Install and enable the [Datadog Jenkins plugin][3] v3.3.0 or newer:
 
 1. In your Jenkins instance web interface, go to **Manage Jenkins > Configure System**.
 2. Go to the `Datadog Plugin` section, scrolling down the configuration screen.
-3. Select the `Datadog Agent` mode. Data collection using Datadog API URL and API key is **not supported**.
+3. Select the `Datadog Agent` mode. CI Visibility is **not supported** using Datadog API URL and API key.
 4. Configure the `Agent` host.
 5. Configure the `Traces Collection` port (default `8126`).
 6. Click on `Enable CI Visibility` checkbox to activate it.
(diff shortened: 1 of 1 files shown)


content/en/continuous_integration/setup_pipelines/jenkins.md
```diff
@@ -44,7 +44,7 @@ Install and enable the [Datadog Jenkins plugin][3] v3.3.0 or newer:
 
 1. In your Jenkins instance web interface, go to **Manage Jenkins > Configure System**.
 2. Go to the `Datadog Plugin` section, scrolling down the configuration screen.
-3. Select the `Datadog Agent` mode.
+3. Select the `Datadog Agent` mode. Data collection using Datadog API URL and API key is **not supported**.
```
drodriguezhdez, line 47:
```suggestion
3. Select the `Datadog Agent` mode. CI Visibility is **not supported** using Datadog API URL and API key.
```

content/en/continuous_integration/setup_pipelines/jenkins.md
```diff
@@ -172,7 +172,7 @@ Second, enable job log collection on the Datadog Plugin:
 
 1. In the web interface of your Jenkins instance, go to **Manage Jenkins > Configure System**.
 2. Go to the `Datadog Plugin` section, scrolling down the configuration screen.
-3. Select the `Datadog Agent` mode.
+3. Select the `Datadog Agent` mode. Data collection using Datadog API URL and API key is **not supported**.
```
drodriguezhdez, line 175: I have my doubts on this one because technically you can send logs to Datadog without an agent, but as you need to select the "Agent" mode to send CI Visibility data, this is a consequence. 
```suggestion
3. Select the `Datadog Agent` mode.
```

drodriguezhdez (commented):
Thanks @alxch- for the PR!

I dropped some comments. I'm wondering if we could use the [Prerequisite](https://docs-staging.datadoghq.com/alexc/CIAPP-jenkins-http/continuous_integration/setup_pipelines/jenkins/?tab=usingui#prerequisite) section that we already have to add this disclaimer. WDYT?

content/en/continuous_integration/setup_pipelines/jenkins.md
alxch-: Should we be more specific then?

```suggestion
3. Select the `Datadog Agent` mode. CI Visibility is **not supported** using Datadog API URL and API key. In order for log collection to work along CI Visibility, the Datadog API URL and API key should not be selected.
```

drodriguezhdez: You cannot select both "Datadog Agent" and "Datadog API URL & API Key" at the same time, so that's why I find it redundant.

drodriguezhdez (approved):

sarina-dd (approved):

PR #13243, opened 2022-03-01
Java app prerequisites update

robertpi:
### What does this PR do?
Updates the installations instructions for Java App Sec. It warns clients that read-only file systems are not supported.

### Motivation
Clients have complained that the solution doesn't work for read-only file systems, so we should make them aware of this up front.

---

### Reviewer checklist
- [ ] Check images for PII


content/en/security_platform/application_security/getting_started/java.md
```diff
@@ -31,13 +31,15 @@ further_reading:
 
    Or one of the following methods, depending on where your application runs:
 
+**Note:** Read-only file systems are not currently supported. The application must have access to a writable `/tmp` directory.
```
cataphract, line 34: It need not necessarily be `/tmp`, though that is the usual default. It can be overridden by setting the `java.io.tmpdir` system property).

jtappa (approved):

PR #13920, opened 2022-05-04
Update preview_links action

hestonhoffman:
Users are confused when the preview links in the GitHub comment don't immediately work. They're only available once the `build_preview` check passes.

Screen Shot 2022-05-04 at 9 53 26 AM

~~Note: I'm going to make a small edit to a markdown file to check if this is working. I'll back the change out when I'm done.~~

### DIFF
--- a/.github/workflows/preview_link.yml
+++ b/.github/workflows/preview_link.yml
@@ -42,7 +42,6 @@ jobs:
         comment-id: ${{ steps.find_comment.outputs.comment-id }}
         edit-mode: replace
         body: |
-          Preview links:
+          Preview links (active after the `build_preview` check completes):
           ${{ env.COMMENT_BODY }}
 
-          Links are active after the `build_preview` check completes.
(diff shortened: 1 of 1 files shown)


.github/workflows/preview_link.yml
```diff
@@ -44,3 +44,5 @@ jobs:
         body: |
           Preview links:
```
kayayarai, line 45:
```suggestion
          Preview links (active after the `build_preview` check completes):
```

.github/workflows/preview_link.yml
```diff
@@ -44,3 +44,5 @@ jobs:
...
           ${{ env.COMMENT_BODY }}
+
+          Links are active after the `build_preview` check completes.
```
kayayarai, line 48:
```suggestion
```

kayayarai (changes requested):
My faith in humanity is less than yours

kayayarai (approved):

PR #14018, opened 2022-05-12
Updating to use hugo js build

davidejones:
### What does this PR do?

This PR removes webpack and uses hugos ability to build javascript instead. The goal is for feature parity with the current js build and to make the minimum amount of changes required to support this work.

The changes:
- moved javascript from src to assets (so hugo can read them)
- `<script>` tags no longer use webpack generated manifest and instead use hugos js.build (so hugo can compile and reference javascript)
- scripts moved to `header-scripts.html` or `footer-scripts.html`. These partials already existed and now we supply some compile options this seemed the perfect place to avoid duplication.
- post deploy sourcemap upload job is updated to find files correctly after the changes
- hugo security config updated so we can read a specific environment variable we were using for version

### Motivation

Faster local development and build step

### Preview

QA the entire site as best as possible especially areas with known javascript specifics.
https://docs-staging.datadoghq.com/david.jones/js/

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII


davidejones:
@stanleegoodspeed Good point, I've removed webpack and babel, there might be some lingering packages but i think this is the majority for now until we do a better audit.

cswatt:
One thing I just caught in QA: there's a link in the left nav **Developers** > **Service Checks** > **Submission - API** which links to `https://docs.datadoghq.com/api/v1/service-checks/` and should redirect to `https://docs.datadoghq.com/api/latest/service-checks/`, but in the preview, it doesn't redirect, instead it 404s.

davidejones:
Hey @cswatt I've looked into this and the missing redirect is unrelated to these changes. This currently happens on any preview site as we don't setup the same kind of redirects for each one that prod has.

cswatt (approved):

stanleegoodspeed (approved):

PR #14117, opened 2022-05-23
Embed agent installation and agentless instructions

fermayo:
@drodriguezhdez proposal to embed agent installation and agentless instructions on each language, for simplicity for the user, which probably will only look at one language at a time.

This is because we have now 3 languages that support agent-based and agentless, 2 languages that only support agent-based, 1 language that only supports agentless, and JUnit which is also agentless but a slightly different case. I feel like trying to separate the instructions causes a lot of indirections and confusion.


drodriguezhdez:
Yeap, it works for me. That was my initial direction in the first proposal, but having it integrated into the instrumentation section. Everything that can reduce the indirections for users is welcome!

PR #14145, opened 2022-05-26
Replace all occurencies of true|false in in configuration docs with 1|0

labbati:
### What does this PR do?
Replaces `false` with `0` and `true` with `1` in all PHP configuration examples.

### Motivation
PHP-FPM won't start if a directive `env[My_ENV] = false` is present. Tested in PHP 7.4 and PHP 8.0.
Error:
```
ERROR: [/etc/php-fpm.d/www.conf:72] empty value
ERROR: Unable to include /etc/php-fpm.d/www.conf from /etc/php-fpm.conf at line 72
ERROR: failed to load configuration file '/etc/php-fpm.conf'
```

For this reason, I replaced all values in configuration and examples, so that users do not get confused.

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/tracing/setup_overview/setup/php.md
+++ b/content/en/tracing/setup_overview/setup/php.md
@@ -134,7 +134,7 @@ Automatic instrumentation captures:
 * Unhandled exceptions, including stacktraces if available
 * A total count of traces (for example, web requests) flowing through the system
 
-**Note**: If your application does not use Composer nor an autoloader registered with `spl_autoload_register()`, set the environment variable, `DD_TRACE_NO_AUTOLOADER=true`, to enable automatic instrumentation.
+**Note**: If your application does not use Composer nor an autoloader registered with `spl_autoload_register()`, set the environment variable, `DD_TRACE_NO_AUTOLOADER=1`, to enable automatic instrumentation.
 
 ## Configuration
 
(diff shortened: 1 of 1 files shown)


content/en/tracing/setup_overview/setup/php.md
```diff
@@ -134,7 +134,7 @@ Automatic instrumentation captures:
 * Unhandled exceptions, including stacktraces if available
 * A total count of traces (for example, web requests) flowing through the system
 
-**Note**: If your application does not use Composer nor an autoloader registered with `spl_autoload_register()`, set the environment variable, `DD_TRACE_NO_AUTOLOADER=true`, to enable automatic instrumentation.
+**Note**: If your application does not use Composer nor an autoloader registered with `spl_autoload_register()`, set the environment variable, `DD_TRACE_NO_AUTOLOADER=1`, to enable automatic instrumentation.
```
bwoebi, line 137: I was curious what this is, but that env seems to be nowhere referenced in our current code? Sounds obsolete?

morrisonlevi:
Do you think it's worth opening a PR for php-fpm to handle this case in the future?

labbati:
I am not well aware of the consequences of that and why `false` is interpreted that way (it might have a special meaning).
Bringing it up the conversation internally. Good idea, though

bwoebi:
@morrisonlevi Yes, it makes sense to fix it. It needs some preserving of the original string alongside it in INI lexer:
```
<INITIAL,ST_VALUE>("true"|"on"|"yes"){TABS_AND_SPACES}* { /* TRUE value (when used outside option value/offset this causes parse error!) */
	RETURN_TOKEN(BOOL_TRUE, "1", 1);
}

<INITIAL,ST_VALUE>("false"|"off"|"no"|"none"){TABS_AND_SPACES}* { /* FALSE value (when used outside option value/offset this causes parse error!)*/
	RETURN_TOKEN(BOOL_FALSE, "", 0);
}
```

bwoebi (approved):
Looks good to me.

urseberry (approved):

PR #14187, opened 2022-06-01
Improving language for Multiple Indexes

timmld:
### What does this PR do?
Improves language and clarifies the Multiple Indexes section.

### Motivation
Several users have been reaching out because they misunderstood the fact that multiple indexes were already enabled by default and that they did not need our help to add new indexes.

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/logs/log_configuration/indexes.md
+++ b/content/en/logs/log_configuration/indexes.md
@@ -28,24 +28,24 @@ You can use indexed logs for [faceted searching][2], [patterns][3], [analytics][
 
 ## Multiple indexes
 
-By default, each account has a single index representing a monolithic set of all your logs. Datadog also offers multiple indexes if you require:
+By default, each new account gets a single index representing a monolithic set of all your logs. Datadog recommends using multiple indexes if you require:
 
 * Multiple [retention periods](#update-log-retention)
 * Multiple [daily quotas](#set-daily-quota), for finer budget control.
 
 The Log Explorer supports [queries across multiple indexes][7].
 
-<div class="alert alert-info">
-<a href="/help">Contact Datadog support</a> to enable multiple-indexes for your account.
-</div>
-
 ### Add indexes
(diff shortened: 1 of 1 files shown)


content/en/logs/log_configuration/indexes.md
```diff
@@ -28,25 +28,25 @@ You can use indexed logs for [faceted searching][2], [patterns][3], [analytics][
...
 The Log Explorer supports [queries across multiple indexes][7].
 
-<div class="alert alert-info">
-<a href="/help">Contact Datadog support</a> to enable multiple-indexes for your account.
-</div>
-
 ### Add indexes
 
-If multiple indexes is activated, use the "New Index" button to create a new index.
+Use the "New Index" button to create a new index. There is a maximum number of indexes you can create for each account.
```
jtappa, line 40: What's the maximum? If we're mentioning that it exists, we should be explicitly stating it here.

jtappa (changes requested):

content/en/logs/log_configuration/indexes.md
timmld: It is intended not to give a number, as it is not fixed and keeps changing over time. The idea here is just to let users know they cannot create an infinite number of indexes

apigirl: Hey Tim! I understand not wanting to give a definite number since it is quickly evolving, but I still think it's useful to give a ballpark so that users will know how to think about it (is it one? is it tens? hundreds? this changes how I approach indexes).

timmld: Actually, thinking about it bit more, it is fine to give a number, it is not planned to change any time soon and there are many reasons to keep it the way it currently is. So I added a number

jtappa (approved):

PR #14243, opened 2022-06-08
DOCS-3549, Issue 14104 - update Ruby example

kayayarai:
### What does this PR do?
Updates Ruby example code

### Motivation
Issue 14104

### Additional Notes
Should we also update /tracing/setup_overview/setup/ruby.md line 2306?

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/tracing/guide/ignoring_apm_resources.md
+++ b/content/en/tracing/guide/ignoring_apm_resources.md
@@ -235,15 +235,15 @@ The Ruby tracer has a post-processing pipeline that deletes traces that meet cer
 
 For example, if the resource name is `Api::HealthchecksController#index`, use the `trace.spans.delete_if` method to delete traces that contain the resource name. This filter can also be used to match on other metadata available for the [span object][2].
 
-```
-Datadog::Tracing::Pipeline.before_flush do |trace|
-    trace.spans.delete_if { |span| span.resource =~ /Api::HealthchecksController#index/ }
-    trace
+```ruby
+Datadog::Tracing.before_flush do |trace|
+  trace.spans.delete_if { |span| span.resource =~ /Api::HealthchecksController#index/ }
+  trace
 end
 ```
 
 [1]: /tracing/setup_overview/custom_instrumentation/ruby/?tab=activespan#post-processing-traces
-[2]: /tracing/setup_overview/setup/ruby/#manual-instrumentation-2
+[2]: /tracing/setup_overview/setup/ruby/#manual-instrumentation
 {{< /programming-lang >}}
 
 {{< programming-lang lang="python" >}}
--- a/content/en/tracing/guide/ignoring_apm_resources.md
+++ b/content/en/tracing/guide/ignoring_apm_resources.md
@@ -233,11 +233,12 @@ Some of the language-specific tracers have an option to modify spans before they
 
 The Ruby tracer has a post-processing pipeline that deletes traces that meet certain criteria. More information and examples can be found in [Post-processing traces][1].
 
-For example, if the resource name is `Api::HealthchecksController#index`, use the `trace.delete_if` method to delete traces that contain the resource name. This filter can also be used to match on other metadata available for the [span object][2].
+For example, if the resource name is `Api::HealthchecksController#index`, use the `trace.spans.delete_if` method to delete traces that contain the resource name. This filter can also be used to match on other metadata available for the [span object][2].
 
 ```
 Datadog::Tracing::Pipeline.before_flush do |trace|
-  trace.delete_if { |span| span.resource =~ /Api::HealthchecksController#index/ }
+    trace.spans.delete_if { |span| span.resource =~ /Api::HealthchecksController#index/ }
+    trace
 end
 ```
 


content/en/tracing/guide/ignoring_apm_resources.md
```diff
@@ -233,11 +233,12 @@ Some of the language-specific tracers have an option to modify spans before they
...
-For example, if the resource name is `Api::HealthchecksController#index`, use the `trace.delete_if` method to delete traces that contain the resource name. This filter can also be used to match on other metadata available for the [span object][2].
+For example, if the resource name is `Api::HealthchecksController#index`, use the `trace.spans.delete_if` method to delete traces that contain the resource name. This filter can also be used to match on other metadata available for the [span object][2].
 
 ```
 Datadog::Tracing::Pipeline.before_flush do |trace|
-  trace.delete_if { |span| span.resource =~ /Api::HealthchecksController#index/ }
+    trace.spans.delete_if { |span| span.resource =~ /Api::HealthchecksController#index/ }
+    trace
 end
 ```
```
ivoanjo, line 243: This new version is correct but it can be improved/simplified a bit:

````markdown
```ruby
Datadog::Tracing.before_flush do |trace|
  trace.spans.delete_if { |span| span.resource =~ /Api::HealthchecksController#index/ }
  trace
end
```
````

(I've added syntax hl for Ruby + used the shorter `Datadog::Tracing.before_flush` + used 2-spaces indentation common for Ruby)

File: content/en/tracing/guide/ignoring_apm_resources.md [hunk 2]
```diff
@@ -233,11 +233,12 @@ Some of the language-specific tracers have an option to modify spans before they
...
```
ivoanjo, line 244: ⬇️ Hey @kayayarai unfortunately GitHub **still** doesn't allow me to comment on unchanged lines but I noticed that on line 246 link 2 is actually broken. I think the hook should go to `#manual-instrumentation` instead of `#manual-instrumentation-2`?

ivoanjo (approved):
> Should we also update /tracing/setup_overview/setup/ruby.md line 2306?

You're correct -- this is wrong on that doc as well. I'll open a PR to update it, since that's on the dd-trace-rb side :)

ivoanjo:
> > Should we also update /tracing/setup_overview/setup/ruby.md line 2306?
>
> You're correct -- this is wrong on that doc as well. I'll open a PR to update it, since that's on the dd-trace-rb side :) 

Actually, scratch that! I, like you probably, had an old copy of `./content/en/tracing/setup_overview/setup/ruby.md` -- that file is actually in `.gitignore` because it gets copied from https://github.com/DataDog/dd-trace-rb/blob/master/docs/GettingStarted.md but that version no longer has the incorrect example.

(And so if you check the actual generated live docs, they don't have it either)

So other than the change in your PR, I think we're good!

content/en/tracing/guide/ignoring_apm_resources.md
kayayarai: Love this, thank you.

kayayarai:
> Actually, scratch that! I, like you probably, had an old copy of 

That is exactly what happened! Nice catch on the link fix too. Thank you for your help with this @ivoanjo

jtappa (approved):

content/en/tracing/guide/ignoring_apm_resources.md
```diff
@@ -233,16 +233,17 @@ Some of the language-specific tracers have an option to modify spans before they
 
 The Ruby tracer has a post-processing pipeline that deletes traces that meet certain criteria. More information and examples can be found in [Post-processing traces][1].
 
-For example, if the resource name is `Api::HealthchecksController#index`, use the `trace.delete_if` method to delete traces that contain the resource name. This filter can also be used to match on other metadata available for the [span object][2].
+For example, if the resource name is `Api::HealthchecksController#index`, use the `trace.spans.delete_if` method to delete traces that contain the resource name. This filter can also be used to match on other metadata available for the [span object][2].
 
-```
-Datadog::Tracing::Pipeline.before_flush do |trace|
-  trace.delete_if { |span| span.resource =~ /Api::HealthchecksController#index/ }
```
delner, line 240: I don't think this example is correct. It would delete only one span (a child span) from the trace. The trace will still come through, just without the one span, or any of its descendants.

I assume the intent is to filter out unwanted traces. Unfortunately, this behavior is broken right now as [detailed here](https://github.com/DataDog/dd-trace-rb/issues/2038#issuecomment-1137696303).

Best we can offer as a workaround right now is setting a `before_action` in the Rails controller that flags the active trace to not be sampled. e.g. `Datadog::Tracing.active_trace.reject!`.

delner (changes requested):
Don't think this is accurate, and will likely result in unwanted behavior.

kayayarai:
Closing based on conversation. If the fix to the bug changes the example, please let docs know. Opened https://github.com/DataDog/documentation/pull/14272 to catch the link we fixed in this PR

PR #14347, opened 2022-06-20
[developers/dogstatsd] fix dogstatsd_queue_size value

keisku:
### What does this PR do?

Set 384 to `dogstatsd_queue_size` in the document.

### Motivation

Screen Shot 2022-06-20 at 15 09 39

"This example decreases the max memory usage of DogStatsD to approximately 384MB" is inconsistent with the value set in the code snippet because it is 512. It should be 384.

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/developers/dogstatsd/high_throughput.md
+++ b/content/en/developers/dogstatsd/high_throughput.md
@@ -378,7 +378,7 @@ Another thing to look at to limit the maximum memory usage is to reduce the buff
 This example decreases the max memory usage of DogStatsD to approximately 384MB:
 
 ```yaml
-dogstatsd_queue_size: 512
+dogstatsd_queue_size: 384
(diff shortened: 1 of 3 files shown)


kayayarai:
I'm not sure your change is correct. Line 372 says: `its default value of 1024 induces an approximate maximum memory usage of 768MB`, which leads me to believe setting it to 512 would induce usage of 384 as stated. What are you basing the change on?

Also, the files in github are not the source for the FR and JA file -- this change would have to go through our translation tools. Don't bother changing the two translated files here.

kayayarai:
Confirmed with an engineer that setting the queue size to 512 does set the memory usage to 384, as the docs originally stated. Correct as-is

PR #14382, opened 2022-06-22
Add test.bundle to documented test configuration tags in junit upload

AdrianLC:
### What does this PR do?
Add `test.bundle` tag to the list of test configuration tags that users can specify manually in junit upload.

### Motivation
We had this temporarily hidden while we did some internal changes, but now it works like the others and should be documented.

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/continuous_integration/setup_tests/junit_upload.md
+++ b/content/en/continuous_integration/setup_tests/junit_upload.md
@@ -206,6 +206,10 @@ You can specify these special tags using the `--tags` parameter when calling `da
 
 All of these tags are optional, and only the ones you specify will be used to differentiate between environment configurations.
 
+`test.bundle`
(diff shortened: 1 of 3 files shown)


buraizu (commented):
Updates to French and Japanese documentation are generally handled through Transifex. Could you confirm if the French and Japanese updates have gone through that process already?

AdrianLC:
> Updates to French and Japanese documentation are generally handled through Transifex. Could you confirm if the French and Japanese updates have gone through that process already?

We had it hidden with as a comment, but they translated the comment when we added it months ago. Now I just moved it above but did not change the text. I thought this would save unnecessary translation work but open to just remove it if you think it's safer.

AdrianLC:
ping @buraizu

PR #14384, opened 2022-06-22
DOCS-3666 Fix docker-compose integration installation doc

themouette:
### What does this PR do?

When installing the agent on my personal host, using docker-compose, I used the provided snippet, but it was not working.

This PR add `privileged: true` to allow collection of other containers data.
In the [Docker agent installation doc](https://docs.datadoghq.com/agent/docker/?tab=standard) it states that agent should be run with `--cgroupns host`, which is `privileged: true` in compose.

I wanted to change `build: datadog` which is supposed to be used to build a local container to `image: datadog/agent` which pulls the image from docker repository, but did not.
I'm balanced between the "it works out of the box" approach (`image: datadog/agent`) and the "in the real world you will customise it anyway" one (`image:` + `Dockerfile`).

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/agent/guide/compose-and-the-datadog-agent.md
+++ b/content/en/agent/guide/compose-and-the-datadog-agent.md
@@ -40,6 +40,7 @@ services:
     image: redis
   datadog:
     build: datadog
+    privileged: true
     environment:
      - DD_API_KEY=${DD_API_KEY}
      - DD_SITE={{< region-param key="dd_site" >}}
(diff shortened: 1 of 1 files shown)


buraizu:
Thanks, created DOCS-3666 to track this update!

vboulineau:
We don't expect our users to have to customize the `datadog/agent` image.

About the `cgroupns host`, the issue is https://github.com/compose-spec/compose-spec/issues/148
Using `privileged` is a bit too much, better to use `pid: host` as a workaround (`pid: host` implies `cgroupns: host`)

hestonhoffman (approved):

PR #14391, opened 2022-06-22
Update support go version to 1.16+

ajgajg1134:
dd-trace-go now supports only the 3 latest go versions which is currently 1.16-1.18

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/tracing/setup_overview/setup/go.md
+++ b/content/en/tracing/setup_overview/setup/go.md
@@ -27,7 +27,7 @@ further_reading:
 
 ## Compatibility requirements
 
-The Go Tracer requires Go `1.12+` and Datadog Agent `>= 5.21.1`.  For a full list of supported libraries, visit the [Compatibility Requirements][1] page.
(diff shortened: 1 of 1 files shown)


katiehockman (approved):
Is there documentation somewhere on what we have to update when we upgrade our version of Go every 6 months? (Assuming that there is something we have to do there)
It would be nice to include this doc as something that needs to be regularly updated.

ajgajg1134:
> Is there documentation somewhere on what we have to update when we upgrade our version of Go every 6 months? (Assuming that there is something we have to do there) It would be nice to include this doc as something that needs to be regularly updated.

Good idea, I wasn't sure quite where to create the docs, so to start I've added a comment within the dd-trace-go README so anyone updating the documentation within our repository will be reminded to update the public docs as well: https://github.com/DataDog/dd-trace-go/pull/1353

jtappa (approved):

PR #14411, opened 2022-06-23
Add links to .NET APM blog

kayayarai:


jonbodner:
This is a bigger conversation, but should we consider putting links to things like tutorials earlier in the document? Maybe open up each APM library documentation page with a section that links to tutorials, how-to guides, and common questions, and then have the detailed documentations immediately following. That way it's clear to people just getting started what the additional resources are without scrolling down a bunch.

alai97 (approved):

PR #14447, opened 2022-06-28
Update PHP APM configuration documentation for version 0.76.0

Anilm3:
### What does this PR do?
This PR updates the PHP APM configuration section to include new environment variables which the user should be aware of in order to opt-out of the new tags introduced for backend ASM.

### Motivation
Provide the user the option to disable some of the new tags introduced for backend ASM.

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/tracing/setup_overview/setup/php.md
+++ b/content/en/tracing/setup_overview/setup/php.md
@@ -394,7 +394,7 @@ Set an application’s version in traces and logs, for example: `1.2.3`, `6c44da
 `DD_TRACE_HTTP_URL_QUERY_PARAM_ALLOWED`
 : **INI**: `datadog.trace.http_url_query_param_allowed`<br>
 **Default**: `*`<br>
-A comma-separated list of query parameters to be collected as part of the URL, this can be empty to prevent collecting any parameters or `*` to collect all parameters. Added in version `0.74.0`
+A comma-separated list of query parameters to be collected as part of the URL, this can also be empty to prevent collecting any parameters or `*` to collect all parameters. Added in version `0.74.0`.
 
 `DD_TRACE_CLIENT_IP_HEADER_DISABLED`
 : **INI**: `datadog.trace.client_ip_header_disabled`<br>
--- a/content/en/tracing/setup_overview/setup/php.md
+++ b/content/en/tracing/setup_overview/setup/php.md
@@ -391,6 +391,21 @@ Enable URL's as resource names (see [Map resource names to normalized URI](#map-
 **Default**: `null`<br>
 Set an application’s version in traces and logs, for example: `1.2.3`, `6c44da20`, `2020.02.13`. Added in version `0.47.0`.
 
+`DD_TRACE_HTTP_URL_QUERY_PARAM_ALLOWED`
+: **INI**: `datadog.trace.http_url_query_param_allowed`<br>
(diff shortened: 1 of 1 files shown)


content/en/tracing/setup_overview/setup/php.md
```diff
@@ -391,6 +391,26 @@ Enable URL's as resource names (see [Map resource names to normalized URI](#map-
 **Default**: `null`<br>
 Set an application’s version in traces and logs, for example: `1.2.3`, `6c44da20`, `2020.02.13`. Added in version `0.47.0`.
 
+`DD_TRACE_HTTP_URL_QUERY_PARAM_ALLOWED`
+: **INI**: `datadog.trace.http_url_query_param_allowed`<br>
+**Default**: `*`<br>
+A comma-separated list of query parameters to be collected as part of the URL, this can also be empty to prevent collecting any parameters or `*` to collect all parameters. Added in version `0.74.0`.
```
kayayarai, line 397:
```suggestion
A comma-separated list of query parameters to be collected as part of the URL. Set to empty to prevent collecting any parameters, or `*` to collect all parameters. Added in version `0.74.0`.
```

kayayarai (changes requested):
Small wording change. Also, I'm going to see if I can improve the format of the regexp default value

kayayarai (approved):

content/en/tracing/setup_overview/setup/php.md
```diff
@@ -391,6 +391,29 @@ Enable URL's as resource names (see [Map resource names to normalized URI](#map-
 **Default**: `null`<br>
 Set an application’s version in traces and logs, for example: `1.2.3`, `6c44da20`, `2020.02.13`. Added in version `0.47.0`.
 
+`DD_TRACE_HTTP_URL_QUERY_PARAM_ALLOWED`
+: **INI**: `datadog.trace.http_url_query_param_allowed`<br>
+**Default**: `*`<br>
+A comma-separated list of query parameters to be collected as part of the URL. Set to empty to prevent collecting any parameters, or `*` to collect all parameters. Added in version `0.74.0`.
+
+`DD_TRACE_CLIENT_IP_HEADER_DISABLED`
```
labbati, line 399: We tend to use the `_ENABLED` pattern rather than the `_DISABLED` pattern. E.g. `DD_TRACE_ENABLED` and `DD_TRACE_CLI_ENABLED`, in this case with a default value of 1. Would you agree with this change?

labbati (commented):
Just a small comment, suggestion for change.

content/en/tracing/setup_overview/setup/php.md
robertpi: I think "disabled" make sense in this case, because the ip collection is enabled by default. So if the user is looking for a config value relating to ip collection, it's probably because they decided they want to disable this setting.

labbati: I am fine with proceeding with "disabled", especially since this has already been discussed and implemented cross team.
However, I would recommend for future params to use the "enable" pattern, not only for consistency, but because we learned from past experience with users that the double negation to enable something ("disabled" + "false/0") causes confusion.

labbati (approved):

labbati:
Than you

PR #14534, opened 2022-07-05
fix a subtle link bug

cswatt:
hey! so i'm not sure if everyone else knew this already, but here's a thing i didn't know you had to do in markdown!

basically: remember that when you are using anchor links, it has to all be lowercase

if your heading is something like `We love APM`, your anchor link is `#we-love-apm`. (it is probably documented somewhere that all heading ids in markdown are lowercased, but i'm gonna be honest here, i have not read the markdown documentation.) urls are case-sensitive, so linking to `#we-love-APM` will not work.

actually now that i'm thinking about it i suspect all urls have to be lowercased


Hesperide (approved):
good find! thank you :)

urseberry (approved):

PR #14563, opened 2022-07-07
Specify that using subdomain for API is misleading

HansFalkenberg-Visma:
### What does this PR do?
Clarifies that subdomains do not affect API usage.

### Motivation
It was not clear whether introducing a subdomain would affect API URLs. And that subdomain for API was _allowed_ but ignored was especially confusing.

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII


apigirl:
Hey @HansFalkenberg-Visma, thanks so much for the update here! We're actually working on a whole series of similar clarifications right now and this might conflict with them. I will let you know if we end up moving forward with this, and if not, what we go with instead!

PR #14603, opened 2022-07-11
Add flag to allow avoiding re-pulling dependent repos on each build

nmuesch:
### What does this PR do?

* Adds some extra logging of exceptions when they occur to have a better idea of the underlying issue
* Renames the `clean-integration` make command to `clean-external-repos`; this better reflects what its cleaning up. While I believe this used to be integration folders, its grown a lot and now includes a lot of other cloned repositories.
* The biggest part of this PR is the addition of a new `CLEAN_EXTRACTED_REPOS` flag. Before PR, every local build would re-pull every dependent repository and then build the docs site. If this is flipped to `false` in your Makefile.config, it'll reuse what the last build pulled. Locally for me, this cut the build down to about 9 seconds. 

### Motivation

While working on adding a different change, I ran into some longer local build times and timeouts from pulling all the external repositories this build depends on. 

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII


hestonhoffman (approved):
LGTM! Tried it out, and it's zippy!

nmuesch:
I'm going to close this with all the recent great work about the build system. Though I'll try to open a new PR that has the same behavior with the new paradigm (will likely be quicker than trying to handle the conflicts here :) )

PR #14629, opened 2022-07-13
[DOCS-3709] Add dependencies for obs pipelines docs

maycmlee:
### What does this PR do?

The Observability Pipelines are currently under Integrations, and because of how the Integrations are setup, those docs don't have an Edit button. I added dependencies to the frontmatter so that there is an Edit button for Obs Pipelines docs.

Also made the Guide index private and disabled the sidebar, because it looks like that's what it should be based on other Guide indexes.

### Motivation

DOCS-3709

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/integrations/observability_pipelines/guide/_index.md
+++ b/content/en/integrations/observability_pipelines/guide/_index.md
@@ -1,6 +1,8 @@
 ---
 title: Observability Pipelines Guides
 kind: guide
+private: true
+disable_sidebar: true
 ---
 
 {{< whatsnext desc="General guides:" >}}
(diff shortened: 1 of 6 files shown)


buraizu (approved):
Hi May, this looks good to me! I looked a little bit more into some other indexes, and I noticed that some others have only `private` or `disable_sidebar`, but not both:

https://github.com/DataDog/documentation/blob/master/content/en/integrations/_index.md?plain=1
https://github.com/DataDog/documentation/blob/master/content/en/dashboards/guide/_index.md?plain=1

That said, I couldn't find any other information about the recommended approach for guide index pages, so not sure what that might be.

maycmlee:
Thanks @buraizu! After the slack discussion, I've added front matter info for guide index pages [here](https://datadoghq.atlassian.net/wiki/spaces/docs4docs/pages/1182958054/Markdown#Guide-index-pages).

PR #14734, opened 2022-07-21
[DOCS-3518] Clarify logs archive upload time

maycmlee:
### What does this PR do?

Clarifies that logs archive upload time can vary.

### Motivation

DOCS-3518

---

### Reviewer checklist
- [ ] Check images for PII


avirishu:
hi @maycmlee I agree that we should not mention a fixed value of 15 mins in the doc, as the upload times can clearly change based on volume, etc. However, we also don't need to mention exactly how often the uploads are done - as that would be too much information and not adding much value to the customer.

So, simply saying that "the upload times can vary" seems good in the documentation.

cswatt (approved):

PR #14741, opened 2022-07-22
make more clear/obvious that service parameter needs to be adjusted

davidcam-nwse:
### What does this PR do?

Fix and make more obvious/clear that the service parameter in conf.yaml (for PHP) needs to be adjusted. 

### Motivation

I copied the configuration but didn't realized that this parameter needed adjustment. As it is written is looks like a "good default" instead of "you need to adjust this" kind of parameter.

So instead of 

> service: "php"

I propose

> service: "<SERVICE_NAME>"

to make it clearer.

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/logs/log_collection/php.md
+++ b/content/en/logs/log_collection/php.md
@@ -192,7 +192,7 @@ logs:
 
   - type: file
     path: "/path/to/your/php/application-json.log"
-    service: php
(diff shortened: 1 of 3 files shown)


cswatt:
Great suggestion! Thanks!

cswatt (approved):

PR #14764, opened 2022-07-26
Avoid continuously repulling dependent repos

nmuesch:
### What does this PR do?

Followup to #14603 with a slightly different approach. 

Update the pre-build setup to check if dependent repos are available locally. (This was partially done before by checking if the user manually cloned them, but this can do one better by using the version of the repo pulled from the last local build)

There's an environment variable check that will ensure that all external repos get pulled on each build, so this behavior should only impact local dev

Updates the error handling during local dev to print out the underlying exception to aid in debugging

Updates the Makefile to make cleaning external repos its own target command, and updates the README to note this behavior. 

### Motivation

Speed up local builds when trying to build with external dependencies

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII


davidejones (approved):

hestonhoffman (approved):
LGTM! I think we're on a merge freeze ATM, so not merging just yet.

PR #14768, opened 2022-07-26
[DOCS-3636] Add Integrate Vector with Datadog 

maycmlee:
### What does this PR do?

- Adds new doc
- Adds to nav
- Redirects "Aggregating Multiple Agents Using Vector" doc to this new one.

### Motivation

DOCS-3636

---

### Reviewer checklist
- [ ] Check images for PII


barieom (commented):
Is there a way to delineate sections more clearly? There's a lot of sections to this and the page looks pretty dense. It'd be great to have more space or maybe even a line whenever there is a Header 2

Copy-wise, looks good to me! I'll likely come back with a small section to add around schemas in a couples days

maycmlee:
> Is there a way to delineate sections more clearly? There's a lot of sections to this and the page looks pretty dense. It'd be great to have more space or maybe even a line whenever there is a Header 2
> 
> Copy-wise, looks good to me! I'll likely come back with a small section to add around schemas in a couples days

Thanks for the feedback @barieom! Hm, so the spacing is automatically done. I can ask the team about line for each H2 sections, but it's not something we usually do. I did reformat the page a bit by taking out the Prerequisite H2 and adding sub-headers to both breakup the info and also so users can link directly to the info. Does that help?

What do you think about moving the `Sink configuration` section to be after the `Source configuration` section and then having `Add Vector-sepcific tag` after `Sink configuration`? It seems like that might flow better in terms of setting up Vector.

Also, is the plan to merge this after the schema info has been added?

urseberry:
This link, which should redirect, gives me a 404. Can you check it? https://docs-staging.datadoghq.com/may/integration-vector-datadog/agent/vector_aggregation

content/en/integrations/observability_pipelines/integrate_vector_with_datadog.md
```diff
@@ -0,0 +1,608 @@
...
+  text: "Learn more about Vector configurations"
+- link: "/integrations/observability_pipelines/working_with_data/"
+  tag: "Documentation"
+  text: "Working with data using Vector"
+---
+
+## Overview
+
+Vector integrates with Datadog to aggregate logs, metrics, and traces from Datadog Agents and route collected telemetry to Datadog.
+
```
urseberry, line 27:
```suggestion

Data flows along the following path:
```

content/en/integrations/observability_pipelines/integrate_vector_with_datadog.md
```diff
@@ -0,0 +1,608 @@
...
+  }
+}
+```
+
+{{% /tab %}}
+{{< /tabs >}}
+
+#### Disk space
+
+You should provision at least 36 GiB per vCPU * of disk space. If you follow the recommendation of 8 vCPUs, you would provision 288 GiB of disk space (10 MiB * 60 seconds * 60 minutes * 8 vCPUs), allocating ​​48 GiB for metrics and 240 GiB for logs. You can add a volume to the Vector instances to hold the buffer in your Helm chart:
```
urseberry, line 541:
```suggestion
You should provision at least 36 GB per vCPU * of disk space. If you follow the recommendation of 8 vCPUs, you would provision 288 GB of disk space (10 MB * 60 seconds * 60 minutes * 8 vCPUs), allocating ​​48 GB for metrics and 240 GB for logs. You can add a volume to the Vector instances to hold the buffer in your Helm chart:
```

maycmlee: Thanks @urseberry! It seems like there is a slight difference between GiB and GB measurements: https://kb.wisc.edu/researchdata/page.php?id=101840

I'll defer to @barieom on whether this should be changed to GB.

urseberry (approved):

barieom (approved):

PR #14823, opened 2022-07-29
Avoid duplicating the RBAC setup

mfpierre:
### What does this PR do?
Avoid duplicating the RBAC setup

### Motivation
Avoid forgetting or updating the wrong RBAC section

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/infrastructure/livecontainers/configuration.md
+++ b/content/en/infrastructure/livecontainers/configuration.md
@@ -56,7 +56,7 @@ In some setups, the Process Agent and Cluster Agent cannot automatically detect 
 
     **Note**: For the `apps` apiGroups, Live Containers need permissions
     to collect common kubernetes resources (`pods`, `services`, `nodes`, etc.),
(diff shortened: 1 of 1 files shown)


cswatt:
hi @mfpierre! While I'm a big fan of deduplicating, I don't think this page is the best use-case. It'll create more confusion for users than is really ideal. However, to make the difference more clear, we could move the "Instructions for previous Agent and Cluster Agent versions" to a different page, maybe called "Legacy configuration" or something. What do you think?

mfpierre:
@cswatt thanks for the review, I'm open to any suggestion that would allow us to avoid maintaining twice the RBAC list as it's prone to errors, maybe we could keep it as it's today but use some kind of an embed/include mechanism? (didn't look further if it was possible)

urseberry:
Created DOCS-3900 to track Cecilia's review.

cswatt:
Closing in favor of https://github.com/DataDog/documentation/pull/15093/

PR #14901, opened 2022-08-04
3833 - adding containers link to agent section

cswatt:
### What does this PR do?

Just adds an extra link in the left nav to the containers section for extra discoverability

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/config/_default/menus/menus.en.yaml
+++ b/config/_default/menus/menus.en.yaml
@@ -247,6 +247,11 @@ main:
     url: agent/basic_agent_usage/source/
     weight: 116
     parent: basic_agent_usage
+  - name: Containers
+    url: containers/
+    parent: agent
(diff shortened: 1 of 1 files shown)


jtappa:
This duplication intermittently gives some weird results of the left nav bar snapping to not at all the right location for me, I think because it's trying to scroll to both locations of "Containers" in left nav at once and so it ends up somewhere in the middle. We might need websites' help on this? Unless we're ok with this behavior.

urseberry (commented):
Adding an official "review" so this drops from the on-call filter. Looks like Jorie already started this one, but let me know if you want more eyes on it.

cswatt:
i don't think this is a good idea anymore. closing

PR #14970, opened 2022-08-10
docs(apm): add python configuration for trace sampling rules

majorgreys:
### What does this PR do?

Add documentation for `DD_TRACE_SAMPLING_RULES` for Python tracing configuration in all languages

### Additional Notes

For translations, in addition to adding `DD_TRACE_SAMPLING_RULES` above for the Python configs, these `DD_TRACE_SAMPLE_RATE` and `DD_TRACE_RATE_LIMIT` are missing from the French and Japanese translations.

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/tracing/trace_collection/library_config/python.md
+++ b/content/en/tracing/trace_collection/library_config/python.md
@@ -55,6 +55,10 @@ Propagation styles to use when extracting tracing headers. When multiple values 
 `DD_TRACE_SAMPLE_RATE`
 : Enable trace volume control
 
+`DD_TRACE_SAMPLING_RULES`
+**Default**: `[]` <br>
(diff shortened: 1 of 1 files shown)


buraizu (changes requested):
Thanks for the PR, but it should be sufficient to update the English documentation. This begins a translation process which you can find outlined here:
https://datadoghq.atlassian.net/wiki/spaces/docs4docs/pages/1189675170/Translation?src=search

Let me know if there's any context I might be unaware of in this regard.

buraizu (commented):
Thanks, this seems fine, but it looks like the `link_checks_preview` is not passing in GitLab. Would you be able to merge in the latest `master` branch and push up once again? Apologies if that's inconvenient or otherwise not feasible -- if so, I can follow up with my team about whether this check is critical under these circumstances.

buraizu (approved):

PR #15043, opened 2022-08-18
API tab labels only say deprecated when x-deprecated flag set

davidejones:
### What does this PR do?

It can be confusing when a v2 api endpoint goes GA as the docs will suddenly display the existing v1 tab as `v1 (deprecated)`. This kind of labeling can cause customers to believe the api they are using is soon to be disabled which isn't the case.

This PR keeps the deprecated label on endpoints using `x-deprecated` but anything else will just display the version.

### Motivation

API discussions

### Preview

https://docs-staging.datadoghq.com/david.jones/version-tablabel/api/latest/metrics/#submit-metrics

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII


devindford (approved):
LGTM, worth noting as I was QAing this, the v1 link completed disappeared not only from the preview but from production 🤔

apigirl (approved):

PR #15091, opened 2022-08-23
3966 - clarity on npm windows support

cswatt:
### What does this PR do?
Adds a parenthetical so non-Server Windows systems are addressed.

### Motivation
Jira ticket from support request.

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/network_monitoring/performance/setup.md
+++ b/content/en/network_monitoring/performance/setup.md
@@ -48,7 +48,7 @@ Data collection is done using eBPF, so Datadog minimally requires platforms that
 
 #### Windows OS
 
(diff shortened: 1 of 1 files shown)


buraizu (changes requested):
I wasn't sure myself, so I looked it up, and it seems like OSs is the way to pluralize OS:
https://www.techtarget.com/whatis/feature/Which-is-correct-OSs-OSes-or-OSs
I'm not especially fond of it, but I'm not finding any other examples of `OSes` in the docs. Maybe it would be better to just say operating system?

buraizu (approved):

PR #15111, opened 2022-08-24
Fix broken build (seen via translated files)

DavidWeid:
### What does this PR do?
- Replaces all instances of `{{% shortcode %}}` with `{{< shortcode >}}`
- Updates `tabs.html` shortcode to have the `markdownify` hugo pipe

### Motivation
The problem is with differing syntax for shortcodes and their nested shortcodes — ie a **tabs and its tab**
Mixing {{% %}} with {{< >}} is the problem, so changing all of the 1st to the 2nd shown here, and thus having all shortcodes just use {{< >}} , fixes things.

Markdownify is needed to ensure proper rendering.


davidejones:
So there was a reason for this being mixed I believe it was to overcome some rendering issues as we upgraded hugo https://github.com/DataDog/documentation/pull/7504. 

If the changes in this PR work then this is great but i know there have been multiple issues with rendering the contents of these nested tabs so it needs some good qa.

Another related issue here https://datadoghq.atlassian.net/browse/WEB-851

DavidWeid:
Moved to this branch:

https://github.com/DataDog/documentation/pull/15236

PR #15177, opened 2022-09-01
[podman] Add note about daemonless

davidor:
### What does this PR do?
Adds a note about running Podman in daemonless mode.

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/agent/guide/podman-support-with-docker-integration.md
+++ b/content/en/agent/guide/podman-support-with-docker-integration.md
@@ -28,11 +28,12 @@ $ podman run -d --name dd-agent \
     gcr.io/datadoghq/agent:latest
 ```
 
-The two importants differences are:
+The two important differences are:
 * `-v /run/podman/podman.sock:/run/podman/podman.sock:ro` to mount the Podman socket instead of the Docker socket.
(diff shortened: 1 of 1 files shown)


content/en/agent/guide/podman-support-with-docker-integration.md
```diff
@@ -28,11 +28,12 @@ $ podman run -d --name dd-agent \
     gcr.io/datadoghq/agent:latest
 ```
 
-The two importants differences are:
+The two important differences are:
 * `-v /run/podman/podman.sock:/run/podman/podman.sock:ro` to mount the Podman socket instead of the Docker socket.
 * `-e DOCKER_HOST=unix:///run/podman/podman.sock` to configure the Agent communication with the Podman socket.
 
-
+When running Podman in daemonless mode, instead of those 2 options, we need to mount the directory where the Podman DB is located, which is `/var/lib/containers` by default:
```
jtappa, line 35:
```suggestion
When running Podman in daemonless mode, instead of these options, you must mount the directory where the Podman database is located, which is `/var/lib/containers` by default:
```

jtappa (changes requested):

jtappa (approved):

PR #15325, opened 2022-09-20
Update Monolog Message Format

datpmwork:
### What does this PR do?
```php
sprintf(
      ' [dd.trace_id=%d dd.span_id=%d]',
      $context['trace_id'],
      $context['span_id']
);
```
- By using this `%d` specifier, the `$context['trace_id']` will be considered as an integer.
- In case the `trace_id` is greater than `9223372036854775807` (the max value), it always return `9223372036854775807` in the output message
- The result is the Log can't be connected with the APM Trace, because actual `trace_id` is different than `9223372036854775807`
image

### Motivation
- I had an issue when connecting the log and APM Trace in some cases, and found out that there are duplicated trace_id appeared with the id `9223372036854775807`.
- Hope this PR can help.

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/tracing/other_telemetry/connect_logs_and_traces/php.md
+++ b/content/en/tracing/other_telemetry/connect_logs_and_traces/php.md
@@ -43,7 +43,7 @@ For instance, you would append those two attributes to your logs with:
   <?php
   $context = \DDTrace\current_context();
   $append = sprintf(
-      ' [dd.trace_id=%d dd.span_id=%d]',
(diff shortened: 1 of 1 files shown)


bwoebi (approved):
That's true, we moved a while back from 63 bit unsigned integers to 64 bit unsigned integers, not fitting into %d anymore.

Thanks for noticing and correcting!

kayayarai (approved):

PR #15335, opened 2022-09-21
tracing/trace_collection/open_standards: correct mistake in exporter YAML

gbbr:


content/en/tracing/trace_collection/open_standards/otel_collector_datadog_exporter.md
```diff
@@ -69,6 +66,14 @@ processors:
         static_configs:
         - targets: ['0.0.0.0:8888']
 
+processors:
+  batch:
+    # Datadog APM Intake limit is 3.2MB. Let's make sure the batches do not
+    # go over that.
+    send_batch_max_size: 1000
```
pavolloffay, line 73: This is really a nit comment, but something I was thinking about in the past: 

The batch processor works on all signals and the default value of max size is `8192`. Should the PR consider splitting the batch setting per signal to make sure the ideal size is used? E.g. metric data points could be sent more than spans.

* https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/README.md

dineshg13 (approved):

content/en/tracing/trace_collection/open_standards/otel_collector_datadog_exporter.md
gbbr: That does make sense and would be much more efficient and accurate. I'll try to figure out better numbers and consider doing that as a follow-up PR. For now, I want to prioritise merging this fix because the current state of the documentation is broken.

apigirl (approved):

PR #15496, opened 2022-10-06
DOCS-4152 - kube dictionary format

cswatt:
kube dictionary example for first mention

expanding kube dictionary example in second mention to match contents of the AD1 version


alai97 (approved):
Looks good!

PR #15586, opened 2022-10-14
[OTel] Remove explicit tags for Collector image

mx-psi:
### What does this PR do?

Remove requirement for using a pinned version of the Collector. In general, we want users to use the latest version.

### Motivation

open-telemetry/opentelemetry-collector-releases/issues/73 was fixed.

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/tracing/trace_collection/open_standards/otel_collector_datadog_exporter.md
+++ b/content/en/tracing/trace_collection/open_standards/otel_collector_datadog_exporter.md
@@ -126,18 +126,13 @@ To get better metadata for traces and for smooth integration with Datadog:
 
 Run an Opentelemetry Collector container to receive traces either from [localhost](#receive-traces-from-localhost), or from [other containers](#receive-traces-from-other-containers).
 
-<div class="alert alert-info">
-The latest tag of the OpenTelemetry Collector Contrib distro <a href="https://github.com/open-telemetry/opentelemetry-collector-releases/issues/73">is not updated on every release</a>.
-Pin the Collector to the latest version to pick up the latest changes.
-</div>
-
 ### Receive traces from localhost
 
 To run the OpenTelemetry Collector as a Docker image and receive traces from the same host:
 
 1. Create a `collector.yaml` file. You may use the [example template above](#configuring-the-datadog-exporter) to get started.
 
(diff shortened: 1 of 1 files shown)


content/en/tracing/trace_collection/open_standards/otel_collector_datadog_exporter.md
```diff
@@ -313,7 +308,7 @@ To use the OpenTelemetry Operator:
    spec:
      mode: daemonset
      hostNetwork: true
-     image: otel/opentelemetry-collector-contrib:0.59.0
+     image: otel/opentelemetry-collector-contrib
```
gbbr, line 311: I assume this then defaults to latest, based on the description of the PR.

gbbr (approved):
Nice!

content/en/tracing/trace_collection/open_standards/otel_collector_datadog_exporter.md
mx-psi: Yup!

dineshg13 (changes requested):
Using latest tag in k8s environment is dangerous. It depends on https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy , whether the collector is using latest image or not. And also it can happen that multiple nodes will have multiple versions running.

mx-psi:
> Using latest tag in k8s environment is dangerous. It depends on https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy , whether the collector is using latest image or not. And also it can happen that multiple nodes will have multiple versions running.

Telling people to use `latest` is consistent with how we tell people to [use the Agent in Docker](https://docs.datadoghq.com/containers/docker/?tab=standard#setup) (the `7` tag gets autoupdated to the latest 7.x version), and [the Agent operator](https://docs.datadoghq.com/containers/kubernetes/installation?tab=operator#deploy-an-agent-with-the-operator), I am just trying to follow the established convention. If people want to pin the version they are free to do it, we are not making it harder to do so with this change. Do you think the situation here is different from that of the Agent?

urseberry (approved):
Ping #documentation on Slack if you need this merged. I'm not merging now, since I see there is still some discussion going on.

dineshg13 (approved):
As discussed offline, we should be ok following the convention established by agent.

PR #15628, opened 2022-10-19
Suggest the Jetbrains Marketplace for IntelliJ

bric3:
### What does this PR do?

@vidurkhannadatadog Received some feedback t DASH on the documentation, where a backlink to the market place was missing. Ultimately the proposed change was to use the related UI elements of the JetBrains Market place by introducing this piece of code.

image

As I wasn't sure if this was OK to

1. use UI material from Jetbrains there,
2. use an external script in this website

So I went with a simple URI. Feel free to iterate on this.

### Motivation

CC @@vidurkhannadatadog

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/developers/ide_integrations/idea/_index.md
+++ b/content/en/developers/ide_integrations/idea/_index.md
@@ -47,6 +47,8 @@ In addition to code highlighting, you can use the Datadog plugin for Intellij ID
 
 {{< img src="/developers/ide_integrations/idea/datadog-plugin1.png" alt="The Datadog plugin" style="width:100%;" >}}
 
+Alternatively you can install the plugin from the [Jetbrains Marketplace](https://plugins.jetbrains.com/plugin/19495-datadog).
+
 ### Add Datadog credentials
 
 After installing the Datadog plugin and restarting IDEA, add your Datadog API and application keys to authenticate with Datadog:


content/en/developers/ide_integrations/idea/_index.md
```diff
@@ -47,6 +47,8 @@ In addition to code highlighting, you can use the Datadog plugin for Intellij ID
 
 {{< img src="/developers/ide_integrations/idea/datadog-plugin1.png" alt="The Datadog plugin" style="width:100%;" >}}
 
+Alternatively you can install the plugin from the [Jetbrains Marketplace](https://plugins.jetbrains.com/plugin/19495-datadog).
```
buraizu, line 50:
```suggestion
Alternatively, you can install the plugin from the [Jetbrains Marketplace][6].
```

content/en/developers/ide_integrations/idea/_index.md
```diff
@@ -147,4 +149,4 @@ Let us know what you think about the plugin! Provide feedback on our [discussion
 [1]: https://github.com/DataDog/datadog-for-intellij/discussions
 [3]: /getting_started/profiler/
 [4]: https://www.datadoghq.com/
-[5]: /account_management/api-app-keys/
\ No newline at end of file
+[5]: /account_management/api-app-keys/
```
buraizu, line 152:
```suggestion
[5]: /account_management/api-app-keys/
[6]: https://plugins.jetbrains.com/plugin/19495-datadog
```

buraizu (changes requested):
Thanks for the PR, just made one suggestion to modify the link to use the reference format, as mentioned in our style guide:
https://github.com/DataDog/documentation/blob/master/CONTRIBUTING.md#links
Also, it should be fine to use the `Get from Marketplace` UI element and script mentioned in the conversations tab. We do this in other places as well, such as here for the Run in Postman button:
https://docs.datadoghq.com/api/latest/

bric3:
@buraizu Hi, thanks for the feedback.

> Also, it should be fine to use the Get from Marketplace UI element and script mentioned in the conversations tab. We do this in other places as well, such as here for the Run in Postman button:
> https://docs.datadoghq.com/api/latest/

So the run in postman button is actually pure markdown.

https://github.com/DataDog/documentation/blob/0fc62732af247af28cc49611e08f884f53882f3b/content/en/api/latest/_index.md?plain=1#L24

But the resources the button asset or its behavior is controlled by an external script.
And I noticed that passing raw HTML is permitted:

https://github.com/DataDog/documentation/blob/0fc62732af247af28cc49611e08f884f53882f3b/config/_default/config.yaml#L27-L31

So as I understand it is ok to add these lines ?

```
<script src="https://plugins.jetbrains.com/assets/scripts/mp-widget.js"></script>
<script>
  // Please, replace #yourelement with a real element id on your webpage
  MarketplaceWidget.setupMarketplaceWidget('install', 19495, "#yourelement");
</script>
```

buraizu:
@bric3 Thanks for looking into that and following up with those additional questions. Would you be able to rename your branch or create a new one in the format `<name>/<branch-name>` with the button and script in place? 

Having the branch name in that format allows for the creation of a documentation preview site where we can make sure that the button works as expected in our docs. As long as it works, I've confirmed with my team that it should be fine to include the button and script in the docs.

PR #15656, opened 2022-10-21
WEB-2855 - curl generate api example content encoding

davidejones:
### What does this PR do?

Updates the generated curl examples to use the header params provided too.

We will try and use the schema.example for the header value. Note that because we try and sync the curl generated examples accordion tab with a python,ruby etc. the title might not match the intention of the example exactly. In this case i've checked for the compression content-encoding in the accordion title to determine what compression to show to help avoid this.

### Motivation

slack / https://datadoghq.atlassian.net/browse/DOCS-3960

### Preview

https://docs-staging.datadoghq.com/david.jones/content-encoding/api/latest/logs/?code-lang=curl#send-logs
https://docs-staging.datadoghq.com/david.jones/content-encoding/api/latest/metrics/?code-lang=curl#submit-metrics

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/layouts/partials/api/curl.html
+++ b/layouts/partials/api/curl.html
@@ -98,13 +98,21 @@
 <span class="o">-H</span> <span class="s1">"{{ $securitySchemes.appKeyAuth.name }}: <span class="n">${ {{- index $securitySchemes.appKeyAuth "x-env-name" -}} }</span>"</span>
 {{- end -}}
 
-{{- /* Render binary data (usually compressed) @<(echo "Uncompressed data" | gzip) */ -}}
+{{- /* Render binary data (usually compressed) */ -}}
 {{- if gt (len $contentEncoding) 0 -}}
+{{- if eq $contentEncoding "gzip" -}}
 &nbsp;\
-<span class="o">--data-binary</span> <span class="n">@<</span><span class="o">({{ $contentEncoding }} <<</span> <span class="n">EOF</span>
+<span class="o">--data-binary</span> <span class="n">@<</span><span class="o">(gzip <<</span> <span class="n">EOF</span>
 <span class="s1">{{ $body | default .exampleValue | default (.dollarScratch.Get "jsonRequestBody") }}</span>
 <span class="n">EOF</span>
 <span class="n">)</span>
+{{- else if eq $contentEncoding "deflate" -}}
+&nbsp;\
+<span class="o">--data-binary</span> <span class="n">@<</span><span class="o">(echo $(cat <<</span> <span class="n">EOF</span>
+<span class="s1">{{ $body | default .exampleValue | default (.dollarScratch.Get "jsonRequestBody") }}</span>
+<span class="n">EOF</span>
+<span class="n">) | gzip --no-name | tail --bytes=+11 | head --bytes=-8)</span>
+{{- end -}}
 {{ else if (or ($body) ($endpoint.action.requestBody)) }}
 {{- /* Render requestBody example JSON "-d @- << EOF ... EOF" */ -}}
 &nbsp;\
--- a/layouts/partials/api/curl.html
+++ b/layouts/partials/api/curl.html
@@ -157,5 +157,5 @@
 {{- end -}}
 <br/>
 {{- else -}}
-<span class="c1">See one of the other client libraries for examples of deflate compression.</span><br/><br/>
+<span class="c1">See one of the other client libraries for an example of sending deflate-compressed logs.</span><br/><br/>
 {{- end -}}
--- a/layouts/partials/api/curl.html
+++ b/layouts/partials/api/curl.html
@@ -26,14 +26,14 @@
   {{- end -}}
 {{- end -}}
 
-{{- if ne $contentEncoding "deflate" -}}
-
 {{- with $exampleDefinition -}}
 ## {{ default $exampleId .summary }}
 {{- range split .description "\n" }}
 # {{ . }}
 {{- end }}
+<br/>
 {{- end -}}
+{{- if ne $contentEncoding "deflate" -}}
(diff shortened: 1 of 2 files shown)


fitzage (approved):

therve:
I'm not exactly sure what were my issues with the PR at the time, but the examples with submit metrics are wrong, as it uses deflate for both IIUC: https://docs-staging.datadoghq.com/david.jones/content-encoding/api/latest/metrics/?code-lang=curl#submit-metrics

davidejones:
@maycmlee & @therve  I've made some changes here, let me know what you think?

therve:
> @maycmlee & @therve I've made some changes here, let me know what you think?

Looks good! One thing maybe is that it seems it's not directly pastable? If we use the echo first I think it works, do you think it's possible? Eg:

```
echo $(cat << EOF
[
  {
    "ddsource": "nginx",
    "ddtags": "env:staging,version:5.1",
    "hostname": "i-012345678",
    "message": "2019-11-19T14:37:58,995 INFO [process.name][20081] Hello World",
    "service": "payment"
  }
]
EOF
) | gzip --no-name | tail --bytes=+11 | head --bytes=-8 | curl -X POST "https://http-intake.logs.datadoghq.com/api/v2/logs" \
-H "Accept: application/json" \
-H "Content-Type: application/json" \
-H "Content-Encoding: deflate" \
-H "DD-API-KEY: ${DD_API_KEY}" \
--data-binary @-
```

davidejones:
I've reordered this @maycmlee & @therve let me know what you think

therve:
The metrics examples don't seem to work :/. Fails with an uncompress error.

davidejones:
@therve copying and pasting the logs deflate example worked. 
It seems like it might be a spec issue or the metrics endpoint doesn't actually accept deflate?

Everything is defaulting to deflate in the spec but none of the other code examples python/java etc use deflate they all use zstd1 https://github.com/DataDog/datadog-api-spec/blob/master/spec/v2/metrics_intake.yaml#L147

therve:
I think the gzip trick is not exactly correct. Trying deflate on the clients work for sure, but I don't why it would work for logs and not metrics.

davidejones:
@therve I think you are right there might be limits with this trick when sending larger payloads.
Although i'm not sure we hit any limit yet in our examples. When I compressed the payloads with this trick and decompressed it with python and it seemed to work fine?

```bash
echo $(cat << EOF
[
  {
    "ddsource": "nginx",
    "ddtags": "env:staging,version:5.1",
    "hostname": "i-012345678",
    "message": "2019-11-19T14:37:58,995 INFO [process.name][20081] Hello World",
    "service": "payment"
  }
]
EOF
) | gzip --no-name | tail --bytes=+11 | head --bytes=-8 > logs.bin
```

```bash
echo $(cat << EOF
{
  "series": [
    {
      "metric": "system.load.1",
      "type": 0,
      "points": [
        {
          "timestamp": 1636629071,
          "value": 0.7
        }
      ]
    }
  ]
}
EOF
) | gzip --no-name | tail --bytes=+11 | head --bytes=-8 > metrics.bin
```

```python
import zlib

def inflate(data):
    d = zlib.decompressobj(-zlib.MAX_WBITS)
    i = d.decompress(data)
    i += d.flush()
    return i

with open('./metrics.bin', mode='rb') as f:
    print(inflate(f.read()))

with open('./logs.bin', mode='rb') as f:
    print(inflate(f.read()))
```

It outputs

```bash
b'{ "series": [ { "metric": "system.load.1", "type": 0, "points": [ { "timestamp": 1636629071, "value": 0.7 } ] } ] }\n'
b'[ { "ddsource": "nginx", "ddtags": "env:staging,version:5.1", "hostname": "i-012345678", "message": "2019-11-19T14:37:58,995 INFO [process.name][20081] Hello World", "service": "payment" } ]\n'
```

therve:
> ```python
>     d = zlib.decompressobj(-zlib.MAX_WBITS)
> ```

I believe raw deflate is just using `zlib.decompress` directly, which doesn't work with your data.

therve:
If I try something like `zlib-flate -compress` on linux, it's compatible with `zlib.decompress` and seems to work with the metrics API as well. So that head/tail trick probably doesn't work as we expect.

davidejones:
I see, thanks for helping to confirm @therve. At this point i'm not sure what we should display for these curl examples then as anything to correctly deflate would require other third party library installs and at that point its likely better to recommend one of the language specific implementations instead? 

Should we only display non compressed and gzip compressed curl examples only? @maycmlee

layouts/partials/api/curl.html
```diff
@@ -83,11 +139,23 @@
...
 {{- /* Render requestBody example JSON "-d @- << EOF ... EOF" */ -}}
-{{- if (or ($body) ($endpoint.action.requestBody)) -}}
 &nbsp;\
 <span class="o">-d</span> <span class="n">@-</span> <span class="o"><<</span> <span class="n">EOF</span>
 <span class="s1">{{ .exampleValue | default $body | default (.dollarScratch.Get "jsonRequestBody") }}</span>
 <span class="n">EOF</span>
 {{- end -}}
 <br/>
+{{- else -}}
+<span class="c1">See one of the other client libraries for examples of deflate compression.</span><br/><br/>
```
maycmlee, line 160:
```suggestion
<span class="c1">See one of the other client libraries for an example of sending deflate-compressed logs.</span><br/><br/>
```

pmbauer:
I think this has some typos 
https://docs.datadoghq.com/api/latest/metrics/#submit-distribution-points
CleanShot 2023-04-17 at 09 21 51@2x

PR #15708, opened 2022-10-25
AD back into JSON format

cswatt:
puts format back into one that worked


alai97 (approved):

chendo:
I spent an entire afternoon trying to figure out why the ADv2 format wasn't working and Kubernetes didn't want any kind of YAML and was digging in Go source code figuring out how this was ever meant to work :(

Thank you for fixing the documentation.

PR #15789, opened 2022-11-02
[RUM][REPLAY]: Rephrase mention about service worker

acorretti:
Explained in detail on Jira card https://datadoghq.atlassian.net/browse/REPLAY-1102

Some customers were misled by the wording about a "service worker", which is a (very specific concept)[https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API] that we don't use. Instead we should explicitly talk about a "dedicated web worker".

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII


buraizu (approved):

buraizu:
Verified in local build

PR #15816, opened 2022-11-04
fix: change deprecated links to their new paths

MatissJanis:
### What does this PR do?

Changing a few deprecated links to their new paths. We do have redirects in place for the deprecated links so they will continue working as before. However, we prefer to have the direct links from docs to be pointing to the non-deprecated paths.

### Motivation

See above

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII


kayayarai (approved):
The fr, ja, and kr links might get overwritten by the next update from the translation service we use, but they will also eventually catch up as the updates move through the translation pipeline.

PR #15845, opened 2022-11-08
[CI Visibility] Adapt CI monitors documentation for GA

liashenko:
### What does this PR do?

We're releasing CI monitors to GA, in this PR removing `Alpha` section, and updating the notification section up to date.

JIRA issue: https://datadoghq.atlassian.net/browse/CIAPP-5265

### Motivation

CI monitors GA.

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/monitors/create/types/ci.md
+++ b/content/en/monitors/create/types/ci.md
@@ -141,11 +141,12 @@ For detailed instructions on the **Say what's happening** and **Notify your team
 
 #### Notifications behavior when there is no data
 
-A monitor that uses an event count for its evaluation query after the specified period with no data will evaluate to 0, potentially resolving immediately.
+A monitor that uses an event count for its evaluation query will resolve after the specified evaluation period with no data, triggering a notification. For example, a monitor configured to alert on the number of pipeline errors with an evaluation window of five minutes will automatically resolve after five minutes without any pipeline executions.
 
-In this case Datadog recommends using a formula like `(number of failures)/(number of all events)`, because if `(number of all events) = 0`, the division `x/0` cannot be evaluated and the monitor will keep the last known state.
+As an alternative, Datadog recommends using rate formulas. For example, instead of using a monitor on the number of pipeline failures (count), use a monitor on the rate of pipeline failures (formula), such as `(number of pipeline failures)/(number of all pipeline executions)`. In this case, when there's no data, the denominator `(number of all pipeline executions)` will be `0`, making the division `x/0` impossible to evaluate. The monitor will keep the previous known state instead of evaluating it to `0`.
 
-For example, use a monitor on pipeline error rate rather than on pipeline error count.
+This way, if the monitor triggers because there's a burst of pipeline failures that makes the error rate go above the monitor threshold, it will not clear until the error rate goes below the threshold, which can be at any time afterwards.
+
 
 ## Further Reading
 
--- a/content/en/monitors/create/types/ci.md
+++ b/content/en/monitors/create/types/ci.md
@@ -141,19 +141,11 @@ For detailed instructions on the **Say what's happening** and **Notify your team
 
 #### Notifications behavior when there is no data
 
-A monitor that uses an event count for its evaluation query will resolve after the specified evaluation period with no data, triggering a notification. For example, a monitor on pipeline error count with an evaluation window of five minutes will automatically resolve after five minutes without any data.
+A monitor that uses an event count for its evaluation query after the specified  period with no data will evaluate to 0, potentially resolving immediately.
 
-As CI pipeline data is usually sparse and can have relatively long periods with no data, this can result in monitor recovery notifications that might not be desired.
+In this case Datadog recommends using a formula like `(number of failures)/(number of all events)`, because if `(number of all events) = 0`, the division `x/0` cannot be evaluated and the monitor will keep the last known state.
 
-In these cases, Datadog recommends configuring the monitor notification to trigger only for alerts by wrapping the full message with the `{{#is_alert}}` and `{{/is_alert}}` directives.
-
-```text
-{{#is_alert}}
-This notification will only be sent for monitor alerts!
-{{/is_alert}}
-```
-
-**Note**: A monitor that uses a formula for its evaluation query will keep the last state (for example, **Alert**) if there is no data by default.
+For example, use a monitor on pipeline error rate rather than on pipeline error count.
 
 ## Further Reading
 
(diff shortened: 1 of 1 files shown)


content/en/monitors/create/types/ci.md
```diff
@@ -158,6 +155,8 @@ This notification will only be sent for monitor alerts!
 {{/is_alert}}
 ```
 
+**Note**: A monitor that uses a formula for its evaluation query will keep the last state (e.g. Alert) if there is no data by default.
```
Dalje-et, line 158: That's not really true, is it? I definitely applies if I do `a / b * 100` but what if I do `a + b`? Not saying that it makes sense creating a monitor with `a + b` but I think it's good to be specific here.

liashenko: Based on this change https://github.com/DataDog/dd-go/pull/76284, all F&F monitors will keep the last state if there is no data. 

I've tested `a+b` and it works like this https://admin.us1.prod.dog/monitors/cluster/realtime/org/2/monitor/101181165?from_ts=1667917463516&to_ts=1667918363516

content/en/monitors/create/types/ci.md
```diff
@@ -158,6 +153,8 @@ This notification will only be sent for monitor alerts!
...
```
jtappa, line 156:
```suggestion
**Note**: A monitor that uses a formula for its evaluation query will keep the last state (for example, **Alert**) if there is no data by default.
```

jtappa (changes requested):

content/en/monitors/create/types/ci.md
fermayo: It's not on "no data", but when the formula cannot be evaluated (i.e. `a/b` and `b=0`). If there's no data and the formula evaluates to `0`, the query will not keep the last state but use `0` for evaluation.

I think we should rewrite this whole section. We should state that if the query is a count and there is no data, it will evaluate to 0, potentially resolving immediately. Because of this, Datadog recommends using a formula like `(number of failures)/(number of all events)`, because if `(number of all events) = 0`, the division `x/0` cannot be evaluated and the monitor will keep the last known state.

content/en/monitors/create/types/ci.md
```diff
@@ -146,17 +141,11 @@ For detailed instructions on the **Say what's happening** and **Notify your team
 
 #### Notifications behavior when there is no data
 
-A monitor that uses an event count, or a formula for its evaluation query will resolve after the specified evaluation period with no data, triggering a notification. For example, a monitor using a formula to alert on pipeline error rate with an evaluation window of five minutes will automatically resolve after five minutes without any data.
-
-As CI pipeline data is usually sparse and can have relatively long periods with no data, this can result in monitor recovery notifications that might not be desired.
+A monitor that uses an event count for its evaluation query after the specified period with no data will evaluate to 0, potentially resolving immediately.
```
fermayo, line 144:
```suggestion
A monitor that uses an event count for its evaluation query will resolve after the specified evaluation period with no data, triggering a notification. For example, a monitor configured to alert on the number of pipeline errors with an evaluation window of five minutes will automatically resolve after five minutes without any pipeline executions.
```

content/en/monitors/create/types/ci.md
```diff
@@ -146,17 +141,11 @@ For detailed instructions on the **Say what's happening** and **Notify your team
...
 
-In these cases, Datadog recommends configuring the monitor notification to trigger only for alerts by wrapping the full message with the `{{#is_alert}}` and `{{/is_alert}}` directives.
+In this case Datadog recommends using a formula like `(number of failures)/(number of all events)`, because if `(number of all events) = 0`, the division `x/0` cannot be evaluated and the monitor will keep the last known state.
 
-```text
-{{#is_alert}}
-This notification will only be sent for monitor alerts!
-{{/is_alert}}
-```
+For example, use a monitor on pipeline error rate rather than on pipeline error count.
```
fermayo, line 148:
```suggestion
```

content/en/monitors/create/types/ci.md
```diff
@@ -146,17 +141,11 @@ For detailed instructions on the **Say what's happening** and **Notify your team
 
 #### Notifications behavior when there is no data
 
-A monitor that uses an event count, or a formula for its evaluation query will resolve after the specified evaluation period with no data, triggering a notification. For example, a monitor using a formula to alert on pipeline error rate with an evaluation window of five minutes will automatically resolve after five minutes without any data.
-
-As CI pipeline data is usually sparse and can have relatively long periods with no data, this can result in monitor recovery notifications that might not be desired.
+A monitor that uses an event count for its evaluation query after the specified period with no data will evaluate to 0, potentially resolving immediately.
 
-In these cases, Datadog recommends configuring the monitor notification to trigger only for alerts by wrapping the full message with the `{{#is_alert}}` and `{{/is_alert}}` directives.
+In this case Datadog recommends using a formula like `(number of failures)/(number of all events)`, because if `(number of all events) = 0`, the division `x/0` cannot be evaluated and the monitor will keep the last known state.
```
fermayo, line 146:
```suggestion
As an alternative, Datadog recommends using rate formulas. For example, instead of using a monitor on the number of pipeline failures (count), use a monitor on the rate of pipeline failures (formula), such as `(number of pipeline failures)/(number of all pipeline executions)`. In this case, when there's no data, the denominator `(number of all pipeline executions)` will be `0`, making the division `x/0` impossible to evaluate. The monitor will keep the previous known state instead of evaluating it to `0`.

This way, if the monitor triggers because there's a burst of pipeline failures that makes the error rate go above the monitor threshold, it will not clear until the error rate goes below the threshold, which can be at any time afterwards.
```

@liashenko not sure about the "any time afterwards". Can you check with the monitors team to see how long will the monitor remember its last state?

fermayo (approved):

content/en/monitors/create/types/ci.md
```diff
@@ -106,11 +103,9 @@ The following example is a test error rate monitor using a formula that calculat
 
 {{< img src="monitors/monitor_types/ci_tests/define-the-search-query-fnf.png" alt="Monitor being defined with steps a, b, and c, where steps a and b are queries and step c calculates the rate from them." style="width:80%;" >}}
 
-<div class="alert alert-info"><strong>Note</strong>: A maximum of two queries can be used to build the evaluation formula per monitor.</div>
```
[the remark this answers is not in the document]
liashenko: This is no longer a restriction.

content/en/monitors/create/types/ci.md
liashenko: it's correct, it will not be recovered until there is `OK` data.

jtappa (approved):

PR #15869, opened 2022-11-09
Update rds.md

matt-slalom:
The `create extension` command will create the extension in whichever schema you happened to be logged into. Since this is being done as a superuser, there's no guarantee that the `datadog` user will be able to find it. That, in turn, leads to missing metrics. Putting this in pg_catalog means it's in everyone's path.

### What does this PR do?
Attempts to correct the documentation

### Motivation
Errors I encounter when installing.

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/database_monitoring/setup_postgres/rds.md
+++ b/content/en/database_monitoring/setup_postgres/rds.md
@@ -80,7 +80,7 @@ CREATE SCHEMA datadog;
 GRANT USAGE ON SCHEMA datadog TO datadog;
 GRANT USAGE ON SCHEMA public TO datadog;
 GRANT pg_monitor TO datadog;
-CREATE EXTENSION IF NOT EXISTS pg_stat_statements schema pg_catalog;
+CREATE EXTENSION IF NOT EXISTS pg_stat_statements schema public;
 ```
 
 {{% /tab %}}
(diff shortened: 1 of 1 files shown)


maycmlee:
Thanks @matt-slalom for opening this PR! We're reviewing this and will be back in touch.

matt-slalom:
> Hey @matt-slalom thanks for the contribution! I understand where you're coming from, but I don't think we want to advise creation in pg_catalog for customers. Modifying pg_catalog requires higher permissions and has a wider footprint than Datadog intends to have.
> 
> Following the instructions should create `pg_stat_statements` in the `public` schema by default, which the datadog user will have access to. Can you explain the steps you took which resulted in the extension being created in a different schema?

Putting it in `public` is fine with me. The default `search_path` with Postgres (at least in AWS RDS) prefers the user's schema over public, so if you don't specify the schema, the extension gets created in the administrator's path, where the `datadog` user won't be able to find it.

matt-slalom:
Was there something else needed for this fix to the documentation?

jtappa:
Hi @matt-slalom, apologies for the wait. I've pinged our DBM team to confirm this change is good to go into the docs.

xcoderzach (approved):

jtappa (approved):

xcoderzach:
Thanks for making this update!

PR #15871, opened 2022-11-09
API page updates.

bgdeutsch:
### What does this PR do?
- Updates `build-api-pages` script to output the summary and description as markdown to the related `_index.md` file.   API page content is controlled by layout, but the proposed indexing strategy relies on markdown content.   These changes allow us to continue building the search index with Hugo and include the API pages.

- After running the script locally this updates markdown files under `content/en/api/latest/` 

- Introduced a frontmatter parameter, `render_content`.   This flag controls whether a specific API page will render the markdown contents of `_index.md`.

### Motivation
API results were broken in our Algolia test preview index.

### Preview
There are no visual changes to API pages.
https://docs-staging.datadoghq.com/brian.deutsch/api-updates/api/latest/ci-visibility-pipelines/
https://docs-staging.datadoghq.com/brian.deutsch/api-updates/api/latest/notebooks/

The following API pages do have markdown content that needs to be rendered.  These should also appear unchanged:
https://docs-staging.datadoghq.com/brian.deutsch/api-updates/api/latest/
https://docs-staging.datadoghq.com/brian.deutsch/api-updates/api/latest/scopes/
https://docs-staging.datadoghq.com/brian.deutsch/api-updates/api/latest/using-the-api/

### Additional Notes
I do have concerns this could cause confusion amongst contributors now that some markdown content will exist in the `api/latest/_index.md` files.    API pages are mostly auto-generated, does it make sense to propose gitignoring those files?  @davidejones what do you think?

We may also need to make some changes to support translation efforts.
---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/assets/scripts/build-api-pages.js
+++ b/assets/scripts/build-api-pages.js
@@ -6,7 +6,6 @@ const marked = require('marked');
 const slugify = require('slugify');
 const $RefParser = require('@apidevtools/json-schema-ref-parser');
 const safeJsonStringify = require('safe-json-stringify');
-
 const supportedLangs = ['en'];
 
 /**
--- a/assets/scripts/build-api-pages.js
+++ b/assets/scripts/build-api-pages.js
@@ -129,6 +129,7 @@ const createPages = (apiYaml, deref, apiVersion) => {
 
     // create markdown containing summary and description for each endpoint
     // this is for compiling the algolia search index, it is not used for rendering content.
+    // api page content is controlled by layout in most cases.
     const data = Object.keys(deref.paths)
       .filter((path) => isTagMatch(deref.paths[path], tag.name))
       .map((path) => deref.paths[path]);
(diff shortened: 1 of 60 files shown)


davidejones:
> I do have concerns this could cause confusion amongst contributors now that some markdown content will exist in the api/latest/_index.md files. API pages are mostly auto-generated, does it make sense to propose gitignoring those files? @davidejones what do you think?

@bgdeutsch Yea I could see that being confusing, that could be managed but not ideal. I’d also prefer not to gitignore api files just because we would then have to generate them on every build where as now it only happens on spec change also the layout of files might not play so nice in hugo when files don’t exist by default.

Another alternative is to break the logic out for that section and read from the spec file instead similar to the partials on corp that are reading data files e.g https://github.com/DataDog/corp-hugo/blob/master/site/layouts/partials/algolia/resources-inthewild.json So instead of using the `.Content` page variable we would build a content variable by extracting the same data out of the spec that you are doing now with the javascript but just with hugo.

PR #16060, opened 2022-11-28
Index API pages

bgdeutsch:
### What does this PR do?
Pulls API pages into the algolia index at build time.

### Motivation
API pages are not controlled by markdown so they were not being properly included

### Preview
https://www.algolia.com/apps/K8XL4ROVCR/explorer/browse/docs_preview?refinementList%5Blanguage%5D=&refinementList%5Btags%5D%5B%5D=api_latest&page=1&searchMode=search

in the algolia dashboard filter results by tag `api_latest`,  results should be surfacing in english, french, and japanese.

cc @davidejones @DavidWeid 

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII


layouts/_default/list.algolia.json
```diff
@@ -35,6 +35,9 @@
         {{ end }}
     {{ end }}
 
+    {{- /* api pages are controlled by layout and not markdown, so we need to index them separately */ -}}
+    {{ partial "algolia/api-pages-full-index.json" (dict "context" $hugo_context) }}
+
```
bgdeutsch, line 40: here we are ignoring all api pages in the main loop and later iterating over all the `api/latest` pages only and adding to the main `algoliaindex` object.   i think we can make this process more efficient by including the api `v1` and `v2` pages in with the main loop and build them up as `api/latest` entries in the index.

buraizu (approved):
Doesn't look like a docs review is needed here, but just wanted to mention that I wasn't able to validate this in the algolia dashboard (perhaps due to inexperience with algolia or lack of permissions)

layouts/partials/algolia/api-page-index.json
```diff
@@ -0,0 +1,54 @@
...
+
+{{/* Add the resulting object to the index */}}
+{{
+    $hugo_context.Scratch.Add "algoliaindex" (
+        dict "objectID" (md5 $full_url)
+        "title" $title
+        "section_header" $section_header
+        "content" $section_content
+        "type" $type
+        "rel_permalink" $rel_permalink
```
DavidWeid, line 48: this should be key:value should be { "relpermalink": value } instead of `rel_permalink` to match the object defined in `list.algolia.json`

PR #16070, opened 2022-11-29
Cucumber versions using JUnit5 are not supported

albertvaka:
### What does this PR do?
Cucumber v2 and later, which use JUnit v5, don't work at the moment despite JUnit v5 being supported.

### Motivation
We never tested Cucumber with JUnit 5. I tested it and our instrumentation doesn't work.

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/continuous_integration/tests/java.md
+++ b/content/en/continuous_integration/tests/java.md
@@ -23,7 +23,7 @@ further_reading:
 
 Supported test frameworks:
 * JUnit >= 4.10 and >= 5.3
-  * Also includes any test framework based on JUnit, such as Spock Framework and Cucumber-Junit (note: Cucumber is only supported with JUnit v4)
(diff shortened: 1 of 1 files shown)


content/en/continuous_integration/tests/java.md
```diff
@@ -23,7 +23,7 @@ further_reading:
 
 Supported test frameworks:
 * JUnit >= 4.10 and >= 5.3
-  * Also includes any test framework based on JUnit, such as Spock Framework and Cucumber-Junit
+  * Also includes any test framework based on JUnit, such as Spock Framework and Cucumber-Junit (note: Cucumber is only supported with JUnit v4)
```
maycmlee, line 26:
```suggestion
  * Also includes any test framework based on JUnit, such as Spock Framework and Cucumber-Junit. **Note**: Cucumber is only supported with JUnit v4.
```

maycmlee (changes requested):
Just a couple of suggestions!

maycmlee (approved):

PR #16097, opened 2022-12-01
security_platform URL update #1

drichards-87:
### What does this PR do?
Ensures product naming consistency in our docs by renaming the path for the security articles (security_platform > security).

### Preview
https://docs-staging.datadoghq.com/deforest/primary_security_platform_url/security/default_rules/ 

### Motivation
DOCS-4327


local/bin/py/build/actions/security_rules.py
```diff
@@ -108,7 +108,6 @@ def security_rules(content, content_dir):
             if p.exists():
                 logger.info(f"removing file {p.name}")
                 global_aliases.append(f"/security_monitoring/default_rules/{p.stem}")
-                global_aliases.append(f"/security_platform/default_rules/{p.stem}")
                 p.unlink()
```
bgdeutsch, line 111: We shouldn't need this as an alias in Hugo since we will have S3 level redirects as a result of: https://github.com/DataDog/cloudops/pull/33471

davidejones: So this is part of the logic to deal with removals or rules without giving users a bad experience. What this is currently doing is when a rule is deleted its adding its url as an alias to `_index.md`. That way if someone still tries to access a deleted rule it will go to the list of rules instead of 404'ing.

I think we should update this code to just have the one entry using the new path
```python
global_aliases.append(f"/security/default_rules/{p.stem}")
```

bgdeutsch: Makes sense, thanks @davidejones!

kayayarai (approved):

davidejones (approved):

PR #16108, opened 2022-12-02
Fixes serverless source code integration syntax error

DanRowe:
### What does this PR do?

Syntax fix in the example for enabling source code integration. The code snippet and text display the wrong syntax for enabling source code integration using the cli.

Attempting to use the current documentation's examples produces the following error.

```sh
Unknown Syntax Error: Extraneous positional argument ("true").
```

### Motivation

I followed the incorrect docs and it caused this error. I don't want others to have to find out the same way.

### Additional Notes

We could also suggest using `-s` or not including true at all `--source-code-integration`. 

Mentioned in https://docs.datadoghq.com/serverless/libraries_integrations/cli/#arguments

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/serverless/configuration/_index.md
+++ b/content/en/serverless/configuration/_index.md
@@ -465,7 +465,7 @@ If you are using a runtime or custom logger that isn't supported, follow these s
 {{< tabs >}}
 {{% tab "Datadog CLI" %}}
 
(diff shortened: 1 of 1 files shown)


drichards-87 (approved):
Approving from docs as the update looks good on our end. I'll let the serverless team review/approve for accuracy. Thanks for submitting this update!

rpelliard (approved):

PR #16235, opened 2022-12-16
Add public folder

devindford:
### What does this PR do?

Adds the public folder so that it is always in the directory (still excludes all the files except for the `.gitkeep` file)
### Motivation

A possible workaround for how we handle static assets. This was a [hugo issue](https://github.com/gohugoio/hugo/issues/8166) that was somewhat similar to what we have. Due to how we place the `static` directory into `assets` I'm wondering if this may possibly be the issue. This change may mitigate our problem or it may not, but this will not effect anything negatively having the empty directory in the repo prior to build.

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Hugo
-public/
+public/*
+!public/.gitkeep
 data/service_checks/
 data/npm
 integrations_data
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -60,7 +60,7 @@ before_script:
 
 # ================== templates ================== #
 .base_template: &base_template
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -1,5 +1,6 @@
 # Hugo
-public/
+public/*
+!public/.gitkeep
```
devindford, line 3: This will allow the public directory to be in the repo but ignore all other files and subdirectories. 

The `!public/.gitkeep` is to allow the gitkeep file to be there and in turn makes git keep the empty directory

bgdeutsch (approved):

urseberry (approved):

PR #16256, opened 2022-12-20
adding tms

cswatt:
the ™ checker is broken cos of a transifex thing. this fixes it.

note: this has also been fixed in transifex, this PR is purely to make all the lights green


cswatt:
duplicated by 16255

PR #16275, opened 2022-12-21
additional security aliases

kayayarai:
Catching security aliases for pages whose name changed in the past


drichards-87 (approved):
Looks good!

PR #16277, opened 2022-12-21
Clarify Required Configurations for AKS

stzou:
### What does this PR do?
Removed references to host and hostCAPath as those are actually not required when disabling tlsVerify.

### Motivation
Having host and hostCAPath wth tlsVerify shouldn't actually be an issue since the CA isn't going to be used if tls verify is disabled. However, it seems that it had some side effects where the security agent was attempting tls verification due to the presence of hostCAPath. References to host and hostCAPath should be removed as it is misleading and not required.

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII


kayayarai:
opened DOCS-4594 to track technical review.

kayayarai (approved):

PR #16367, opened 2023-01-05
shadow-DOM-beta-callout

jamiemilstein:
Added a beta tag  for customers to request access to Shadow DOM


rtrieu (approved):
Verified locally the callout is there.

PR #16435, opened 2023-01-12
Update admission_controller.md

elee0625:
### What does this PR do?
Final changes to the doc before GA

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/tracing/trace_collection/admission_controller.md
+++ b/content/en/tracing/trace_collection/admission_controller.md
@@ -2,7 +2,6 @@
 title: Injecting Libraries Using Admission Controller
 kind: documentation
 description: "Inject instrumentation libraries into applications using Cluster Agent and Admission Controller"
-is_beta: false
 ---
 
 ## Overview
--- a/content/en/tracing/trace_collection/admission_controller.md
+++ b/content/en/tracing/trace_collection/admission_controller.md
@@ -84,7 +84,7 @@ To select your pods for library injection, annotate them with the following, cor
 
 The available library versions are listed in each container registry.
 
-**Note**: As mentioned above, library injection is applied on new pods only and does not have any impact on running pods. However, if you already have an application instrumented using version X of the library, and then use library injection to instrument using version Y of the same tracer library, the tracer does not break. Rather, the library version loaded first is used.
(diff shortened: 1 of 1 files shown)


content/en/tracing/trace_collection/admission_controller.md
```diff
@@ -89,6 +85,8 @@ To select your pods for library injection, annotate them with the following, cor
 
 The available library versions are listed in each container registry.
 
+**Note**: As mentioned above, library injection is applied on new pods only and does not have any impact on running pods. However, if you already have an application instrumented using version X of the library, and then use library injection to instrument using version Y of the same tracer library, the tracer will not break. Rather, the library version to be first loaded will be used.
```
maycmlee, line 88:
```suggestion
**Note**: As mentioned above, library injection is applied on new pods only and does not have any impact on running pods. However, if you already have an application instrumented using version X of the library, and then use library injection to instrument using version Y of the same tracer library, the tracer does not break. Rather, the library version loaded first is used.
```

content/en/tracing/trace_collection/admission_controller.md
```diff
@@ -2,13 +2,9 @@
 title: Injecting Libraries Using Admission Controller
 kind: documentation
 description: "Inject instrumentation libraries into applications using Cluster Agent and Admission Controller"
-is_beta: true
+is_beta: false
```
maycmlee, line 5:
```suggestion
```

maycmlee: Can be removed completely.

maycmlee (commented):
Left some feedback!

content/en/tracing/trace_collection/admission_controller.md
yoannrobin, line 88:
```suggestion
**Note**: If you already have an application instrumented using version X of the library, and then use library injection to instrument using version Y of the same tracer library, the tracer will not break. Rather, the library version to be first loaded will be used.
```
In your example with vX used in the application and lib injection on vY, which one would be used in the end? The current statement is not explicit

yoannrobin (changes requested):

content/en/tracing/trace_collection/admission_controller.md
elee0625: It would most likely be Y. I would assume that the customer would use Lib Injection for their latest setup.

yoannrobin (approved):

kayayarai (approved):

PR #16465, opened 2023-01-13
[variables.md] expanding on recommendations for conditional variable

cahillsf:
… usage

### What does this PR do?

Opening this PR based on customer feedback that their users are misconfiguring conditional variables.  Adding the [the ticket](https://datadog.zendesk.com/agent/tickets/1052503) here for reference

### Motivation

https://datadog.zendesk.com/agent/tickets/1052503

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/monitors/notify/variables.md
+++ b/content/en/monitors/notify/variables.md
@@ -228,7 +228,7 @@ This is the escalation message @dev-team@company.com
 {{< /tabs >}}
 
 
-<div class="alert alert-info"><strong>Note</strong>: Any text or notification handle placed outside of the configured conditional variables will be invoked with every monitor state transition.  Any text or notification handle placed inside of configured conditional variables will only be invoked with that specific monitor state transition.
+<div class="alert alert-info"><strong>Note</strong>: Any text or notification handle placed outside of the configured conditional variables are invoked with every monitor state transition. Any text or notification handle placed inside of configured conditional variables are only invoked with that specific monitor state transition.
 </br></br>
 If you configure a conditional block for a state transition into `alert` or `warning` condition with an **@-notifications** handle, it is recommended to configure a corresponding `recovery` condition in order for a recovery notification to be sent to the handle.
 </div>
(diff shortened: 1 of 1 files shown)


content/en/monitors/notify/variables.md
```diff
@@ -228,6 +228,11 @@ This is the escalation message @dev-team@company.com
 {{< /tabs >}}
 
 
+<div class="alert alert-info"><strong>Note</strong>: Any text or notification handle placed outside of the configured conditional variables are invoked with every monitor state transition. Any text or notification handle placed inside of configured conditional variables are only invoked with that specific monitor state transition.
+</br></br>
+If you configure a conditional block for a state transition into `alert` or `warning` condition with an **@-notifications** handle, it is recommended to configure a corresponding `recovery` condition in order for a recovery notification to be sent to the handle.
+</div>
```
buraizu, line 234:
```suggestion
If you configure a conditional block for a state transition into `alert` or `warning` conditions with an **@-notifications** handle, it is recommended to configure a corresponding `recovery` condition in order for a recovery notification to be sent to the handle.

**Note**: Any text or notification handle placed **outside** the configured conditional variables is invoked with every monitor state transition. Any text or notification handle placed **inside** of configured conditional variables is only invoked if the monitor state transition matches its condition.
```

buraizu (changes requested):
Thanks for the PR!
The new content seems slightly too long for a `div`, so this review suggests adding it in as regular text and moving it above the `Examples` section to be more prominent, and also reverses the order of the sections. 

I do see that a `div` is used for a similar note on line `338`, however. If you'd like to keep the `div`, would you be open to enclosing only the **Note** and leaving the section about recovery notifications outside the `div`?

Also, markdown syntax doesn't work inside of HTML elements -- if you decide to keep the `div`, the markdown will need to be updated to HTML (for example, \<code\>TEXT\</code\> instead of \`code\`).

Let me know if you have any questions!

content/en/monitors/notify/variables.md
cahillsf: cool -- that works for me!  thanks for the notes and feeback

buraizu (approved):

PR #16655, opened 2023-01-27
WEB-3095 | Update CI to use Rules

devindford:
### What does this PR do?

Convert our CI from using `only` and `except` to the newer `rules` convention

### Motivation

https://datadoghq.atlassian.net/browse/WEB-3095

### Additional Notes

Carefully review the changes made, and ensure that jobs will run as they did prior by comparison with only and except

---

### Reviewer checklist
- [ ] Check images for PII


.gitlab-ci.yml
```diff
@@ -61,20 +63,43 @@ before_script:
...
 .base_template: &base_template
   image: public.ecr.aws/x2b9z2t7/webops/site-build:latest
   tags:
     - "runner:main"
-  except:
-    - tags
-  only:
-    - branches
+  rules:
+    - if: $CI_PIPELINE_SOURCE == "tags"
```
StefonSimmons, line 92: Should the value for For  the $CI_PIPELINE_SOURCE be `push` or  instead use `if: $CI_COMMIT_TAG`? https://docs.gitlab.com/ee/ci/jobs/job_control.html#common-if-clauses-for-rules

.gitlab-ci.yml
```diff
@@ -61,20 +63,43 @@ before_script:
...
+      when: never
+    - if: $CI_COMMIT_BRANCH != null # Variable doesnt exists on merge request pipelines or tag pipelines.
```
StefonSimmons, line 94: can this just be `if: $CI_COMMIT_BRANCH` instead of `if: $CI_COMMIT_BRANCH != null`?

devindford: @StefonSimmons we probably could, I would need to do some digging but a lot of this is being purposefully explicit for readability for people who may not be fully familiar with how CI setups work. 

We also are doing something similar on the corp side for a rollback variable, makes the code more readable without the need for comments.

.gitlab-ci.yml
devindford: Good catch, the variable should be the tag one here, with a similar `!= null` to ensure it does not run on tagging.

StefonSimmons: i didnt see 'tags' as a possible value for the PIPELINE SOURCE

devindford: Yea it wasn't it was a mistake on my part, the commit tag you listed only exists when it's a tag job, which we want to prevent so by check for it not equaling null we will be able to prevent the run.

devindford:
@StefonSimmons made the changes we discussed, if you can give this another look when you get a chance please

StefonSimmons (approved):
👍 👍

urseberry (approved):

PR #17067, opened 2023-02-23
(DOCS-4786) AWS connection example and edits

hestonhoffman:
- Adds AWS connection example
- Removes references to "custom" connections. I think is more of an internal way of talking about these connections. It isn't really used in the UI (I could only find it in one tooltip), so I think our use of it in the docs could lead to a bit of confusion. 
- Moves some information from the authentication page to the connection page. Looking at this again, I think it makes sense to move all documentation that deals with action-level credentials to the connection page.

---

### Reviewer checklist
- [ ] Check images for PII


content/en/workflows/access.md
```diff
@@ -73,35 +72,9 @@ Because workflow [actions][5] connect with external software systems, you may ne
...
 
-Use a custom [connection][7] to authenticate a [generic action][8] or any action for which the integration tile does not offer authentication. The workflow connection object allows you to provide credentials to connect to an integration. The parameters vary by integration.
-
-To connect to an arbitrary service, use the HTTP Connection type and choose from two authentication options:
-- Token-based authentication
-- A username and password combination
-
-### Add credentials to an action
-
-Whether you configured credentials in the integration tile or created a custom connection, Workflows makes your credentials available for each workflow action. To add credentials to an action, follow the steps below:
```
hestonhoffman, line 97: I removed this section because I don't think is true now. Actions that use integration tile for credentials don't have a **Connection** drop-down section (I think the **Authentication** drop-down was renamed **Connection**

urseberry (approved):
A+ reorganization

PR #17151, opened 2023-03-02
dbm_apm: add table for supported tracers

EmmaOnThursday:
### What does this PR do?
Adds table of supported tracers with library and version information.

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/database_monitoring/guide/connect_dbm_and_apm.md
+++ b/content/en/database_monitoring/guide/connect_dbm_and_apm.md
@@ -16,12 +16,6 @@ This guide assumes that you have configured [Database Monitoring][1] and are usi
 
 ## Before you begin
 
-Supported tracers
-: [dd-trace-go][3] >= 1.44.0 (support for [database/sql][4] and [sqlx][5] packages)<br />
-[dd-trace-rb][6] >= 1.8.0 (support for [mysql2][7] and [pg][8] gems)<br />
-[dd-trace-js][9] >= 3.13.0 or >= 2.26.0 (support for [postgres][10], [mysql][13], and [mysql2][14] clients)<br />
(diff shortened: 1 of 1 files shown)


apigirl (approved):
looks good for docs

content/en/database_monitoring/guide/connect_dbm_and_apm.md
```diff
@@ -31,6 +25,30 @@ Supported Agent versions
 Data privacy
 : Enabling SQL comment propagation results in potentially confidential data (service names) being stored in the databases which can then be accessed by other third-parties that have been granted access to the database.
 
+
+**Supported tracers**
+
+| Language | Library                        | Postgres    | MySQL       |
```
alexandre-normand, line 31: I think this looks great but one thing that came up as we were doing the support matrix was the `library` terminology. I think maybe `library` is the most sensible term but some of the items in that column are actually drivers. I thought about using the generic term `technology` at some point but it's probably imperfect as well. Did you check how the rest of documentations refers to those _things_? I think as long as we align with that, this should be good.

EmmaOnThursday: i did not - great callout, i'll look around other docs.

EmmaOnThursday: oof we basically use different words for different languages. `library` and `framework` are the two most common. It seems like in the land of tracers, we don't reference `drivers` at all. I can update the column name to `Library or Framework`?

alexandre-normand: Yeah, that sounds good to me. Thanks for checking!

alexandre-normand (approved):

PR #17193, opened 2023-03-07
WEB-3284 - remove invalid integration categorization

davidejones:
### What does this PR do?

This PR:
- Uses the classifiers in the assetlib as a reference to only allow valid categorization of integrations
- Write to the `layouts/shortcodes/integration_categories.md` file with the classifiers so that they can be used by the `https://docs.datadoghq.com/developers/integrations/check_references/` page

### Motivation

https://datadoghq.atlassian.net/browse/WEB-3284

### Preview

- Test https://docs-staging.datadoghq.com/david.jones/validate-categories/integrations/ loads the same integrations.
- Test local build still works despite installing lib that isn't always available

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/layouts/shortcodes/integration_categories.md
+++ b/layouts/shortcodes/integration_categories.md
@@ -1,0 +1,71 @@
+- Category::Alerting
+- Category::Automation
+- Category::AWS
+- Category::Azure
+- Category::Caching
+- Category::Cloud
+- Category::Collaboration
+- Category::Compliance
+- Category::Configuration & Deployment
+- Category::Containers
+- Category::Cost Management
+- Category::Data Store
+- Category::Developer Tools
+- Category::Event Management
+- Category::Exceptions
+- Category::Google Cloud
+- Category::Incidents
+- Category::IOT
+- Category::ISP
+- Category::Issue Tracking
+- Category::Kubernetes
+- Category::Languages
+- Category::Log Collection
+- Category::Mainframe
+- Category::Marketplace
+- Category::Messaging
+- Category::Metrics
+- Category::Mobile
+- Category::Monitoring
+- Category::Network
+- Category::Notification
+- Category::Oracle
+- Category::Orchestration
+- Category::OS & System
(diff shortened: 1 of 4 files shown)


davidejones:
cc @hestonhoffman for visibility from our discussion

devindford (approved):

hestonhoffman:
Should we add `layouts/shortcodes/integration_categories.md` to .gitignore?

layouts/shortcodes/integration_categories.md
```diff
@@ -0,0 +1,71 @@
...
+- Supported OS::HP-UX
+- Supported OS::IBM i/OS
+- Supported OS::IBM z/OS
+- Supported OS::iOS
+- Supported OS::Linux
+- Supported OS::macOS
+- Supported OS::Solaris
+- Supported OS::Windows
+- Supported OS::Mac OS
+- Category::OS System
```
kayayarai, line 70: Is this different from line 34 `OS & System`?

davidejones:
> Should we add `layouts/shortcodes/integration_categories.md` to .gitignore?

@hestonhoffman  So this was intentional but it really depends on how this shortcode is used

- if its used in committed files we need to make sure `layouts/shortcodes/integration_categories.md` exists so it doesn't throw non existant errors for users who won't have access to the assetlib
- If its only used in autogenerated files we can gitignore it as you said as it will generate with the other files

It would be better to do the latter if you can confirm that its only going to be used in that way?

layouts/shortcodes/integration_categories.md
davidejones: I believe this is temporary while they are being corrected in upstream repos but yes this is a good point we don't want this to appear as a valid choice for new integrations

hestonhoffman:
@davidejones Okay, it will be used in committed files, so let's keep it.

hestonhoffman (approved):

PR #17228, opened 2023-03-09
WEB-3203 | Add redirect metadata job to Docs

devindford:
### What does this PR do?

- Adds redirect logic to documentation
- Removes redundant redirects
- Adds logic for integrations to remove redundant redirects that may exist 
- Removes logic that was creating approx 1500 redirects for OOTB security rules that were not in used (verified via logs that the redirects have not been hit)

I'm open to input, but would like to just roll the CI job out on previews for the time being to ensure we are not missing anything before we push this to live (maybe Monday?)

### Motivation

In an effort to get the correct status code of `301` for redirects from aliases. Hugo handles it with data in a temporary page that redirects to the alias, but gives a status code of `200`

### Preview
https://docs-staging.datadoghq.com/devin.ford/WEB-3203/

### Additional Notes

 Check different areas of the site and ensure redirects are behaving as we expect

The pipeline output during an integrations build will notify what aliases have been removed from integrations on intatke

Screenshot 2023-03-07 at 9 38 53 AM

This most recent "failing" [job](https://gitlab.ddbuild.io/DataDog/documentation/-/jobs/239309153) is working as intended. If there is a redundant redirect, the pipeline will complete adding the rest of the redirects to S3, but will not add the redundant ones identified, and instead will output them at the end of the job and have a failure status in order to have them addressed. In this case specifically, it is happening because the cache of integrations comes from the latest master branch, which does **NOT** have the new integrations logic ran to exclude these, but will be fixed once the master deploy runs and runs the new integrations code. 

You can see what that will look like in this [old job](https://gitlab.ddbuild.io/DataDog/documentation/-/jobs/239017332) where the cache was disabled, and the subsequent [passing job](https://gitlab.ddbuild.io/DataDog/documentation/-/jobs/239017346) for redirects.

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/local/bin/py/build/actions/security_rules.py
+++ b/local/bin/py/build/actions/security_rules.py
@@ -125,8 +125,8 @@ def security_rules(content, content_dir):
                 "type": "security_rules",
                 "disable_edit": True,
                 "aliases": [
-                    f"{data.get('defaultRuleId', '').strip()}",
-                    f"/security_monitoring/default_rules/{data.get('defaultRuleId', '').strip()}",
+                    # f"{data.get('defaultRuleId', '').strip()}",
+                    # f"/security_monitoring/default_rules/{data.get('defaultRuleId', '').strip()}",
                     f"/security_monitoring/default_rules/{p.stem}"
                 ],
                 "rule_category": [],
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -41,6 +41,7 @@ before_script:
   - export DATADOG_API_KEY=$(get_secret 'dd-api-key')
   - export DATADOG_APP_KEY=$(get_secret 'dd-app-key')
   - export BRANCH=${CI_COMMIT_REF_NAME}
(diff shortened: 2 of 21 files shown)


local/bin/py/build/actions/security_rules.py
```diff
@@ -125,8 +125,8 @@ def security_rules(content, content_dir):
                 "type": "security_rules",
                 "disable_edit": True,
                 "aliases": [
-                    f"{data.get('defaultRuleId', '').strip()}",
-                    f"/security_monitoring/default_rules/{data.get('defaultRuleId', '').strip()}",
+                    # f"{data.get('defaultRuleId', '').strip()}",
+                    # f"/security_monitoring/default_rules/{data.get('defaultRuleId', '').strip()}",
```
devindford, line 129: Commenting these out vs removing for now as an over precaution

davidejones (approved):

drichards-87 (approved):

PR #17302, opened 2023-03-15
WEB-3346 | Add redirect CI job for live

devindford:
### What does this PR do?

Add the redirect meta job to the live site in CI

### Motivation

Now that we have been running the redirect meta job in preview for a few days, and haven't seen odd behavior on redirects, we want to roll it out to the live site. 

For context:
This job runs post deploy, hugo will use the current alias pages while this job is still working to make sure redirects are always live on site. As the job works, it replaces the redirects in the bucket with metadata that will redirect as the link is hit and serve the proper `301` code. This job has built in logic to not allow redirects to the same page (causing redirect loops, hugo bypasses these for alias pages to avoid the same) and will inform the user at the end of the job what aliases should be removed since they are redundant. If you have any questions or want further clarification please feel free to slack me.

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -450,3 +450,18 @@ set_redirect_metadata_preview:
   script:
     - in-isolation update_redirect_metadata
   interruptible: true
+
+set_redirect_metadata_live:
+  <<: *base_template
+  <<: *live_rules
+  variables:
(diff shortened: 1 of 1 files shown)


urseberry (approved):
Thank you very much!

PR #17325, opened 2023-03-16
use public definition classifiers

davidejones:
### What does this PR do?

Use `get_non_deprecated_classifier_names` data instead when writing to shortcode

### Motivation

Currently we are outputting all our classifiers in the shortcode however some of these exist temporarily while we transition off them. We don't want them to appear in the docs and perpetuate its usage so switching to this different dataset should only include the items we want to publically display.

### Preview

Nothing to preview as this shortcode isn't used yet. 

### Additional Notes

Prerequisite to merging:
- [This PR is merged ](https://github.com/DataDog/dd-source/pull/28235)
- The assetlib version used in this pr is bumped to the version including above changes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/local/bin/py/build/actions/integrations.py
+++ b/local/bin/py/build/actions/integrations.py
@@ -28,7 +28,7 @@ from os.path import (
 from actions.format_link import format_link_file
 
 try:
-    from assetlib.constants import CLASSIFIER_TAGS
+    from assetlib.constants import CLASSIFIER_TAGS, get_public_classifiers
 except ImportError:
     CLASSIFIER_TAGS = []
     if getenv("CI_COMMIT_REF_NAME"):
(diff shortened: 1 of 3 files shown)


local/bin/py/build/actions/integrations.py
```diff
@@ -28,7 +28,8 @@
 from actions.format_link import format_link_file
 
 try:
-    from assetlib.constants import CLASSIFIER_TAGS
+    from assetlib.classifiers import get_all_classifier_names, get_non_deprecated_classifier_names
+    CLASSIFIER_TAGS = get_all_classifier_names()
```
nmuesch, line 32: Is the importError needed here? Why not instead import at the top of the file and let python fail if the assetlib isn't there?

davidejones: So the only reason I did that was just because some people may not have the library installed locally but still want the rest of the script to process.

nmuesch (approved):
LGTM from the APW side.

hestonhoffman (approved):
LGTM

bgdeutsch (approved):

PR #17349, opened 2023-03-17
Checkpoint on Friday's work importing details from Vector's CUE files.

silversupreme:
### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII


binarylogic (changes requested):
⚠️ I've been working to improve our JSON schema and I think I can improve it in such a way that we don't need the CUE data at all. I _really_ (really) do not want to add another dependency to our CUE data if we can avoid it. And we can use this PR as an opportunity to shift what we find useful out of that and into our JSON schema files.

PR #17354, opened 2023-03-20
Revert "Add UI image and info on slo calendar view"

estherk15:
This reverts commit 059f4d4ee19056ea97ff08ed61994984fa44f562.

### What does this PR do?
- This is not a publicly available feature, only available to a few customers

### Motivation
- Customers were confused about this, PM requested to remove this.

---

### Reviewer checklist
- [ ] Check images for PII


static/images/monitors/service_level_objectives/slo_status_page.png
maycmlee, line 1: We should probably keep the image, otherwise it will break for the translated docs (at least the JA one, the FR doesn't seem to have the image added yet) until the revert changes goes through the translation process.

estherk15: This isn't available and the image and info is causing more confusion as it is. What do you think about removing altogether. Will it break the full translation or just the image on the page?

maycmlee (approved):

PR #17373, opened 2023-03-21
Add documentation for running Gradle/Maven with DD Java tracer

nikita-tkachenko-datadog:
### What does this PR do?

This PR updates CI Visibility Tests product documentation for Java language.
The updated instructions explain how to run Maven/Gradle builds with DD Java Tracer attached.

### Motivation

A new version of DD Java Tracer is about to be released.
The new version supports instrumenting Maven/Gradle builds.
This instrumentation is needed to support CI Test Session Level Visibility feature, which is a part of CI Visibility offering.

### Additional Notes

In "Configuring reporting method" section agentless and agent-ful running modes swapped places.
Also a note saying that agent-ful method is the preferred one was removed.
Now that Test Session Level Visibility is about to be released, agentless mode is the preferred one.

---

### Reviewer checklist
- [ ] Check images for PII


content/en/continuous_integration/tests/java.md
```diff
@@ -287,10 +280,51 @@ test {
...
+{{< tabs >}}
+{{% tab "Maven" %}}
+
+Run your tests using the `MAVEN_OPTS` environment variable to specify a path to the Datadog Java Tracer JAR.
+
+In the tracer arguments, specify the following:
+
+* CI Visibility is enabled by setting the `dd.civisibility.enabled` property to `true`.
+* The environment where tests are being run (for example, `local` when running tests on a developer workstation or `ci` when running them on a CI provider) is defined in the `dd.env` property.
+* The name of the service or library under `test` is defined in the `dd.service` property.
```
nikita-tkachenko-datadog, line 297: To me highlighting `test` as code here seems slightly confusing.

In this case "library under test" means "library which is being tested". So "test" is used more as a general term, rather than as a specific command (the command itself might be different, for instance with Maven it can be `verify`).

Perhaps we could change the phrasing here then, for example to something like "the name of the service or library that is being tested is defined in the `dd.service` property"?

The same applies to the Gradle section below.

@alai97 please let me know what you think.

content/en/continuous_integration/tests/java.md
```diff
@@ -69,18 +57,31 @@ Additionally, configure which [Datadog site][2] to which you want to send data.
 
 [1]: https://app.datadoghq.com/organization-settings/api-keys
 [2]: /getting_started/site/
+
 {{% /tab %}}
+{{% tab "On-Premises CI provider (Datadog Agent)" %}}
+
+If you are running tests on an on-premises CI provider such as Jenkins or self-managed GitLab CI, you can install the Datadog Agent on each worker node by following the [Agent installation instructions][1]. Datadog recommends running CI Visibility with Agentless mode.
```
nikita-tkachenko-datadog, line 64: I wonder if we can somehow make it more obvious to the customers that it is possible to use Agentless mode both with cloud CI providers and with on-premise providers, i.e. they do not need to migrate from on-premise CI to cloud CI if they want to switch from using the Agent to Agentless.

I thought that "Cloud CI provider (Agentless)" tab name was a bit misleading in this sense, so I renamed it to "Agentless", but maybe there is a better way to convey it.

@alai97 what do you think?

alai97: @nikita-tkachenko-datadog My intention was to keep the tabs consistent across the Test pages (such as for [.NET](https://docs.datadoghq.com/continuous_integration/tests/dotnet/)).

```suggestion
If you are running tests on an on-premises CI provider such as Jenkins or self-managed GitLab CI, you can install the Datadog Agent on each worker node by following the [Agent installation instructions][1]. You can also run CI Visibility on an on-premises CI provider with Agentless mode.
```

content/en/continuous_integration/tests/java.md
alai97:
```suggestion
* The name of the service or library that is being tested is defined in the `dd.service` property.
```

content/en/continuous_integration/tests/java.md
```diff
@@ -287,10 +280,51 @@ test {
...
+{{% /tab %}}
+{{% tab "Gradle" %}}
+
+Run your tests using the `org.gradle.jvmargs` system property to specify a path to the Datadog Java Tracer JAR.
+
+In the tracer arguments, specify the following:
+
+* CI Visibility is enabled by setting the `dd.civisibility.enabled` property to `true`.
+* The environment where tests are being run (for example, `local` when running tests on a developer workstation or `ci` when running them on a CI provider) is defined in the `dd.env` property.
+* The name of the service or library under `test` is defined in the `dd.service` property.
```
alai97, line 318:
```suggestion
* The name of the service or library that is being tested is defined in the `dd.service` property.
```

nikita-tkachenko-datadog:
I completely reverted the change to Agentless/Agentful reporting method tabs, as there was a technical aspect that I missed. In the end both reporting methods are identical in terms of supported CI Visibility features, so we should not be changing this part at all.

albertvaka (approved):
I hope we can simplify the setup 😅 but as of now this LGTM.

nikita-tkachenko-datadog:
> I hope we can simplify the setup 😅 but as of now this LGTM.

Yep, the simplification is on the way, it will be done as a separate PR :)

content/en/continuous_integration/tests/java.md
```diff
@@ -287,13 +283,124 @@ test {
...
+./gradlew cleanTest test -Pdd-civisibility --rerun-tasks -Dorg.gradle.jvmargs=\
+-javaagent:$DD_TRACER_FOLDER/dd-java-agent-$DD_TRACER_VERSION.jar=\
+dd.civisibility.enabled=true,\
+dd.env=ci,\
+dd.service=my-java-app
+{{< /code-block >}}
+
+**Important:** Specifying `org.gradle.jvmargs` in the command-line will override the value specified elsewhere. If you have this property specified in a `gradle.properties` file, be sure to replicate the necessary settings in the command-line invocation
+
+**Note:** Currently CI Visibility is not compatible with [Gradle Configuration Cache][1], so do not enable the cache when running your tests with the tracer
```
alai97, line 401:
```suggestion
Specifying `org.gradle.jvmargs` in the command line overrides the value specified elsewhere. If you have this property specified in a `gradle.properties` file, be sure to replicate the necessary settings in the command line invocation.

**Note:** CI Visibility is not compatible with [Gradle Configuration Cache][1], so do not enable the cache when running your tests with the tracer.
```

PR #17444, opened 2023-03-28
Update log.md

jmyprasetyo:
Remove the "-" from the log config as this will resulted in creating 3 separate configs.

### What does this PR do?

This will make sure that we're targeting the right configuration. The original configuration is incorrect as it will creates 3 different configurations as shown here:
https://a.cl.ly/7KuzPAmN

Removing the - will resulted as the valid configuration as shown below:
https://a.cl.ly/2NubwPl0

### Motivation

Will reduce the number of ticket raised by the customer regarding configuration mistake.

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/containers/kubernetes/log.md
+++ b/content/en/containers/kubernetes/log.md
@@ -486,8 +486,8 @@ The following configuration defines the integration template for Redis container
         - redis
       logs:
         - source: redis
-        - service: redis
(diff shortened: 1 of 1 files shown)


urseberry (approved):
💯

PR #17726, opened 2023-04-14
use the "jq" command in dd-opentracing-cpp build instructions

dgoffredo:
The JSON endpoint `https://api.github.com/repos/DataDog/dd-opentracing-cpp/releases/latest` used to return pretty-printed JSON. We could then parse the output using standard line-based unix tools.

It is now returning a single line of JSON. This revision modifies the build example to parse the JSON using the non-standard `jq` command line tool instead.

### DIFF
--- a/content/en/tracing/trace_collection/dd_libraries/cpp.md
+++ b/content/en/tracing/trace_collection/dd_libraries/cpp.md
@@ -86,23 +86,29 @@ After the Agent is installed, follow these steps to add the Datadog tracing libr
 ### Compile against dd-opentracing-cpp
 
 ```bash
-# Gets the latest release version number from GitHub.
-get_latest_release() {
-  wget -qO- "https://api.github.com/repos/$1/releases/latest" |
-    grep '"tag_name":' |
-    sed -E 's/.*"([^"]+)".*/\1/';
-}
-DD_OPENTRACING_CPP_VERSION="$(get_latest_release DataDog/dd-opentracing-cpp)"
-# Download and install dd-opentracing-cpp library.
-wget https://github.com/DataDog/dd-opentracing-cpp/archive/${DD_OPENTRACING_CPP_VERSION}.tar.gz -O dd-opentracing-cpp.tar.gz
-mkdir -p dd-opentracing-cpp/.build
-tar zxvf dd-opentracing-cpp.tar.gz -C ./dd-opentracing-cpp/ --strip-components=1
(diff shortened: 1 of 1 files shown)


dgoffredo:
@wantsui and @cgilmour for visibility. I don't know Mohammad or Cécile's github handles :)

content/en/tracing/trace_collection/dd_libraries/cpp.md
```diff
@@ -86,23 +86,29 @@ After the Agent is installed, follow these steps to add the Datadog tracing libr
...
+  # Download and install dd-opentracing-cpp library.
+  wget https://github.com/DataDog/dd-opentracing-cpp/archive/${DD_OPENTRACING_CPP_VERSION}.tar.gz -O dd-opentracing-cpp.tar.gz
+  mkdir -p dd-opentracing-cpp/.build
+  tar zxvf dd-opentracing-cpp.tar.gz -C ./dd-opentracing-cpp/ --strip-components=1
+  cd dd-opentracing-cpp/.build
+  # Download and install the correct version of opentracing-cpp, & other deps.
+  ../scripts/install_dependencies.sh
+  # Configure the project, build it, and install it.
+  cmake ..
+  make -j
```
dgoffredo, line 109: `make -j` will spawn a lot of compiler processes, which might cause problems in resource constrained virtual environments (like a CI job).

I measured it on my laptop, and the max resident memory usage was less than 600 megabytes (presumably because there isn't that much code), so I think the default speed gain of `-j` is worth it.

content/en/tracing/trace_collection/dd_libraries/cpp.md
```diff
@@ -86,23 +86,29 @@ After the Agent is installed, follow these steps to add the Datadog tracing libr
...
-../scripts/install_dependencies.sh
-cmake ..
-make
-make install
+# Requires the "jq" command, which can be installed via
+# the package manager, for example "apt install jq",
+# "apk add jq", "yum install jq".
+if ! command -v jq >/dev/null 2>&1; then
+  >&2 echo "jq command not found. Install using the local package manager."
+else
```
dgoffredo, line 94: I put the entire real body of the script in an `else` branch so that I don't need to call `exit` above. If a user is copy/pasting this code into their shell, `exit` would exit the command shell, which would be rude.

cswatt (approved):

PR #17766, opened 2023-04-19
fix: specify full rum intake origin urls

relsunkaev:
### What does this PR do?
Specifies the full intake origin base urls. It seems they're missing the `rum` subdomain which is very confusing for readers (such as myself). I personally use us5 so I'm not 100% sure that all intake origins use the `rum` subdomain. Feel free to correct me.

### Motivation
Spent too much time figuring this out.

---

### Reviewer checklist
- [ ] Check images for PII


bcaudan:
Hello @relsunkaev,

We are in the process of removing product specific subdomain (`rum`, `logs`, ...) in the intake domains so we are already recommending to use the subdomain less version in this proxy setup documentation.
Cf [site documentation](https://docs.datadoghq.com/getting_started/site/#sdk-domains), mobile SDKs are already using those and browser SDK should do the same in the next major version.

PR #17867, opened 2023-04-25
config-update

jamiemilstein:
We are updating the configuration of heatmaps and as such, I removed the snippet from the docs


drichards-87 (approved):
The PR looks good. I'll get it merged now.

PR #17880, opened 2023-04-26
[.NET Tracing] Added note that we don't support partial trust

andrewlock:
### What does this PR do?
Adds a note that the .NET Tracer doesn't support partial trust.

### Motivation
We don't support running in partial trust, so want to document it publicly

### Additional Notes
Partial trust only applied to .NET Framework, so don't need to duplicate on .NET Core page

---

### Reviewer checklist
- [ ] Check images for PII


rtrieu (approved):
Looks good!

PR #18056, opened 2023-05-09
Removing new_group_delay section

SimoneT-DD:
The new_group_delay option is not supported for composite monitors so we are removing the doc section


rtrieu (approved):
Looks good!

hestonhoffman (approved):

PR #18099, opened 2023-05-11
semantic color build error

davidejones:
### What does this PR do?

semantic-color.md is generated in the build but it also used to exist locally to avoid local build errors such as cannot find file.

Recent changes mean that this file no longer exists locally by default, in order to avoid the build errors we now instead use a shortcode that will include the file if it exists.

### Motivation

Local build errors due to recent reconfigurations

### Preview

Check this page is still working
https://docs-staging.datadoghq.com/david.jones/fix-semantic-error/dashboards/guide/compatible_semantic_tags/

Check local build doesn't throw any errors

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/layouts/shortcodes/semantic-color.html
+++ b/layouts/shortcodes/semantic-color.html
@@ -1,0 +1,4 @@
+{{- $file := "layouts/shortcodes/semantic-color.md" -}}
+{{- if (fileExists $file) -}}
+  {{- $file | readFile | markdownify -}}
+{{- end -}}


bgdeutsch (approved):
i tested this by removing the semantic-color.md shortcode and building with yarn run start, and the build didn't error out.

StefonSimmons (approved):
The build doesnt error out locally running make start or yarn start. the table is still not being successfully from web-ui

alai97 (approved):

PR #18371, opened 2023-06-01
(maint) Fix Vale

hestonhoffman:
Fixes the Vale linter. Unfortunately, the config file doesn't like it when we use a file path glob. It only expects file extensions there. So the action will check everything, which takes a little longer. The action itself is still confined to content/en, so it shouldn't run on the translation PRs. And it should only fail on errors added in the PR.

Should be good to merge. 

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/.vale.ini
+++ b/.vale.ini
@@ -1,7 +1,7 @@
 StylesPath = "../datadog-vale/styles"
 MinAlertLevel = suggestion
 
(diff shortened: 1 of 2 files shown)


rtrieu (approved):
LGTM

PR #18391, opened 2023-06-05
Stefon.simmons/integrations vid img carousel - [WEB-2951]

StefonSimmons:
### What does this PR do?

1. adds infinity carousel and modal components to integration pages with a non empty media array
   - written using alpine.js
   - adds global alpine store for methods shared between the main carousel and modal.
   - no modal appears on mobile view (<768px)

2. remove cache from parent "footer/footer.htm" partial and cache child partial "global-modals/global-modals.html" instead
   - allows for dynamic integrations carousel modal partial in "footer/footer.html"

### Motivation

https://datadoghq.atlassian.net/jira/software/projects/WEB/boards/102?selectedIssue=WEB-2951

### Preview

https://docs-staging.datadoghq.com/stefon.simmons/integrations-vid-img-carousel/integrations/

ex of multiple: https://docs-staging.datadoghq.com/stefon.simmons/integrations-vid-img-carousel/integrations/buoyant_inc_buoyant_cloud/

ex of multiple w/ video in carousel: https://docs-staging.datadoghq.com/stefon.simmons/integrations-vid-img-carousel/integrations/bigpanda_saas/

ex of single slide carousel: https://docs-staging.datadoghq.com/stefon.simmons/integrations-vid-img-carousel/integrations/agentil_software_services_5_days/

ex of carousel w/ less 3 or less slides: https://docs-staging.datadoghq.com/stefon.simmons/integrations-vid-img-carousel/integrations/avmconsulting_workday/

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII


layouts/partials/footer/footer.html
```diff
@@ -13,7 +13,13 @@
     {{ $.Scratch.Set "data" $dot.Page.Site.Data.partials.footer }}
 {{ end }}
 
+<!--- footer from websites-modules --->
 {{ partial "footer.html" . }}
 
 <!--- global modals --->
-{{ partial "global-modals/global-modals.html" . }}
+{{ partialCached "global-modals/global-modals.html" . }}
```
davidejones, line 20: This seemed fine on first inspection but looking closer won't this cause the i18n calls in the html to be cached? we might end up with the ja output used for all or the en output used for all etc.

layouts/partials/footer/footer.html
```diff
@@ -13,7 +13,13 @@
...
+
+<!--- integrations carousel modal --->
+{{ if in .Params.categories "marketplace" }}
+{{ partial "integrations-carousel/integrations-carousel-modal.html" . }}
+{{ end }}
```
davidejones, line 25: I'm also wondering if we removed the partialcached from footer just so we could add this dynamic piece, should this just live somewhere else instead? or should we keep partialcached on footer and pass the integration variant  or something?

layouts/integrations/single.html
```diff
@@ -1,12 +1,18 @@
 {{ define "main" }}
+    {{$title := .Params.integration_title | default .Title}}
     <div class="row">
         <div class="col-12 order-1">
-            <h1 id="pagetitle">{{ if .Params.integration_title }}{{ .Params.integration_title }}{{ else }}{{.Title}}{{ end }}</h1>
+            <h1 id="pagetitle">{{$title}}</h1>
```
davidejones, line 5: Much nicer! i'm just thinking we don't need to assign this to $title if we aren't using it anywhere else. So maybe just ```<h1 id="pagetitle">{{.Params.integration_title | default .Title}}</h1>```

layouts/partials/footer/footer.html
davidejones: Also i don't think we can scope this to just marketplace. There are other integrations that have media e.g 
https://github.com/DataDog/integrations-extras/blob/master/1e/manifest.json#L15
https://app.datadoghq.com/integrations/one-e?search=1e

layouts/partials/integrations-carousel/integrations-carousel.html
```diff
@@ -0,0 +1,205 @@
...
+            </div>
+            <div class="item-caption">
+                <p class="m-0">{{$el.caption}}</p>
+            </div>
+        </div>
+        {{ end }}
+    </div>
+</div>
+
+<script src="https://player.vimeo.com/api/player.js"></script>
```
davidejones, line 205: should we consider including this player.js somewhere else?

davidejones (changes requested):

layouts/partials/integrations-carousel/integrations-carousel.html
davidejones: Maybe something like https://github.com/DataDog/documentation/blob/master/layouts/partials/header-scripts.html#L65-L72 if eq integrations ?

layouts/partials/footer/footer.html
StefonSimmons: The card said to target marketplace but I just enabled the carousel on al integrations with media: ex: https://docs-staging.datadoghq.com/stefon.simmons/integrations-vid-img-carousel/integrations/launchdarkly/

Re: the partialCached I can add it to the bottom of the baseof. I figured the footer/footer was cached originally because of the global-modal partial

StefonSimmons: 1e: https://docs-staging.datadoghq.com/stefon.simmons/integrations-vid-img-carousel/integrations/1e/

layouts/partials/integrations-carousel/integrations-carousel.html
StefonSimmons: I thought this was appropriate since it's only for controlling the integrations-carousel videos and it would load on the page for only pages where the integration-carousel loads. Is there a reason why we wouldnt keep it here?

layouts/partials/footer/footer.html
StefonSimmons: I would be opposed to moving the modal to the baseof instead. I figured the footer/footer was originally cached because global-modals didnt have any conditionals included in the partial

layouts/integrations/single.html
StefonSimmons: that works 👍

layouts/partials/footer/footer.html
StefonSimmons: i moved it to baseof and cached the footer again

layouts/partials/integrations-carousel/integrations-carousel.html
davidejones: Just a script tag can be render blocking. I also think the js lib can be easily forgotten when its included this way. Not a huge deal though, lets leave this as is but can you create another card to update after.

davidejones (approved):

PR #18437, opened 2023-06-07
Update csharp.md

shivinders:
need clarification on what tailing a file with datadog agent mean

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII


hestonhoffman:
Hi @shivinders, thanks for the question. "Tailing" is an industry term that refers to printing the end of a log file (usually the last 10 lines). It comes from the `tail` command in Linux, which is perhaps why it seems out of place in a CSharp file *CSharp is more common in the Windows realm). In PowerShell, you can use something like `Get-Content .\path\to\log\file -Tail 10` to tail a file. I'm going to close this PR, but I'll think about how to clarify the operation in this doc and open a separate PR to fix it.

PR #18485, opened 2023-06-12
Fix links for instrumenting Go tracing

avedmala:
There are two places in the docs where we are trying to point to the Go tab of instrumentation but it instead points to Ruby, this PR fixes those

---

### Reviewer checklist
- [ ] Check images for PII


Chronobreak (approved):
Nice catch! 👍

cswatt (approved):

PR #18506, opened 2023-06-13
[ci-visibility] Add troubleshooting item about JUnit reports timestamps

juan-fernandez:
### What does this PR do?
Add troubleshooting item to reflect that old timestamps will be discarded. 

### Motivation
Make it easier to customer to find issues.

### Preview

https://docs-staging.datadoghq.com/juan-fernandez/add-note-about-junit-timestamps/continuous_integration/troubleshooting/#you-are-uploading-junit-test-reports-with-datadog-ci-but-some-or-all-tests-are-missing

---

### Reviewer checklist
- [ ] Check images for PII


rodrigo-roca:
It looks good to me 👍 
We need to have in mind that this might need to get updated in the near future because the 18h is going to be shortened
Anyhow even if the hour amount is wrong it would still be more useful than no documentation at all in my opinion

AdrianLC (approved):

rtrieu (approved):
Looks good!

PR #18520, opened 2023-06-13
Fix Kubernetes sample manifests and generate script

JacksonDavenport:
### What does this PR do?

The main thing this does is fix the `generate.sh` script and regenerate the manifests, as in Helm chart `3.5.1` the default values.yaml file was updated to remove the dummy API Key. Without an API Key the Helm chart doesn't generate the manifests. So the generate script was creating manifests without the actual Agent DaemonSet. The API Key still gets replaced with the placeholder script, so this is just to help generation.

Minor changes to the DaemonSet instructions to:
- Cleanup the table formatting (just in the raw form so its easier to read) and name of DaemonSet. 
- Remove the Operator instructions for unprivileged as we have the actual DaemonSet instructions lower down. 
- Remove instructions for the legacy KSM steps as we are favoring the KSM Core integration. 

### Motivation

The manifests did not contain the actual DaemonSet needed

### Additional Notes

Its probably worth re-visting this doc and Cluster Agent Setup doc to combine the steps for the manual Agent and Cluster Agent instructions. Similar to how the `guide/kubernetes_daemonset` doc was created after stripped out the DaemonSet instructions from the main installation page with all Helm, Operator, DaemonSet. But that is going to be a much bigger change than this. 

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/containers/guide/kubernetes_daemonset.md
+++ b/content/en/containers/guide/kubernetes_daemonset.md
@@ -27,14 +27,14 @@ To install the Datadog Agent on your Kubernetes cluster:
 
 2. **Create the Datadog Agent manifest**. Create the `datadog-agent.yaml` manifest out of one of the following templates:
 
-    | Metrics                   | Logs                      | APM                       | Process                   | NPM                       | Security                       | Linux                   | Windows                 |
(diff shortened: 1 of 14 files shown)


rtrieu (approved):
Looks good!

PR #18543, opened 2023-06-14
fix some OpenTelemetry branding

lucaspimentel:
### What does this PR do?

Replace somes uses of "OTel" in public documentation with "OpenTelemetry", which is the preferred branding.
See https://opentelemetry.io/

### Motivation

"OTel" is a common nickname internally, but we should use correct branding of 3rd party entities, projects, libraries, etc in our public documentation.

### Additional Notes

We should probably make these same changes to the other language pages in the same path, and anywhere else where "OTel" or "Open Telemetry" (with a space) is used.

---

### Reviewer checklist
- [ ] Check images for PII


buraizu (approved):
LGTM

PR #18644, opened 2023-06-22
python tracer incorrectly designed "PHP tracer"

deronnax:
it's a very simple 5 characters fix.


estherk15 (approved):
Thanks for the fix @deronnax!

PR #18695, opened 2023-06-26
DOCS-4953 azure support matrix

cswatt:
adds an azure support matrix as an faq-type page so it doesn't get search indexed.


estherk15 (approved):
Nice! How are users going to find this page?

PR #18822, opened 2023-07-06
Document JIT vs allocation profiling

realFlowControl:
### What does this PR do?

Document the fact that an active JIT disables PHP allocation profiling.

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/profiler/profile_types.md
+++ b/content/en/profiler/profile_types.md
@@ -209,9 +209,11 @@ CPU
 
 Allocations (v0.88+)
 : The number of allocations by each function during the profiling period (default: 67s), including allocations which were subsequently freed. Stack allocations are not tracked.
+_Note: Not availble when JIT is active_
 
 Allocated memory (v0.88+)
 : The amount of heap memory allocated by each function during the profiling period (default: 67s), including allocations which were subsequently freed. Stack allocations are not tracked.
+_Note: Not availble when JIT is active_
 
 [1]: /profiler/enabling/php/#requirements
 {{< /programming-lang >}}
--- a/content/en/profiler/profile_types.md
+++ b/content/en/profiler/profile_types.md
@@ -213,7 +213,7 @@ _Note: Not available when JIT is active_
 
 Allocated memory (v0.88+)
 : The amount of heap memory allocated by each function during the profiling period (default: 67s), including allocations which were subsequently freed. Stack allocations are not tracked.<br />
-_Note: Not availble when JIT is active_
+_Note: Not available when JIT is active_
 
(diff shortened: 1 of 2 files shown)


content/en/profiler/profile_types.md
```diff
@@ -208,10 +208,12 @@ CPU
 : Shows the time each function spent running on the CPU.
 
 Allocations (v0.88+)
-: The number of allocations by each function during the profiling period (default: 67s), including allocations which were subsequently freed. Stack allocations are not tracked.
+: The number of allocations by each function during the profiling period (default: 67s), including allocations which were subsequently freed. Stack allocations are not tracked.<br />
+_Note: Not availble when JIT is active_
```
buraizu, line 212:
```suggestion
_Note: Not available when JIT is active_
```

content/en/tracing/trace_collection/library_config/php.md
```diff
@@ -120,8 +120,8 @@ Whether to enable the endpoint data collection in profiles. Added in version `0.
 `DD_PROFILING_ALLOCATION_ENABLED`
 : **INI**: `datadog.profiling.allocation_enabled`. INI available since `0.88.0`.<br>
 **Default**: `1`<br>
-Enable the allocation size and allocation bytes profile type. Added in version `0.88.0`.<br>
-**Note**: This supersedes the `DD_PROFILING_EXPERIMENTAL_ALLOCATION_ENABLED` environment variable (`datadog.profiling.experimental_allocation_enabled` INI setting), which was available since `0.84`. If both are set, this one takes precedence.
+Enable the allocation size and allocation bytes profile type. Added in version `0.88.0`. When an active JIT is detected, allocation profiling will be turned off due to a bug in the ZendEngine.<br>
```
buraizu, line 123:
```suggestion
Enable the allocation size and allocation bytes profile type. Added in version `0.88.0`. When an active JIT is detected, allocation profiling is turned off due to a limitation of the ZendEngine.<br>
```

buraizu: We don't typically list bugs in the documentation -- is this something that is expected to be resolved/changed in the near future? If so, we can create an additional card to track that, and update this section again when it's resolved.
If this is not expected to change in the near term, do you think the language suggested above would be sufficient?

realFlowControl: It should be fixed in the upcoming PHP versions 8.2.8 and 8.1.21 as soon as they are released. But we need to double check this when the release has been tagged

buraizu (approved):

PR #19254, opened 2023-08-03
Add metadata so 'inbox' finds CSM and Case Management pages

kayayarai:


brett0000FF (approved):
Very nice. Tried out the "inbox" search on your preview site, and the results look much better.

PR #19271, opened 2023-08-04
Fix .NET logo, redux

lucaspimentel:
### What does this PR do?

Follow up from #19266.

- Switch the .NET logo used in several additional pages from either the ".NET Framework" logo or the ".NET Core" to use the ".NET" logo instead
- Update the .NET logo image file to add more padding around the logo so it more closely matches the other language logos (Java, Go, Ruby, etc)

| | Before | After |
|-|--------|--------|
| Image | Before | After |
| Dimensions | 200x128 | 200x128 |
| File size | 1,110 bytes | 1,006 bytes | 

Pages changed:

| Before | After (staging preview) |
|--------|--------|
| https://docs.datadoghq.com/serverless/installation/ | https://docs-staging.datadoghq.com/lpimentel/dotnet-logo/serverless/installation/ |
| https://docs.datadoghq.com/data_streams/ | https://docs-staging.datadoghq.com/lpimentel/dotnet-logo/data_streams/ |

Note 1: I couldn't find usages of the OpenTracing or Logs partials shown in the file diff.
Note 2: The staging previews show the original logo image, not the newer image with more padding.

### Motivation

The reason for switching the logo is that some pages reference only one .NET and don't need to reference both ".NET Framework" and ".NET Core" separately. Also, ".NET" is the correct branding since .NET 5.

The reason for add padding to the image is to be more visually consistent with the other language logos. Before this change, .NET logo looked larger than the others because it had less white padding around the logo.

### Additional Notes

I optimized the new image file size as done in #18791.

As note above, the staging preview links use the original image file. I couldn't find a way to link to a preview that uses the new image file. It seems that staging previews keep referencing the original image locations.

---

### Reviewer checklist
- [ ] Check images for PII


urseberry (approved):
Regarding the new images, our static content system caches images. The new ones will show up eventually on the live site. There is a delay, though. If it's important to you to show the new images right away, it's best to create a new one rather than overwriting the existing one.

Leaving this unmerged in case you'd prefer to create a new image. Let me know by leaving a comment on the PR!

lucaspimentel:
@urseberry

> our static content system caches images

Thanks for clarifying!

> If it's important to you to show the new images right away

Not urgent at all. We can merge the PR. Thanks!

PR #19413, opened 2023-08-17
connect_traces_and_profiles: Go timeline tags and facets

felixge:
### What does this PR do?

CleanShot 2023-08-17 at 16 12 35@2x

### Motivation

A customer asked for more documentation, specifically around stuff like this.

### Additional Notes

---

### Reviewer checklist
- [ ] Check images for PII

### DIFF
--- a/content/en/profiler/connect_traces_and_profiles.md
+++ b/content/en/profiler/connect_traces_and_profiles.md
@@ -75,13 +75,22 @@ os.Setenv("DD_PROFILING_EXECUTION_TRACE_PERIOD", "15m")
 
 Setting these variables will record up to 1 minute (or 5 MiB) of execution tracing data [every 15 minutes][2].
 
-While recording execution traces, your application may observe an increase in CPU usage similar to a garbage collection. Although this should not have a significant impact for most applications, Go 1.21 includes [patches][3] to eliminate this overhead.
+You can find this data:
+
+- In the [Profile List][3] by adding `go_execution_traced:yes` to your search query. Click on a profile to view the [Profile Timeline][4]. To go even deeper, download the profile and use `go tool trace` or [gotraceui][5] to view the contained `go.trace` files.
+- In the [Trace Explorer][6] by adding `@go_execution_traced:yes` (note the `@`) to your search query. Click on a span and then select the `Code Hotspots` tab to view the [Span Timeline](#span-execution-timeline-view).
+
+While recording execution traces, your application may observe an increase in CPU usage similar to a garbage collection. Although this should not have a significant impact for most applications, Go 1.21 includes [patches][7] to eliminate this overhead.
 
 This capability requires `dd-trace-go` version 1.37.0+ (1.52.0+ for timeline beta) and works best with Go version 1.18 or later (1.21 or later for timeline beta).
 
 [1]: /profiler/enabling/go
 [2]: https://github.com/DataDog/dd-trace-go/issues/2099
-[3]: https://blog.felixge.de/waiting-for-go1-21-execution-tracing-with-less-than-one-percent-overhead/
+[3]: /profiler/profile_visualizations/#single-profile
+[4]: /profiler/profile_visualizations/#timeline-view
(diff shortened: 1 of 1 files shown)


content/en/profiler/connect_traces_and_profiles.md
```diff
@@ -75,13 +75,22 @@ os.Setenv("DD_PROFILING_EXECUTION_TRACE_PERIOD", "15m")
 
 Setting these variables will record up to 1 minute (or 5 MiB) of execution tracing data [every 15 minutes][2].
 
-While recording execution traces, your application may observe an increase in CPU usage similar to a garbage collection. Although this should not have a significant impact for most applications, Go 1.21 includes [patches][3] to eliminate this overhead.
+You can find this data:
+
+- In the [Profile List][3] by adding `go_execution_traced:yes` to your search query. Click on a profile to view the [Profile Timeline][4]. To go even deeper, download the profile and use `go tool trace` or [gotraceui][5] to view the contained `go.trace` files.
```
nsrip-dd, line 80: Do we need to suggest using another tool here? Not that I'm opposed to it, those are both fine tools, it just seems to me that our documentation should focus on how to use our product. e.g. we don't suggest that Java users hop into JMC for in-depth JFR analysis, or to use `go tool pprof` to view pprof-formatted data locally, even though those are reasonable things to do sometimes.

content/en/profiler/connect_traces_and_profiles.md
```diff
@@ -75,13 +75,22 @@ os.Setenv("DD_PROFILING_EXECUTION_TRACE_PERIOD", "15m")
...
+- In the [Trace Explorer][6] by adding `@go_execution_traced:yes` (note the `@`) to your search query. Click on a span and then select the `Code Hotspots` tab to view the [Span Timeline](#span-execution-timeline-view).
+
+While recording execution traces, your application may observe an increase in CPU usage similar to a garbage collection. Although this should not have a significant impact for most applications, Go 1.21 includes [patches][7] to eliminate this overhead.
```
nsrip-dd, line 83: I realize this line hasn't actually changed, but I don't see any reference to what an "execution trace" is in this documentation. Should we provide a link? And maybe a quick reference to what it is in-line, like "While recording execution traces, the data powering Go timelines, ..."?

nsrip-dd (approved):
Thanks Felix! Looks good to me. I left two comments (meant to submit them together...) but I consider them both non-blocking.

content/en/profiler/connect_traces_and_profiles.md
felixge: I agree that we should not try to explain how to use 3rd party tools for viewing our data. But we're not doing that, we're just pointing out their existence.

IMO this is the right thing to do here for a few reasons:

1. It really clarifies what kind of data we're collecting. There is a lot of public information you can find with these keywords.
2. It explains why we have a download feature. Our current documentation only mentions that we have it, but not how we expect people to use it.
3. It may help users to understand the data in more depth, and even make good suggestions for improvements we could make in our product.

On the other side, I don't see compelling reasons for withholding this information.

nsrip-dd: The only reason to not include would be if it's out of scope for our documentation. I agree with what you said, just wanted to consider whether it's in scope. But, indeed it's good to know that our data is the same data you get straight from the runtime, that there's nothing proprietary or special about it, and that it's available if you need it :)

content/en/profiler/connect_traces_and_profiles.md
nsrip-dd: Sure! I'm happy to do that

content/en/profiler/connect_traces_and_profiles.md
felixge: Let's keep it as-is for now. Unless the docs team has a strong view on this, I think it's okay.

rtrieu (approved):
LGTM

PR #19453, opened 2023-08-21
Update dual-shipping docs for orchestrator to include helm instuctions

DaniCalifornia:
Adding helm instructions for dual shipping orchestrator data. There is no case where you'd add config to the `datadog.yaml` so I removed it to avoid confusion


hestonhoffman (approved):
Perfect, thanks @DaniCalifornia!

PR #19655, opened 2023-09-07
Add search terms for ksm

hestonhoffman:
Need to test before merging

---

### Reviewer checklist
- [ ] Check images for PII
- [ ] Check that `cache_enabled` is set to `true` in the `pull_config_preview.yaml` file

### DIFF
--- a/.vale.ini
+++ b/.vale.ini
@@ -1,6 +1,4 @@
-StylesPath = 'styles'
-Packages = https://github.com/DataDog/datadog-vale/archive/refs/tags/0.1.1.zip
-
+StylesPath = "../datadog-vale/styles"
 MinAlertLevel = suggestion
 
 [*.md]
--- a/.vale.ini
+++ b/.vale.ini
@@ -1,4 +1,6 @@
-StylesPath = "../datadog-vale/styles"
+StylesPath = 'styles'
+Packages = https://github.com/DataDog/datadog-vale/archive/refs/tags/0.1.1.zip
+
 MinAlertLevel = suggestion
 
 [*.md]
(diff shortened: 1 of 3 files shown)


nsollecito:
should we look at defining a synonym for "ksm" as well?

hestonhoffman:
@nsollecito Yes, I think that's a good idea. I'll add it to the algolia file. I just realized I haven't been doing this to other abbreviations 😨 I'll do a search for those later

hestonhoffman:
@nsollecito Does the object ID need to be one word?

nsollecito:
> @nsollecito Does the object ID need to be one word?

yes, I think it should be.. according to the docs, the object id is a unique identifier for the entry: https://www.algolia.com/doc/api-reference/api-methods/save-synonym/#createupdate-a-regular-synonym-where-all-the-words-are-equivalent

The array containing the synonyms in this config treats all of the values as if they were equal, so it's not a mapping between the object id and the synonym values. If you want to include a one-way mapping, it looks like we need to use a different structure. Let me know if you want us to take a closer look. 

cc @bgdeutsch

hestonhoffman:
> > @nsollecito Does the object ID need to be one word?
> 
> yes, I think it should be.. according to the docs, the object id is a unique identifier for the entry: https://www.algolia.com/doc/api-reference/api-methods/save-synonym/#createupdate-a-regular-synonym-where-all-the-words-are-equivalent
> 
> The array containing the synonyms in this config treats all of the values as if they were equal, so it's not a mapping between the object id and the synonym values. If you want to include a one-way mapping, it looks like we need to use a different structure. Let me know if you want us to take a closer look.
> 
> cc @bgdeutsch

Gotcha - Not sure a one-way mapping is really necessary in this case, so I folded all of the terms into the single entry. A closer look would be great :)

nsollecito (approved):

rtrieu (approved):
Looks great, thank you so much!

PR #19744, opened 2023-09-13
[php tracing] update url to library releases

irabinovitch:
[php tracing] update url to library releases

update url to library releases for php tracing libraries. Old link pointed at service list rather than installation bits.


cswatt:
thanks ilan

cswatt (approved):

PR #18483, opened 2023-06-12
Add new APT and RPM signing keys

chouetz:
### What does this PR do?
Update GPG keys for key rotation

### Motivation
Key rotation process

### Additional Notes

---

### Reviewer checklist
- [ ] Review the changed files.
- [ ] Review the URLs listed in the [Preview](#preview) section.
- [ ] Check images for PII
- [ ] Review any mentions of "Contact Datadog support" for internal support documentation.

PR #19784, opened 2023-09-18
(DOCS-6224) apt-key is deprecated

hestonhoffman:

The `apt-key` command is deprecated. If I understand this correctly, `apt-key` is a wrapper for the `gpg` command and we're already running `gpg`, so I don't think `apt-key` is needed.

Mimics the manual instructions on the [Ubuntu installation page](https://app.datadoghq.com/account/settings/agent/latest?platform=ubuntu) in the app.

### Merge instructions

### DIFF
--- a/content/en/agent/guide/linux-agent-2022-key-rotation.md
+++ b/content/en/agent/guide/linux-agent-2022-key-rotation.md
@@ -77,12 +77,11 @@ Datadog encourages you to use one of the [install methods](#install-methods-that
 
 Run the following commands on the host:
 
-```bash
-$ curl -o /tmp/DATADOG_APT_KEY_F14F620E https://keys.datadoghq.com/DATADOG_APT_KEY_F14F620E.public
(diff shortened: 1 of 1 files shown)


content/en/agent/guide/linux-agent-2022-key-rotation.md
chouetz, line 1: Hi, I'm not sure we should edit this file (see [previous comment](https://github.com/DataDog/documentation/pull/18483#discussion_r1235161417)). It was made only for support of the 2022 linux key rotation, so eventually we'd rather remove the whole file, I guess

chouetz (commented):
I think we can discard this update, and rather propose the suppression of the whole file

content/en/agent/guide/linux-agent-2022-key-rotation.md
hestonhoffman: No probs, I'll close the PR. This feedback came straight from a customer, so I'm hesitant to remove it completely. I'll follow up with the reporter and see if I can gather more info on what the customer was doing

PR #19790, opened 2023-09-18
Update interpolation-the-fill-modifier-explained.md

JakeYankovich:
Previously, the "How to control interpolation?" section said, "Interpolation is disabled by the .as_count() and .as_rate() modifiers when used on any metric type."

This seems to imply that this works for *any metric type*. Our metric type modifiers doc directly contradicts this: "GAUGE metric types represent the absolute and final value of a metric; as_count() and as_rate() modifiers have no effect on them." https://docs.datadoghq.com/metrics/custom_metrics/type_modifiers/?tab=gauge#in-application-modifiers

After some brief testing, the truth seems to be that gauge type metrics are not affected by those modifiers.

### Merge instructions


maycmlee:
Thanks @JakeYankovich for opening this PR! Created[ DOCS-6242](https://datadoghq.atlassian.net/browse/DOCS-6242) for review.

aliciascott (approved):
Hi @JakeYankovich thanks for adding this! I fixed the link (links are numbered at the bottom and then referenced above), but now you know for next time :) And then cleaned it up some as it had a bit of redundant information. In this case it is fine just to state Gauge is the exception and link back to the page that explains this. Additionally, interpolation can still disabled when using the `.fill(null)`, but this is described in the below table in the docs on this page, so I think it is more clear now. I will approve and merge this.

PR #19843, opened 2023-09-20
Dashboards Querying Sentence Update

alai97:

Flagged by a user from [Hotjar](https://insights.hotjar.com/sites/1021060/feedback/responses?sort_by=-index&frid=95224000): “Hey great doc! This sentence talks about 1 minute buckets, but the screenshot below is set to 1 seconds.”

### Merge instructions


buraizu (commented):
Thanks for flagging @alai97, but it looks like it might make more sense to update the screenshot in this case:
- The associated JSON examples below the screenshot reference `60` in their queries, so it's likely meant to be `1 minute`
- Metrics aren't reported in 1-second intervals by default, so it might be a misleading example for some users

Thanks again, and let me know if you have any insights or feedback around that!

buraizu (approved):
LGTM!

PR #19896, opened 2023-09-25
switch sourcing workflows

davidejones:

This PR sources workflow automation rules via git instead of github api calls as it was not filtering files it required.

```
Note: The limit for the tree array is 100,000 entries with a maximum size of 7 MB when using the recursive parameter.
```
from https://docs.github.com/en/rest/git/trees?apiVersion=2022-11-28#get-a-tree

### Merge instructions

### Additional notes
https://docs-staging.datadoghq.com/david.jones/switch-sourcing-workflows/service_management/workflows/actions_catalog/


local/bin/py/build/configurations/pull_config_preview.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 - config:
-    cache_enabled: true
+    cache_enabled: false
```
davidejones, line 3:
```suggestion
    cache_enabled: true
```

Apply before merging

bgdeutsch (approved):
no issues locally or in preview, and build logs look clean as well.    lgtm 👍

aliciascott (approved):

PR #19904, opened 2023-09-25
Update python.md b3 propagation style configs

ZStriker19:
We've had `b3` propagations style deprecated and suggested `b3multi` for some time. With Python tracer version 2.0.0 we remove `b3` propagation style, therefore we should remove it here as well, and customers should use `b3multi` instead.
### Merge instructions


brett0000FF (approved):
LGTM--thanks!

PR #19933, opened 2023-09-27
Update Intelligent Test Runner docs: remove application key setup requirement

nikita-tkachenko-datadog:

Updates CI Visibility Intelligent Test Runner setup documentation.
ITR endpoints have been updated recently and no longer require application key to be included with the requests.
The docs are updated accordingly.

### Merge instructions


tonyredondo (approved):

juan-fernandez (changes requested):
thanks for this! The PR has the do not merge label but I'll block just as extra precaution: I'm not 100% sure what version of dd-trace-js will include the removal of the app key so I can't update that yet

juan-fernandez (approved):
I've added the compatibility for javascript so I'm unblocking

content/en/continuous_integration/intelligent_test_runner/javascript.md
```diff
@@ -54,9 +52,7 @@ NODE_OPTIONS="-r dd-trace/ci/init" DD_ENV=ci DD_SERVICE=my-javascript-app yarn t
 
 {{% tab "Cloud CI provider (Agentless)" %}}
 
-{{% ci-itr-agentless %}}
-
-After setting these environment variables, run your tests as you normally do:
+After completing setup, run your tests as you normally do:
 
 {{< code-block lang="shell" >}}
 NODE_OPTIONS="-r dd-trace/ci/init" DD_ENV=ci DD_SERVICE=my-javascript-app DD_CIVISIBILITY_AGENTLESS_ENABLED=true DD_API_KEY=$DD_API_KEY DD_APPLICATION_KEY=$DD_APPLICATION_KEY yarn test
```
romainkomorndatadog, line 58:
```suggestion
NODE_OPTIONS="-r dd-trace/ci/init" DD_ENV=ci DD_SERVICE=my-javascript-app DD_CIVISIBILITY_AGENTLESS_ENABLED=true DD_API_KEY=$DD_API_KEY yarn test
```

content/en/continuous_integration/intelligent_test_runner/javascript.md
```diff
@@ -15,17 +15,17 @@ further_reading:
 Intelligent Test Runner is only supported in the following versions and testing frameworks:
 
 * `jest>=24.8.0`
-  * From `dd-trace>=3.16.0` or `dd-trace>=2.29.0`.
+  * From `dd-trace>=4.16.0` or `dd-trace>=3.37.0`.
```
romainkomorndatadog, line 18: This is very weird to me.

ITR support's available much earlier than this. We're suddenly bumping up the minimum version requirement just so we can remove the app key from the docs?

I guess we can say "these docs only apply from version X and later" but mayve we should keep the older docs around and say that the requirement for the app key was removed in version X.

romainkomorndatadog: If we're going to do things this way, then maybe what we should actually be doing is documenting the tracer configuration in each tracer's versioned docs (eg: https://ddtrace.readthedocs.io/en/v1.18.0/integrations.html#pytest for `dd-trace-py==1.18.0`).

content/en/continuous_integration/intelligent_test_runner/javascript.md
juan-fernandez: good catch!

content/en/continuous_integration/intelligent_test_runner/javascript.md
juan-fernandez: this is something that bugs me too and we have no good solution for it. If we wanted to do it properly we'd have to have a huge matrix of versions and features. 

I think that the doc site is set up to promote people to bump to the highest possible version without breaking their setup. I think yours is a fair point though

romainkomorndatadog: I guess it's not a big deal because it's not a breaking change (in the sense that someone who still has an app key won't find themselves in a broken state). The only risk is someone with an older version if a tracer trying out ITR and running into a missing app key error, at which point I think the discoverability of what's actually going on isn't great, but maybe that's not something to really worry about.

content/en/continuous_integration/intelligent_test_runner/python.md
```diff
@@ -17,7 +17,7 @@ further_reading:
 Intelligent Test Runner is only supported in the following versions and testing frameworks:
 
 * `pytest>=6.8.0`
-  * From `ddtrace>=1.18.0`.
+  * From `ddtrace>=2.0.0`.
```
romainkomorndatadog, line 20:
```suggestion
  * From `ddtrace>=2.0.1`.
```
I think. :(

content/en/continuous_integration/intelligent_test_runner/python.md
```diff
@@ -17,7 +17,7 @@ further_reading:
 Intelligent Test Runner is only supported in the following versions and testing frameworks:
 
 * `pytest>=6.8.0`
-  * From `ddtrace>=1.18.0`.
+  * From `ddtrace>=2.0.1`.
```
romainkomorndatadog, line 20:
```suggestion
  * From `ddtrace>=2.1.0`.
```

romainkomorndatadog (approved):
App key was removed in `dd-trace-py` version `2.1.0` so this PR is correct.

romainkomorndatadog (approved):
LGTM

juan-fernandez:
there has been a bad merge, sorry! I'll fix it (not so many teams should've been pinged)

estherk15 (approved):
Looks good!

PR #19953, opened 2023-09-28
Update exclusion.md

Faysal1357:
https://docs.datadoghq.com/dashboards/functions/exclusion/#cutoff  In addition, the functions do not delete datapoints from Datadog entirely; it only removes them from your visualization. Disabling the function brings your datapoints back.

The above statement, specially "remove them from your visualization" might be misunderstood by customer. Some customers assumed that these functions removes the groups from their widget but in reality it still shows the groups that are above/below the threshold but the values is shown as "-"

### Merge instructions

### DIFF
--- a/content/en/dashboards/functions/exclusion.md
+++ b/content/en/dashboards/functions/exclusion.md
@@ -22,24 +22,16 @@ When you graph this metric as a timeseries, you would have 3 x 4 = 12 lines on y
 | `clamp_min()` | Set any metric values _under_ a threshold value to equal that value. | `clamp_min(avg:system.load.1{*}, 100)` |
 | `clamp_max()` | Set any metric values _over_ a threshold value to equal that value.  | `clamp_max(avg:system.load.1{*}, 100)` |
 
-The `clamp_min()` and `clamp_max()` functions take one parameter:
-
--   `THRESHOLD`: the threshold value you've specified.
-    -   `clamp_min()` sets all datapoints below the threshold to equal that value, while `clamp_max()` limits datapoints above the threshold.
+Add a threshold value. The `clamp_min()` sets all datapoints below the threshold to equal that value, while `clamp_max()` limits datapoints above the threshold.
 
 ## Cutoff
 
 | Function       | Description                                     | Example                                 |
 | -------------- | ----------------------------------------------- | --------------------------------------- |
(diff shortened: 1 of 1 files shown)


content/en/dashboards/functions/exclusion.md
```diff
@@ -41,7 +41,7 @@ The `cutoff_min()` and `cutoff_max()` functions take one parameter:
 
 The cutoff functions do not remove values that are equal to the threshold value.
 
-In addition, the functions do not delete datapoints from Datadog entirely; it only removes them from your visualization. Disabling the function brings your datapoints back.
+In addition, the functions do not delete datapoints from Datadog entirely; it only removes them from your visualization. Disabling the function brings your datapoints back. Since, both cutoff_min() and cutoff_max() are functions, they apply after the raw_query has returned the result. In the case of a grouped query, this means that they can and will change an entire timeseries to NaN (-) but are unable to remove a group as the raw_query has already returned it.
```
lesam, line 44: This is an odd thing to say in our docs I think - dashboard functions never actually change data, only the view of the data - so it's odd to me to specifically call out this one function (we don't, for example, call out that `top()` specifically doesn't delete data, or that `default_zero` doesn't actually add zeroes to the underlying data store).

Also, is `raw_query` a word our users understand here?

lesam: > are unable to remove a group as the raw_query has already returned it.

I don't think this is right either - cutoff_min and cutoff_max work on individual points, so it would be strange if they removed a group.

Other functions (like top) do remove groups, because they explicitly are about keeping/removing groups as a whole (and don't touch individual points)

lesam: i.e. it's not a technical limitation that we don't remove a group, it's because the definition of these functions is 'sets points below a cutoff equal to Nan', which doesn't say anything about removing groups, only points.

lesam: This is still a strange place to document this - it's a bit like a restaurant documenting 'in addition, our chefs are careful never to spit in the pasta-based dishes'.

It's true that these functions don't change the customer's data, but _no functions_ in F&F change customer data, so it's weird (to me) to document specifically that these two don't. But if you feel it's necessary for clarity it's fine to say this, it is correct.

lesam: If I was a datadog customer and came across this in the documentation, I would suddenly be very worried that there is some other dashboard function that _does_ delete data or mutate it somehow, which is not true.

PR #19963, opened 2023-09-28
add wrapper to tables in imported libraries

fitzage:

Certain types of imported content had tables that were too wide for the page, and the nature of the generated markdown meant that they didn’t get wrapped in the div necessary to make the tables side-scroll: https://datadoghq.atlassian.net/browse/WEB-4102

It seemed appropriate to me to try to resolve this at build time instead of using front-end scripting, so I used regex on the appropriate template to add in the wrapping div.

https://docs-staging.datadoghq.com/fitzage/table-fix/tracing/trace_collection/dd_libraries/ruby/#grpc

### Merge instructions

### DIFF
--- a/layouts/multi-code-lang/single.html
+++ b/layouts/multi-code-lang/single.html
@@ -16,9 +16,9 @@
             {{ partial "breadcrumbs.html" . }}
         </div>
     </div>
-    
+
     {{ partial "translate_status_banner/translate_status_banner.html" . }}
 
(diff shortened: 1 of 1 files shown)


devindford (approved):
Checked multiple pages with tables, horizontal scroll only appears when the table is wider than the container.

brett0000FF (approved):
Awesome, thank you for this!

PR #19966, opened 2023-09-28
Add host-level tag details to KSM Core doc

LawrenceLin690:

Add the details for host-tag inheritance for the KSM metrics. As the cluster centric metrics like `kubernetes_state.deployment.replicas_available` don't have anything to do with a host, so they aren't tagged by a host, so they don't inherit host-level tags. 

Can add `DD_TAGS` to add tags globally.

Additionally re-ordered the global and features config just for consistency with other operator docs. Technically too App Key isn't needed here so removed for simplicity. 

### Merge instructions


JacksonDavenport:
From containers side definitely good context to add. Ideally there would be the tabs for Helm and Operator showing the config samples for this like:

Helm
```yaml
datadog:
  tags: 
    - "<TagName>:<TagValue>"
```    
Operator
```yaml
kind: DatadogAgent
apiVersion: datadoghq.com/v2alpha1
metadata:
  name: datadog
spec:
  features:
    kubeStateMetricsCore:
      enabled: true
  global:
    tags:
      - "<TagName>:<TagValue>"
``` 

I think it would also make sense to include an extra header now on line 153 like `### Legacy check`. 

As previously this section was broken up with the table above. But now with this section here it just sort of blends into it awkwardly.

LawrenceLin690:
Thanks @JacksonDavenport made those changes recommended

PR #19984, opened 2023-09-29
Leave a note about the httpd selinux restrictions

bwoebi:
Traces might be missing, crashes might happen, appsec may not work properly, ... unless that line is added to SELinux config.
### Merge instructions


aliciascott (approved):
Hey @bwoebi , this looks good, I just had to fix the code block within the callout as standard code blocks won't work in callouts; we have to either just use plain ticks or the `lang` param, elected to just use ticks here to be consistent with the format in the other callouts on the page.

PR #20017, opened 2023-10-02
[intr-24] Add Email Change by Support Audit Event

tylerreinecke:

This PR adds documentation for a new event carried out by Support users via SupportDog. Specifically, it is for when Datadog Support changes a user's email.

**Note:** in the code itself (`pluto` backend) this event is referred to as 'email change'. In this PR I chose to use the term update instead of change, as I think update implies that the action was taken due to customer request, whereas `email change` sounds like Support just made the change (with or without a formal request).

### Merge instructions

I am writing this PR ahead of time, we still have some internal review that needs to be done, so please do not merge this PR immediately without confirmation, or I am happy to merge it myself after approval :)


maycmlee (approved):
LGTM!

PR #20019, opened 2023-10-02
AWS lambda: clarifying source type per transifex feedback

aliciascott:
This PR adds the word identifiers to the source type column in the table here:
https://docs.datadoghq.com/serverless/aws_lambda/configuration/?tab=datadogcli#rename-upstream-service-names

This came from feedback from transifex where it was causing confusion in our Japanese translated pages.

From the SME:
```
General is meant to be “General Identifiers” and Specific is meant to be “Specific Identifiers”. For instance if the call always goes through the "xyz" gateway before reaching the downstream service, this would be better to use instead of the lambda_api_gateway key
```

### Merge instructions


alai97 (approved):
Looks good!

PR #20020, opened 2023-10-02
apm: add proxies to data security

majorgreys:
### Merge instructions


alai97 (approved):

dgoffredo:
The proxy tables look correct to me and @cgilmour.

In looking at this, we noticed that we should probably omit the query parameters from the "http.url" tag currently added by nginx. We will probably make this change and update the docs accordingly. For now, these docs look good to us.

PR #20048, opened 2023-10-04
Fix ignoring apm resources example for kubernetes

dsem:

The correct syntax for ignoring multiple apm resources for kubernetes is to use a string of comma-separated values. The current example, an array, is an invalid definition for Kubernetes environment variable values; values must be a string.

### Merge instructions


brett0000FF:
@dsem - Thanks for submitting this fix! We are looking into this.

brett0000FF (approved):
Looks good. Thanks again for submitting this @dsem !

PR #20051, opened 2023-10-04
import all sub folders of `rulesets` - [WEB-4099]

StefonSimmons:

modifies `pull-and-push-folder` script to pull in all sub folders of `datadog-static-analyzer-rule-docs/rulesets`

https://datadoghq.atlassian.net/jira/software/projects/WEB/boards/102?selectedIssue=WEB-4099
### Merge instructions

- [x] Please merge after CorpWeb reviews

### Additional notes

nothing should be changed on this page or the rule detail pages.
preview -> https://docs-staging.datadoghq.com/stefon.simmons/modify-pull-push-folder-action/continuous_integration/static_analysis/rules/#overview


hestonhoffman (approved):

davidejones (changes requested):
A couple of things come to mind here that might be issues, can you look into them?

- if another rule in pullconfig has "rulesets" string or in the future we add it then its behavior might be different than expected?
- **/* usually means any number of subdirectories, so what happens if we have `rulesets/foo/bar/baz.md` will it preserve the correct paths too?

local/bin/py/build/actions/pull_and_push_folder.py
```diff
@@ -80,13 +86,10 @@ def pull_and_push_folder(content, content_dir):
             if file_name.endswith("README.md"):
                 file_name = "_index.md"
         # Writing the new content to the documentation file
-        dirp = "{}{}".format(
-            content_dir,
-            content["options"]["dest_dir"][1:],
-        )
-        makedirs(dirp, exist_ok=True)
+        dirp = (Path(content_dir) / Path(content["options"]["dest_dir"].lstrip('/')) / directory).resolve()
+        dirp.mkdir(exist_ok=True)
```
StefonSimmons, line 90: @davidejones  if we use `mkdir` instead of `makedirs` wouldnt the script throw a `FileNotFoundError` when trying to create possible nested directories under, for instance, `static_analysis/rules`: eg: `static_analysis/rules/python-best-practices/some-other-dir/some-markdown.md`?

davidejones: @StefonSimmons So it should work because mkdir on a pathlib object is the same as doing os.makedirs() 

Here is an example showing the equivalents https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module

davidejones (approved):

PR #20059, opened 2023-10-04
[DOCS-6328] clarifying lower helm versions

aliciascott:
This PR addresses hotjar feedback where the reader pointed out that the `--name` flag was deprecated on Helm `v3`, the docs stated "for Helm < v3.. use.." (Helm version _less than v3_) the use of the greater than; less than signs can cause confusion, so added clarification that if the reader is using Helm `v2.x` they need to use the `--name` flag. 

Helm `v1.x` is deprecated so no need to say "lower than v3".

### Merge instructions

### DIFF
--- a/content/en/containers/kubernetes/installation.md
+++ b/content/en/containers/kubernetes/installation.md
@@ -158,7 +158,7 @@ Using the Datadog Operator requires the following prerequisites:
 - `<TARGET_SYSTEM>`: The name of your OS. For example, `linux` or `windows`.
 
 
-**Note**: If you are using Helm v2.x, run the following:
+**Note**: If you are using Helm `2.x`, run the following:
    ```bash
    helm install --name <RELEASE_NAME> \
     -f datadog-values.yaml \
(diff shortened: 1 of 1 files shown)


content/en/containers/kubernetes/installation.md
```diff
@@ -158,7 +158,7 @@ Using the Datadog Operator requires the following prerequisites:
 - `<TARGET_SYSTEM>`: The name of your OS. For example, `linux` or `windows`.
 
 
-**Note**: If you are using Helm < v3, run the following:
+**Note**: If you are using Helm v2.x, run the following:
```
jhgilbert, line 161: I feel like we usually use `2.x` with no v for a version number, but I leave it up to you, since it was already like that and I'm not entirely sure. :)

jhgilbert (approved):
Approved with more of a question than a suggestion, handle it as you like!

content/en/containers/kubernetes/installation.md
aliciascott: You are right, good catch! I like that better too.

PR #20078, opened 2023-10-05
(DOCS-6330) Adjust some Korean strings

hestonhoffman:
I've adjusted these in Transifex as well, but the file is pretty far from completion, so fixing them here as well in the short term.

### Merge instructions


aliciascott (approved):
looks good

PR #20091, opened 2023-10-06
Ruby profiler: Remove `google-protobuf` as needed dependency

ivoanjo:

Starting with `ddtrace` 1.15.0, the Ruby profiler no longer needs the `google-protobuf` dependency. Thus, I've updated the docs to reflect that.

I did keep a reference to it, assuming that some customers may already have an older `ddtrace` and may want to add profiling to it without upgrading -- feedback welcome on the wording there (or if we don't usually do this kind of reference, we could drop it as well -- the library does have a good error message saying that `google-protobuf` is needed, so customers would spot it quickly).

### Merge instructions

### Additional notes

The 1.15 review is planned to go out on monday, so I've not checked the "merge after reviewing" box.

### DIFF
--- a/content/en/profiler/enabling/ruby.md
+++ b/content/en/profiler/enabling/ruby.md
@@ -46,12 +46,13 @@ To begin profiling applications:
 
 1. If you are already using Datadog, upgrade your agent to version [7.20.2][2]+ or [6.20.2][3]+.
 
-2. Add the `ddtrace` and `google-protobuf` gems to your `Gemfile` or `gems.rb` file:
+2. Add the `ddtrace` gem to your `Gemfile` or `gems.rb` file:
 
     ```ruby
-    gem 'ddtrace', '~> 1.0'
-    gem 'google-protobuf', '~> 3.0'
+    gem 'ddtrace', '~> 1.15'
     ```
+
+    (If you're running a version of `ddtrace` older than 1.15.0, you'll also need to add the `google-protobuf` gem as a dependency.)
 
 2. Install the gems with `bundle install`.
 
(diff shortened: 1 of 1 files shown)


content/en/profiler/enabling/ruby.md
```diff
@@ -46,13 +46,14 @@ To begin profiling applications:
...
-2. Add the `ddtrace` and `google-protobuf` gems to your `Gemfile` or `gems.rb` file:
+2. Add the `ddtrace` gem to your `Gemfile` or `gems.rb` file:
 
     ```ruby
-    gem 'ddtrace', '~> 1.0'
-    gem 'google-protobuf', '~> 3.0'
+    gem 'ddtrace', '~> 1.15'
     ```
 
+    (If you're running a version of `ddtrace` older than 1.15.0, you'll also need to add the `google-protobuf` gem as a dependency.)
```
hestonhoffman, line 55:
```suggestion
    If you're running a version of `ddtrace` older than 1.15.0, add the `google-protobuf` gem (version ~> 3) as a dependency.
```
Rephrased slightly and adding the version.

hestonhoffman (approved):
LGTM - I rephrased slightly and added the version number

content/en/profiler/enabling/ruby.md
ivoanjo: (Sounds good -- I've tweaked it to say `~> 3.0` instead of `~> 3`)

ivoanjo:
Hey @hestonhoffman, version 1.15.0 has now been released, we're good to merge this one :)

PR #20144, opened 2023-10-12
DOCS-4841 Profiler features for language and tracer versions

kayayarai:
DOCS-4841, DOCS-5273

### Merge instructions


content/en/profiler/enabling/supported_versions.md
```diff
@@ -0,0 +1,48 @@
...
+- **Minimum versions** are the minimum required to get access to a particular feature-If you have an earlier version, that feature is not supported. 
+- **Recommended versions** are the runtime and tracer versions that give you access to **all** the supported features. Generally speaking, it's best if you update to the latest version of all tracers.
+
+More details about runtime support and profile types is available on each language's profiler setup page (click the language heading in the table to access).
+
+|                                   | [Java][1]  | [Python][2]  |  [Go][3]   |  [Ruby][4] |   [Node.js][5]  |  [.NET][6]   |   [PHP][7]  | [Rust/C/C++][8] |
+|-----------------------------------|:-------:|:-------:|:------------:|:------:|:---------:|:-------:|:------:|:----------:|
+| Operating systems                 | | | | Linux only | | | CentOS&nbsp;7+ | Linux&nbsp;v4.17+ |
+| Minimum&nbsp;runtime&nbsp;version | JDK&nbsp;8+ | Python&nbsp;2.7+ | Go&nbsp;1.12+ | Ruby&nbsp;2.3+ | Node&nbsp;14+ | .NET&nbsp;Core&nbsp;2.1+, .NET&nbsp;5+, .NET&nbsp;Framework&nbsp;4.6.1+ | PHP&nbsp;7.1+ |  |
```
nsrip-dd, line 20: How do we decide what to put in this row? I'm not sure if this is the case for other profiler client libraries, but for Go we have tended to drop support for older runtime versions as time goes on. So, for a sufficiently old tracer library version, Go 1.12 might work, but it won't work for any Go profiler client library version released in at least the last ~2 years. And the Go maintainers themselves only support Go 1.20 and Go 1.21.

content/en/profiler/enabling/supported_versions.md
```diff
@@ -0,0 +1,48 @@
...
+| {{< ci-details title="I/O" >}}The time each method spent reading from and writing to files and sockets.{{< /ci-details >}}   | {{< X >}} |       |       |       |       |       |       |       |
+| {{< ci-details title="Exceptions" >}}The number of exceptions raised, including those caught.{{< /ci-details >}}   | {{< X >}} | Python 3.7+ |       |       |       | .NET 5+<br>tracer&nbsp;2.31.0 |  beta<br>tracer&nbsp;0.92.0  |       |
+| **Other features:** |
+| {{< ci-details title="Code Hotspots" >}}Find specific lines of code related to performance issues.{{< /ci-details >}}   | {{< X >}} | {{< X >}} | {{< X >}} | {{< X >}} | beta | {{< X >}} | tracer&nbsp;0.71.0 |      |
+| {{< ci-details title="Endpoint Profiling" >}}Identify endpoints that are bottlenecks or responsible for heavy resource consumption.{{< /ci-details >}}   | {{< X >}} | {{< X >}} | {{< X >}} | {{< X >}} | beta | tracer&nbsp;2.15.0 | tracer&nbsp;0.79.0 |      |
```
nsrip-dd, line 34: For Go, Code Hotspots and Endpoint Profiling were added in client library version 1.37.0

nsrip-dd (commented):
Thanks! I'm one of the Go profiler maintainers so I left comments for the bits pertaining to Go

content/en/profiler/enabling/supported_versions.md
brett0000FF: @nsrip-dd  Got it. Good catch. For minimum versions, we should generally list the oldest version that aligns with the support policy. If that's Go 1.20, I think we should list that as the minimum version.

I also realized that the recommended runtime version is 1.12+ for some reason. I suppose that should be 1.21?

content/en/profiler/enabling/supported_versions.md
```diff
@@ -0,0 +1,48 @@
...
+
+More details about runtime support and profile types is available on each language's profiler setup page (click the language heading in the table to access).
+
+|                                   | [Java][1]  | [Python][2]  |  [Go][3]   |  [Ruby][4] |   [Node.js][5]  |  [.NET][6]   |   [PHP][7]  | [Rust/C/C++][8] |
+|-----------------------------------|:-------:|:-------:|:------------:|:------:|:---------:|:-------:|:------:|:----------:|
+| Operating systems                 | | | | Linux only | | | CentOS&nbsp;7+ | Linux&nbsp;v4.17+ |
+| Minimum&nbsp;runtime&nbsp;version | JDK&nbsp;8+ | Python&nbsp;2.7+ | Go&nbsp;1.20+ | Ruby&nbsp;2.3+ | Node&nbsp;14+ | .NET&nbsp;Core&nbsp;2.1+, .NET&nbsp;5+, .NET&nbsp;Framework&nbsp;4.6.1+ | PHP&nbsp;7.1+ |  |
+| Recommended runtime version | JDK&nbsp;11+ | Python&nbsp;3.6+ | Go&nbsp;1.21+ | Ruby&nbsp;2.3+ | Node&nbsp;14+ | .NET&nbsp;6+ | PHP&nbsp;7.1+ | |
+| Minimum tracer version   | | 0.35.0 | 1.23.0 | 0.48.0 | 0.23.0 | 2.7.0| | |
+| Recommended tracer version |    | 2.3.2  | 1.57.0 | 1.15.0 | 2.12.0 | 2.31.0  | 0.92.0 | |
```
realFlowControl, line 23: Isn't the recommended version always the latest?
I think we did not have this version number anywhere in the docs, and this would mean to update this section with every release from now one.

realFlowControl (commented):
This is just my personal opinion, but I would rather like to have the features and versions and stuff at the language page and not merged together. IDK if we had any customers complaining about this, but I personally would like to see this grouped by language (as a PHP customer I am not really interested in what other languages can do, but I'd like a quick overview of the PHP profiler features)

content/en/profiler/enabling/supported_versions.md
```diff
@@ -0,0 +1,48 @@
...
+
+The following table shows a summary of the features available for each language runtime. 
+- **Minimum versions** are the minimum required to get access to a particular feature-If you have an earlier version, that feature is not supported. 
+- **Recommended versions** are the runtime and tracer versions that give you access to **all** the supported features. Generally speaking, it's best if you update to the latest version of all tracers.
+
+More details about runtime support and profile types is available on each language's profiler setup page (click the language heading in the table to access).
+
+|                                   | [Java][1]  | [Python][2]  |  [Go][3]   |  [Ruby][4] |   [Node.js][5]  |  [.NET][6]   |   [PHP][7]  | [Rust/C/C++][8] |
+|-----------------------------------|:-------:|:-------:|:------------:|:------:|:---------:|:-------:|:------:|:----------:|
+| Operating systems                 | | | | Linux only | | | CentOS&nbsp;7+ | Linux&nbsp;v4.17+ |
```
realFlowControl, line 19: PHP supports other Linux versions as well, I do think this is not the right way of representing this information

content/en/profiler/enabling/supported_versions.md
```diff
@@ -0,0 +1,48 @@
...
+| Minimum&nbsp;runtime&nbsp;version | JDK&nbsp;8+ | Python&nbsp;2.7+ | Go&nbsp;1.20+ | Ruby&nbsp;2.3+ | Node&nbsp;14+ | .NET&nbsp;Core&nbsp;2.1+, .NET&nbsp;5+, .NET&nbsp;Framework&nbsp;4.6.1+ | PHP&nbsp;7.1+ |  |
+| Recommended runtime version | JDK&nbsp;11+ | Python&nbsp;3.6+ | Go&nbsp;1.21+ | Ruby&nbsp;2.3+ | Node&nbsp;14+ | .NET&nbsp;6+ | PHP&nbsp;7.1+ | |
```
nsavoire, line 21: For Node, recommended runtime version would be 18+ (there is some extra latency between profiling periods on node < 18).

jbachorik:
Regarding Java - my main issue here is that the version matrix for Java is quite complicated and simplifications like this matrix can confuse our users.
I can give a few examples:
### Java version
While it is true that the minimum version is JDK 8, there are three flavours we are supporting with various level of functionality:
- OpenJDK 8u262 and newer - all listed features
- OpenJ9 8 - a form of CPU profiling only (not true CPU profiling)
- Oracle JDK 8 - a form of CPU profiling with the possibility to opt in for allocation profiling

For JDK 11 and 17 we have the minimum update number since which we can support code hostpost/endpoing profiling/timeline OOTB.

### Operating system
In this case we support all features only on Linux and we have some requirements on GLIBC version. For other OSes the feature set is limited but consistent across those OSes.

---

I wonder - would it be possible to have a discoverable link to a detailed Java feature matrix for each item such that a user can get the precise information about the required versions for a particular feature? Otherwise, I am afraid, this will generate a lot of frustration and support cases :(

content/en/profiler/enabling/supported_versions.md
AlexJF: +1 on this being incomplete/over-simplified. Missing things like supported architectures and GNU libc vs MUSL libc issues for profilers with native code that will require compilation.

AlexJF (commented):
I appreciate some of the advantage of the table view, and can see how it may be useful especially for multi-language customers.

I don't think it is is realistic or practical to move information from the language-specific pages into it or have these pages link to it as part of the normal flow. Instead it should serve as a standalone page or entrypoint into the language-specific ones with all the required level of detail.

content/en/profiler/enabling/supported_versions.md
brett0000FF: I went ahead and just linked to the releases page in the repo so readers can quickly find the latest version without us needing to maintain this.

content/en/profiler/enabling/supported_versions.md
brett0000FF: Thanks for the feedback. I went ahead and removed the OS row for now.

brett0000FF:
> Regarding Java - my main issue here is that the version matrix for Java is quite complicated and simplifications like this matrix can confuse our users. I can give a few examples:

@jbachorik - I definitely see where you are coming from. I tried to simplify the table more and added links back to the Java content specifically. Still thinking this through. I also removed the OS row for now.

brett0000FF:
> * I'm not sure this is a good idea to remove the matrix from .NET because a .NET customer will look at the .NET documentation.
> * the first lines of the shared matrix do not really make sense because the requirements are different PER profile type and cannot be "merged" on a single line
> * for recommended tracer version: always the latest one so the line should be removed (like the other ones before Profile types:)

@chrisnas - Appreciate the feedback! I put the matrix back on the .NET page. I also cleaned this up some more, including links to the latest tracer version and so on. I'm hoping to chat with @gturbat about our holistic approach for this soon too.

content/en/profiler/enabling/php.md
```diff
@@ -20,7 +20,7 @@ aliases:
 
 ## Requirements
 
-The Datadog Profiler requires at least PHP 7.1, on 64-bit Linux.
```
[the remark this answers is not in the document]
brett0000FF: Good catch. I meant to add this back.

content/en/profiler/enabling/go.md
```diff
@@ -22,7 +22,7 @@ The profiler is shipped within Datadog tracing libraries. If you are already usi
 
 ## Requirements
 
-The Datadog Profiler requires Go 1.12+.
+For a summary of the minimum and recommended runtime and tracer versions across all languages, read [Supported Language and Tracer Versions][15].
 
 For [Code Hotspots][2] and [Endpoint Profiling][3], use Go version 1.18+ and `dd-trace-go` version 1.37.0+.
```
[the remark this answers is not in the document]
brett0000FF: Confirmed with the team that the minimum version should actually be Go 1.19.

brett0000FF (approved):

content/en/profiler/enabling/supported_versions.md
```diff
@@ -0,0 +1,76 @@
...
+
+|                                   |  [Java][1]   |   [Python][2]    |    [Go][3]    |   [Ruby][4]    | [Node.js][5]  |  [.NET][6]  |   [PHP][7]    | [Rust/C/C++][8] |
+|-----------------------------------|:------------:|:----------------:|:-------------:|:--------------:|:-------------:|:-----------------------------------------------------------------------:|:-------------:|:---------------:|
+| <strong>Minimum&nbsp;runtime&nbsp;version</strong> | [JDK&nbsp;8+][17]  | Python&nbsp;2.7+ | Go&nbsp;1.19+ | Ruby&nbsp;2.3+ | Node&nbsp;14+ | .NET&nbsp;Core&nbsp;2.1+, .NET&nbsp;5+, .NET&nbsp;Framework&nbsp;4.6.1+ | PHP&nbsp;7.1+ |                 |
+| <strong>Recommended runtime version</strong>       | [JDK&nbsp;11+][17] | Python&nbsp;3.6+ | Go&nbsp;1.21+ | Ruby&nbsp;2.3+ | Node&nbsp;18+ |                              .NET&nbsp;6+                               | PHP&nbsp;8.0+ |                 |
+| <strong>Minimum tracer version</strong>            |              |      0.35.0      |    1.23.0     |     0.48.0     |    0.23.0     |                                  2.7.0                                  | 0.71.0        |                 |
```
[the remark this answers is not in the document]
brett0000FF: Noted--I appreciate this feedback. I'll bring this up with Guillaume and see if we should adjust how we can better frame this concept.

The intention is that the minimum version is still a viable option. If the minimum versions we are recommending are essentially unviable, I think we should bump them to a newer version.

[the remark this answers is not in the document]
ivoanjo: Thanks for looking into it. I can see how saying that a release from 6-12 months ago may be a bit more palatable, although it's something can easily get out-of-date... let's see what Guillaume thinks ;)

PR #20179, opened 2023-10-13
[CIAPP-8034] Update ITR conf image

calvarjorge:

The image shows tracked files above the excluded branches, while in-app the opposite is true.

### Merge instructions


andrea-mosk (approved):
The changes look good, but are you sure we need to modify non-english content as well? Usually that is automatically generated (probably the documentation team can confirm though)

brett0000FF (approved):
Looks good, thanks! @andrea-mosk is correct--we don't need to make updates to the files for other languages. It won't hurt anything, it's just that our Transifex service will override the updates.

PR #20186, opened 2023-10-13
Add some better installation notes for OP in K8s

silversupreme:
This PR adds two important things to the installation of OPW in Kubernetes environments:

* In EKS environments, the EBS driver is also required for buffers, as well as a `StorageClass` that is compatible with the OPW charts. I've included links to the necessary resources here.
* In addition, for remote configuration installs, ports must be manually specified in the Helm charts. Otherwise, no Load Balancer will be created at all.

### Merge instructions


brett0000FF (approved):
Looks great!

PR #20267, opened 2023-10-19
[DOCS-6354] Rename duplicate anchor to fix page scroll

brett0000FF:

- There are two "Configure Docker injection" headings on this page, which prevents us from linking to the sections correctly.
- The links function, but the page doesn't scroll to the heading.
- As a workaround, we can change the anchor so that each section has a unique anchor id.

### Merge instructions

### Additional notes

- I already verified that we only link to this heading in the "Agent and app in separate containers" tabs, so it should be ok to change the one in "Agent on host, app in containers".


aliciascott (approved):
Great catch! looks good

PR #20303, opened 2023-10-23
Spans that match a retention filter but are dropped can still be retained by the Datadog intelligent filter

cecile75:
Change
"Note: The order of the retention filter list changes indexing behavior. If a span matches a retention filter early in the list, the span is either kept or dropped. Any matching retention filter lower on the list does not catch the already-processed span."
to
"Note: The order of the retention filter list changes indexing behavior. If a span matches a retention filter early in the list, the span is either kept or dropped. Any matching custom retention filter lower on the list does not catch the already-processed span."
so it’s a little clearer that the spans that matched a retention filter but are dropped can still be retained by the Datadog intelligent filter.

### Merge instructions


dussault-antoine (approved):
thanks for the change

hestonhoffman (approved):

PR #20313, opened 2023-10-23
[workflow-automation] internal and hidden actions should not appear in action catalog

dd-gplassard:
### Merge instructions

### DIFF
--- a/local/bin/py/build/actions/workflows.py
+++ b/local/bin/py/build/actions/workflows.py
@@ -37,7 +37,7 @@ def workflows(content, content_dir):
                         data = json.loads(f.read())
                     except:
                         logger.warn(f"Error parsing {file_name}")
-            if data and data.get('stability', '') == 'stable':
+            if data and data.get('stability', '') == 'stable' and not data.get('internal') and not data.get('hidden'):
                 p = Path(file_name)
                 for action_name, action_data in data.get('actions', {}).items():
                     action_stability = action_data.get('stability')


natashadada (approved):

natashadada:
Should we do a check for `deprecated` as well?

buraizu (approved):

local/bin/py/build/actions/workflows.py
```diff
@@ -37,7 +37,7 @@ def workflows(content, content_dir):
                         data = json.loads(f.read())
                     except:
                         logger.warn(f"Error parsing {file_name}")
-            if data and data.get('stability', '') == 'stable':
+            if data and data.get('stability', '') == 'stable' and not data.get('internal') and not data.get('hidden') and not data.get('deprecated'):
```
hestonhoffman, line 40: Is this needed? Currently it's set up to only publish if it finds a `stability` of `stable`. For child actions (line 44), it's set up to only publish child actions that do not list a `stability` or are set to `stable`. My understanding is that any `stability` outside of `stable`/null is covered by those rules.

hestonhoffman: @natashadada explained this one to me. Thanks Natasha!

hestonhoffman, line 40: Based on Slack convos, I think we need to move this logic down to line 44 to check the child actions

dd-gplassard: 🤦 you're absolutely right, thanks for the catch !

local/bin/py/build/actions/workflows.py
```diff
@@ -41,7 +41,7 @@ def workflows(content, content_dir):
                 p = Path(file_name)
                 for action_name, action_data in data.get('actions', {}).items():
                     action_stability = action_data.get('stability')
-                    if not action_stability or action_stability == 'stable':
+                    if not action_stability or action_stability == 'stable' and not data.get('internal') and not data.get('hidden') and not data.get('deprecated'):
```
StefonSimmons, line 44: @dd-gplassard in your in-progress docs, you define a `visibility`  field instead of `internal`. should this be accounted for as well?

dd-gplassard: Yes sorry about that, actually in our code we use a `@visible false` annotation but it gets translated into a `"hidden": true` in our manifest :/
So the check should be on hidden

Code annotation : https://github.com/DataDog/dd-source/blob/5f528e423c17084c4e9e1eec907f8b40a9959901/domains/workflow/actionplatform/apps/wf-actions-worker/src/runner/bundles/com.datadoghq.internal/actions.ts#L34
Manifest field : https://github.com/DataDog/dd-source/blob/5f528e423c17084c4e9e1eec907f8b40a9959901/domains/workflow/actionplatform/apps/wf-actions-worker/src/runner/bundles/com.datadoghq.internal/manifest.json#L222

And on a side note we might be able to remove this concept so maybe a cleanup PR will follow shortly :)

hestonhoffman (approved):

StefonSimmons (approved):
looks good. based on a slack convo, the new boolean fields (deprecated, hidden, and internal  will only be applied at the action child level and not for the action parent/bundle level.

e.g. an action can have `stability: stable` and `internal: true`. in this case the child action would not show.

new documentation regarding these new fields is in progress

PR #20351, opened 2023-10-25
Add estimated usage metric for Observability Pipelines to the docs

jszwedko:
Adding the estimated usage metric for Observability Pipelines to our usage documentation. I think this will need to be added to the translated pages. Who would take care of that?

### Merge instructions


buraizu (approved):

buraizu:
Hi @jszwedko, the translated docs are managed through a 3rd party service and there's no need to loop anyone in, although there may be some delay before the update is published in the translated docs. The docs team has ways to make that happen faster in some cases, so feel free to reach out internally if that might be needed for this update.

jszwedko:
Thanks @buraizu !

PR #20363, opened 2023-10-25
Remove unneeded checkout on review status job

nkzou:

The approved status github action takes much longer than expected (1min - 2min instead of 10-20s) because it's checking out the entire documentation repo when it shouldn't need to. This removes that step.

### Merge instructions

### Additional notes

I took a look at the associated [github action](https://github.com/DataDog/github-actions/blob/master/src/post-review-status.ts) to make sure the checkout wasn't needed but would appreciate a sanity check on this.


davidejones (approved):
Thanks! I can't see any reason for needing the checkout either, from what i can tell this should be good to go.

PR #20393, opened 2023-10-27
Clarify the beta feature is only for GRACE-supported resources not le…

sophiah123:
…gacy granular access

I want to be more clear that monitors & synthetics tests aren’t supported just yet for this "beta gain access feature" though they will be in a few qtrs once they migrate from the legacy granular access control system to GRACE. It’s only available on dashboards, notebooks, security rules, and SLOs (see the table in the docs above the section I edited. If all 3 types of restrictions are supported, then it's on the new GRACE system. If it only supports role restrictions, then it's on the legacy granular access control system which this feature doesn't touch). 

### Merge instructions

### DIFF
--- a/content/en/account_management/rbac/granular_access.md
+++ b/content/en/account_management/rbac/granular_access.md
@@ -27,7 +27,7 @@ Use the different principals to control access patterns in your organization and
 The feature to gain access to individual resources is in private beta. To request access, <a href="/help">contact Support</a>.
 </div>
 
-A user with the `user_access_manage` permission can gain edit access to any individual resource that has restricted access. To get access, click the **Gain Edit Access** button in the granular access control modal.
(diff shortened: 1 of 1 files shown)


content/en/account_management/rbac/granular_access.md
```diff
@@ -27,7 +27,7 @@ Use the different principals to control access patterns in your organization and
 The feature to gain access to individual resources is in private beta. To request access, <a href="/help">contact Support</a>.
 </div>
 
-A user with the `user_access_manage` permission can gain edit access to any individual resource that has restricted access. To get access, click the **Gain Edit Access** button in the granular access control modal.
+A user with the `user_access_manage` permission can gain edit access to any individual resource that supports team, role, and user / service account-based restrictions. Resources with only role-based access restrictions are not supported. To get access, click the **Gain Edit Access** button in the granular access control modal.
```
cswatt, line 30:
```suggestion
A user with the `user_access_manage` permission can gain edit access to any individual resource that supports restrictions based on team, role, and user or service account. Resources with only role-based access restrictions are not supported. To get access, click the **Gain Edit Access** button in the granular access control modal.
```

cswatt (approved):

PR #20442, opened 2023-11-02
set artifact compression flags

davidejones:

Sets some build flags so that the artifacts uploaded at the end of a job are completed faster
Completing in roughly 30s versus the current 2-3mins

### Merge instructions

### Additional notes

We've had the same flags on corp for almost 2yrs without any negative impact


StefonSimmons (approved):
seeing results of faster compression time for artifacts in this pipeline (24sec `Uploading artifacts for successful job` from ~2.5mins in previous builds)


info on new variables: https://docs.gitlab.com/ee/ci/runners/configure_runners.html#artifact-and-cache-settings

PR #20469, opened 2023-11-03
set default algolia rank for all integrations

davidejones:

Because of hugo cascade consistency issues we cannot set general frontmatter for *.md in a section and then also target specific files as it seems it can differ build to build.

So instead we are looking for a default_algolia_rank param and using that when there is no rank set.

This should be temporary lets create an issue in hugo to try to resolve this inconsistency

### Merge instructions


local/bin/py/build/configurations/pull_config_preview.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 - config:
-    cache_enabled: true
+    cache_enabled: false
```
davidejones, line 3: apply suggestion before merging

```suggestion
    cache_enabled: true
```

local/bin/py/build/configurations/pull_config.yaml
```diff
@@ -192,6 +192,9 @@
               dependencies: ["https://github.com/DataDog/heroku-buildpack-datadog/blob/master/README.md"]
               aliases:
               - /developers/faq/how-do-i-collect-metrics-from-heroku-with-datadog
+              algolia:
+                tags: ["heroku", "heroku buildpack"]
+                rank: 1
```
apigirl, line 197:
```suggestion
                rank: 80
```

jhgilbert:

PR #20474, opened 2023-11-05
Update advanced_log_collection.md

jcayetano:
Includes an example for Linux and Windows exclusion paths.
Added a second note to highlight the path are case sensitive.

Clarify use cases for customers using Log exclusion for both Linux and Windows.

### Merge instructions


maycmlee (approved):
Thanks!

PR #20503, opened 2023-11-07
Add files via upload

CyrusBelsoi:

Update public docs to fix discrepancies in images vs text

### Merge instructions


estherk15 (commented):
Hi @CyrusBelsoi, does the metric make a difference? What if we changed the example to match the image rather than change the image?

estherk15 (approved):

PR #20540, opened 2023-11-08
hugo micro perf updates

davidejones:

This PR improves performance of the hugo build step

- Cache logo lookups 18s down to 10s cumulative
- left nav api lookups 15s down to <1s cumulative
- defer algolia json creation on preview
- using fileexists over getpage on security_rules/single saves 5s cumulative
- Cache header main nav from websites modules. On docs the top main nav doesn't change page to page.
- Sets `disableSVG: true` for hugo minification of html. This is due to the fact it was modifying svg values in the html unnecessarily this in turn causing us to deploy html files that hadn't really changed.

### Merge instructions

#### What difference is it making?

Showing the hugo build at around 1 min instead of 3-4 mins

#### What to test?

- Check logos appear normal on https://docs-staging.datadoghq.com/david.jones/micro-perf/integrations/
- Check left nav on api appears normal https://docs-staging.datadoghq.com/david.jones/micro-perf/api/latest/
- Spot check some security rule detailed pages link to the integration e.g `Set up the 1password integration.`  on https://docs-staging.datadoghq.com/david.jones/micro-perf/security/default_rules/1password-item-exfiltration-by-user/ works
- Check the header nav (the one thats shared with corpsite) appears normal on various pages

example build
https://gitlab.ddbuild.io/DataDog/documentation/-/jobs/367648287

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -224,7 +224,20 @@ algolia_sync_preview:manual:
   allow_failure: true
   interruptible: true
   timeout: 2h
+  cache:
+    - *hugo_cache
+    - *yarn_cache
+    - *pip_cache
+  dependencies:
+    - build_preview
   script:
+    # build the site to get algolia.json ondemand
+    - touch Makefile.config preview_error.log
+    - make dependencies
+    - make config
+    - yarn run prebuild
+    - build_site
+    # end build
     - yarn add atomic-algolia@https://s3.amazonaws.com/origin-static-assets/corp-node-packages/master/atomic-algolia-v0.3.29.tgz
     - >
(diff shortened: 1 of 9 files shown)


.gitlab-ci.yml
```diff
@@ -224,7 +224,20 @@ algolia_sync_preview:manual:
...
   interruptible: true
   timeout: 2h
+  cache:
+    - *hugo_cache
+    - *yarn_cache
+    - *pip_cache
+  dependencies:
+    - build_preview
   script:
+    # build the site to get algolia.json ondemand
```
bgdeutsch, line 234: @davidejones could we have `algolia.json` output as an artifact from the `build_preview` job and somehow use that here instead of building the site again?

bgdeutsch (approved):

PR #20601, opened 2023-11-13
Move Azure setup instructions to be a subpage of SCIM

urseberry:
Datadog now supports SCIM with Okta. When there was only 1 identity provider (Azure), it made sense for all SCIM info to be on one page. Now that there are two, this PR puts Azure on its own page so it's easier to add Okta next.

No new text, just menu updates and moving text onto the new page. I did also add a Further reading section to the main SCIM page.

### Merge instructions


brett0000FF (approved):
Looks good!

PR #20638, opened 2023-11-15
CSM resource evaluation filters

drichards-87:

Moves the existing resource evaluation filter docs from the CSM setup docs to a standalone article. This will make it easier to maintain long-term as right now they're currently duplicated across three different shortcodes.

### Merge instructions


config/_default/menus/menus.en.yaml
```diff
@@ -3293,6 +3293,11 @@ main:
     parent: cspm
     identifier: resource_catalog
     weight: 114
+  - name: Resource Evaluation Filters
```
urseberry, line 3296: What's the reason for including this guide in the left nav? I assume it's important or commonly used? Or the "Misconfigurations" navigation item is used to collect common misconfigurations that might appear in different parts of the docs?

content/en/security/cloud_security_management/guide/resource_evaluation_filters.md
```diff
@@ -0,0 +1,78 @@
...
+    tag: "Documentation"
+    text: Setting Up Cloud Security Management
+---
+
+You can use resource tags to create filters that include or exclude resources from being evaluated by Cloud Security Management (CSM). The filters must be specified as a comma-separated list of `key:value` pairs.
+
+**Notes**:
+
+- Resource evaluation filters can only be used with hosts that are scanned by cloud integrations.
+- Tags must be applied directly to the resource. User tags added in Datadog are not taken into account by the filters. The only exception is for tags added on the integration tiles for AWS and Google Cloud Platform.
```
urseberry, line 18:
```suggestion
- Tags must be applied directly to the resource. The filters do not take into account user tags added in Datadog. The only exception is for tags added on the integration tiles for AWS and Google Cloud Platform.
```

urseberry (approved):
Hi! Doc looks good. Left a couple of comments, but they are not blockers. Approving in case I forget to come back to this later.

PR #20644, opened 2023-11-15
[APM][.NET] Socument support for .NET 8

pierotibou:

.NET has been released yesterday and we support it officialy since a week. Let the doc reflect that.

### Merge instructions

- [ X] Please merge after reviewing

### DIFF
--- a/content/en/tracing/trace_collection/compatibility/dotnet-core.md
+++ b/content/en/tracing/trace_collection/compatibility/dotnet-core.md
@@ -31,6 +31,7 @@ The .NET Tracer supports automatic instrumentation on the following .NET Core ve
 
 | Version              | Microsoft End of Life | Support level        | Package version      |
 | -------------------- | --------------------- | -------------------- | -------------------- |
+| .NET 8               |                       | [GA](#support-ga)    | latest (>= 2.41.0)   |
 | .NET 7               |                       | [GA](#support-ga)    | latest (>= 2.20.0)   |
 | .NET 6               |                       | [GA](#support-ga)    | latest (>= 2.0.0)    |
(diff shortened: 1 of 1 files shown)


zacharycmontoya (approved):

content/en/tracing/trace_collection/compatibility/dotnet-core.md
```diff
@@ -31,6 +31,7 @@ The .NET Tracer supports automatic instrumentation on the following .NET Core ve
 
 | Version              | Microsoft End of Life | Support level        | Package version      |
 | -------------------- | --------------------- | -------------------- | -------------------- |
+| .NET 8               |                       | [GA](#support-ga)    | latest (>= 2.41.0)   |
```
andrewlock, line 34: _Next_ release I would say, as that's when we're actually running on GA bits 🙂 

```suggestion
| .NET 8               |                       | [GA](#support-ga)    | latest (>= 2.42.0)   |
```

hestonhoffman: We probably want to hold off on publishing this until that version releases if that's the case

hestonhoffman (commented):
Left a comment above, but it didn't review as the docs team for some reason

content/en/tracing/trace_collection/compatibility/dotnet-core.md
pierotibou: So we've discussed this internally. We support .NET 8 with 2.41 even though we were testing on the RC build only and we haven't added new features. That said nothing should prevent customers to use 2.41, and this will prevent people to ask the support teams.
We will update to 2.42 when we release to be even safer but we can merge as is.

PR #20731, opened 2023-11-21
DOCS-6573 Split out Service Catalog first page

kayayarai:
Service Catalog index page was getting heavy. Split out two major sections into their own pages. The actual text did not change, just updated link references and added a couple screen shots to the browsing page.

### Merge instructions


amlakhanpal:
This looks great and definitely a _huge_ improvement from what we have right now. It still feels a bit text heavy in places (which is naturally going to happen), so maybe adding screenshots on the sub-pages will help. Let me know if I can support on that at all! Thank you, Kari!!

estherk15 (approved):
Looks good!

PR #20734, opened 2023-11-21
Updating the documentation for dd-dotnet

kevingosse:

Update the diagnostic doc to mention the new dd-dotnet tool. Update the expected outputs.

### DIFF
--- a/content/en/tracing/troubleshooting/dotnet_diagnostic_tool.md
+++ b/content/en/tracing/troubleshooting/dotnet_diagnostic_tool.md
@@ -5,7 +5,15 @@ kind: documentation
 
 If after installing the .NET tracer your application does not produce traces as expected, run the diagnostic tool for basic troubleshooting. It can help you figure out issues with your setup such as missing environment variables, incomplete installation, or unreachable Agent.
 
-Install the tool one of the following ways:
+The diagnostic tool is bundled with the tracer since version 2.42.0. It's located in the tracer folder, and automatically added to the PATH to be invoked from anywhere.
+
+## Installing dd-trace
+
+**This sections is for versions of the tracer older than 2.42.0.**
+
+Older versions of the tracer aren't shipped with dd-dotnet. Instead, you can install the dd-trace tool. The features and syntax are similar to dd-dotnet.
+
+You can install dd-trace in one of the following ways:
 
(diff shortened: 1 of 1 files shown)


content/en/tracing/troubleshooting/dotnet_diagnostic_tool.md
```diff
@@ -48,22 +58,51 @@ No issue found with the target process.
...
 
 Running checks on process 35888
-Target process is running with .NET Framework
-Profiler is not loaded into the process
-Tracer is not loaded into the process
-The environment variable COR_ENABLE_PROFILING should be set to '1' (current value: '0')
-```
+Process name: SimpleApp
+
+---- STARTING TRACER SETUP CHECKS -----
```
andrewlock, line 66: Where are the "issues"? 🤔

content/en/tracing/troubleshooting/dotnet_diagnostic_tool.md
```diff
@@ -48,22 +58,51 @@ No issue found with the target process.
 
 Example output with issues:
 ```bash
-$ dd-trace check process 35888
+$ dd-dotnet check process 35888
```
andrewlock, line 61: I think you need to update the previous example too

andrewlock (commented):
LGTM, except I think you missed an update/updated one with the wrong grab

link04 (changes requested):
It looks like the example session the process check is using the older No Issues output instead of the new one and the output with issues has no issues, for the IIS output they both  got no issues. Could we add an example where they haven't set a key for it to show up on the `---- STARTING TRACER SETUP CHECKS -----` section?

kevingosse:
> It looks like the example session the process check is using the older No Issues output instead of the new one and the output with issues has no issues, for the IIS output they both got no issues. Could we add an example where they haven't set a key for it to show up on the `---- STARTING TRACER SETUP CHECKS -----` section?

My bad, fixed in https://github.com/DataDog/documentation/pull/20734/commits/ca9e467639d64b2f0ff5d43ff9b0128382f31bc4

andrewlock (approved):
LGTM thanks

link04 (approved):

urseberry (approved):

PR #20752, opened 2023-11-22
Update nginx tracing module with new naming convention

dmehala:

This PR updates the nginx installation script to support the new naming convention we recently introduce. As one of our customer mentionned in https://github.com/DataDog/nginx-datadog/issues/52, the current installation script is broken.

### Merge instructions

### Additional notes

Even though it's not about this PR, having a 'copy to clipboard' button for code snippets would be really helpful.


dgoffredo (approved):
Thanks, that'll work.

urseberry:
We do have a special code partial that provides a copy button and some other nice features, but it doesn't work within tabs, which this page has. It's documented internally here: to the right of the data-type drop-down menu / to the left of the metric name drop-down menu

urseberry (approved):

PR #20834, opened 2023-11-29
Add note about java prepared statements

jmeunier28:

Adds a note to our APM-DBM link java tracer docs about prepared statement support for trace propagation. This is a response to some customers filing support tickets after enabling `full` mode, but not seeing any traces in the DBM UI for their samples. 

### Merge instructions

### DIFF
--- a/content/en/database_monitoring/connect_dbm_and_apm.md
+++ b/content/en/database_monitoring/connect_dbm_and_apm.md
@@ -1,7 +1,7 @@
 ---
 title: Connect Database Monitoring and Traces
 kind: documentation
-aliases: 
+aliases:
 - /database_monitoring/guide/connect_dbm_and_apm/
 further_reading:
   - link: 'https://www.datadoghq.com/blog/link-dbm-and-apm/'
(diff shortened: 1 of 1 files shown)


content/en/database_monitoring/connect_dbm_and_apm.md
```diff
@@ -181,6 +181,8 @@ public class Application {
 }
 ```
 
+**Note:** Prepared statements are not supported in `full` mode, and all JDBC API calls that use prepared statements will automatically be downgraded to `service` mode. Since the vast majority of java SQL libraries use prepared statements by default, this means that **most** java applications will only be able to take advantage of `service` mode.
```
buraizu, line 184:
```suggestion
**Note**: Prepared statements are not supported in `full` mode, and all JDBC API calls that use prepared statements are automatically downgraded to `service` mode. Since most Java SQL libraries use prepared statements by default, this means that **most** Java applications are only able to use `service` mode.
```

buraizu (approved):

PR #20915, opened 2023-12-05
Fitzage/og image defaults

fitzage:

1. Adds 5 default og images
2. When an og image isn’t specified using another method, assign one of these 5 images
    1. Random-ish
    2. Shouldn’t change every build because it’s based on the markdown file id
3. If all else fails (really seems to only be on 404 pages where it’s not necessary anyway), fall back to the fifth default image.

https://datadoghq.atlassian.net/browse/WEB-4329

Preview: https://docs-staging.datadoghq.com/fitzage/og-image-defaults/

### Merge instructions


DavidWeid (approved):
@fitzage Docs has the same params files with `meta_image: datadog_logo_share_tt.png` in case you need to change those

buraizu (approved):
LGTM

PR #20983, opened 2023-12-12
istio tracing: remove table of version correspondence

dgoffredo:
A recent correspondence with a customer over an issue with our Istio tracing integration brought to light that our documentation is outdated.

We have a project planned to rewrite our Istio tracing documentation entirely, but for now I think we can remove unnecessary out-of-date information.

This revision removes a table that related Istio's release version to the corresponding Datadog C++ tracing library release version. That correspondence is no longer relevant, the table was out of date, and the context about environment variables included a broken link.

Our support policy for Istio is "when a new Istio release comes out, we support it." I think it's best to leave that implicit, and avoid customers inferring it from a table about library versions.


cswatt (approved):

cswatt:
I'm going to go ahead and merge this, as out-of-date information isn't of any use.

PR #21029, opened 2023-12-14
Remove private beta designation for AP1 PrivateLink

bobmshannon:

PrivateLink for AP1 is now generally available. As such, we can remove the banner indicating that it is in private beta.

### Merge instructions


brett0000FF (approved):
Thanks!

PR #18592, opened 2023-06-16
Update ddsketch_trace_metrics.md

crunchypistachiobutter:
removing outdated content from ~2 years back

### Additional Notes

---

### Reviewer checklist
- [ ] Review the changed files.
- [ ] Review the URLs listed in the [Preview](#preview) section.
- [ ] Check images for PII
- [ ] Review any mentions of "Contact Datadog support" for internal support documentation.

PR #21075, opened 2023-12-18
Remove incorrect information and sync with English doc

Johanna-Seif:

This removes the section in the French doc saying that we do not support ddsketch metrics for European customer. This is out of dates, the metrics are available. Copied the exact modificaiton of the English version:
* https://github.com/DataDog/documentation/pull/18592

### Merge instructions


jhgilbert:
Hi Johanna, after checking with the team, they've confirmed that the French docs will automatically be synced with the English version at the next available opportunity -- the changes you've made here would unfortunately get overwritten quickly, since this repo is only the source of truth for English.

I went into our translation tool and marked this file as urgent, so it should be translated promptly. Thank you!

PR #21096, opened 2023-12-19
Delete unnecessary strings in Japanese documentation of Browser RUM

takeshiemoto:

I have found an unnecessary string in the Japanese document and would like to remove it.

### Merge instructions


brett0000FF (approved):
Thanks for the catch and for contributing to our documentation! I applied this fix via our translation service too so this won't get overridden.

PR #21170, opened 2023-12-27
Fix configmap name in autodiscovery confirmap example in integration.md

Rbillon59:

Configmap name from the `volumes` part of the pod definition does not match the example configmap name in the `configmap` definition. Taking the example as is does not work.

Current example :

`Configmap` definition example : 

```yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: redis-config-map (note the confirmap name here)
```

`pod` definition update example :

```yaml
volumes:
      # [...]
        - name: redisdb-config-map
          configMap:
            name: redisdb-config-map (note the configmap name here)
```

### Merge instructions

### DIFF
--- a/content/en/containers/kubernetes/integrations.md
+++ b/content/en/containers/kubernetes/integrations.md
@@ -454,7 +454,7 @@ The following ConfigMap defines the integration template for `redis` containers:
 kind: ConfigMap
 apiVersion: v1
 metadata:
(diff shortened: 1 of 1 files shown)


apigirl (approved):

apigirl:
hey @Rbillon59, thank you so much for taking the time to make this fix

PR #21176, opened 2023-12-27
Update ODBC documentation in dbm-sqlserver-agent-config-examples.md

qian-steven:
This fixes the following issues in our DSN/ODBC documentation:
- User/Password are not valid fields for `odbc.ini`, the DD agent doesn't pull credentials from here, and it is generally not best practice to store credentials here. These fields were removed from the example.
- The Server and Port fields are unnecessary. The same information is required to be set in the `conf.yaml` file, so having it in `odbc.ini` is redundant. These fields were also removed from the example.
- There is now explicit clarification that the `driver` field in `conf.yaml` pulls from `odbcinst.ini` with an additional example.

### Merge instructions


apigirl:
Added [DOCS-6997](https://datadoghq.atlassian.net/browse/DOCS-6997) to validate

lu-zhengda (approved):

hestonhoffman (approved):

PR #21183, opened 2023-12-28
MNTS-90344 Clarify Downtime V1 output

pierreln-dd:

This PR clarifies that the Downtimes V1 api will only return V1 downtimes (and currently active V2 downtimes)

### Merge instructions


pierreln-dd:
I just saw that V1 will be marked as Deprecated, so I'm closing the PR, as this is redundant

PR #21213, opened 2024-01-03
remove 'csp-report.' subdomain for CSP report intake URLs

BenoitZugmeyer:

* the `csp-report.` subdomain hasn't been setup when the AP1 region was created, so the doc is mentioning an inexistent domain `csp-report.browser-intake-ap1-datadoghq.com`

* starting from the latest major release, the Browser SDK doesn't use intake subdomain anymore (`logs.`, `rum.` and `replay.`)

  * this allows a simpler setup when deploying new regions
  * also it allows a simpler, more secure proxy setup, as all requests can be forwarded to a single domain

Given those two information let's remove the `csp-report.` subdomain as well. New regions will be easier to setup, and it will follow the same pattern as other intake usages.

### Merge instructions

### DIFF
--- a/content/en/integrations/content_security_policy_logs.md
+++ b/content/en/integrations/content_security_policy_logs.md
@@ -52,7 +52,7 @@ Before you add a directive to a CSP header, [generate a client token in your Dat
 You need a URL where browsers can send policy violation reports. The URL must have the following format:
 
 ```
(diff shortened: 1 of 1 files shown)


alai97 (approved):

bcaudan (approved):

BenoitZugmeyer:
After some tests, I confirm that using the apex domains `browser-intake-datadoghq.com` (for US1) and `browser-intake-ap1-datadoghq.com` (for AP1) works as expected.

This PR is ready to be merged.

PR #21297, opened 2024-01-09
Document Ruby memory-related profile types

ivoanjo:

The upcoming dd-trace-rb 1.19.0 release introduces three new profile types for Ruby. These are similar to other profilers, and I've mostly borrowed their description for these profile types, with minimal tweaks.

### Merge instructions

Ideally, this PR should only be merged after we release 1.19.0, which we plan to do this week. I'll drop a note on this PR once the release is out so it can be merged :)

### Additional notes

N/A

### DIFF
--- a/content/en/profiler/profile_types.md
+++ b/content/en/profiler/profile_types.md
@@ -160,17 +160,18 @@ Wall Time
 
 Allocations (alpha, v1.19.0+)
 : The number of objects allocated by each method during the profiling period (default: 60s), including allocations which were subsequently freed. This is useful for investigating garbage collection load.<br />
-_Requires: Ruby 2.7+_
+_Requires: Ruby 2.7+_ and [needs to be enabled][2]
 
 Heap Live Objects (alpha, v1.19.0+)
 : The number of objects allocated by each method in heap memory that have not yet been garbage collected. This is useful for investigating the overall memory usage of your service and identifying potential memory leaks.<br />
-_Requires: Ruby 2.7+_
+_Requires: Ruby 2.7+_ and [needs to be enabled][2]
 
 Heap Live Size (alpha, v1.19.0+)
 : The amount of heap memory allocated by each method that has not yet been garbage collected. This is useful for investigating the overall memory usage of your service and identifying potential memory leaks.<br />
-_Requires: Ruby 2.7+_
+_Requires: Ruby 2.7+_ and [needs to be enabled][2]
 
 [1]: /profiler/enabling/ruby/#requirements
+[2]: https://github.com/DataDog/dd-trace-rb/releases/tag/v1.19.0#:~:text=You%20can%20enable%20these%20features%3A
 {{< /programming-lang >}}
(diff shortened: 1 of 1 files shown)


content/en/profiler/profile_types.md
```diff
@@ -158,6 +158,18 @@ CPU
 Wall Time
 : The elapsed time used by each function. Elapsed time includes time when code is running on CPU, waiting for I/O, and anything else that happens while the function is running.
 
+Allocations (alpha, v1.19.0+)
+: The number of objects allocated by each method during the profiling period (default: 60s), including allocations which were subsequently freed. This is useful for investigating garbage collection load.<br />
+_Requires: Ruby 2.7+_
```
AlexJF, line 163: The framing before the list suggests these are on by default. Shall we mention here the env variables that are also required or something along the lines of `Requires: Ruby 2.7+ and [manual enablement](link to release notes)`

ivoanjo: That's a good point. One thing I spotted is that we seem to make the same mistake everywhere -- e.g. for the other languages we don't have enablement info and we make it seem like all features show up by default.

Which means I guess we get to be the first ones? ;) 

I slightly hesitate before linking to the release notes, though. I guess https://docs.datadoghq.com/profiler/enabling/ruby would be a nicer spot to link to?

AlexJF: Mentioning alpha features on what looks/works essentially as a quickstart page feels dangerous 😅 

Didn't we have an "Advanced" section somewhere in our docs before?

urseberry: Instructions for specific configurations might be appropriate for a new guide: https://docs.datadoghq.com/profiler/guide/

urseberry (approved):
This looks fine to me. Please re-request my review if you make significant changes and would like me to look at them.

content/en/profiler/profile_types.md
ivoanjo: Ok, I propose that for now we go with the quick link off to the release notes; and once we go beta we'll give it a second pass and add it to the docs.

I'll wait until the actual release is out to have the correct link/anchor to the release notes to add here.

AlexJF: > Instructions for specific configurations might be appropriate for a new guide: [docs.datadoghq.com/profiler/guide](https://docs.datadoghq.com/profiler/guide/)

Not really sure that would be a good approach. Based on the existing guide, it seems we're after a more learning-focused "tutorial" format rather than "here's how to turn it on". Left unchecked we could end up with so many random guides that it'd hide the real gems.

OTOH, it'd actually be great to have a more general "how to profile memory" tutorial with subsections for how to enable it on different languages. It's just that it'd be a lot of work and a major stepping stone in the way of just announcing a new alpha feature 😅

ivoanjo:
@urseberry I've added the links to the release notes, this should be good to go :)

urseberry (approved):

PR #21378, opened 2024-01-16
Clarify the usage of regular expressions in included/excluded job lists in DD Jenkins plugin

nikita-tkachenko-datadog:

In the settings of Datadog Jenkins plugin one can specify which jobs to include/exclude from monitoring.
The included/excluded lists can contain regular expressions, but not glob patterns.
The two have different syntax.
A user was trying to add a glob pattern instead of a regular expression.

### Merge instructions

### DIFF
--- a/content/en/continuous_integration/pipelines/jenkins.md
+++ b/content/en/continuous_integration/pipelines/jenkins.md
@@ -857,7 +857,7 @@ You can configure the Jenkins Plugin to include or exclude some pipelines:
 **Environment variable**: `DATADOG_JENKINS_PLUGIN_INCLUDED`<br/>
 **Example**: `susans-job,johns-.*,prod_folder/prod_release`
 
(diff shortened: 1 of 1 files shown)


content/en/continuous_integration/pipelines/jenkins.md
```diff
@@ -857,6 +857,8 @@ You can configure the Jenkins Plugin to include or exclude some pipelines:
 **Environment variable**: `DATADOG_JENKINS_PLUGIN_INCLUDED`<br/>
 **Example**: `susans-job,johns-.*,prod_folder/prod_release`
 
+Keep in mind that the lists of included and excluded jobs can contain regular expressions and not glob patterns (so in order to include a job with a specific prefix you need to use `prefix-.*` and not `prefix-*`).
```
cswatt, line 860:
```suggestion
Lists of included and excluded jobs can contain regular expressions, but not glob patterns. To include a job with a specific prefix, use `prefix-.*`—not `prefix-*`.
```

cswatt (approved):

PR #21394, opened 2024-01-17
[SCICM-1575] Add note about version tag uniqueness

rpelliard:
This clarifies the intended usage of the `version` tag in Unified Service Tagging. Currently, some users are setting `version` tags that never change, or that aren't bumped for every code change.

### Merge instructions


Letiste (approved):
Should we also mention the dependency on Unified Service Tagging in the SCI page?

samarcan (approved):

aliciascott (approved):
Looks good, let me know if you end up adding additional info based on @Letiste last comment and need additional review.

PR #21396, opened 2024-01-17
Add PHP troubleshooting

realFlowControl:

We had a customer reaching out to us that they would like to enable the profiler but they are concerned about a possible overhead and would like to activate the profiler in a minimal configuration/overhead mode. Then I saw that the troubleshooting page for PHP is slim, so I added some stuff.

### Merge instructions


morrisonlevi (approved):
On the PHP side of things, it looks good to me.

hestonhoffman (approved):

PR #21411, opened 2024-01-17
Address feedback for Language and Tracer Versions for Profiler

brett0000FF:

Updating new Profiler Compatibility table to address feedback. [Preview site with changes](https://docs-staging.datadoghq.com/brett0000FF/profiler-table-tweaks/profiler/enabling/supported_versions).

Updates:
- Remove minimum tracer versions because it's not useful.
- Replace _Recommended_ version with _Feature complete_ language to make it more clear that this version gives access to all features.
- Add line breaks after alpha and beta tags to improve readability.
- Streamline wording in definitions and table intro sentences.
- Add alert box emphasize that users can click language headings to view specific setup pages.
- Clarify that minimum runtime versions are the requirement if they are missing from the Profile types table.
- Fix Feature-complete runtime version to be .NET 7 instead of .NET 6.
- Remove trailing .0 on tracers versions; replace with `+` to improve readability.

### Merge instructions


hestonhoffman (approved):
LGTM!

content/en/profiler/enabling/supported_versions.md
```diff
@@ -9,46 +9,46 @@ further_reading:
...
-| <strong>Recommended tracer version</strong>        | [latest][9]  |   [latest][10]   | [latest][11]  |  [latest][12]  | [latest][13]  |                              [latest][14]                               | [latest][15]  |  [latest][16]   |
+| <strong>Feature-complete runtime version</strong>       | [JDK&nbsp;11+][17] | Python&nbsp;3.6+ | Go&nbsp;1.21+ | Ruby&nbsp;3.1+ | Node&nbsp;18+ |                              .NET&nbsp;7+                               | PHP&nbsp;8.0+ |                 |
+| <strong>Feature-complete tracer version</strong>        | [latest][9]  |   [latest][10]   | [latest][11]  |  [latest][12]  | [latest][13]  |                              [latest][14]                               | [latest][15]  |  [latest][16]   |
 
 ## Profile types
 
-To collect profile types for each language, use the minimum versions summarized in the following table:
+To collect profile types, use at least the minimum versions summarized in the following table. If a runtime isn't specified, the profile type requires the minimum runtime version in [Runtime and tracer versions](#runtime-and-tracer-versions).
```
ivoanjo, line 30: > use at least the minimum versions summarized

I'm... still concerned and don't think we should do this :(

We're stating that customers can use a 3-year-old Ruby library version and a 4-year-old Python version. And technically... they can. Probably. Nobody's really checking that such versions still work. And even if they do still "work", they'll provide an incredibly sub-par experience.

Do we really want to tell such customers in 2024+ to go ahead and use these versions?

content/en/profiler/enabling/supported_versions.md
```diff
@@ -9,46 +9,46 @@ further_reading:
 ---
```
ivoanjo, line 9: :arrow_up: Minor: Re: the page name, it's a bit weird using the term "tracer" here for the libraries, as in most other onboarding pages we use the term library or similar:

* https://docs.datadoghq.com/profiler/enabling/ => tracing libraries
* https://docs.datadoghq.com/tracing/trace_collection/custom_instrumentation/dd_libraries/ => datadog libraries
* https://docs.datadoghq.com/tracing/trace_collection/automatic_instrumentation/dd_libraries/ => tracing libraries

ivoanjo (commented):
Left a few notes!

content/en/profiler/enabling/supported_versions.md
brett0000FF: Noted, that's good feedback. I updated the title and some of the other language to be more consistent with "tracing libraries".

content/en/profiler/enabling/supported_versions.md
brett0000FF: Thanks for the feedback! I think it's less telling users to go ahead and use these versions, and it's more so the bare minimum that a user would need in order to use at least some profiling features.

This is for a user coming in and wanting to see at a glance if they are able to use profiling.

@chrisnas,  @gturbat , and I had a discussion this week, and after some back and forth, we agreed that there is value including this info. I'll let them chime in if they have more to say from their perspective too.

gturbat (approved):

PR #21433, opened 2024-01-18
Provide explanation for why instance endpoints are recommended for DBM

jhgilbert:
DOCS-7029

### Merge instructions


lu-zhengda (approved):

content/en/database_monitoring/setup_mysql/aurora.md
```diff
@@ -36,7 +36,7 @@ Performance impact
...
 
 Proxies, load balancers, and connection poolers
-: The Agent must connect directly to the host being monitored, preferably through the instance endpoint. The Agent should not connect to the database through a proxy, load balancer, connection pooler, or the **Aurora cluster endpoint**. While this can be an anti-pattern for client applications, each Agent must have knowledge of the underlying hostname and should stick to a single host for its lifetime, even in cases of failover. If the Datadog Agent connects to different hosts while it is running, the values of metrics will be incorrect.
+: The Agent must connect directly to the host being monitored, preferably through the instance endpoint. The Agent should not connect to the database through a proxy, load balancer, connection pooler, or the **Aurora cluster endpoint**. The Agent relies on the statistics table of each instance, so each Agent must have knowledge of the underlying hostname and should stick to a single host for its lifetime, even in cases of failover. If the Datadog Agent connects to different hosts while it is running, the values of metrics will be incorrect.
```
caseycull, line 39: thinking about it more I think "stats table" sounds a bit clearer here, this phrasing is what we use in customer conversations most often

caseycull (approved):

content/en/database_monitoring/setup_mysql/aurora.md
justiniso: I think this is good, but if we're revamping, might be a good time to improve the overall message? We wrote this a long time ago, but I think we can be better about describing the tradeoffs of connecting to the cluster endpoint.

The key takeaways:
* if you connect to the cluster endpoint, the Agent will connect to one _random_ replica and only provide visibility to that replica. Most of the time this is not what customers want
* if there is a failover event, the Agent will calculate the difference in statistics between two different hosts, which can produce wildly incorrect stats
* for these reasons, we recommend using the instance endpoint

content/en/database_monitoring/setup_mysql/aurora.md
```diff
@@ -36,7 +36,7 @@ Performance impact
 Database Monitoring runs as an integration on top of the base Agent ([see benchmarks][1]).
 
 Proxies, load balancers, and connection poolers
-: The Agent must connect directly to the host being monitored, preferably through the instance endpoint. The Agent should not connect to the database through a proxy, load balancer, connection pooler, or the **Aurora cluster endpoint**. While this can be an anti-pattern for client applications, each Agent must have knowledge of the underlying hostname and should stick to a single host for its lifetime, even in cases of failover. If the Datadog Agent connects to different hosts while it is running, the values of metrics will be incorrect.
```
jhgilbert, line 39: @justiniso Thanks for your feedback, it was very helpful to understand the exact reasoning here. I've updated the verbiage to just this file, any additional feedback before I make similar updates elsewhere?

justiniso (approved):
Nice, thanks for the update! Very clear 👍

urseberry (approved):
Thanks for taking the time to make sure the wording is super clear and consistent!

PR #21446, opened 2024-01-18
Update widget_colors.md

distantshape:
Clarifying that the Classic color palette does not guarantee adjacent colors will be distinct. There are rare conditions where adjacent colors are the same when using the Classic palette. We cannot avoid collisions in all cases. I received internal feedback that these lines were confusing because they see adjacency collisions. There is also general confusion around this.

### Merge instructions


alai97 (approved):
Looks good for docs!

PR #21467, opened 2024-01-22
fix: trace_id_128_bit_generation_enabled default value according to the actual code

sho-he:

Detault value of trace_id_128_bit_generation_enabled was different from the actual code. 
So it shoud be modified default value to prevent misunderstandings.

https://github.com/DataDog/dd-trace-rb/blob/3becb40d148b4ec16c346eec6378d3e6b9886671/lib/datadog/tracing/configuration/settings.rb#L199

### Merge instructions

### Additional notes

Need to fix [dd-trace-rb docs](https://github.com/DataDog/dd-trace-rb/blob/master/docs/GettingStarted.md)


sho-he:
@jhgilbert 
> these changes would get overwritten if we merged them, because our non-English docs are generated from another source
I think so, thank you for your reaction and considering to update document!

PR #21471, opened 2024-01-22
Fix image

edrevo:

Fix ITR example image.

### Merge instructions


jhgilbert:
Hi @edrevo, using the same name as the previous image can cause caching issues, so I created #21474 using the image you provided, if you would like to take a look. Thank you!

PR #21639, opened 2024-02-01
Add downtime example with boolean OR across two different groups

gmckeever:
This includes an additional example that shows how to perform a boolean or with 2 or more keys. OR is not needed in this scenario which may be confusing without an example.

Release notes on this feature: https://app.datadoghq.com/release-notes/tackle-expected-maintenance-periods-proactively-with-more-flexible-downtimes


estherk15 (approved):
Thanks for adding this!

PR #21650, opened 2024-02-02
trace-agent log level can not be updated from cli

ajgajg1134:
This implies this config change is possible outside dedicated containers which isn't accurate.

The log level _can_ be changed by remote config but NOT via the CLI currently.

### Merge instructions


aliciascott (approved):
Looks good!

PR #21654, opened 2024-02-02
Revert "Add downtime example with boolean OR across two different groups"

estherk15:
Reverts DataDog/documentation#21639

Found out that this is not supported, this example evaluates as an AND not an OR. 
`Top level ORs are not supported, for example, service:A OR host:X. This requires two separate Downtimes.`


aliciascott (approved):
looks good!

PR #21703, opened 2024-02-07
Fix broken link

codyborders:


drichards-87 (approved):
Thanks for finding/fixing this! FYI, if you include a forward slash in your branch name, it'll generate a staging preview of the docs site with your changes. Not really necessary for a small change such as this, but it can be useful for larger updates.

PR #21711, opened 2024-02-07
Update single-step-apm.md

zarirhamza:

Mention explicitly that applications running in the same namespace as the agent will not be instrumented by single step

### Merge instructions

### DIFF
--- a/content/en/tracing/trace_collection/automatic_instrumentation/single-step-apm.md
+++ b/content/en/tracing/trace_collection/automatic_instrumentation/single-step-apm.md
@@ -167,10 +167,12 @@ docker run -d --name dd-agent \
 
 You can enable APM by installing the Agent with the Datadog Helm chart. This deploys the Datadog Agent across all nodes in your Linux-based Kubernetes cluster with a DaemonSet.
 
+Note that single step instrumentation is deployed to the `default` namespace and will not instrument any applications in the same namespace. 
+
 ### Requirements
 
 - Make sure you have [Helm][13] installed.
-
+  
 ### Installation
 
 To enable single step instrumentation with Helm:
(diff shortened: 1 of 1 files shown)


content/en/tracing/trace_collection/automatic_instrumentation/single-step-apm.md
```diff
@@ -167,6 +167,8 @@ docker run -d --name dd-agent \
 
 You can enable APM by installing the Agent with the Datadog Helm chart. This deploys the Datadog Agent across all nodes in your Linux-based Kubernetes cluster with a DaemonSet.
 
+**Note:** Single step instrumentation is deployed to the `default` namespace and does not instrument any applications in the same namespace.
```
brett0000FF, line 170:
```suggestion
**Note:** Single step instrumentation is deployed to the `default` namespace, but this doesn't automatically instrument applications in this namespace. To specify which namespaces to instrument, read [enabling or disabling instrumentation for namespaces](#enabling-or-disabling-instrumentation-for-namespaces).
```
Seems like we should point people to the `enabledNamespaces` configuration. Do you agree with this? Thanks for the update!

zarirhamza: I think it takes away a bit from the point of the note which is to caution away from deploying applications to the default namespace. Even if you specify the namespace in the config file it will not instrument it if its in the same namespace. The point of specifying namespaces in the config is to minimize the range of instrumentations to a select few namespaces, but if nothing is done it will automatically instrument every non-default namespace. I think we should keep the note as is.

brett0000FF (approved):

PR #21717, opened 2024-02-07
[DBMON-3565] Add DBM integration overheads

lu-zhengda:
This PR adds agent overhead for DBM integrations for PostgreSQL, MySQL and SQL Server.

### Merge instructions


drichards-87:
Added DOCS-7253 for editorial review.

content/en/database_monitoring/setup_mysql/_index.md
```diff
@@ -23,4 +23,20 @@ For setup instructions, select your hosting type:
 
 <br>
 
-
+### Agent integration overhead
+
+Agent integration overhead tests were made on an Amazon EC2 machine `c5.xlarge` instance (4 VCPU/ 8GB RAM). The database used for the tests was a MySQL 8.0 instance running on an Amazon RDS `db.m5.large` instance (2 VCPU/ 8GB RAM). The database was running a TPC-C workload with 20 warehouses.
```
jmeunier28, line 28: Does this stuff deserve its own page? Also, can we formally recommend to customers to run the agent on `c5.xlarge` hardware if running in AWS?

lu-zhengda: > Does this stuff deserve its own page?
I thought about putting all integrations results into one page but after looking at how benchmark results are presented on other places, they are usually on the page of the product https://docs.datadoghq.com/agent/basic_agent_usage/?tab=httpuncompressed#agent-overhead? 

> can we formally recommend to customers to run the agent on c5.xlarge hardware if running in AWS?
i have mixed feeling about this - means we can only recommend for EC2 but not other setups.

jmeunier28: > i have mixed feeling about this - means we can only recommend for EC2 but not other setups.

True, but a on of our customers deploy agents this way so it will provide value for a lot of people. Other customers could also extrapolate how to provision their hosts in other clouds or on-prem simply by looking up what the CPU/memory allocation is for that instance type in the AWS docs

I don't feel super strongly, but I think people would appreciate this level of direction, what do you think?

nenadnoveljic: > can we formally recommend to customers to run the agent on c5.xlarge hardware if running in AWS?

Since the sizing depends on the number of configured databases, in lots of cases, this machine can be an overkill and incur unnecessary costs.

PR #21843, opened 2024-02-16
profiler: explain how to profile third-party Go services

nsrip-dd:

We have an Agent integration for collecting profiles from Go services
through a Go standard library debug HTTP endpoint. Document its
existence. Customers can use this integration to profile Go services for
which they don't control the source code and can't use our client
library. For example, Kubernetes cluster administrators can use this to
profile the Kubernetes API server. We still prefer that customers use
our client library if they can, to get access to the full range of
features.

### Merge instructions


nsrip-dd:
Not going to get around to finishing this.

PR #21896, opened 2024-02-22
Fix links

edrevo:
### What does this PR do? What is the motivation?
Fix broken links

### Merge instructions

- [x] Please merge after reviewing

PR #21902, opened 2024-02-22
alias cleanup, leading slash

davidejones:

In relation to #21896 we discovered we can get in a temporary state of 404 links between redirect passes that run.
This PR fixes current aliases without a leading slash.

### Merge instructions


aliciascott (approved):
thanks, looks good

aliciascott (approved):

PR #21975, opened 2024-02-28
Mention there's no OTLP in custom OTEL instrumentation

ikonst:
(Am I actually right here?)


buraizu:
@ikonst  Thank you for the PR, we're validating this internally and will follow up with any updates.

brett0000FF:
@ikonst - Thanks for the contribution! To answer your question, yes, your note is correct. With this method, there's no need for OTLP ingestion because the traces are sent to Datadog with the Agent automatically.

I'm going to close this PR, so our team can look into adding a similar note to a few other pages. Thanks!

PR #21989, opened 2024-02-29
Fix loading_time description

ypresto:

- Fix inverted description in "SPA Route Change": original Japanese text was "page HAS activity", fixed is "page has NO activity".
- Fix broken in-page link

### Merge instructions


drichards-87:
Hi @ypresto 👋. Thanks for letting us know about the issue. I've let our translators know so they can fix it.

PR #21992, opened 2024-02-29
Move call to set dd_service annotation

andrea-mosk:

Setting the service is a very important part of the setup for Argo CD applications monitoring in Datadog. Currently, it's in a separate section after the default setup. This PR moves it to be part of the main setup to ensure that customers know about it and try to set it when possible.

### Merge instructions

### DIFF
--- a/content/en/continuous_delivery/deployments/argocd.md
+++ b/content/en/continuous_delivery/deployments/argocd.md
@@ -111,34 +111,23 @@ metadata:
   annotations:
     notifications.argoproj.io/subscribe.cd-visibility-trigger.cd-visibility-webhook: ""
     dd_env: <YOUR_ENV>
+    dd_service: <YOUR_SERVICE>
 ```
 
-There are two annotations:
+There are three annotations:
 1. The notifications annotation subscribes the Argo CD application to the notification setup created above.
 2. The `dd_env` annotation configures the environment of the application. Replace `YOUR_ENV` above with the environment
    to which this application is deploying (for example: `staging` or `prod`). If you don't set this annotation,
    the environment defaults to `none`.
+3. The `dd_service` annotation configures the service of the application. Replace `YOUR_SERVICE` above with the service
+   that the Argo CD application is deploying (for example: `transaction-service`). When this annotation is used, the service
+   name is added to all the deployment executions generated from the application. Moreover, if your service is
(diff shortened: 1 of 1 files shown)


content/en/continuous_delivery/deployments/argocd.md
```diff
@@ -111,35 +111,24 @@ metadata:
...
+There are three annotations:
 1. The notifications annotation subscribes the Argo CD application to the notification setup created above.
 2. The `dd_env` annotation configures the environment of the application. Replace `YOUR_ENV` above with the environment
    to which this application is deploying (for example: `staging` or `prod`). If you don't set this annotation,
    the environment defaults to `none`.
+3. The `dd_service` annotation configures the service of the application. Replace `YOUR_SERVICE` above with the service
+   that the Argo CD application is deploying (for example: `transaction-service`). When this annotation is used, the service
+   name is added to all the deployment executions generated from the application. Moreover, if your service is
+   registered in [Service Catalog][13], the team name is also added to all the deployment executions. Omit this annotation
+   if your Argo CD application is configured to deploy more than one service.
```
alxch-, line 126: I think people will not read until then end except if we make it bold or at the start of that point

andrea-mosk: Mmm until which part? The part in which we say to omit it in some cases?

alxch-: yes, which is at the very end

andrea-mosk: Right, but if they have a multi-service Argo CD application, my guess is that they will either:

- Do not finish reading the paragraph as they do not have a single service being deployed. In that case they won't specify anything which is the correct behavior
- Read the paragraph and understand that they don't need to specify anything

I don't want to start the paragraph by having something like "Omit this if", because it might discourage them from adding the annotation IMO. In future, we will have support for multi-service application so what I want to do is to point them at an additional section in that case instead of saying "omit this", and in that case we could have it at the beginning. WDYT?

alxch-: Maybe something along the lines of
> If your ArgoCD application deploys a single service, replace `YOUR_SERVICE` above with the service...

Without using "omit" which I understand can be a blocker. I'm fine with this anyway as this is temporary.

alxch- (approved):

drichards-87 (approved):

PR #22042, opened 2024-03-04
Update integration-management with note about partner ints

eho1307:

Add a clarifying note that partner/community integrations need to be re-installed after the agent has been upgraded

### Merge instructions

### DIFF
--- a/content/en/agent/guide/integration-management.md
+++ b/content/en/agent/guide/integration-management.md
@@ -9,7 +9,7 @@ algolia:
 
 The Agent comes with a set of bundled official Datadog integrations to allow users to start monitoring their applications quickly. These integrations are available as single Python packages, and you can upgrade them separately.
 
-**Note:** Non-core integrations such as those built by partners or community members are not retained when the Agent is upgraded. Please be mindful that these integrations will need to be re-installed upon upgrading the Agent version.
+**Note:** Non-core integrations such as those built by partners or community members are not retained when the Agent is upgraded. Please be mindful that these integrations need to be re-installed upon upgrading the Agent version.
 
 For Agent v6.8+, the `datadog-agent integration` command allows users to manage the official Datadog integrations that are available for the Agent. It has the following subcommands:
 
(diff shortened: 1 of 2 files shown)


content/en/agent/guide/integration-management.md
```diff
@@ -9,6 +9,8 @@ algolia:
 
 The Agent comes with a set of bundled official Datadog integrations to allow users to start monitoring their applications quickly. These integrations are available as single Python packages, and you can upgrade them separately.
 
+**Note:** Non-core integrations such as those built by partners or community members are not retained when the Agent is upgraded. Please be mindful that these integrations need to be re-installed upon upgrading the Agent version.
```
cswatt, line 12:
```suggestion
**Note**: Community and Marketplace integrations are not retained when the Agent is upgraded. These integrations need to be re-installed upon upgrading the Agent version.
```

content/en/agent/troubleshooting/integrations.md
```diff
@@ -19,6 +19,8 @@ Datadog integrations are configured through the Datadog Agent using YAML configu
 
 If an integration that you've configured is not showing up in Datadog, run the [`status` CLI command][2] and look for the integration under the *Running Checks* heading.
 
+**Note:** Non-core integrations such as those built by partners or community members are not retained when the Agent is upgraded. Please be mindful that these integrations need to be re-installed upon upgrading the Agent version.
```
cswatt, line 22:
```suggestion
**Note**: Community and Marketplace integrations are not retained when the Agent is upgraded. These integrations need to be re-installed upon upgrading the Agent version.
```

cswatt (changes requested):
I think we should use "Community and Marketplace integrations" rather than "non-core integrations"—the word "core" might have meaning to us internally since the repo is called `integrations-core`, but users may not understand what it means.

Also removing use of "please" as per docs style guide.

content/en/agent/guide/integration-management.md
eho1307: @cswatt what do you think of **Note**: Community, Partner and Marketplace integrations are not retained when the Agent is upgraded. These integrations need to be re-installed upon upgrading the Agent version.

cswatt (approved):

PR #22123, opened 2024-03-08
set `site` param to selected region - [WEB-4546]

StefonSimmons:

set `site` url param with selected region whether the param already exists or not.

motivation: https://datadoghq.atlassian.net/browse/WEB-4546
preview: https://docs-staging.datadoghq.com/stefon.simmons/update-site-region-url-state/account_management/saml/
   - select a region from the region dropdown on the right.
   - the `site` param should be appended to the url and the `site` Cookie should be set to the selected region
   - the `site` param should not persist when navigating othes pages but the Cookie and dropdown should.
### Merge instructions

- [x] Please merge after CorpWeb review

### DIFF
--- a/assets/scripts/region-redirects.js
+++ b/assets/scripts/region-redirects.js
@@ -19,7 +19,6 @@ document.addEventListener('DOMContentLoaded', () => {
     hideTOCItems()
 
     if (regionSelector) {
-        regionOnChangeHandler(currentUserSavedRegion)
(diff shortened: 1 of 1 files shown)


aliciascott (approved):
good for docs

devindford (approved):
Working as expected, param in in url on region selection, but not when navigating away as intended. Also the region persists when in the URL in a private browser

PR #22158, opened 2024-03-12
fix opentelemetry docs

michael-mader:
The original paragraph mentions 3 scenarios but the image and the text above only mention two scenarios.

### Merge instructions


urseberry (approved):
Thank you very much!

PR #22239, opened 2024-03-18
[RUM-2702] Documenting the session-application-cookie relation

acorretti:

Based on troubleshooting experience we want to clarify this technical limitation in the docs.

### Merge instructions

### DIFF
--- a/content/en/real_user_monitoring/browser/troubleshooting.md
+++ b/content/en/real_user_monitoring/browser/troubleshooting.md
@@ -58,7 +58,7 @@ The RUM Browser SDK relies on cookies to store session information and follow a 
 
 ## Session IDs, Cookies and RUM Applications
 
-There is a one-to-one relation between a RUM session and the RUM application it belongs to. Therefore, the domain set for the `_dd_s` cookie holds a direct relation to the RUM application it is monitoring and cannot monitor any additional application.
+There is a one-to-one relation between a RUM session and the RUM application it belongs to. Therefore, the domain set for the `_dd_s` cookie is fully dedicated to the RUM application it is monitoring and cannot monitor any additional applications.
 
 ## Technical limitations
 
--- a/content/en/real_user_monitoring/browser/troubleshooting.md
+++ b/content/en/real_user_monitoring/browser/troubleshooting.md
@@ -56,7 +56,7 @@ The RUM Browser SDK relies on cookies to store session information and follow a 
 
 **Note**: The `_dd_l`, `_dd_r`, and `_dd` cookies have been replaced with `_dd_s` in recent versions of the RUM Browser SDK.
 
-## Session IDs, Cookies and RUM Applications
+## Session IDs, cookies and RUM applications
 
 There is a one-to-one relation between a RUM session and the RUM application it belongs to. Therefore, the domain set for the `_dd_s` cookie is fully dedicated to the RUM application it is monitoring and cannot monitor any additional applications.
 
--- a/content/en/real_user_monitoring/browser/troubleshooting.md
+++ b/content/en/real_user_monitoring/browser/troubleshooting.md
@@ -56,6 +56,10 @@ The RUM Browser SDK relies on cookies to store session information and follow a 
 
 **Note**: The `_dd_l`, `_dd_r`, and `_dd` cookies have been replaced with `_dd_s` in recent versions of the RUM Browser SDK.
 
+## Session IDs, Cookies and RUM Applications
+
+There is a one-to-one relation between a RUM session and the RUM application it belongs to. Therefore, the domain set for the `_dd_s` cookie holds a direct relation to the RUM application it is monitoring and cannot monitor any additional application.
+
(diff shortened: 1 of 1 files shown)


BenoitZugmeyer (approved):

amortemousque (approved):

content/en/real_user_monitoring/browser/troubleshooting.md
```diff
@@ -56,6 +56,10 @@ The RUM Browser SDK relies on cookies to store session information and follow a
 
 **Note**: The `_dd_l`, `_dd_r`, and `_dd` cookies have been replaced with `_dd_s` in recent versions of the RUM Browser SDK.
 
+## Session IDs, Cookies and RUM Applications
+
+There is a one-to-one relation between a RUM session and the RUM application it belongs to. Therefore, the domain set for the `_dd_s` cookie holds a direct relation to the RUM application it is monitoring and cannot monitor any additional application.
+
```
bcaudan, line 62: **💬 suggestion:** To be more precise, the one-to-one relationship is actually between the RUM session and the _dd_s cookie.
Given the cookie is configured on a specific domain (or top domain), the session is then tied to this domain.
The RUM data model expect that one session is linked to a single application, so we should advise to:

- not have multiple applications on the same domain, to avoid having the session data scattered across multiple applications
- not have an application with multiple domains, to avoid to have a single user journey split in multiple sessions

‏

acorretti: How would you suggest to phrase it? Feel free to edit.

To me, the session<>application 1:1 relation is implicit, because the RUM UI does handle crossing multiple applications correctly. We've had that in the past and it was very confusing.

bcaudan: > To me, the session<>application 1:1 relation is implicit, because the RUM UI does not handle crossing multiple applications correctly

As described in the two scenarios in my comment, there are cases outside this 1:1 relation that are technically possible and where the setup could make sense for customers.
IMO, clarifying those behaviours/limitations and our expectation could be helpful for customers.

acorretti: > there are cases outside this 1:1 relation that are technically possible and where the setup could make sense for customers.
> IMO, clarifying those behaviours/limitations and our expectation could be helpful for customers.

I respectfully disagree. While technically possible, the RUM UI has not been designed to handle those scenarios, therefore I believe we must clarify what the intended and supported designed is like.

We can of course discuss with @nhientrandatadog when she's back from PTO, but I don't feel comfortable taking your suggestion in this PR.

bcaudan: > I respectfully disagree [...] I don't feel comfortable taking your suggestion in this PR.

Not sure to understand which point is bothering you though 🤔

> While technically possible, the RUM UI has not been designed to handle those scenarios, therefore I believe we must clarify what the intended and supported design is like.

I agree, however I found the current wording misleading as it could easily be interpreted as the actual behaviour instead of the expected behaviour.
Given observed configurations that were outside of our expectations, listing unsupported use cases seems valuable to me.

content/en/real_user_monitoring/browser/troubleshooting.md
```diff
@@ -56,6 +56,10 @@ The RUM Browser SDK relies on cookies to store session information and follow a
 
 **Note**: The `_dd_l`, `_dd_r`, and `_dd` cookies have been replaced with `_dd_s` in recent versions of the RUM Browser SDK.
 
+## Session IDs, Cookies and RUM Applications
```
rtrieu, line 59:
```suggestion
## Session IDs, cookies and RUM applications
```

content/en/real_user_monitoring/browser/troubleshooting.md
acorretti: > Not sure to understand which point is bothering you though 🤔

Given the considerations, I disagree with "could be helpful for customers". I think we must clarify the intended and supported design.

> I agree, however I found the current wording misleading as it could easily be interpreted as the actual behaviour instead of the expected behaviour.
> Given observed configurations that were outside of our expectations, listing unsupported use cases seems valuable to me.

But that's the reason why I'm making this change: To clarify what is RUM's intended use case so that customers don't fall into unsupported use cases due to an unexpected (yet technically possible) setup.

If any customer reads our intended use case and thinks we should support an alternative one, that's interesting feedback for the PM.

rtrieu (approved):
Looks good, thanks!

content/en/real_user_monitoring/browser/troubleshooting.md
bcaudan: >  I think we must clarify the intended and supported design.

I still agree 🙂 
IMO, expliciting our intentions and pointing out the limitations of the implementation would help to clarify that.

It would have been nice to reach a common understanding but no big deal, this section could be revisited if needed.

PR #22244, opened 2024-03-18
Team handles can be used in montior notifcations

SimoneT-DD:
Datadog teams can now support notification channels. When a notification channel is set and the team handle is used in a monitor message the notification will be redirected to that notification channel for that team

### DIFF
--- a/content/en/monitors/notify/_index.md
+++ b/content/en/monitors/notify/_index.md
@@ -112,6 +112,10 @@ Disk space is low @ops-team@company.com
 
 {{% notifications-email %}}
 
+#### Teams
+
+If a [notification channel][17] is set, you can route notifications to a specific Team. Monitor alerts targeting @team-handle will be redirected to the selected communication channel.
+
 #### Integrations
 
 {{% notifications-integrations %}}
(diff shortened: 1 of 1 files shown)


content/en/monitors/notify/_index.md
```diff
@@ -112,6 +112,10 @@ Disk space is low @ops-team@company.com
 
 {{% notifications-email %}}
 
+#### Teams
+
+If a [notification channel][17] is set, you can route notifications to a specific Team. Monitor alerts targeting @team-handle will be redirected to the selected communication channel.
```
urseberry, line 117: Link 17 goes to the Teams documentation, so it's confusing to me to link the "notification channel" text. Consider moving the link to "Team"?
```suggestion
If a notification channel is set, you can route notifications to a specific [Team][17]. Monitor alerts targeting @team-handle are redirected to the selected communication channel.
```

urseberry (changes requested):

content/en/monitors/notify/_index.md
estherk15: [@estherk15 To do in a separate PR] Need to add a note on the Teams page about how to add a notification channel, and then link to those instructions instead of the general Teams page.

estherk15 (approved):

PR #22253, opened 2024-03-19
Update locked deprecation guide

SimoneT-DD:
As we are moving on with the locked deprecation, we updated the public doc explaining that locked monitors will no longer be supported after March 27, 2024.

### DIFF
--- a/content/en/monitors/guide/how-to-set-up-rbac-for-monitors.md
+++ b/content/en/monitors/guide/how-to-set-up-rbac-for-monitors.md
@@ -128,13 +128,13 @@ Before Datadog released the feature allowing restriction of monitor editing to s
 
 {{< img src="/monitors/guide/monitor_rbac_locked.jpg" alt="RBAC Locked Monitor" style="width:70%;">}}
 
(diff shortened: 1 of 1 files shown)


jhgilbert (changes requested):
Hi @SimoneT-DD, thanks for this update! We usually don't include deprecation dates in documentation for products we own (versus third-party libraries). The preferred approach is to communicate these deprecations directly to customers, since most of them are unlikely to see the docs update. Then, when the deprecation happens, we push a docs update saying that the feature is deprecated.

Is it possible to take that approach here?

SimoneT-DD:
Hi @jhgilbert, we actually communicated this to customer already. I wanted to update the doc because right now we are saying that the locked attribute is deprecated but still supported. However we would like customers to migrate to the new attribute and we will start refusing all the APIs call that are still using the deprecated one. That's the reason why I added the date in the doc. 

I'm open for other suggestion that would deliver the same message: locked is deprecated and we would like everyone to migrate to the new attribute because the old one won't be supported anymore soon.

jhgilbert:
@SimoneT-DD Thanks for the context, I just edited your changes for style. I can approve and merge this when the checks have run. I'll put in a ticket for the docs team to follow up with a docs update on March 27, unless you would like to do the update?

jhgilbert (approved):

PR #22267, opened 2024-03-19
Data Jobs Monitoring - Add tags to EMR init script

paul-laffon-dd:

Adding the two tags:
- `init_script_version:1` to track which version of the script is used 
- `cluster_id` with the same value as `job_flow_id`, to have a unified tagging for cluster_id with other technologies

### Merge instructions


rtrieu (approved):
LGTM!

PR #22271, opened 2024-03-19
Note how method parameters are named in the expression language

deejgregor:

The documentation suggests that method parameters can be accessed, but does not state now this done in the expression langauage.

### Merge instructions

### Additional notes

Please have some familiar with the expression language and/or implementation review this as _I have familiarity with neither_ . I only discovered method parameters appeared to be named by using "Capture method parameters and local variables" on a log probe.

### DIFF
--- a/content/en/dynamic_instrumentation/expression-language.md
+++ b/content/en/dynamic_instrumentation/expression-language.md
@@ -15,7 +15,7 @@ In log templates and tag values, expressions are delimited from the static parts
 Probe conditions must evaluate to a Boolean, for example: `startsWith(user.name, "abc")`, `len(str) > 20` or `a == b`.
 
 Generally, the Expression Language supports:
-* Accessing local variables, method parameters, and deeply nested fields and attributes within objects.
+* Accessing local variables, method parameters (`p0`, `p1`, etc.), and deeply nested fields and attributes within objects.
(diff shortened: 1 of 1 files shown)


OmerRaviv:
Hi @deejgregor! Thank you very much for sending this PR.

The syntax for accessing parameter values is indeed to just use their names directly, which is no different than you would use for local variables. I've added a suggestion to your PR that would maybe convey this point. Does that help clarify the thing you found confusing? Please let me know if you think this would clarify the confusion here?

jpbempel:
Hi @deejgregor, 
`p0`, `p1`, ... argument names are a default naming when probes cannot find any names (or an issue with the mapping prevent to correctly named them).
We are not relying on `-parameters` javac flag to map argument names as this is not widely used in the Java ecosystem and not by default.
However most build tools (Maven, gradle) compile the code with `-g` javac flag which includes local variable table and therefore include argument names. 
Some third-party libraries may not compile their artifacts with `-g` (it includes JDK itself) and therefore no information are available for having correct naming for arguments. in that case we end up with this default naming generated by the datadog library.

jhgilbert:
Hi @deejgregor, I see that this PR is now a draft -- I'm going to cancel the internal review for now. If you'd like to resubmit this in the future, we'll be happy to review it! Thank you!

jhgilbert:

PR #22298, opened 2024-03-21
Update Specifying library versions section for k8s

brett0000FF:

The latest Agent release changed behavior of `libVersions`.

Previously, all tracing libraries were injected, even if only one library was configured in `libVersions`. Now, only the libraries a user configures are injected.

Updated docs to reflect the new functionality. Additionally, moved container registry table from Injecting libraries locally so users can find the latest versions better.

### Merge instructions

- [ ] Please merge after @vidurkhannadatadog approves

### DIFF
--- a/content/en/tracing/trace_collection/automatic_instrumentation/single-step-apm.md
+++ b/content/en/tracing/trace_collection/automatic_instrumentation/single-step-apm.md
@@ -262,9 +262,9 @@ To disable instrumentation for specific namespaces, add the `disabledNamespaces`
 
 ### Specifying tracing library versions
 
-<div class="alert alert-info">Injecting specific tracing libraries is only available for Datadog Cluster Agent version 7.52.0+.</div>
+<div class="alert alert-info">Starting with Datadog Cluster Agent v7.52.0+, you can inject a subset of tracing libraries into your applications.</div>
 
-Specify Datadog tracing library versions to control which versions and libraries are injected into your applications. You can configure this in two ways, which are applied in the following order of precedence:
+Specify Datadog tracing libraries and their versions to inject into your applications. You can configure this in two ways, which are applied in the following order of precedence:
 
 1. [Specify at the service level](#specifying-at-the-service-level), or
 2. [Specify at the cluster level](#specifying-at-the-cluster-level).
@@ -293,8 +293,6 @@ Replace <CONTAINER IMAGE TAG> with the desired library version. Available versio
 
 <div class="alert alert-warning">Exercise caution when using the <code>latest</code> tag, as major library releases may introduce breaking changes.</div>
 
-If an application is already instrumented using a different library version, the version loaded first takes precedence. Library injection happens at the admission controller level before runtime, so it overrides manually configured libraries.
-
 For example, to inject a Java library:
 
 {{< highlight yaml "hl_lines=12" >}}
@@ -315,6 +313,8 @@ spec:
         - # ...
 {{< /highlight >}}
 
+If an application is already instrumented using a different library version, the version loaded first takes precedence. Library injection happens at the admission controller level before runtime, so it overrides manually configured libraries.
+
 #### Specifying at the cluster level
 
 If you don't inject tracing libraries from the pod spec, you can specify tracing libraries for the entire cluster with Single Step Instrumentation configuration. When `apm.instrumentation.libVersions` is set, only the specified libraries and versions are injected.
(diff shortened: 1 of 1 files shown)


vidurkhannadatadog:
Can we add

Docker Hub is subject to image pull rate limits. If you are not a Docker Hub customer, Datadog recommends that you update your Datadog Agent and Cluster Agent configuration to pull from GCR or ECR. For instructions, see [Changing your container registry](https://docs.datadoghq.com/agent/guide/changing_container_registry).


Datadog publishes instrumentation libraries images on gcr.io, Docker Hub, and Amazon ECR:



on top of the registry table? otherwise lgtm

brett0000FF:
> Can we add
> 
> Docker Hub is subject to image pull rate limits. If you are not a Docker Hub customer, Datadog recommends that you update your Datadog Agent and Cluster Agent configuration to pull from GCR or ECR. For instructions, see [Changing your container registry](https://docs.datadoghq.com/agent/guide/changing_container_registry).
> 
> Datadog publishes instrumentation libraries images on gcr.io, Docker Hub, and Amazon ECR:
> 
> on top of the registry table? otherwise lgtm

Can do! Just pushed a change to include that additional language. @hestonhoffman will merge this for us tomorrow. 👍

vidurkhannadatadog:
I would remove the "If you include a library but don’t specify its version, it defaults to the latest version.". Else lgtm.

hestonhoffman (approved):

liliyadd:
For libVersions change, I think it's important to include the following information:

1. Injecting some tracing libraries and not all 5 of them is only available for DCA >= 7.52.0
2. That users can still overwrite libraries and their versions for specific deployments with the annotation `admission.datadoghq.com/<language>-lib.version`
3. If users have `apm.instrumentation.enabled=true` and `apm.instrumentation.libVersions` is not set, the SSI will inject latest version of all 5 libraries.

Maybe having some kind of precedence table or diagram would make it easier to understand? What do you think @vidurkhannadatadog ?

brett0000FF:
@liliyadd @vidurkhannadatadog - I included the extra info from Liliya in this PR. Please let me know if you have any feedback before I merge. Thanks!

vidurkhannadatadog:
You can also overwrite libraries and their versions for specific deployments using the annotation admission.datadoghq.com/<language>-lib.version.

1. Where?
2. How?

this needs to be expanded upon with an example.

liliyadd (commented):
I'm not sure what should be the wording , but this note sounds wrong to me

`Injecting specific tracing libraries is only available for Datadog Cluster Agent version 7.52.0+`

Before 7.52.0 users could have injected specific versions of tracing libraries. What changed in 7.52.0 is that users can inject a subset of tracing libraries, for example only Java and Python, instead of injecting all available libs.

liliyadd:
Also, for me it's very hard to understand the precedence and combination of library version configurations in the format they are described. I was thinking about adding some structure to the configuration options.

### Specifying tracing libraries and their versions per service
This way of configuring injected libraries has highest precedence and will overwrite library version configuration via Single Step Instrumentation.
<Add most of the content from https://docs.datadoghq.com/tracing/trace_collection/library_injection_local/?tab=kubernetes#step-2---annotate-your-pods-for-library-injection>
Note: Overwriting tracing libraries set via `apm.instrumentation.libVersions` is only available for Datadog Cluster Agent version 7.52.0+.

### Specifying tracing libraries and their versions via Single Step Instrumentation configuration
If injected tracing libraries are not configured via pod spec, they could be configured for the whole cluster via `apm.instrumentation.libVersions`. When `apm.instrumentation.libVersions` is set, only specified libraries and versions will be injected.
<Example>
Note: Injecting specific tracing libraries is only available for Datadog Cluster Agent version 7.52.0+.

### Injecting `latest` version of all available tracing libraries via Single Step Instrumentation configuration
If `apm.instrumentation.enabled=true` and `apm.instrumentation.libVersions` is unset, all supported tracing libraries will be injected into the apps running in namespaces where SSI is enabled.
<Example>

brett0000FF:
> Also, for me it's very hard to understand the precedence and combination of library version configurations in the format they are described. I was thinking about adding some structure to the configuration options.

@liliyadd - Appreciate your feedback and keen eye! It's always welcome.

I took another pass to better structure the content around the hierarchy of configuring through the pod spec vs. Single Step config.

Please let me know what you think. Thanks! 🙏 

CC @vidurkhannadatadog

hestonhoffman (approved):
LGTM!

PR #22301, opened 2024-03-22
[NO-TICKET] Bump minimum Ruby profiler versions for allocation, heap and timeline

ivoanjo:
As of dd-trace-rb 1.21.1:

* Allocation is now beta => changed table to reflect that
* Timeline is now on by default

In the past few releases we did a number of important performance improvements, so I'm bumping the minimum versions we state for allocation + heap + timeline as otherwise running older versions means running with known performance pitfalls.

### Merge instructions

### Additional notes

N/A


hestonhoffman (approved):
LGTM!

PR #22325, opened 2024-03-25
[NO-TICKET] Update Ruby profile types

ivoanjo:

This PR updates the Ruby profile types documentation to match the same changes in https://github.com/DataDog/documentation/pull/22301:

* Allocation is now in beta
* Bumped the minimum versions to v1.21.1+
* Allocation works in all Rubies

In the past few releases we did a number of important performance improvements, so I'm bumping the minimum versions we state for allocation + heap + timeline as otherwise running older versions means running with known performance pitfalls.

Finally, I've clarified that allocation profiling actually works on all Rubies we support, not only Ruby 2.7. This was an oversight on my earlier PR to add this documentation, and doesn't actually reflect the code. Oops :)

### Merge instructions

### Additional notes

N/A

### DIFF
--- a/content/en/profiler/profile_types.md
+++ b/content/en/profiler/profile_types.md
@@ -160,7 +160,7 @@ Wall Time
 
 Allocations (beta, v1.21.1+)
 : The number of objects allocated by each method during the profiling period (default: 60s), including allocations which were subsequently freed. This is useful for investigating garbage collection load.<br />
-_Requires: [Manual enablement][3]
+_Requires:_ [Manual enablement][3]
 
 Heap Live Objects (alpha, v1.21.1+)
 : The number of objects allocated by each method in heap memory that have not yet been garbage collected. This is useful for investigating the overall memory usage of your service and identifying potential memory leaks.<br />
--- a/content/en/profiler/profile_types.md
+++ b/content/en/profiler/profile_types.md
@@ -158,20 +158,21 @@ CPU
 Wall Time
 : The elapsed time used by each function. Elapsed time includes time when code is running on CPU, waiting for I/O, and anything else that happens while the function is running.
 
-Allocations (alpha, v1.19.0+)
+Allocations (beta, v1.21.1+)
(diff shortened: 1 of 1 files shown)


content/en/profiler/profile_types.md
```diff
@@ -158,20 +158,21 @@ CPU
 Wall Time
 : The elapsed time used by each function. Elapsed time includes time when code is running on CPU, waiting for I/O, and anything else that happens while the function is running.
 
-Allocations (alpha, v1.19.0+)
+Allocations (beta, v1.21.1+)
 : The number of objects allocated by each method during the profiling period (default: 60s), including allocations which were subsequently freed. This is useful for investigating garbage collection load.<br />
-_Requires: Ruby 2.7+_ and [needs to be enabled][2]
+_Requires: [Enabling manually][3]
```
AlexJF, line 163: Minor: "Requires manual enablement" sounds slightly more correct?

content/en/profiler/profile_types.md
```diff
@@ -158,20 +158,21 @@ CPU
...
 
-Heap Live Objects (alpha, v1.19.0+)
+Heap Live Objects (alpha, v1.21.1+)
 : The number of objects allocated by each method in heap memory that have not yet been garbage collected. This is useful for investigating the overall memory usage of your service and identifying potential memory leaks.<br />
 _Requires: Ruby 2.7+_ and [needs to be enabled][2]
```
AlexJF, line 167: If we change the above from `needs to be enabled`, I'd say we should update these as well?

AlexJF (approved):

content/en/profiler/profile_types.md
ivoanjo: Fixed in 933e1a7066b1afe3c7e30f840178daba183d6f3c

content/en/profiler/profile_types.md
ivoanjo: Fixed in 933e1a7066b1afe3c7e30f840178daba183d6f3c

buraizu (approved):

PR #22344, opened 2024-03-26
update supported versions of Red Hat Linux

tetsuyamashima:

Added rows corresponding to version 5.x and 6.x of Red Hat Linux.

### Merge instructions


estherk15 (commented):
Thanks for contributing to the docs @tetsuyamashima! Can you include supporting docs or motivation for this change?

spencergilbert (approved):
I'm not sure if we can provide `X.51.*` to represent that any patch version should support CentOS/RHEL 6 - but that may save us an update here if we release another patch.

Otherwise this seems correct for the CentOS 6 deprecation that happened with the `7.52` release. I'm not certain about the CentOS 5 support - but I believe it's correct.

estherk15 (approved):

PR #22389, opened 2024-03-28
[SYNTHETICS MOBILE] Remove error codes that can't happen in synthetics mobile

RemiGonin:
The error codes for synthetics mobile in the documentation contained errors that can happen for a synthetics browser test, but not a mobile one.
We also find elements in a XML representation instead of a HTML representation.

Changes:
- Remove error codes that don't belong to mobile
- Change HTML to XML
- My linter also removed trailing spaces. 

### Merge instructions


ClaraJacintho (approved):
Looks good for Synthetics Mobile

rtrieu (approved):
Looks good, thank you for the update!

PR #22412, opened 2024-03-28
add region link-element functionality with DOCS-7443 use case

cswatt:
Previously, using the `region-param` shortcode could be used to create a `<code>` or `<span>` element that contains the desired region-specific parameter. 

In DOCS-7443, there came a request where a link should change depending on the site selector. I've added functionality so the `region-param` shortcode can also create an `<a>` element where the `href` attribute is populated by the region-specific parameter. This is done by adding a `link="true"` attribute.

This PR makes the following changes:
- Add a case for `link="true"` in `layout/shortcodes/region-param.html`
- Write functionality in `assets/scripts/region-redirects.js`
- Add `ip_ranges_url` dictionary to `regions.config.js`
- Implement on a docs page, `forwarding_custom_destinations`

### Usage (for Docs4Docs after merge)

This shortcode:

```
{{< region-param key="PARAM_KEY" link="true" text="LINK_TEXT" >}}
```

Renders as:

```html
<a href="PARAM_KEY_VALUE">LINK_TEXT</a>
```

### Don't in-app links already get redirected by region?
Yes. This is an extra case for **non-in-app links that are region-dependent**. 

### Merge instructions
I've tested this locally, but this could probably use more testing...


brett0000FF (approved):
Very cool! I tested this with a new param, and it worked like a charm. Thanks!

PR #22168, opened 2024-03-12
Initial overhaul of widget overview

estherk15:
### What does this PR do? What is the motivation?

- Add more information where possible to Widget intro section
- [DOCS-7202](https://datadoghq.atlassian.net/browse/DOCS-7202)

### Merge instructions

- [ ] Please merge after reviewing

### Additional notes

[DOCS-7202]: https://datadoghq.atlassian.net/browse/DOCS-7202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #22439, opened 2024-04-01
Update slo.md

roxanne-moslehi:
Follow up from https://github.com/DataDog/documentation/pull/22168

### Merge instructions


roxanne-moslehi:
@estherk15 can we take out this sentence from the overview? It's not up to date since we support all groups for all monitor types and it feels weird to call out that it was previously limited to 20 groups

estherk15 (approved):

PR #22474, opened 2024-04-03
Remove duplicate key

etnbrd:
This PR removes a duplicate identifier preventing CI to pass.


maycmlee (approved):
Thanks for fixing this!

PR #22509, opened 2024-04-04
[WIP] Add GraalVM native-image profiler setup instructions

jbachorik:
### Merge instructions


content/en/profiler/enabling/java.md
```diff
@@ -131,7 +132,22 @@ java \
...
+   ```
+
+   **Note**: Only JFR based profiling is supported for the GraalVM native-image applications. None of the 'DDPROF'
+related configuration options is effective.
+
+{{% /collapse-content %}}
+
+   **Note**: The `-javaagent` argument needs to be before `-jar`, adding it as a JVM option rather than an application argument. For more information, see the [Oracle documentation][6]:
 
     ```shell
```
brett0000FF, line 152: This code block isn't rendering properly in the preview anymore. Deleting one space before each line should fix it. Side note: if we change the code fence to `bash` instead of `shell` it won't include the `$` on the first line.

content/en/profiler/enabling/java.md
```diff
@@ -131,7 +132,22 @@ java \
 {{% /tab %}}
 {{< /tabs >}}
 
-    **Note**: The `-javaagent` argument needs to be before `-jar`, adding it as a JVM option rather than an application argument. For more information, see the [Oracle documentation][6]:
+{{% collapse-content title="Build and run native-image" level="h4" %}}
```
brett0000FF, line 135: Curious what you think, but I think it might flow better if we move the GraalVM collapsed section to around line 158, between the example and step 4? That way the note about `-javagent` more closely relates to step 3.

Also, I think we could change the collapsed section title to include `(Optional)` at the beginning.

brett0000FF (approved):
@jbachorik  Looks good! Thanks for writing this up. I made a few style suggestions, and I added a couple of comments for you to consider. Please let me know if anything is unclear. Happy to merge this after you take a look! 🚀

content/en/profiler/enabling/java.md
jbachorik: I was considering this - the downside is that the `-javaagent` note actually applies also to the GrallVM native-image setup. It's just that it isn't explicit.
TBH, I would leave this up to you - if you feel that it flows better, let's do it, I have no strong opinion about that.

Regarding `(Optional)` - sure, that's perfectly fine.

Will you do the adds? You can directly push to this branch ...

brett0000FF: Added everything (left the section where it was given that the note should apply to both)!

PR #22532, opened 2024-04-05
[RUM] Replace mentions of metrics with telemetry

rtrieu:

As per discussion with @MaelNamNam to improve accuracy of the terms used - in some cases, we don't offer metrics OOTB in RUM, so we are technically referring to custom metrics
- Removes mentions of "metrics" in RUM docs and replaces them with "telemetry" or "data points" unless we're actually talking about metrics

### Merge instructions


MaelNamNam (approved):
LGTM, thanks a lot @rtrieu

brett0000FF (approved):
Nice, this is a helpful change!

PR #22639, opened 2024-04-11
(maint) Fix local build loop issue

hestonhoffman:
We've been running into the local build loop issue a lot more recently. The problem seems to stem from macOS triggering multiple CHMOD changes, possibly in an interaction with VSCode. This seems to fix the issue locally, but I'm not sure what the broader implications are.

### Merge instructions

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
         "preinstall": "bash local/bin/sh/preinstall.sh",
         "prestart": "rimraf public && npx hugo mod clean",
         "docker:start": "/usr/local/bin/hugo server -D -d ./public -s ./ --port 1313 --navigateToChanged --noHTTPCache --bind 0.0.0.0",
-        "start": "./node_modules/.bin/hugo server -D -d ./public -s ./ --port 1313 --navigateToChanged --noHTTPCache --verbose --poll 2s",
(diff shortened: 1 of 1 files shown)


hestonhoffman:
Just noticed a `--noChmod` option in the [docs](https://gohugo.io/commands/hugo_server/#options). I wonder if that's a better way to go.

package.json
```diff
@@ -8,7 +8,7 @@
         "preinstall": "bash local/bin/sh/preinstall.sh",
         "prestart": "rimraf public && npx hugo mod clean",
         "docker:start": "/usr/local/bin/hugo server -D -d ./public -s ./ --port 1313 --navigateToChanged --noHTTPCache --bind 0.0.0.0",
-        "start": "./node_modules/.bin/hugo server -D -d ./public -s ./ --port 1313 --navigateToChanged --noHTTPCache",
+        "start": "./node_modules/.bin/hugo server -D -d ./public -s ./ --port 1313 --navigateToChanged --noHTTPCache --noChmod",
```
hestonhoffman, line 11: Originally I was using `--poll 2s` which also seemed to work. But this targets the problem more directly.

bgdeutsch (approved):
tested your branch running vscode, no issues w/ either full `make start` or `yarn run start`.   no issues around saving files, no random changes detected.   it seems low risk since it's a local only change, so i think we should be good to release and keep an eye on things.

jhgilbert (approved):

PR #22688, opened 2024-04-16
[DOCS-7636] moving browser tests MFA doc

aliciascott:
This PR moves the Browser tests "Monitor An Application That Requires Authentication With Browser Tests" _from_ Synthetics -> Guides _to_ Synthetics - Browser Tests so it's easier to find.

### Merge instructions

### DIFF
--- a/config/_default/menus/menus.en.yaml
+++ b/config/_default/menus/menus.en.yaml
@@ -4523,6 +4523,11 @@ main:
     url: synthetics/browser_tests/advanced_options
     parent: synthetics_browser_tests
     weight: 303
+  - name: Multi-factor authentication methods 
+    identifier: synthetics_browser_tests_toto_mfa
+    url: synthetics/browser_tests/browser-tests-totp
+    parent: synthetics_browser_tests
+    weight: 303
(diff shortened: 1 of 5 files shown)


config/_default/menus/menus.en.yaml
```diff
@@ -4523,6 +4523,11 @@ main:
     url: synthetics/browser_tests/advanced_options
     parent: synthetics_browser_tests
     weight: 303
+  - name: Authentication in Browser Tests 
+    identifier: synthetics_browser_tests_authentication
+    url: synthetics/browser_tests/app-that-requires-login
+    parent: synthetics_browser_tests
+    weight: 303
```
cswatt, line 4530:
```suggestion
    weight: 304
```

looks like the previous item in the menu is also 303, so this one should be incremented!

cswatt (changes requested):

config/_default/menus/menus.en.yaml
aliciascott: oh my gosh.. 🤦 good catch :)

PR #22731, opened 2024-04-18
(DOCS-6444) Make AlwaysOn information more prominent

neko-dd:

Puts AlwaysOn information into a shortcode so that it can be in all the tabs (reducing the chance that people skip over the info and go straight to the tabs), and makes the most important part bolder.

Note: Doesn't do the same thing on all pages because some of them need different language, or the content is already in a shortcode. In the case of Azure and GCSQL, I just made more of the text bold.

### Merge instructions

### DIFF
--- a/content/en/database_monitoring/setup_sql_server/azure.md
+++ b/content/en/database_monitoring/setup_sql_server/azure.md
@@ -128,7 +128,7 @@ For [SQL Server on Windows Azure VM][1] follow the [Setting Up Database Monitori
 
 ## Install the Agent
 
-Since Azure does not grant direct host access, the Datadog Agent must be installed on a separate host where it is able to talk to the SQL Server host. There are several options for installing and running the Agent.
+**Because Azure does not grant direct host access, the Datadog Agent must be installed on a separate host where it is able to talk to the SQL Server host**. There are several options for installing and running the Agent.
 
 {{< tabs >}}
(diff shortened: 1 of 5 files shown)


content/en/database_monitoring/setup_sql_server/azure.md
```diff
@@ -128,7 +128,7 @@ For [SQL Server on Windows Azure VM][1] follow the [Setting Up Database Monitori
 
 ## Install the Agent
 
-Since Azure does not grant direct host access, the Datadog Agent must be installed on a separate host where it is able to talk to the SQL Server host. There are several options for installing and running the Agent.
+**Because Azure does not grant direct host access, the Datadog Agent must be installed on a separate host where it is able to talk to the SQL Server host**. There are several options for installing and running the Agent.
```
jhgilbert, line 131: I would leave this unbolded here, since we aren't bolding it elsewhere -- I think Casey was just worried about the AlwaysOn verbiage specifically.

jhgilbert (approved):
Approved with the recommendation to remove bold on one part, thank you!

content/en/database_monitoring/setup_sql_server/azure.md
neko-dd: Okay, happy to defer to you on this one.

PR #22786, opened 2024-04-22
Cross-link Error Tracking documentation

jhgilbert:
DOCS-7608

### Merge instructions

Do not merge this, thanks!

### DIFF
--- a/content/en/error_tracking/rum.md
+++ b/content/en/error_tracking/rum.md
@@ -3,11 +3,12 @@ title: Error Tracking for Web and Mobile Applications
 kind: documentation
 ---
 
-{{< whatsnext desc="To get started with Error Tracking for RUM, see the corresponding documentation to upload source maps for your framework:" >}}
+{{< whatsnext desc="To get started with Datadog Error Tracking for RUM, see the corresponding documentation:" >}}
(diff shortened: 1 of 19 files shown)


content/en/error_tracking/rum.md
```diff
@@ -0,0 +1,13 @@
+---
+title: Error Tracking for Web and Mobile Applications
+kind: documentation
+---
+
+{{< whatsnext desc="To get started with Error Tracking for RUM, see the corresponding documentation to upload source maps for your framework:" >}}
```
hestonhoffman, line 6: Should this include Roku?

hestonhoffman (approved):
So, this LGTM. I'll admit the RUM section (in the RUM docs) kinda threw me a little at first, cos some of the entries jump to elsewhere in the RUM docs, and they don't do that in the other error tracking docs. As long as that's expected (and I can see why it would be), then this LGTM

content/en/error_tracking/rum.md
jhgilbert: Great catch -- the docs I copied from RUM omitted Roku, so I added it to both, thank you!

And yeah, the jumps within RUM are unfortunate but expected, they were already there and I just don't have a great solution for those, especially since it's apparently preferred to reroute mobile folks over to the mobile section of the docs at the first opportunity. So I'm going to leave them, but I'm with you, I find it jarring.

PR #22806, opened 2024-04-23
Adds a console output example for Java logs

petems:
* Currently only have a file example
* K8s/containers more likely to use console output

Currently there's no specific example of how to do console/stdout output for Java logs in the docs, only file. In a K8s and container world, we're more likely to use console output and it was a bit of trial and error for me to figure it out, it'd be nice to have that directly in the docs. 

### Merge instructions


petems:
Closing in favour of https://github.com/DataDog/documentation/pull/22808 as the preview builder doesnt work nicely with forks

PR #22834, opened 2024-04-24
[CIVIS-9871] Remove gitlab `resumed` and `paused` partial statuses

rodrigo-roca:
We are removing the gitlab pipelines with partial statuses `blocked` and `resumed`. You can check on org2 that we do not have those values for gitlab pipelines.

### Merge instructions
Ready to be merged!

### Additional notes
* [Jira card](https://datadoghq.atlassian.net/browse/CIVIS-9871)


TovRudyy (approved):

buraizu (changes requested):
Thanks for the PR, this looks good overall. However, could you please make the following adjustment:
- Upload the new image with a slightly modified name (for example, static/images/ci/partial_retries_facet_panel_2.png)
- Update the image shortcode on line `182` to point to the new image name

The reason for this request is to prevent caching issues with images. We have some internal docs related to this which I'm happy to share with you if you'd like. Let me know if you have any questions!

PR #22863, opened 2024-04-25
[web-4507] first websites-sources integration

bgdeutsch:
sources two integrations, `ably` and `akamai_mpulse`, from `websites-sources` as an initial onboarding of content from `websites-sources`.   the data for these integrations are being sourced from [Asset Publishing Workflow](https://datadoghq.atlassian.net/wiki/spaces/APW/overview).

additionally this PR builds support for importing integration logo image data directly from `websites-sources` into docs repo, eliminating the extra network call to s3 at build time.

jira: https://datadoghq.atlassian.net/browse/WEB-4507

### Merge instructions
- [ ] Please merge after websites reviews

note cache is disabled to simulate what would happen in a production build.   this solution accounts for preview builds with the cache enabled as well.   i will re-enable the cache after review.

### Additional notes
https://docs-staging.datadoghq.com/brian.deutsch/first-apw-integration/integrations/

- confirm all integration logos look correct.
- confirm `ably` and `akamai_mpulse` specifically - the tiles exist and match what's currently in live.

https://docs-staging.datadoghq.com/brian.deutsch/first-apw-integration/integrations/ably
https://docs-staging.datadoghq.com/brian.deutsch/first-apw-integration/integrations/akamai_mpulse

- these integration detail pages match what's in live currently, including content, carousel/media, etc.
- confirm all other integration details pages are unaffected by these changes.


layouts/partials/integrations-logo.html
```diff
@@ -36,7 +36,7 @@
 {{- $extension := .extension | default ".svg" -}}
 {{- $fallback := .fallback | default "none" -}}
 {{- $img_url := $ctx.Site.Params.integrations_img_url -}}
-{{- $imagesJSON := getJSON "https://s3.amazonaws.com/origin-static-assets/websites-integrations/images.json" -}}
+{{- $imagesJSON := site.Data.images -}}
```
bgdeutsch, line 39: will rename this to something more descriptive soon

assets/styles/components/_integrations.scss
```diff
@@ -35,6 +35,19 @@ $screen-md-max: ($screen-lg-min - 1) !default;
...
+            display: none;
+        }
+
+        td {
+            word-break: break-word;
+            width: 50%;
+        }
+    }
+  }
+
```
bgdeutsch, line 50: need the CSS changes here as the metrics tables used to be driven by a shortcode but are now pure markdown (so cant drop in custom classes/no html)

layouts/integrations/single.html
```diff
@@ -1,4 +1,12 @@
 {{ define "main" }}
+    {{ $media := slice }}
+
+    {{ if and (.Params.tile) (.Params.tile.media) }}
+        {{ $media = .Params.tile.media }}
+    {{ else if .Params.media }}
+        {{ $media = .Params.media }}
+    {{ end }}
```
davidejones, line 8: maybe we can use default instead i think something like this might work `{{ $media := .Params.tile.media | default .Params.media | default "" }}`

local/bin/py/build/update_pre_build.py
```diff
@@ -42,6 +42,17 @@ def __init__(self, tempdir):
...
+        self.apw_integrations = [
+            {
+                'extract_path': f'{self.extract_dir}/dogweb/integration/akamai_mpulse',
+                'cache_path': 'temp/content/en/integrations/akamai_mpulse.md'
+            },
+            {
+                'extract_path': f'{self.extract_dir}/integrations-extras/ably',
+                'cache_path': 'temp/content/en/integrations/ably.md'
+            }
+        ]
```
davidejones, line 55: I'm wondering if we can just build a list of the names instead and do any path work in the usage logic instead.
For example the cache path will always be the same also the extract path may need to be removed from all repos not just a specific integrations-extras etc as we still have duplicates

```
self.apw_integrations = [
'akamai_mpulse',
'ably'
]
```

bgdeutsch: thanks @davidejones!    refactored this

hestonhoffman:
I'm going to give this one a review tomorrow!

davidejones (approved):

brett0000FF:
Added under review label to take it out of the queue, since @hestonhoffman called dibs on this. 😄

local/bin/py/build/configurations/pull_config_preview.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 - config:
-    cache_enabled: true
+    cache_enabled: false
```
hestonhoffman, line 3:
```suggestion
    cache_enabled: true
```

PR #22873, opened 2024-04-25
Update serverless and cloud cost with new nav references

estherk15:

- Removed images with outdate nav reference
- Updated docs with navigation instructions

### Merge instructions

### Additional notes

This page also has a lot of images showing the nav, but I'm not sure if removing it is the way to go. If you think these images should be removed, I can add that work to this PR!


cswatt (approved):
Thank you so much for doing this, it looks great!

I'm inclined towards keeping the images (at least for now) since the old nav is at least minimized, and they're nice blog-quality screenshots.

cswatt (approved):

PR #22890, opened 2024-04-26
[CCV] specify Preview Portal

klara-kr:

We previously changed the documentation to clarify that we don't support the new improved Azure export experience. This is adding to that to inform customers they need to use the preview portal to create exports in the old way.

### Merge instructions


jonathankauch (approved):

crousseaux (approved):
lgtm, thanks for the quick add

PR #22922, opened 2024-04-29
(DOCS-7884) Adds documentation for App Builder components

neko-dd:

* Adds a reference page for App Builder components.
* Links to said page from the Build page

### Merge instructions


content/en/service_management/app_builder/components.md
```diff
@@ -0,0 +1,767 @@
...
+
+### General
+
+Label
+: The text that displays at the top of the component.<br>
+**Value**: string or expression
+
+Options
+: The list of checkboxes that a user can select from. Minimum number of options is 1.<br>
+**Value**: expression<br>
```
hestonhoffman, line 142: For the values that take expressions, it's great that we're giving an example, but I think we should also describe the expected result, because in this case the expression has to conform to the example below - an array of objects, each object consisting of a "label" and "value" key-value pair.

content/en/service_management/app_builder/components.md
```diff
@@ -0,0 +1,767 @@
...
+
+### General
+
+Label
+: The text that displays at the top of the component.<br>
+**Value**: string or expression
+
+Options
+: The list of radio button options that a user can select from.<br>
+**Value**: string or expression<br>
```
hestonhoffman, line 394: I think the tooltip is incorrect for this one. It didn't accept a string for me. Same comment as above re expression

content/en/service_management/app_builder/components.md
```diff
@@ -0,0 +1,767 @@
...
+: The text that displays at the top of the component.<br>
+**Value**: string or expression
+
+Placeholder text
+: The text that displays when no value is entered.<br>
+**Value**: string or expression
+
+Options
+: The list of select options that a user can select from.<br>
+**Value**: string or expression<br>
```
hestonhoffman, line 503: As above, this one doesn't accept a string + expression stuff

hestonhoffman (commented):
This is looking good - I left some suggestions about expressions, and also a note on fields that don't accept strings (I think this is a tooltip bug - I posted more about it in the Slack channel).

content/en/service_management/app_builder/components.md
neko-dd: PTAL, and this text is reproduced in the other two locations as well, as well as having the others only say they accept expressions (like this one already said)

hestonhoffman (approved):

PR #22985, opened 2024-05-02
Fix the generation of the Accept/Content-Type header based on 204 responses

clementgbcn:

The 'Content-Type' header in a request specifies the media type of the request body being sent to the server. It tells the server what kind of data is being sent in the request. On the other hand, the 'Accept' header in a request specifies the media types that the client can process in the response. It tells the server what kind of response content types are acceptable to the client.

In the context of a cURL request to an endpoint returning a 204 status code, setting the 'Content-Type' header is relevant if you are sending a request body with content. The server will use the 'Content-Type' header to determine how to process the data in the request body.

On the other hand, setting the 'Accept' header in the request tells the server what kind of response content types are acceptable to the client. Even if the server is responding with a 204 status code, specifying the 'Accept' header can help communicate to the server the preferred format of the response data in case there is a range of possible content types that could be returned.

In summary, the 'Content-Type' header pertains to the request body being sent by the client, while the 'Accept' header pertains to the response content types that the client can process and prefers.

The issue was detected with the [security monitoring rule validation endpoint](https://docs.datadoghq.com/api/latest/security-monitoring/#validate-a-detection-rule) (and validated with the delete dashboard one). The `cURL` example does not work, customers have to replace  `-H "Accept: application/json"` \ by `-H "Content-Type: application/json" \` to make the request work. However, they encounter the error `{"errors":["Invalid JSON input"]}` since the `Content-Type` is not defined.

### Testing
Before
image

After
image

### Merge instructions


therve (approved):

davidejones (approved):

estherk15 (approved):
Thanks for adding this!

PR #23024, opened 2024-05-06
Update dotnet-core.md to remove misleading statement in installation …

jessicagamio:
…Note

While the Datadog products (APM and Cont. Profiler) use the same API, we (Datadog) have a way to share this internally so they can both work at the same time.

clarifies the following - While the Datadog products (APM and Cont. Profiler) use the same API, we (Datadog) have a way to share this internally so they can both work at the same time.

### Merge instructions

### DIFF
--- a/content/en/tracing/trace_collection/automatic_instrumentation/dd_libraries/dotnet-core.md
+++ b/content/en/tracing/trace_collection/automatic_instrumentation/dd_libraries/dotnet-core.md
@@ -61,7 +61,7 @@ For a full list of Datadog's .NET Core library and processor architecture suppor
 </div>
 
 <div class="alert alert-warning">
-  <strong>Note:</strong> Datadog's automatic instrumentation relies on the .NET CLR Profiling API. This API allows only one subscriber (for example, Datadog's .NET Tracer with Profiler enabled). To ensure maximum visibility, run only one APM solution in your application environment.
+  <strong>Note:</strong> Datadog's automatic instrumentation relies on the .NET CLR Profiling API. This API allows only one subscriber, with exception of our Datadog APM and the Continous Profiler. To ensure maximum visibility, run only one APM solution in your application environment. 
 </div>
 
 <div class="alert alert-info">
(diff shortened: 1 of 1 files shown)


content/en/tracing/trace_collection/automatic_instrumentation/dd_libraries/dotnet-core.md
```diff
@@ -61,7 +61,7 @@ For a full list of Datadog's .NET Core library and processor architecture suppor
 </div>
 
 <div class="alert alert-warning">
-  <strong>Note:</strong> Datadog's automatic instrumentation relies on the .NET CLR Profiling API. This API allows only one subscriber (for example, Datadog's .NET Tracer with Profiler enabled). To ensure maximum visibility, run only one APM solution in your application environment.
+  <strong>Note:</strong> Datadog's automatic instrumentation relies on the .NET CLR Profiling API. This API allows only one subscriber, with exception of our Datadog APM and the Continous Profiler. To ensure maximum visibility, run only one APM solution in your application environment.
```
urseberry, line 64:
```suggestion
  <strong>Note:</strong> Datadog's automatic instrumentation relies on the .NET CLR Profiling API. This API allows only one subscriber, with the exception of APM and the Continuous Profiler. To ensure maximum visibility, run only one APM solution in your application environment. 
```

urseberry (changes requested):

content/en/tracing/trace_collection/automatic_instrumentation/dd_libraries/dotnet-core.md
zacharycmontoya: What do you think of this suggestion? I think it might be confusing to say that there is are exceptions. Additionally, whatever we end up with in this file, can you please copy the contents over to the dotnet-framework page as well?
```suggestion
  <strong>Note:</strong> Datadog's automatic instrumentation relies on the .NET CLR Profiling API. This API allows only one subscriber (for example, Datadog APM). To ensure maximum visibility, run only one APM solution in your application environment. 
```

jessicagamio: We can go with that to keep it simple. I agree.

urseberry (approved):

zacharycmontoya:
@jessicagamio before merging, can you copy the updated text to  the corresponding dotnet-framework.md page?

jessicagamio:
I added the PR for .Net framework - https://github.com/DataDog/documentation/pull/23055

PR #23028, opened 2024-05-07
[CWS-2092] - Remove beta tags for remote agent rules update

Malo10LeGoff:

Default and custom agent rule will be soon GA so we don't need a reference to a private/public beta

### Merge instructions


aliciascott (approved):
Looks good

PR #23098, opened 2024-05-09
Add image to shortcode

maycmlee:

- Move image from the parent page to general reuse shortcode.
- Add a new processor shortcode for SDS use case.

### Merge instructions


aliciascott (approved):
Looks good! Noted in slack to re-take images without the mouse cursor :)

PR #23108, opened 2024-05-10
Update aws.md

mk161162:

fix typo 
Azure --> 設定 (setup in Japanese)

### Merge instructions


aliciascott:
Hi @mk161162 , typically we don't update the non-English language pages as we have a translation process that does this, however, I can go ahead and merge this through for now, so it is at least correct in the short term.

aliciascott (approved):

PR #23186, opened 2024-05-15
docs: typo directive instead of directory

amardeep2006:
The correct term is directive for www.conf

### Merge instructions


cswatt (approved):
This is a great catch! Thank you so much!

PR #23252, opened 2024-05-17
LLMObs submitting evaluations page

Yun-Kim:

This PR adds a new page `Submitting evaluations` to the LLMObs documentation, duplicating and pulling out the SDK/API instructions for submitting evaluations into the new page to make easier for users to find.

### Merge instructions

### DIFF
--- a/config/_default/menus/main.en.yaml
+++ b/config/_default/menus/main.en.yaml
@@ -2983,21 +2983,26 @@ menu:
       parent: tracing_llm_obs
       identifier: tracing_llm_obs_trace_an_application
       weight: 702
+    - name: Submit Evaluations
+      url: tracing/llm_observability/submit_evaluations/
+      parent: tracing_llm_obs
+      identifier: tracing_llm_obs_submit_evaluations
+      weight: 703
     - name: Span Kinds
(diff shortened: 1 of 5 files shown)


jhgilbert:
Hi @Yun-Kim, thanks for the updates! A few concerns here: 

- This PR is out of date in that it contains changes that have already been merged -- it would be helpful not to include those so the PR is easier to review.
- We should not duplicate the API documentation. Your concern seems to be that the Evaluations API information will be too difficult for the user to find if they're just browsing through the content. In that case, we have a few options:
    - The top of the API documentation can have a bit more verbiage around both API endpoints, and links to them, so both are discoverable from the top level. This is my recommendation, since it still allows us to link to "the API" as one monolithic page. I'm happy to set that up.
    - I can set up the API documentation to have two separate pages, API > Evaluations API and API > Spans API. This is a bit more awkward because the left nav doesn't have enough levels to support this very well, but it can be done if that's really what we need.

Either way, I ask that you limit this PR to just editing the existing pages as you want them (changing "Evaluation metrics" to "Evaluations", for example), so I can merge that first and then do the above work on top of it. Thank you!

cc @barieom

jhgilbert (approved):

PR #23279, opened 2024-05-20
Node.js updates for ESM support

domalessi:
https://datadoghq.atlassian.net/browse/DOCS-7897

Struggled to figure out the right flow/formatting. Main goal of this PR is to clarify requirements for importing and initializing the tracer. 

For non-ESM applications, the steps are to either (a) add tracer in code, or (b) add tracer via command line arguments. 

For ESM applications, the steps are the same -- to either (a) add tracer in code, or (b) add tracer via command line arguments -- BUT with the additional step that regardless of whether (a) or (b) was used, an additional command line argument must be used (`--loader` or `--import` depending on Node.js version). 

I would've like to figure out a way to specify the full command line for ESM applications -- in other words, to merge the "Option 2: Add the tracer via command line arguments" and "ESM applications only: import the loader" sections, and include a single code line sample with everything there. But in order to do this, I would've had to include a call-out under the "Option 1: Add the tracer in code" section to note that ESM apps need to run an additional command line. 

I don't know if this makes sense, but basically I don't love the structure I have here but it seemed like the best approach. If you have thoughts/suggestions, I'd love to hear them!

### Merge instructions

### DIFF
--- a/content/en/tracing/trace_collection/automatic_instrumentation/dd_libraries/nodejs.md
+++ b/content/en/tracing/trace_collection/automatic_instrumentation/dd_libraries/nodejs.md
@@ -108,7 +108,7 @@ Node.js < v20.6
 node --loader dd-trace/loader-hook.mjs entrypoint.js
 ```
 
-Node.js >= v20.6
+**Node.js >= v20.6**
 ```shell
 node --import dd-trace/register.js entrypoint.js
 ```
--- a/content/en/tracing/trace_collection/automatic_instrumentation/dd_libraries/nodejs.md
+++ b/content/en/tracing/trace_collection/automatic_instrumentation/dd_libraries/nodejs.md
@@ -104,7 +104,7 @@ node --require dd-trace/init app.js
 EcmaScript Modules (ESM) applications require an additional command line argument. Run this command regardless of how the tracer is imported and initialized.
 
 Node.js < v20.6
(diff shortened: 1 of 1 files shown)


content/en/tracing/trace_collection/automatic_instrumentation/dd_libraries/nodejs.md
```diff
@@ -92,17 +89,29 @@ initializes in one step.
...
+
+EcmaScript Modules (ESM) applications require an additional command line argument. Run this line regardless of how the tracer was imported and initialized.
+
+Node.js < v20.6
+```sh
+node --loader dd-trace/loader-hook.mjs entrypoint.js
+```
+
+Node.js >= v20.6
+```sh
```
brett0000FF, line 112:
```suggestion
```shell
```

brett0000FF (approved):
Really nice job! Please let me know if you have any questions about my suggestions. I tried to include a note for each of them.

I do like the approach you took to organize it. The only thing I can think that would be nice to add (not a blocker though) is a brief, one-line description for each option for why a user would use that over the other.

For example, maybe for option 1, this is for users who want more precise control over their complex applications. And for option 2, this is for users who want a simpler setup but with consistent setup across environments due to the env variables.

bengl (approved):

PR #23353, opened 2024-05-24
Document ddagentuser is now part of the Performance Log Users group

julien-lebot:

This PR updates the documentation for the `ddagentuser` to include that, starting with 7.51, it is also a member of the "Performance Log Users" group.

### Merge instructions


aliciascott (approved):
looks good

PR #23374, opened 2024-05-27
NDM: Update profiles documentation

FlorianVeaux:
- Remove govcloud banner
- Update the documentation to use the recommended way of writing profiles (files instead of in-line)


content/en/network_monitoring/devices/profiles.md
```diff
@@ -28,36 +24,39 @@ By default, all profiles in the Agent configuration directory are loaded. To cus
...
+      OID: 1.3.6.1.4.1.232.6.2.8.1.0
+      name: cpqHeSysUtilLifeTime
 ```
 
-Then either reference it explicitly by name, or use sysObjectID detection:
+This example profile will be used on any network device where its sysobject id either **is** `1.3.6.1.4.1.232.9.4.10` or **starts with** `1.3.6.1.4.1.232.9.4.2.`.
+
+
+**Advanced**:
+In case you need different metrics for network devices that share the same sysobjectid. You can write profiles without any sysobjectid and force the Datadog Agent to use by configuring the `profile` option in the snmp configuration.
```
kayayarai, line 48:
```suggestion
If you need different metrics for network devices that share the same `sysobjectid`, you can write profiles without any `sysobjectid`, and configure the `profile` option in the SNMP configuration.
```

kayayarai (changes requested):

kayayarai (approved):

PR #23383, opened 2024-05-28
Update github.md

oltruong:
Typo in GitHub

Just fixing a typo.

### Merge instructions


jhgilbert:
Hi @oltruong, thanks for the fix! The repo is not the source of truth for non-English content, so this PR will get overwritten at some point. But I've updated the translation in the SOT as well, so the change should stick eventually. But if you see it revert in the meantime, that's what's going on.

jhgilbert (approved):

PR #23426, opened 2024-05-30
removing banner from French docs

aliciascott:
PM would like the beta and gov banner removed from the french copy of these docs, and meanwhile I'll flag this as more urgent in transifex to get the rest of the updates in sooner.
Thanks!

### Merge instructions


maycmlee (approved):

maycmlee:
You could probably try to remove the banners in transfix if you wanted to.

PR #23429, opened 2024-05-30
[SMTC-161] add http semantic mapping for otel

majorgreys:
We add a mapping table from OpenTelemetry HTTP semantic conventions to the Datadog span tag.

This is done as a separate file as the intent is to have this table sourced from another repository.

* https://opentelemetry.io/docs/specs/semconv/http/ 

### Merge instructions

### Additional notes

image


aliciascott (approved):
@majorgreys this looks good, is this ready to be merged as well ?

majorgreys:
@aliciascott I added someone from the otel team to ensure this table is accurate.

majorgreys:
@aliciascott I also wasn't totally sure about what to do with `http.response.header.<header-name>`. I think the row conveys what I am intending but is there a better way to indicate the mapping is over all `http.response.header.*` attributes?

aliciascott (approved):

layouts/shortcodes/otel-sem-http-table.md
```diff
@@ -0,0 +1,15 @@
...
+| --- | --- |
+| `client.address` | `http.client_ip` | 
+| `http.response.body.size` | `http.response.content_length` |
+| `http.response.header.<header-name>` | `http.response.headers.<header-name>` |
+| `http.response.status_code` | `http.status_code` | 
+| `http.request.body.size` | `http.request.content_length` |
+| `http.request.header.referrer` | `http.referrer` |
+| `http.request.header.<header-name>` | `http.request.headers.<header-name>` |
+| `http.request.method` | `http.method` |
+| `http.route` | `http.route` |
```
dineshg13, line 11: i don't see http.route as otel semantics in https://opentelemetry.io/docs/specs/semconv/http/

majorgreys: You can see it here https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/

majorgreys:
> Why are we not adding whole set of semantics ? Any reasons for targeting this subset.

We are focusing for now on the http semantics which we would generate from other sources.

I agree that this page should encompass all the semantic mappings.

dineshg13 (approved):

fpddog (approved):

bgdeutsch (approved):

PR #23430, opened 2024-05-30
un-bluring the IP's on images for consistency

aliciascott:
Per PM, they would prefer that the IP's are un-blurred in these 2 screenshots, since this is the demo environment, we typically only blur out of "best practices" but don't need to.

### Merge instructions


maycmlee (approved):
LGTM! I'll merge after all the checks have passed.

PR #23540, opened 2024-06-06
Update service_page.md

lee-brian:
Add new section on service health

Here's how the rendered content looks:

image

### Merge instructions


content/en/tracing/services/service_page.md
```diff
@@ -42,6 +43,32 @@ Consult on this page:
...
+|  **Alert**  |               At least one signal is active.               |               Non-muted alerting P1 monitor or monitor with a paging integration setup (PagerDuty or Opsgenie).                | Active incident of any severity | Active faulty deployment, ongoing APM latency/error rate alert. |                   --                    |
+| **Warning** | At least one signal is active and not in a critical state. | Non-muted alerting P2 monitor, warning P1 monitor, or warning monitor with a paging integration setup (PagerDuty or Opsgenie). | Stable incident of any severity |                   Ongoing log anomaly alert.                    | New (within 48 hours) issues for review |
+| **Ok** |    No signal from critical or warning state is active.     |                                                               --                                                               |               --                |                               --                                |                   --                    |
```
cswatt, line 71: I'm having a lot of trouble understanding what's happening in this table. The headers don't seem lined up in a way that makes sense. If you get a monitors alert, it means "Non-muted alerting P1 monitor or monitor with a paging integration setup (PagerDuty or Opsgenie)."? What's the verb there? I also can't tell how the "condition" column is meant to be interpreted. 

Basically I think this table needs to be reformatted, and (I think) each statement within needs to have a clear subject and verb (for example, saying "An incident of any severity is active" vs. "Active incident of any severity")—but I also don't have enough context (or understanding of the table) to be sure about this

cswatt (changes requested):
Some feedback about the status table, plus some smaller suggestions for concision/clarity

cswatt (approved):

PR #23620, opened 2024-06-11
Update Audit Trail Documentation to Include Password Reset events

dmalysza:

Updating documentation for addition of password reset event to Audit Trail as part of this [PR](https://github.com/DataDog/dogweb/pull/119733). Clarifying that the password reset Audit Trail events are emitted to all orgs that the user is active in, even if password authentication isn't enabled for that org. 

### Merge instructions

This feature is being rolled out behind a feature flag. We will merge the documentation once the feature is ready to roll out to all customers. 


content/en/account_management/audit_trail/events.md
```diff
@@ -60,7 +60,7 @@ See the [Audit Trail documentation][2] for more information on setting up and co
 | [Role created or deleted][7] | A role is created or deleted in the org. | `@evt.name:"Access Management" @asset.type:role @action:(created OR deleted)` |
 | [Role access request][8] | A user created, responded to, or deleted an access request for a role, and the value of the access request. | `@evt.name:"Access Management" @asset.type:role_request` |
 | [User's role][6] | A user is added or deleted from a role in the org. | `@evt.name:"Access Management" @asset.type:role @action:modified` |
-| [Password][9] | A user modified their password in the org. | `@evt.name:"Access Management" @asset.type:password @action:modified` |
+| [Password][9] | A user modified or reset their password in the org. Password reset events are delivered to all orgs that user is active in, even if the org does not have password authentication configured | `@evt.name:"Access Management" @asset.type:password @action:modified` |
```
tylerreinecke, line 63: I think we should have two different rows for password modification events: one for update and one for reset. I think it just reads cleanest, as "modified or reset" within a single row could be interpreted as being the same action. It is okay to have the same logs query for two different events (see support role/permission modification events).

Perhaps in the password reset row we want to specify that it happens via forgot password? (@tannerprynn-datadog is there anywhere else the password reset can be triggered?)

Don't forget the period at the end of the disclaimer sentence 😉

tylerreinecke (changes requested):

content/en/account_management/audit_trail/events.md
tannerprynn-datadog: I don't think I have a strong opinion on whether we would want to separate it into two rows or not. I think either way isn't that confusing. Maybe we can just send a quick ping to the Audit Trail PM to review our proposed language

dmalysza: If we go the two row route, I think we would want to remove the "modified" part out of the row that references the password reset event.

tylerreinecke: We still would need to keep the modified part because the audit asset itself is still being modified. https://github.com/DataDog/dogweb/pull/119733/files#diff-2eb130f48aa3d3caa4166445ddeb7a2f9c7b80e44bc13e6b60196c09eec14302R466

dmalysza: Should we then use the term 'reset' at all if we are aligning with the audit trail actions? This might cause more confusion to the end user reading these docs because we have the same modification action described in two audit trails. Maybe something like like "User modified password via a password reset" may work?

tannerprynn-datadog: Finalized language approved by Audit Trail PM on this doc: https://docs.google.com/document/d/1fHCIaHyi3wwtO3XJK49z46X1JzRhHMC2WV1NYe9qsUQ/edit

tylerreinecke: I guess it is okay, and no we should keep the term modified for the asset. The modification action refers to asset itself and what is happening to it, also IIRC there is only like `created`, `modified`, etc available, and not intended to be user defined. The reset term used in the description does not need to directly match the asset action. For example when a user is invited, we use the action created.

brett0000FF (approved):
Looks good--let me know when you'd like me to merge this. 🚀

PR #23627, opened 2024-06-11
Updated rule creator option names

michaelcretzman:
Simple rule creator and Advanced rule creation were changed to Assisted rule creator and Manual rule creator.

Simple rule creator and Advanced rule creation were changed to Assisted rule creator and Manual rule creator.

### Merge instructions


jhgilbert (approved):
These changes look fine, but just wanted to flag that "simple" and "advanced" still appear in the file as general descriptors of these tools. That's probably intentional but wanted to make sure!

PR #23694, opened 2024-06-14
probabilistic sampler requires 128 bit trace ids

ajgajg1134:

Makes it clear that the new probabilistic sampler requires 128 bit trace IDs. We recently ran into a support case where a customer was using an older version of a tracer without this support and had disconnected traces when trying to use the probabilistic sampler.
### Merge instructions


knusbaum (approved):

brett0000FF (approved):
Thanks for making this update 🚀

PR #23713, opened 2024-06-17
Update workflow iteration syntax

natashadada:
The product is a little inconsistent and the correct syntax has capitalized Index and Value. Updating the docs page to reflect that.

It's my first documentation PR - please let me know if this is not the right way to make this change!


hestonhoffman (approved):
Looks great! Thanks @natashadada

hestonhoffman:
Just sorting something out on master and then I'll merge this

PR #23716, opened 2024-06-17
Data Jobs Monitoring update

yiliangzhou:

Update Data Jobs Monitoring setup guide:
* remove beta sign up form at the top
* add missing configuration needed for k8s setup
* update product icon 
* add navigation tile for data jobs monitoring

### Merge instructions


content/en/data_jobs/kubernetes.md
```diff
@@ -65,6 +61,7 @@ You can install the Datadog Agent using the [Datadog Operator][3] or [Helm][4].
      global:
        tags:
          - 'data_workload_monitoring_trial:true'
+         - 'djm_config.enabled:true'
```
warrierr, line 64: @yiliangzhou @paul-laffon-dd  why are we adding this as a host tag instead of as an agent config setting? Don't we expect the agent config to be able to generate the metering metric?

warrierr (changes requested):

content/en/data_jobs/kubernetes.md
yiliangzhou: good catch. This is a mistake.

estherk15 (approved):
Looks good! Thanks for making those updates. Let me know when you would like this merged!

gabe-ochoa (approved):

renteria1010 (approved):

PR #23770, opened 2024-06-19
Update device_topology_map.md

FlorianVeaux:
Based on feedback from support ticket:
1. Explain that LLDP is preferred over CDP
2. Explain that the same protocol (LLDP or CDP) needs to be enabled on adjacent devices otherwise they can't discover each other.
3. Explain that the interface view is different from the NDM interface view

### Merge instructions

### DIFF
--- a/content/en/network_monitoring/devices/device_topology_map.md
+++ b/content/en/network_monitoring/devices/device_topology_map.md
@@ -27,7 +27,7 @@ The Datadog Agent version 7.52 and later automatically collects topology data. N
 
 ### Prerequisites
 
-1. Devices have LLDP (Link Layer Discovery Protocol) and/or CDP (Cisco Discovery Protocol) enabled with SNMP. The same protocol should be used on any two connected devices for them to discover each other. LLDP is generally preferred as it is a more common option.
+1. Devices have LLDP (Link Layer Discovery Protocol) and/or CDP (Cisco Discovery Protocol) enabled with SNMP. Use the same protocol on connected devices so that they can discover each other. LLDP is generally preferred as it is a more common option.
 2. Datadog Agent version 7.52 or later is installed.
 
 ## Navigation options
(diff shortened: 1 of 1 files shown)


content/en/network_monitoring/devices/device_topology_map.md
```diff
@@ -27,7 +27,7 @@ The Datadog Agent version 7.52 and later automatically collects topology data. N
 
 ### Prerequisites
 
-1. Devices have LLDP (Link Layer Discovery Protocol) and/or CDP (Cisco Discovery Protocol) enabled with SNMP.
+1. Devices have LLDP (Link Layer Discovery Protocol) and/or CDP (Cisco Discovery Protocol) enabled with SNMP. The same protocol should be used on any two connected devices for them to discover each other. LLDP is generally preferred as it is a more common option.
```
kayayarai, line 30:
```suggestion
1. Devices have LLDP (Link Layer Discovery Protocol) and/or CDP (Cisco Discovery Protocol) enabled with SNMP. Use the same protocol on connected devices so that they can discover each other. LLDP is generally preferred as it is a more common option.
```

content/en/network_monitoring/devices/device_topology_map.md
```diff
@@ -124,6 +124,7 @@ In addition to providing an overview of your network's physical connections, you
 ### Inspect
 
 Choose **Inspect** to see the device's interface connections. Click on any of the connected interfaces for further investigation.
+This view only shows the physical interfaces that are actually connected to another device. In practice this means that it shows a subset of the total set of interfaces of a network device.
```
kayayarai, line 127:
```suggestion
This view shows only the physical interfaces that are actually connected to another device. This means that it shows a subset of the total set of interfaces of a network device.
```

kayayarai (changes requested):

kayayarai (approved):

PR #23839, opened 2024-06-25
[2024 GPG keys rotation] Add the next future key inside of the documentation

Pythyu:

Following on the 2024 GPG keys rotation we generated a new future key that is going to be used in 2028. We need to update the documentation so  that any user using manual instructions is going to be ready for the next rotation in 2028.

### Merge instructions


content/fr/agent/versions/upgrade_to_agent_v6.md
```diff
@@ -1,4 +1,4 @@
----
```
chouquette, line 1: this looks like an accidental paste

content/ja/tracing/trace_collection/library_injection_local.md
```diff
@@ -1,4 +1,4 @@
----
+06462314-
```
chouquette, line 1: this also looks accidental

Pythyu: Nice catch !

drichards-87:
Created DOCS-8315 for the Docs Team editorial review.

hestonhoffman (approved):

PR #23849, opened 2024-06-25
Add documentation on how to setup tracer for chiseled containers

vandonr:
Chiseled containers contain no shell, so we cannot run the usual commands in there.

### Merge instructions

### Additional notes
I added some instructions, but they feel a little bit out of place since they are docker commands, while other instructions are just a line of bash.
If you think those instructions belong somewhere else, please advise.

I wrote a [personal blog post](https://medium.com/p/00b1db18f4b9) explaining it in more details.

### DIFF
--- a/content/en/tracing/trace_collection/automatic_instrumentation/dd_libraries/dotnet-core.md
+++ b/content/en/tracing/trace_collection/automatic_instrumentation/dd_libraries/dotnet-core.md
@@ -117,6 +117,16 @@ To install the .NET Tracer machine-wide:
    Other distributions
    : `sudo tar -C /opt/datadog -xzf datadog-dotnet-apm<TRACER_VERSION>-tar.gz && /opt/datadog/createLogPath.sh`
 
+   For chiseled or distro-less Docker image (without shell)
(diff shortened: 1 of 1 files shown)


drichards-87:
Hi @vandonr. Thanks for submitting this PR. I went ahead and created DOCS-8314. This is a small change, but your question about whether Docker commands should be mixed with Bash commands is valid. So I want to make sure that someone more familiar with the APM tracing docs takes a look at this.

brett0000FF (approved):
@vandonr - Thanks for adding this content! I agree that the original info was somewhat out of place, so I put it beneath in a new section. Please me know what you think of this change. Thanks!

vandonr:
ok, it is true that those instructions will work for all containers, but copying an empty folder instead of using `mkdir` directly is kind of a hack that is unnecessary in containers that have a shell.

brett0000FF:
> ok, it is true that those instructions will work for all containers, but copying an empty folder instead of using `mkdir` directly is kind of a hack that is unnecessary in containers that have a shell.

@vandonr - Got it. Do you think we could just rename the new section to "Chiseled containers" then to be more specific?

PR #23859, opened 2024-06-25
[reporting] update screenshots | DBNBK-121, CRP-913

jcmaunsell:
Updates stale screenshots on the scheduled reports documentation. The old screenshots included outdated information about the dashboard reporting feature.

### Merge instructions


buraizu (changes requested):
@jcmaunsell thanks for the PR. If possible, could you make a small update, so that the old images are not being deleted, and the new images are added with a different file path? For example, `static/images/dashboards/scheduled_reports/dashboard_permissions_2024.png`, with the image short codes in the documentation likewise updated to reference the new path. 

The reason for this request is that due to our translation process, deleting images results in broken images in our translated docs. Apologies for the hassle, and let me know if you have any questions!

buraizu (approved):

PR #23876, opened 2024-06-26
[CCV] Cloud cost management - Update Azure docs

crousseaux:
CCM now support creating Azure exports using the improved experience.
We are updating the docs to reflect this.

New tabs:
Screenshot 2024-06-26 at 17 57 52

### Merge instructions


crousseaux:
Alternatively, we could have something like this instead of the tabs. But I thought it was too confusing.

Screenshot 2024-06-26 at 17 42 18

goeln:
Thanks for putting this together! A few suggestions:
1) Right below the tab for Regular vs Improved exports, can we add a short note that Improved exports are in Preview and link to the docs? Something like: "The Azure improved exports experience (https://learn.microsoft.com/en-us/azure/cost-management-billing/costs/tutorial-improved-exports) is in Preview and may not be available for all customers"

2) Should we make the tabs for Regular exports vs Improved exports be at the top? Even though steps 1-3 are the same regardless, it might make it more natural to add the links

crousseaux:
@goeln unfortunately links don't work in tabs :/ so kept the old format
Screenshot 2024-07-05 at 17 40 50

goeln (approved):
looks good to me!

alai97 (approved):

PR #23915, opened 2024-06-27
[rum] Break down React Native manual instrumentation options

rtrieu:

- Breaks down the many options available for manual instrumentation of the React Native SDK into separate paragraphs instead of them being in one long snippet, as per https://datadoghq.atlassian.net/browse/DOCS-8151

### Merge instructions

### Additional notes

Not sure if there is enough detail for some of these options, so definitely open to feedback and suggestions.


MaelNamNam (approved):
LGTM! Thanks a lot!

buraizu (approved):
LGTM

PR #23933, opened 2024-06-28
Add the Node SDK to the list of Cloudcraft SDKs

jamesponddotco:
Adds the recently released Node.js SDK to the list of official Cloudcraft SDKs. Examples are not included in this PR, but will come in a future one.

### Merge instructions

### Additional notes
It seems there is an issue where the "Node" string won't show up as a tab option, probably because the language selector expects "TypeScript" or "JavaScript" instead of "Node".

@davidejones Do you know how to fix that?


buraizu:
> It seems there is an issue where the "Node" string won't show up as a tab option, probably because the language selector expects "TypeScript" or "JavaScript" instead of "Node".

My guess is that this is correct, I believe the list of valid parameters is here:
https://github.com/DataDog/documentation/blob/master/config/_default/params.yaml#L12

jamesponddotco:
> > It seems there is an issue where the "Node" string won't show up as a tab option, probably because the language selector expects "TypeScript" or "JavaScript" instead of "Node".
> 
> My guess is that this is correct, I believe the list of valid parameters is here: https://github.com/DataDog/documentation/blob/master/config/_default/params.yaml#L12

Modifying that file didn't seem to fix the issue, so maybe it's that plus something else?

jamesponddotco:
@buraizu With the change made by David, we're ready to review :)

buraizu (approved):

PR #23939, opened 2024-07-01
use install.datadoghq.com CF distribution instead of the full S3 URL

chouquette:
BARX-243

### Merge instructions

### DIFF
--- a/content/en/agent/configuration/agent-fips-proxy.md
+++ b/content/en/agent/configuration/agent-fips-proxy.md
@@ -75,7 +75,7 @@ DD_API_KEY=<DD_API_KEY> \
 DD_SITE="ddog-gov.com" \
 DD_FIPS_MODE=1 \
 bash -c "$(curl -L \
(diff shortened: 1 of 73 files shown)


Pythyu (approved):

f4usto (approved):

jhgilbert (approved):

jhgilbert:
Hi @chouquette, thanks for the PR! I wanted to flag that changes made to non-English files can be overwritten, as GitHub is not the source of truth for those files. But updating just the English files is enough, and those changes will eventually propagate through the rest. Just wanted to mention it in case you saw those non-English changes revert in the future. Thank you!

chouquette:
> Hi @chouquette, thanks for the PR! I wanted to flag that changes made to non-English files can be overwritten, as GitHub is not the source of truth for those files. But updating just the English files is enough, and those changes will eventually propagate through the rest. Just wanted to mention it in case you saw those non-English changes revert in the future. Thank you!

Duly noted, thanks a lot for the headsup!

PR #23988, opened 2024-07-02
Small update to root cert mounting instructions for PL

gsaville12321:
Specifies that `.pem` files mounted to private locations should be mounted to a `.pem` file and not directly to the `/certs` directory to clear up some confusion

### Merge instructions

- [ x] Please merge after reviewing


urseberry (approved):
Thank you!

PR #23994, opened 2024-07-03
Fix IDE name

Vomain:
There's a typo in the name of IntelliJ IDEA. That's the only place we used "IDEA" instead of "IntelliJ IDEA" as far as I can tell.

### Merge instructions


urseberry (approved):
Thank you!

PR #24022, opened 2024-07-05
Resource based sampling supported versions + private beta form link

dussault-antoine:
### Merge instructions


brett0000FF (approved):
Thanks, looks good! I pushed one commit to fix the GitHub links. They were all actually broken because they were missing `https://`.

PR #24069, opened 2024-07-10
Mention that we only support tracing for the client side of Hazelcast

vandonr:

We've had escalations recently from clients who are trying to trace their hazelcast server and don't see anything in APM.

### Merge instructions


vandonr:
The only issue we could have here is that by mentioning that we only trace the client for _this_ integration, clients could get the idea that it means we trace the server side for all other technologies mentioned on this page 🤔 
And I'm not convinced it's the case.

buraizu (approved):

buraizu:
Confirmed in local build

PR #24142, opened 2024-07-15
Add git provider to Cocoapods installs

jhgilbert:
DOCS-8402

### Merge instructions

### DIFF
--- a/content/en/logs/log_collection/ios.md
+++ b/content/en/logs/log_collection/ios.md
@@ -28,9 +28,9 @@ The `dd-sdk-ios` library supports all versions of iOS 11 or later.
 {{< tabs >}}
 {{% tab "CocoaPods" %}}
 
-You can use [CocoaPods][6] to install `dd-sdk-ios`:
+You can use [CocoaPods][6] to install `dd-sdk-ios`. For security reasons, use the Git provider and the hash of a specific commit:
 ```
-pod 'DatadogCore'
+pod 'DatadogCore', :git => 'https://github.com/DataDog/dd-sdk-ios.git', :commit => '<COMMIT_HASH>' # Example hash: '312e7c2f43'
 pod 'DatadogLogs'
 ```
 
--- a/content/en/real_user_monitoring/guide/mobile-sdk-upgrade.md
+++ b/content/en/real_user_monitoring/guide/mobile-sdk-upgrade.md
@@ -143,7 +143,7 @@ let package = Package(
   <summary>CocoaPods</summary>
 
   ```ruby
-  pod 'DatadogCore'
+  pod 'DatadogCore', :git => 'https://github.com/DataDog/dd-sdk-ios.git', :commit => '<COMMIT_HASH>' # Example hash: '312e7c2f43'
   pod 'DatadogLogs'
   pod 'DatadogTrace'
   pod 'DatadogSessionReplay'
--- a/content/en/real_user_monitoring/mobile_and_tv_monitoring/setup/ios.md
+++ b/content/en/real_user_monitoring/mobile_and_tv_monitoring/setup/ios.md
@@ -52,9 +52,9 @@ Declare the library as a dependency depending on your package manager:
 {{< tabs >}}
 {{% tab "CocoaPods" %}}
 
-You can use [CocoaPods][1] to install `dd-sdk-ios`:
+You can use [CocoaPods][1] to install `dd-sdk-ios`. For security reasons, use the Git provider and the hash of a specific commit:
 ```
-pod 'DatadogCore'
+pod 'DatadogCore', :git => 'https://github.com/DataDog/dd-sdk-ios.git', :commit => '<COMMIT_HASH>' # Example hash: '312e7c2f43'
 pod 'DatadogRUM'
 ```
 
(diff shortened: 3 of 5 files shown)


olacin-dd (approved):
Approving regarding the security perspective, thank you ! 🙇

content/en/logs/log_collection/ios.md
```diff
@@ -28,9 +28,9 @@ The `dd-sdk-ios` library supports all versions of iOS 11 or later.
 {{< tabs >}}
 {{% tab "CocoaPods" %}}
 
-You can use [CocoaPods][6] to install `dd-sdk-ios`:
+You can use [CocoaPods][6] to install `dd-sdk-ios`. For security reasons, use the Git provider and the hash of a specific commit:
 ```
-pod 'DatadogCore'
+pod 'DatadogCore', :git => 'https://github.com/DataDog/dd-sdk-ios.git', :commit => '<COMMIT_HASH>' # Example hash: '312e7c2f43'
```
maxep, line 33: /blocker We should not recommend to use the `:commit => '<COMMIT_HASH>'` as it's by-passing our release process. This could lead to customers integrating unstable version of the SDK.

I thought we were not impacted by the incident because our various podspecs specify authors, is this really necessary?

maxep (changes requested):

content/en/logs/log_collection/ios.md
maxep: cc @cedricvanrompay-datadog

olacin-dd: As outlined in the [postmortem](https://app.datadoghq.com/notebook/8970372/-draft-postmortem-ir-28706-multiple-cves-in-cocoapods-supply-chain-attack), the checksum verification of Cocoapods itself is weak (only takes the podspec into account, not the code).
There was a [vulnerability](https://evasec.webflow.io/blog/eva-discovered-supply-chain-vulnerabities-in-cocoapods#3-achieving-zero-click-account-takeover-by-defeating-email-security-boundaries) that allowed account takeover (which would impact people pulling pods from Cocoapods directly, not Git) without any interaction.
Since we cannot provide our customers a safe list of hashes for our pods, the safest way is to use Git:
```
pod 'DatadogCore' # Unsafe example
pod 'DatadogCore', :git => 'https://github.com/DataDog/dd-sdk-ios.git', :tag => '2.14.0' # better
pod 'DatadogCore', :git => 'https://github.com/DataDog/dd-sdk-ios.git', :commit => '312e7c2f43' # even better
```

If we want to stick to the release process (which makes sense obviously), then we could use the tag as a middleground ?

maxep: Thanks for the links @olacin-dd 🙏 
```ruby
pod 'DatadogCore', :git => 'https://github.com/DataDog/dd-sdk-ios.git', :tag => '<VERSION>'
```
This could work but I'm not sure we should state this in doc because:
1. It looks like the vulnerability mentioned has been [fixed](https://blog.cocoapods.org/CocoaPods-Trunk-RCEs-2023/) before the report was released.
2. We have to publish to CocoaPods trunk anyway, for react-native and flutter (which have their own podspecs and depend on ours) and any customer not using the `:git` parameter.

On the checksum verification: Our code is not hosted on CocoaPods trunk, it's only referring to our git repo, see [this example](https://github.com/CocoaPods/Specs/blob/master/Specs/d/2/d/DatadogCore/2.14.1/DatadogCore.podspec.json#L22L25). So they validate the podspec checksum before cloning the source repo, it seems as safe as the alternative proposed here. Unless I'm missing something..

cedricvanrompay-datadog: Referencing a _tag_ on our GitHub repository instead of a _commit_ should work. The only difference is that it won't protect against an attacker updating a tag to point to another revision, but this would require that the attacker gets control of our repository, that's a much stronger threat model and we can assume it won't happen here.

I still think we should advise people to point to our GitHub instead of CocoaPods Trunk (that is, to merge this PR) for every project we can distribute this way. Even if the CocoaPods team say "the problem was fixed", the number and severity of security issues they were affected by should make us prefer avoiding going through them as much as possible. Also I think now that there is Swift Package Manager (SPM), the CocoaPods project will likely receive even less maintenance than before.

For projects we _have_ to distribute through CocoaPods trunk, we should find a way for users to double-check the integrity of what they received.

jhgilbert: Thanks for the discussion, all -- it sounds like I should update these to `pod 'DatadogCore', :git => 'https://github.com/DataDog/dd-sdk-ios.git', :tag => '<VERSION>'`, is that correct?

maxep: @jhgilbert 👋 

I did further testing and this proposal alone won't work, it requires adding `git: tag:` to all our `pod` (including internal one), which will be very verbose and confusing to CocoaPods users.

Instead, I propose to make SPM (Swift Package Mangers) the first and default tab selected, with a recommendation to use SPM. Would that be possible?

jhgilbert: @maxep I think making the Swift Package Manager the default should be okay, let me follow up with the team at our planning meeting tomorrow. Thank you!

PR #24240, opened 2024-07-20
Update python.md

manchicken:
I had a really difficult time performing a simple Ubuntu upgrade because of APM logs flooding the screen. I think that having this front-and-center may help people find this work-around more quickly than I was able to.

This is a document change only

### Merge instructions

### Additional notes

None, just an FYI for the docs.


estherk15:
Thanks for your suggestion @manchicken! This is a very specific use case and I'll consult with the Logs team on how best to convey this. Thank you again for your helping us improve our documentation!

PR #24277, opened 2024-07-23
add new permissions required outside the secaudit

rmshree:
We recently added support for additional resources that require permissions outside the Security Audit policy. We want to add these to our documentation so customers can add these additional permissions when turning on Resource Collection.


buraizu (commented):
Left a suggestion for clarifying that the new permissions are optional but recommended

buraizu:
Closing for now as this was added here:
https://github.com/DataDog/documentation/pull/24286

PR #24300, opened 2024-07-24
Update the payment.failure event to match the rule

Taiki-San:
We updated the Out-of-the-Box rule to use a different event name. This PR reflect this change in the documentation. The change is backward compatible

### Merge instructions


Moutix (approved):
You can probably add payment.success as well

Taiki-San:
> You can probably add payment.success as well

We'll do so later after releasing OOTB rules targeting them

urseberry (approved):

PR #24301, opened 2024-07-24
preview build speed improvements

davidejones:

* ~Sets up a dependencies-preview makefile target so that we can skip `derefs` and load from the cache on preview~
* freeze obs pipelines generation part as we aren't showing that onsite right now (checked with May)
* update deploy to exclude .mov from html deploy and add to asset deploy (we use mp4 and this excluded a larger file)
* update build/render hugo settings for datadog & cloudcraft api sections to reduce code examples output in `/public/` dir such as .py, .go files. They are embedded in html so we don't need them.
* add pybeta,rbbeta,ts to exclude on upload same reason as above

### Merge instructions

### Additional notes

- ~Check actions catalog still exists https://docs-staging.datadoghq.com/david.jones/build-improvements/service_management/workflows/actions_catalog/~
- Check api code examples still appear https://docs-staging.datadoghq.com/david.jones/build-improvements/api/latest/


michaelcretzman (approved):

.gitlab-ci.yml
```diff
@@ -152,15 +152,14 @@ build_preview:
     - touch preview_error.log # Create log file for errors
     - exec 2> $PREVIEW_ERROR_LOG # We want to output any detected error to a log for use in notifications
     - make dependencies
-    - make vector_data
```
StefonSimmons, line 155: @davidejones do we want to remove the related `vector_data` rule and `/assets/scripts/reference-process.js` script?

davidejones: not just yet, that code is still needed and this is running on master still so just removing on preview

StefonSimmons (approved):

PR #24323, opened 2024-07-24
Automatically update Windows private location version

hestonhoffman:

Automatically updates the Windows private location version in a number of different files. I initially wanted to use the existing sdk-versions workflow, but this ended up being more complicated cos the source repo is private and it uses tags instead of versions. Ended up using the method suggested in the ticket, which was scraping the public mirror XML.

This still uses a workflow to open a pull request, so it shouldn't affect the build directly.

### Merge instructions


estherk15 (approved):

marcadev-fc (approved):
LGTM!

data/synthetics_worker_versions.json
```diff
@@ -0,0 +1,6 @@
+[
```
bgdeutsch, line 1: @hestonhoffman we may want to git ignore this file if it's going to be written to by automation.

local/bin/py/synthetics_worker_version.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python3
```
bgdeutsch, line 1: it looks like this is doing something very similar to `version_getter.py` which is referenced in the github workflow.    is this one needed as well?

hestonhoffman: Ah yeah sorry, I thought I'd deleted/renamed this!

data/synthetics_worker_versions.json
hestonhoffman: I think adding this to ignore would cause the build to fail, wouldn't it? Cos the partial reads that file. 

Similar to `data/sdk_versions.json`, I want to use the GitHub workflow to check for version updates and update this file for use with the build. I didn't want to incorporate this code into the build, cos it causes a potential failure if the network request fails for some reason.

bgdeutsch (approved):
lgtm 👍

PR #24409, opened 2024-07-31
Update list.md

sundaylaundry:

Incidents have been publicly available as a datasource for the list widget for a few months now. We recently received a customer support ticket about a customer using the Events datasource to display incidents, unaware that they can display them directly as a list widget. This simply mentions that incidents can be configured as such and the sorting capabilities.

### Merge instructions


aliciascott (approved):
looks good

PR #24410, opened 2024-07-31
Update pci-apm.md

jinatdatadog:
clarify that the traces endpoint works for non-agent traffic as well


aliciascott (approved):
good for docs

PR #24418, opened 2024-07-31
remove https redirect wording

nlopez:
There was security discussion around [this article](https://jviide.iki.fi/http-redirects) re: returning an error rather than redirecting. Let's remove this wording until we have a concrete plan and new date.


aliciascott (approved):
looks good! @nlopez is this ready to merge as well ?

simnyc (approved):

PR #24435, opened 2024-08-01
Modify script for processing yaml metric spec

dhan0779:

We add a method to process `metric-spec.yaml` files instead of the standard `csv` for generating the metric tables for docs. This is for an ongoing effort to replace the usage of csv files for integrations. We also create a method to reformat the metric spec yaml files into yaml files that are used by the `get-metric-from-git` partial.

https://datadoghq.atlassian.net/browse/AWSMC-1286

**Testing:** The `format_metric_spec_yaml()` was tested to make sure the reformatted yaml files were identical to the ones produced when generated from csvs.

### Merge instructions


rtrieu (approved):
Looks good!

PR #24487, opened 2024-08-05
(Workflows) Exclude azure network bundle

hestonhoffman:

We think an issue with the source files in the Azure Network Workflows bundle is causing the build to run out of memory. This excludes the bundle to prevent the issue. Long-term, we'd like to stop generating these files and instead deep link to each action in the app.

I tested this on an older dd-source commit, but once a new PR is up, we should point this PR it at that branch instead for testing.

### Merge instructions

### DIFF
--- a/local/bin/py/build/configurations/pull_config_preview.yaml
+++ b/local/bin/py/build/configurations/pull_config_preview.yaml
@@ -1,6 +1,6 @@
 ---
 - config:
-    cache_enabled: true
+    cache_enabled: false
 
 - data:
     - org_name: jenkinsci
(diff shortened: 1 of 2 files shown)


devindford (approved):
Tested against the breaking change from DD-Source and the build runs as expected and excludes the `azure network` results from the workflows automation actions catalog.

Tested on a separate branch updating the pointer to the PR to reintroduce azure network to the manifest.json file

https://gitlab.ddbuild.io/DataDog/documentation/-/jobs/595842734
https://docs-staging.datadoghq.com/devin.ford/test-python-update/service_management/workflows/actions_catalog/?search=azure

local/bin/py/build/configurations/pull_config_preview.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 - config:
-    cache_enabled: true
+    cache_enabled: false
```
devindford, line 3:
```suggestion
    cache_enabled: true
```

maycmlee (approved):

PR #24514, opened 2024-08-07
CCM add text on cost monitor evaluation delay period

dave-handy:
There was a recent support issue caused because this delay is not documented or otherwise clear to end users. They were comparing the values reported by the monitor to their own dashboards and metric explorer. This is also important if they use templated links to dashboards in their monitor template.

### Merge instructions


crousseaux (approved):
thanks, lgtm

buraizu (approved):

PR #24538, opened 2024-08-07
DOCS-7960 - secrets for dbm

cswatt:
Previously, the DBM docs defaulted to providing passwords in plaintext. After some feedback, it was determined that the docs should default to referencing these passwords as secrets with the `ENC[]` syntax.

This PR:
 - adds a "store your passwords in a secret management solution" setup step to all pages
 - makes config file snippets (not when stuff is passed through command line) use `ENC[]` by default
 - removes sections that had said "you can also use secrets" that make it sound optional

### Merge instructions


cswatt:
Added DOCS-8658 to review because this is a lot

layouts/shortcodes/dbm-secret.md
```diff
@@ -0,0 +1,3 @@
+Store your password using secret management software such as [Vault](https://www.vaultproject.io/). You can then reference this password as `ENC[<SECRET_NAME>]`: for example, `ENC[datadog_user_database_password]`. See [Secrets Management](/agent/configuration/secrets-management/) for more information.
```
jhgilbert, line 1: It's probably worth being specific about where you can use the `ENC` reference -- I'm guessing that's just in the Agent configuration files?

jhgilbert (approved):
Approved with one suggestion, thanks!

ynnadkrap (approved):

PR #24649, opened 2024-08-14
[web-5187] Workflows automation refactoring

bgdeutsch:
refactors how we pull in the `workflows` rules for the action catalog page.

### Merge instructions
- [] Please merge after docs/websites reviews

### Additional notes
preview here


assets/scripts/components/grouped-item-listings.js
```diff
@@ -116,11 +116,12 @@ export function initializeGroupedListings() {
...
         const activeCategoryFilter = (activeCategory) ? activeCategory.dataset.filter : '';
 
-        if(searchValue){
+        if (searchValue) {
             url.searchParams.set('search', searchValue)
-        }else{
+        } else {
             url.searchParams.delete('search')
         }
+
```
StefonSimmons, line 124: @bgdeutsch the search functionality is not working the same way. on live, you can search by the  action name ([data-name]). something probably needs to be updated here

StefonSimmons: it could also be something int he actioncatalog/list.html

local/bin/py/build/configurations/pull_config_preview.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 - config:
-    cache_enabled: true
+    cache_enabled: false
```
StefonSimmons, line 3:
```suggestion
    cache_enabled: true
```

hestonhoffman (approved):
Local build works great (and is much faster 🎉 ). Sandra also noticed the search issue that Stefon pointed out, but everything else looks good!

assets/scripts/components/grouped-item-listings.js
hestonhoffman: Just confirming that we'd like the search to work for action names

bgdeutsch: @hestonhoffman @StefonSimmons this is fixed thank you both for flagging!

StefonSimmons (approved):

assets/scripts/components/grouped-item-listings.js
neko-dd: Should this be fixed in the staging preview as well? I still see the issue there

PR #24674, opened 2024-08-15
SaaS Cost Integrations Edit

alai97:

Adds a tile in the Overview section for Elastic Cloud (one of the SaaS Cost Integrations that we just added). Updates the in-app integration tile link + updates the left navigation menu to include Elastic Cloud.

Chat with Andrea on Slack

### Merge instructions

### Additional notes

https://docs-staging.datadoghq.com/alai97/saas-cost-integrations-add-ec-tile/cloud_cost_management/saas_costs?tab=databricks#overview


csantosdd (approved):
@alai97 just curious it feels like this section should be `range`, or is it necessary for it to repeat and explicitly call each img partial?

doesn't block merge though 👍

aliciascott (approved):
Approved with comment in slack :)

PR #24703, opened 2024-08-16
Code Analysis Documentation Edit

alai97:

No content is being changed in this update. We are shuffling the header sections around to show users the Setup instructions first (like setting up the repository, setting up GitHub integration, using IDE integrations, and then looking at the scan results) as opposed to what it is now (in a different order).

> What I'm hearing is that customers aren't finding info on DD hosted scans, so maybe find a way to push it upward in the page, or put it in a heading?

Chat with Kari on Slack

### Merge instructions


aliciascott (approved):
looks good

PR #24720, opened 2024-08-19
[ASCII-2204] Update the fips-proxy version to latest 1.1.4 release in docs

jeremy-hanna:

fips-proxy has [released 1.1.4](https://github.com/DataDog/fips-proxy/releases/tag/1.1.4) and this PR updates the respective documentation to mention that version instead of 1.1.3

### Merge instructions


michaelcretzman (approved):
@jeremy-hanna in the future, you only need to update the English files under the `content/en` directory, not the other language files under the other content folders (`es`, `fr`, etc.). Our translators will cover those. 

For now, it looks like you only updated a version number. So we should be okay.

PR #24721, opened 2024-08-19
(DOCS-8744) - document App Builder tab and text area components

neko-dd:

* Documents tab and text area components on Components page
* Adds related content to Events page
* Adds event reaction item to misc places on Components page where there is a new option

### Merge instructions


kc-beard (approved):
approving to unblock but I think we should remove `on, off` and clarify that they are booleans

brett0000FF (approved):
🚀

PR #24771, opened 2024-08-20
(Feat) Allow inline images

hestonhoffman:

Sometimes, writers want to use small inline images to demonstrate interface buttons or signpost their text. We've been doing this using raw HTML image tags, but there are some issues:
- The images aren't processed by Hugo, so they don't get the same fingerprinting (etc.) that is applied to the rest of our images.
- Because Hugo doesn't process the images, they don't show up in preview builds. They seem to work fine in prod and local builds.
- It's better to be consistent and not rely on workarounds.

Example PR of one of these not rendering in previews: https://docs-staging.datadoghq.com/neko/DOCS-8757-workflows-screenshots/service_management/workflows/build/#add-an-error-path

Preview of working inline images: https://docs-staging.datadoghq.com/heston/inline-images/tracing/guide/#tutorials-enabling-tracing

Note:
- This implementation doesn't use alt text, cos I think it looks confusing if the image is missing for some reason and the alt text shows up in the middle of a sentence. When authors use these, they're usually adding enough context for the sentence to work without the image. We'll document some guidelines around using these.
- To get this to work, all I really did was remove the div around inline images. This means there's no styling applied besides stuff that's assigned in the shortcode. This didn't seem like an issue, seeing as though we've already been using images like this.
- The Go template is much easier to review without whitespace:
Screenshot 2024-08-21 at 9 36 09 AM

### Merge instructions


content/en/service_management/workflows/build.md
```diff
@@ -190,7 +190,7 @@ You can add an error path for the workflow to follow if it encounters an error.
 
 To add an error path:
 1. Hover over the step where you'd like to add an error path.
-1. Click and drag the **Error path** icon to place a new error path on the canvas.
+1. Click and drag the **Error path** icon {{< img src="service_management/workflows/error-path-icon.png" alt="Your image description" style="width:24px;" inline="True" >}} to place a new error path on the canvas.
```
hestonhoffman, line 193: Just for demo purposes.

jhgilbert (approved):

bgdeutsch (approved):
LGTM 👍

PR #24821, opened 2024-08-22
[DOCS-8298] Add Obs Pipelines destination batch settings

maycmlee:

Adds info on destination batch settings.

[DOCS-8298](https://datadoghq.atlassian.net/browse/DOCS-8298)

### Merge instructions

### Additional notes

[DOCS-8298]: https://datadoghq.atlassian.net/browse/DOCS-8298?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


content/en/observability_pipelines/destinations/_index.md
```diff
@@ -29,7 +29,43 @@ Select and set up your destinations when you [set up a pipeline][2]. This is ste
     {{< nextlink href="observability_pipelines/destinations/#rsyslog-or-syslog-ng" >}}Rsyslog or Syslog-ng{{< /nextlink >}}
     {{< nextlink href="observability_pipelines/destinations/#splunk-http-event-collector-hec" >}}Splunk HTTP Event Collector (HEC){{< /nextlink >}}
     {{< nextlink href="observability_pipelines/destinations/#sumo-logic-hosted-collector" >}}Sumo Logic Hosted Collector{{< /nextlink >}}
-{{< /whatsnext >}}
+{{< /whatsnext >}} 
+
+### Batch settings {#batch-settings-intro}
+
+Observability Pipelines destinations batch events sent to the downstream integration. The following parameters define each destinations' batch settings:
```
neuronull, line 36: I think ...

```suggestion
Observability Pipelines destinations batch events sent to the downstream integration. The following parameters define each destination's batch settings:
```

content/en/observability_pipelines/destinations/_index.md
```diff
@@ -29,7 +29,43 @@ Select and set up your destinations when you [set up a pipeline][2]. This is ste
...
+
+These are the batch settings for each destination:
+
+| Destination                                | Max Events     | Max Bytes       | Timeout (seconds)   |
+|--------------------------------------------|----------------|-----------------|---------------------|
+| Amazon OpenSearch                          | None           | 10,000,000      | 1                   |
+| Amazon S3 (Datadog Log Archives)           | None           | 100,000,000     | 900                 |
+| Azure Storage (Datadog Log Archives)       | None           | 100,000,000     | 900                 |
+| Datadog Logs                               | 1,000          | 4,250,000       | 5                   |
+| Datadog Metrics                            | 100,000        | None            | 2                   |
```
neuronull, line 60: We should omit this one for now because it isn't released yet. But I guess we'll need to periodically come back to this table to add others as we add destinations.

content/en/observability_pipelines/destinations/_index.md
```diff
@@ -43,6 +79,14 @@ Set up the Amazon OpenSearch destination and its environment variables when you
 
 {{% observability_pipelines/configure_existing_pipelines/destination_env_vars/amazon_opensearch %}}
 
+### Batch settings
```
neuronull, line 82: For all of the below sections where the batch settings is being mentioned on a per component basis: I think we should refrain from mentioning them there. 🤔  Perhaps we should mention only in one place and not both ?

neuronull (commented):
I'm curious if Product/Design might have input on the usage of the term "settings" in these changes.

I used that term in the internal confluence page because in the sink configuration, they are truly settings. But in OP we do not expose them as something the user can change. So I'm wondering if we should ajust the terminology and just call it "Batching behavior". , and remove/adjust any language that suggests it _could_ be configurable.

content/en/observability_pipelines/destinations/_index.md
maycmlee: Ah yes, good point. Removed, but noted in the docs draft for to add batch settings.

content/en/observability_pipelines/destinations/_index.md
maycmlee: @neuronull after the Logs-to-metrics template comes out (and I'm assuming there is going to be a Datadog metrics destination for it that is similar to the Datadog Logs destination), I'm going to make each destination it's own page in the docs (like the [Amazon S3 archive](https://docs.datadoghq.com/observability_pipelines/destinations/amazon_s3/?tab=docker) destination), so that's why I put them here. But I'm fine with removing it for this first iteration, because it does seem quite redundant.

Just wondering if you have thoughts about adding the batch settings for each destination once they have their own doc?

maycmlee: I just added the batch settings for the Amazon S3 destination doc as an example.
Preview: https://docs.datadoghq.com/observability_pipelines/destinations/amazon_s3/?tab=docker#batch-settings-intro

Commit: https://github.com/DataDog/documentation/pull/24821/commits/8a52d29c729e2d674dae31c913a381be6193cade

neuronull: > and I'm assuming there is going to be a Datadog metrics destination for it that is similar to the Datadog Logs destination

correct, there will be one

neuronull: > Just wondering if you have thoughts about adding the batch settings for each destination once they have their own doc?

If they each have their own page, yeah makes sense to me. Then each page has its own batch parameters, and a link back to the main description of the batching behavior. Makes sense. (and the main page describing batching behavior wouldn't have the table that contains everything).

So I guess, maybe that could be a solution for this PR? add the individual parameters in each section but not have the large table. WDYT ?

maycmlee: > So I guess, maybe that could be a solution for this PR? add the individual parameters in each section but not have the large table. WDYT ?

@neuronull Yes, I agree, that seems like the better solution here. I kept the individual parameters, and removed the large table.

> I used that term in the internal confluence page because in the sink configuration, they are truly settings. But in OP we do not expose them as something the user can change. So I'm wondering if we should ajust the terminology and just call it "Batching behavior". , and remove/adjust any language that suggests it could be configurable.

I forgot to respond to this earlier. That's a good point and makes sense. I changed the header from `Batch settings` to `Event batching`, and used `batching behavior` or `batching parameters` in the explanation. Does that work?

Commit: https://github.com/DataDog/documentation/pull/24821/commits/64d9e4b2ae6ae6fdb89122d711a13ca678447c8b

And a little rewording: https://github.com/DataDog/documentation/pull/24821/commits/c9351402af51ba8630b2df72f9a21153f9de52da

> I'm curious if Product/Design might have input on the usage of the term "settings" in these changes.

I'll check in with Pratik about all of this when he's back.

neuronull: Yeah, looks great!

neuronull (approved):

content/en/observability_pipelines/destinations/_index.md
maycmlee: Thanks for the review and feedback!

neuronull: yw!

content/en/observability_pipelines/destinations/_index.md
```diff
@@ -34,6 +34,26 @@ Select and set up your destinations when you [set up a pipeline][2]. This is ste
...
+Observability Pipelines destinations send events in batches to the downstream integration. A batch of events is flushed when one of the following parameters is met:
+
+- Maximum number of events
+- Maximum number of bytes
+- Timeout (seconds)
+
+For example, if a destination's parameters are:
+
+- Maximum number of events = 2
+- Maximum number of bytes = 10,0000
```
buraizu, line 48:
```suggestion
- Maximum number of bytes = 100,000
```

content/en/observability_pipelines/destinations/_index.md
```diff
@@ -34,6 +34,26 @@ Select and set up your destinations when you [set up a pipeline][2]. This is ste
...
+- Timeout (seconds) = 5
+
+And the destination receives 1 event in a 5-second window, it flushes the batch at the 5-second timeout.
+
+If the destination receives 3 events within 2 seconds, it flushes a batch with 2 events and then flushes a second batch with the remaining event after 5 seconds. If the destination receives 1 event that is more than 10,0000 bytes, it flushes this batch with the 1 event.
```
buraizu, line 53:
```suggestion
If the destination receives 3 events within 2 seconds, it flushes a batch with 2 events and then flushes a second batch with the remaining event after 5 seconds. If the destination receives 1 event that is more than 100,000 bytes, it flushes this batch with the 1 event.
```

content/en/observability_pipelines/destinations/_index.md
buraizu: Not sure if it's better to move the comma or delete a zero here. Feel free to update as needed.

buraizu (approved):
Left a couple of suggestions, but overall LGTM

content/en/observability_pipelines/destinations/_index.md
maycmlee: Thanks for catching that! Moving the comma is correct. Committed you suggestion.

PR #24830, opened 2024-08-22
DOCS-8724 CD Visibility Partial Update

alai97:

Refactors the CD Visibility partial and adds some spacing in between the tiles.

### Merge instructions

### Additional notes

https://docs-staging.datadoghq.com/alai97/update-cd-partial/continuous_delivery

### DIFF
--- a/content/en/continuous_delivery/deployments/ciproviders.md
+++ b/content/en/continuous_delivery/deployments/ciproviders.md
@@ -26,7 +26,7 @@ CD Visibility for CI providers deployments is in private beta. To request access
 If you are executing deployments using a CI provider, you can monitor your deployments with Deployment Visibility. Ensure the following requirements are met:
 
 1. You are using [Pipeline Visibility][1] to monitor your CI pipelines.
-2. Your CI provider supports the **Custom tags and metrics at runtime** feature, which allows you to add [user-defined text and numerical tags][2] to pipelines and jobs in Pipeline Visibility. 
+2. Your CI provider supports the **Custom tags and measures at runtime** feature, which allows you to add [user-defined text and numerical tags][2] to pipelines and jobs in Pipeline Visibility. 
 3. You are executing deployments in a CI job (or a [related level][9] in your CI provider). The concept of a CI job may vary depending on your provider. For more information about how Datadog defines a CI job, see the [Terminology section][9].
 
 ## Setup
--- a/layouts/partials/continuous_delivery/cd-getting-started.html
+++ b/layouts/partials/continuous_delivery/cd-getting-started.html
@@ -1,56 +1,57 @@
-{{ $dot := . }}
-<div class="ci-cd-integrations">
+{{ $root := . }}
+{{ $providers := slice
+    (dict "href" "/continuous_delivery/deployments/argocd/" "src" "integrations_logos/argocd_small.svg" "alt" "argo cd")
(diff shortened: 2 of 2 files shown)


hestonhoffman (commented):
The Argo tile works, but the other tiles don't go to the listed providers and there aren't pages for them in the sidenav. Is that expected?

alai97:
> The Argo tile works, but the other tiles don't go to the listed providers and there aren't pages for them in the sidenav. Is that expected?

@hestonhoffman Yes, for now. We have a catch-all page for CI Providers in [CD Visibility](https://docs.datadoghq.com/continuous_delivery/deployments/ciproviders/) instead of individual pages. For now, ArgoCD is the only one with a [dedicated doc page](https://docs.datadoghq.com/continuous_delivery/deployments/argocd) for its setup.

hestonhoffman (approved):
> > The Argo tile works, but the other tiles don't go to the listed providers and there aren't pages for them in the sidenav. Is that expected?
> 
> @hestonhoffman Yes, for now. We have a catch-all page for CI Providers in [CD Visibility](https://docs.datadoghq.com/continuous_delivery/deployments/ciproviders/) instead of individual pages. For now, ArgoCD is the only one with a [dedicated doc page](https://docs.datadoghq.com/continuous_delivery/deployments/argocd) for its setup.

Ah, I see. I think the user experience here is a little confusing. Perhaps you could mention the products you're linking from at the top of the page somewhere? I just noticed this is in private beta, so perhaps it's not that big a deal - especially if dedicated pages are on the way for the rest. I'll go ahead and approve and let y'all be the judge of that.

alai97:
> > > The Argo tile works, but the other tiles don't go to the listed providers and there aren't pages for them in the sidenav. Is that expected?
> > 
> > 
> > @hestonhoffman Yes, for now. We have a catch-all page for CI Providers in [CD Visibility](https://docs.datadoghq.com/continuous_delivery/deployments/ciproviders/) instead of individual pages. For now, ArgoCD is the only one with a [dedicated doc page](https://docs.datadoghq.com/continuous_delivery/deployments/argocd) for its setup.
> 
> Ah, I see. I think the user experience here is a little confusing. Perhaps you could mention the products you're linking from at the top of the page somewhere? I just noticed this is in private beta, so perhaps it's not that big a deal - especially if dedicated pages are on the way for the rest. I'll go ahead and approve and let y'all be the judge of that.

We do, the sentence before the partial links to both pages: "Datadog integrates with [CI providers](https://docs.datadoghq.com/continuous_delivery/deployments/ciproviders) and CD providers like [Argo CD](https://docs.datadoghq.com/continuous_delivery/deployments/argocd) to collect metrics and track the execution performance and results of your deployments."

hestonhoffman:
Yeah, I read it. But I suspect that nuance might be lost on most readers 😅 Cos like, there's a big button that says GitLab, and my lizard brain wants to click it and be sent to something about GitLab :)

LisianeTu (approved):

alai97:
@hestonhoffman Based on your feedback and some discussion with folks on the team, I think we can remove the CI provider from the tile selector and have ArgoCD as the only option:

Screenshot 2024-08-27 at 2 26 48 PM

Like so, what do you think?

hestonhoffman:
@alai97 That's definitely better! I kind of wonder whether you even need that logo, cos you have the two links in the paragraph before that already do the job for you. But again, I think this is a big improvement from a user perspective!

alai97:
@hestonhoffman  has confirmed via Slack that we are good to move forward with htis

PR #24886, opened 2024-08-23
[web-5210] remove sections from docs search links

bgdeutsch:
https://datadoghq.atlassian.net/browse/WEB-5210

Currently we show search results down to the section which doesn't always make sense.  For example, [this search](https://docs.datadoghq.com/search/?s=azure) for `Azure`, it's leading the user to the `Troubleshooting` section which doesn't make sense based on their search. 
 Discussed w/ the team and we felt like these results are likely providing more confusion than value to the customer.

This PR is removing all section results in favor of showing full page results in the UI to remove any confusion.

### Merge instructions
- [x] Please merge after websites reviewing

### Additional notes
https://docs-staging.datadoghq.com/brian.deutsch/web-5210-section-results/search/?s=azure
https://docs-staging.datadoghq.com/brian.deutsch/web-5210-section-results/search/?s=unified%20service%20tags

### DIFF
--- a/assets/scripts/components/algolia/getHitData.js
+++ b/assets/scripts/components/algolia/getHitData.js
@@ -4,7 +4,7 @@ import { truncateContent, truncateContentAtHighlight } from "../../helpers/trunc
 export function getHitData(hit, searchQuery = '') {
     const title = hit.title ? hit.title : hit.type;
     const cleanRelPermalink =
-        hit.language == 'en' ? hit.relpermalink : hit.relpermalink.replace(`/${hit.language}/`, '');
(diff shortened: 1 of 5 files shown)


brett0000FF (approved):
Thanks! I compared and contrasted a few searches between the preview and live site. The updated approach seems to work better. 👍

brett0000FF:
@bgdeutsch - Hey! I tested a few more, and I'm wondering if this is expected:

image

Notice that it's pulling back results from the further reading section. I'd expect this to return results from the first few sentences on the Continuous Profiler page instead?

stanleegoodspeed (approved):

PR #24951, opened 2024-08-28
Switch additionalConfigs to string

RLProteus:

Corrects example in documentation to use string type. This matches the example found [here](https://github.com/DataDog/datadog-operator/blob/main/examples/datadogagent/datadog-agent-with-prometheus-autodiscovery-advanced-config.yaml#L10).

### Merge instructions


buraizu (approved):
Thanks @RLProteus!

PR #25020, opened 2024-09-02
Update Timeshift docs with "before" functions deprecation

brebernjak:

Reflecting some product changes in the documentation:
- calendar_timeshift is now available everywhere
- explain better the arguments and their possible values
- reorder the page content to make calendar_timeshift more visible
- "before" functions are being deprecated (except hour_before)

### Merge instructions

### DIFF
--- a/content/en/dashboards/functions/timeshift.md
+++ b/content/en/dashboards/functions/timeshift.md
@@ -30,7 +30,7 @@ timeshift(avg:system.load.1{*}, -1209600)
 |:-------------------|:----------------------------------------------------------------------------------------------|:-----------------------------------|
 | `calendar_shift()` | Graph values from the previous day, week, or month from the current timestamp for the metric. | `calendar_shift(<METRIC_NAME>{*}, "<TIME_SHIFT_STRING>", "<TIME_ZONE_CODE>")` |
 
-To access the `calendar_shift()` function click the **Add function** button, select **Timeshift > Month before**. The calendar shift allows you to compare the same metric across equivalent timeframes. Here is an example of cloud cost metric `aws.cost.net.amortized` with the calendar_shift() value from two weeks ago compared to the current value.
+To access the `calendar_shift()` function click the **Add function** button, select **Timeshift > Month before**. The calendar shift allows you to compare the same metric across equivalent time frames. Below is an example of cloud cost metric `aws.cost.net.amortized` with the calendar_shift() value from two weeks ago compared to the current value.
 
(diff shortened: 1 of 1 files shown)


content/en/dashboards/functions/timeshift.md
```diff
@@ -22,6 +22,24 @@ For example, if you wanted to use this to compare current system load with load
 timeshift(avg:system.load.1{*}, -1209600)
 ```
 
+## Calendar shift
+
+<div class="alert alert-info">The calendar shift feature is now available for all data sources.</div>
```
rtrieu, line 27:
```suggestion
<div class="alert alert-info">The calendar shift feature is available for all data sources.</div>
```

content/en/dashboards/functions/timeshift.md
```diff
@@ -22,6 +22,24 @@ For example, if you wanted to use this to compare current system load with load
...
+
+| Function           | Description                                                                                   | Example                            |
+|:-------------------|:----------------------------------------------------------------------------------------------|:-----------------------------------|
+| `calendar_shift()` | Graph values from the previous day, week, or month from the current timestamp for the metric. | `calendar_shift(<METRIC_NAME>{*}, "<TIME_SHIFT_STRING>", "<TIME_ZONE_CODE>")` |
+
+To access the `calendar_shift()` function click the **Add function** button, select **Timeshift > Month before**. The calendar shift allows you to compare the same metric across equivalent timeframes. Here is an example of cloud cost metric `aws.cost.net.amortized` with the calendar_shift() value from two weeks ago compared to the current value.
```
rtrieu, line 33:
```suggestion
To access the `calendar_shift()` function click the **Add function** button, select **Timeshift > Month before**. The calendar shift allows you to compare the same metric across equivalent time frames. Below is an example of cloud cost metric `aws.cost.net.amortized` with the calendar_shift() value from two weeks ago compared to the current value.
```

rtrieu (changes requested):
A few minor feedback but otherwise looks good!

content/en/dashboards/functions/timeshift.md
brebernjak: The idea behind this banner is to highlight that this support is new, because earlier this was supported only for some data sources. We would be removing the banner after a while.
if you think "now" shouldn't be here, I'd prefer not having a banner at all, as it should be assumed that everything works for every data source, unless stated differently.

rtrieu: Ah then yes, I think in this case we should remove this banner.

rtrieu (approved):
LGTM

PR #25026, opened 2024-09-03
Fix jumping between Metrics and OTel docs

brett0000FF:

- Feedback that the current linking setup was confusing.
- Use include full page markdown shortcode to prevent teleporting and have more specific breadcrumbs.

### Merge instructions


config/_default/menus/main.en.yaml
```diff
@@ -1707,7 +1707,7 @@ menu:
       identifier: dev_tools_metrics_api
       weight: 106
     - name: Send OpenTelemetry Metrics to Datadog
-      url: opentelemetry/otel_metrics/
```
jhgilbert, line 1710: Do you need to add an alias for these two deleted URLs?

brett0000FF: Good point! Thanks for catching. I added these here: 314e6ca9068f906c66146668e2bbd8213616a58f.

Note, I didn't add an alias for the Metrics Types page. Previously, the menu file pointed to the same page from two different locations. Now, if you used the old URL, it'd just take you to where you'd expect to go.

jhgilbert (approved):

PR #25047, opened 2024-09-04
Use correct sample rate range in RUM docs

karllindmark:
Sample rate is specified as a range between 0 to 1 in the best practice article, but the actual code and documentation specifies it as a range between 0 to 100.

This change aligns the numbers to be consistent.

### Merge instructions

### Additional notes
N/A

### DIFF
--- a/content/en/real_user_monitoring/guide/best-practices-for-rum-sampling.md
+++ b/content/en/real_user_monitoring/guide/best-practices-for-rum-sampling.md
@@ -21,7 +21,7 @@ This guide walks you through best practices for RUM sampling so you can capture 
 Sessions are randomly sampled based on the percentage listed in the [SDK configuration][1]. To that end, make sure to use the correct configuration variable names for the SDK version being used.
 
 ### Configure the sampling rate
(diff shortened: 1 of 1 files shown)


karllindmark:
Wait, maybe I'm misunderstanding this completely – does it _actually_ draw a number between 0 and 1, while the user is able to specify a number in the range of 0 to 100? 😅

drichards-87 (approved):
Thanks for submitting this PR @karllindmark. It looks good from a docs perspective, but I've asked the RUM team to take a look as well.

thomas-lebeau (approved):
Thank you @karllindmark, I think your proposal makes sense, otherwise, the session would always be kept.

>  does it actually draw a number between 0 and 1, while the user is able to specify a number in the range of 0 to 100?

Here is the [related code](https://github.com/DataDog/browser-sdk/blob/02e5c3a28618d022f5e81573a65103a37a13d79c/packages/core/src/tools/utils/numberUtils.ts#L6). Technically the random number generated is between 0 and 1 and it is then multiplied by 100 to compare things of the same scale. I think it's fair to think of the whole formula as the random number generated, hence your change correct.

PR #25089, opened 2024-09-05
[test-visibility] do not offer auto instrumentation for cypress

ManuelPalenzuelaDD:
Cypress does not support autoinstrumentation, as it does not work with NODE_OPTIONS, so we are adding a note for this.

### Merge instructions

Screenshot 2024-09-05 at 18 39 10


juan-fernandez (commented):
could we make it a bit more prominent? We can do that with html, like we do in here:

https://github.com/DataDog/documentation/blob/master/content/en/tests/setup/javascript.md?plain=1#L616-L618

PR #25103, opened 2024-09-05
kevin_gleeson/Update azure-logging-guide.md

teapuddles:
Customer from this ticket (https://datadog.zendesk.com/agent/tickets/1832610) faced a few pitfalls due to the following not mentioned in the docs: 

- They were attempting to collect logs from an Azure App Service application, but we didn't state that the Event Hub would need to be created in the same region as the Application they were looking to log from. 

- They also got caught up with the DD_SITE param in the Function Code section (lines 189 - 195). We do mention that they would need to update this param to their respective DD_SITE, so I didn't make any changes here. However, it would be nice if we could add function code versions to our github which are already scoped to the each available site. 

### Merge instructions

- [ X] Please merge after reviewing

### Additional notes
I'll likely make an additional doc update request for the bit about the function code. This is definitely not the first time I've ran into customers facing this problem, but we also make it clear that they would need to update this parameter. The addition of per-site function code would just be a QoL update for customers setting this up.

### DIFF
--- a/content/en/logs/guide/azure-logging-guide.md
+++ b/content/en/logs/guide/azure-logging-guide.md
@@ -127,9 +127,10 @@ The instructions below walk through a basic, initial setup using the Azure Porta
 If you already have an Event Hubs namespace configured with an Event Hub connection string, skip to [Add an Event Hub to your Event Hubs namespace](#add-an-event-hub-to-your-event-hubs-namespace).
 
 1. In the Azure portal, navigate to the [Event Hubs][208] overview and click **Create**.
-2. Fill in the **Project Details** and **Instance Details** sections as desired.  
+2. For logging from an Azure App Service, ensure that the Event Hub is created in the same region as the application. 
+3. Fill in the **Project Details** and **Instance Details** sections as desired.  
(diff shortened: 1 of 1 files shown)


buraizu:
Hi @teapuddles, thanks for letting us know about this and opening a PR. After looking into this a bit further, I've opted instead to add the following note in two places on the doc:
`Note: Resources can only stream to Event Hubs in the same Azure region.`

One is in the `Automated installation` tab, and the other in the `Manual installation` tab. I went with this approach for the following reasons:
- It places the information a bit further up in the doc
- It extends the information to the automated installation instructions
- It uses more general language to clarify that the restriction applies to all resources, not just Azure App Services

As such, I'll close this PR for now. If you have any additional suggestions for improvements, please feel free to open a new PR or reach out to the documentation team in Slack.

As for the function code itself, this is maintained in the `datadog-serverless-functions` repository. Based on the [CODEOWNERS file](https://github.com/DataDog/datadog-serverless-functions/blob/master/.github/CODEOWNERS#L2), it looks like the Azure integrations team may be able to help with that.

Thanks again for helping improve the docs!

PR #25204, opened 2024-09-11
Changes to Tracing Tomcat as a Windows Service

Jungwon1127:
The previous changes suggested were not the recommended way to pass environment variables to Tomcat according to the documentation here: https://github.com/apache/tomcat/blob/main/bin/catalina.bat which says:
WHEN RUNNING TOMCAT AS A WINDOWS SERVICE:
...
The configuration that controls Windows Services is stored in the Windows Registry, and is most conveniently maintained using the "tomcat@VERSION_MAJOR@w.exe" maintenance utility.

I tried my best to format the instructions as previously done but feel free to edit so it is more inline with how we write documentation.

Current instructions on how to trace tomcat services running as Windows Services are incorrect

### Merge instructions

### DIFF
--- a/content/en/tracing/trace_collection/automatic_instrumentation/dd_libraries/java.md
+++ b/content/en/tracing/trace_collection/automatic_instrumentation/dd_libraries/java.md
@@ -117,7 +117,7 @@ To enable tracing when running Tomcat on Linux:
 To enable tracing when running Tomcat as a Windows service:
 
 1. Open the "tomcat@VERSION_MAJOR@w.exe" maintenance utility located in the `./bin` directory of the Tomcat project folder.
-2. Navigate to the Java tab, and add the following to `Java Options`:
+2. Navigate to the **Java** tab, and add the following to `Java Options`:
 ```text
 -javaagent:C:\path\to\dd-java-agent.jar
 ```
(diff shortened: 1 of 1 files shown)


content/en/tracing/trace_collection/automatic_instrumentation/dd_libraries/java.md
```diff
@@ -116,27 +116,12 @@ To enable tracing when running Tomcat on Linux:
...
-To enable tracing when running Tomcat with an environment setup script:
-
-1. Create `setenv.bat` in the `./bin` directory of the Tomcat project folder, if it doesn't already exist.
-1. Add the following to `setenv.bat`:
-   ```text
-   set CATALINA_OPTS=%CATALINA_OPTS% -javaagent:"c:\path\to\dd-java-agent.jar"
-   ```
-If the previous step doesn't work, try adding the following instead:
+1. Open the "tomcat@VERSION_MAJOR@w.exe" maintenance utility located in the `./bin` directory of the Tomcat project folder.
+2. Navigate to the Java tab, and add the following to `Java Options`:
```
brett0000FF, line 120:
```suggestion
2. Navigate to the **Java** tab, and add the following to `Java Options`:
```

brett0000FF (approved):
Thanks for the fix! Approved with one small formatting suggestion.

PR #25271, opened 2024-09-17
Update note on log status remapper

agashjeya:

The initial note didn't clearly explain that if multiple pipelines match a log, only the first status remapper across all applicable pipelines is applied. The updated note clarifies this behavior both within a single pipeline and across multiple pipelines, ensuring users have a correct understanding of how log status remappers are processed.

### Merge instructions


content/en/logs/log_configuration/processors.md
```diff
@@ -163,7 +163,7 @@ Each incoming status value is mapped as follows:
 * Strings beginning with **o** or **s**, or matching **OK** or **Success** (case-insensitive) map to **OK**
 * All others map to **info (6)**
 
-**Note**: If multiple log status remapper processors are applied to a given log within the pipeline, only the first one (according to the pipeline's order) is taken into account.
+**Note**: Note: If multiple log status remapper processors are applied to a log within a pipeline, only the first one in the pipeline’s order is considered. Additionally, across all pipelines that match the log, only the first status remapper encountered (from all applicable pipelines) will be applied.
```
estherk15, line 166:
```suggestion
**Note**: If multiple log status remapper processors are applied to a log within a pipeline, only the first one in the pipeline's order is considered. Additionally, for all pipelines that match the log, only the first status remapper encountered (from all applicable pipelines) is applied.
```

estherk15 (approved):
Small suggestion, looks good.

PR #25282, opened 2024-09-17
[PROF-10127] Ruby profiler allocations and heap config improvements

ivoanjo:

This PR updates the Ruby profiler enabling page to add a "Configuration" section, similar to the ones we have for Java/.net/Go (I've even copy-pasted a few of the descriptions that are the same).

This section has a few standard settings we've supported for a long time, and, crucially, it documents the flags needed to enable Ruby allocation and heap profiling.

We previously were linking to the library release notes to document how to enable these features.

We would like to have this config be more prominent in the docs.

I've also updated the profile types page to link to this new info.

### Merge instructions

### DIFF
--- a/content/en/profiler/enabling/ruby.md
+++ b/content/en/profiler/enabling/ruby.md
@@ -112,6 +112,38 @@ end
 
 7. A minute or two after starting your Ruby application, your profiles will show up on the [Datadog APM > Profiler page][5].
 
+## Configuration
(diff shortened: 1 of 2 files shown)


estherk15 (approved):
Looks good, thanks for adding that information!

ivoanjo:
Thank you for the really fast review :))

PR #25304, opened 2024-09-18
 SCICM-2609

OmerRaviv:
Put the the instructions on .pdb files at the top level of .NET instructions (previously they only appear under SourceLink, which is misleading as without .pdbs, we will not have any file path / line number information available, and so Source Code Integration would provide a degraded experience.

### Merge instructions

Please wait for review from @jminuscula before merging.


buraizu (changes requested):
Thanks for the PR, just requesting to re-add the formatting for the file extensions.

jminuscula (approved):
thank you @OmerRaviv!

OmerRaviv:
Thanks @buraizu ! Ready to merge from my standpoint.

buraizu (approved):

PR #25405, opened 2024-09-23
[ACCESS-2693] Update api spec with permissions information

wangwillson1:

In the API reference, this PR adds the permissions required to access a given endpoint.

https://docs-staging.datadoghq.com/willson.wang/ACCESS-2693-update-api-spec/api/latest/users/#update-a-user

### Merge instructions

- Please wait for a review from aaa-core-access before merging


srosenthal-dd (commented):
Can you help me understand what this changes means from the user perspective?

Will all API endpoints that start to list permissions in the public docs?

I assume the extra OAuth section is because not all permissions are available in OAuth? I do hope that is a temporary situation in most cases.

We (including PM) should be aware that the actual permission logic is occasionally more complicated. In the example you link, "update a user", the first permission check is an OR, but then it looks specifically at the user/service-account being updated and validates that the caller has the appropriate permission ([code](https://github.com/DataDog/dogweb/blob/4f45b5b47b0f4a8d1f82957c8bf948bee57e3832/dogweb/controllers/api/user.py#L1021))

wangwillson1:
> Will all API endpoints that start to list permissions in the public docs?

all non-rapid endpoints, since API clients is working on adding api spec autogen support for them in the coming weeks

> I assume the extra OAuth section is because not all permissions are available in OAuth? I do hope that is a temporary situation in most cases.

yes precisely - ideally we can (one day) just get rid of the whole notion of separate oauth endpoints, but once we get there the docs change will be relatively easy

bi1yeu (approved):

layouts/partials/api/permissions.html
```diff
@@ -0,0 +1,41 @@
...
+      <li><code>{{ . }}</code></li>
+    {{ end }}
+    <br/>
+  {{ end }}
+{{ end }}
+
+
+{{ with $authZ }}
+  {{ $slug := (replaceRE " " "-" ($tag.name | humanize | lower)) }}
+  <p>OAuth apps will require the <code>{{ delimit . ", "}}</code> authorization <a href="{{ "api/latest/scopes/" | absLangURL }}#{{ $slug }}">scope</a> to access this endpoint.</p>
```
urseberry, line 40:
```suggestion
  <p>OAuth apps require the <code>{{ delimit . ", "}}</code> authorization <a href="{{ "api/latest/scopes/" | absLangURL }}#{{ $slug }}">scope</a> to access this endpoint.</p>
```
We use present tense in the public docs.

bgdeutsch (approved):

urseberry (approved):

wangwillson1:
> Willson, do you want me to merge this, or should I wait for Stephen to re-review?

good to merge now, thank you!

urseberry (approved):

PR #25441, opened 2024-09-26
Separate ArgoCD trigger into two conditions

GabrielAnca:

Update the CD Visibility <> Argo CD configuration guide to separate the trigger into two sets of conditions.

This is done to cover an edge case: the Argo CD notification deduplication logic will not send the notification when the application transitions between two states that match the conditions. For example, if the application transitions from `phase: Running, health: Healthy` to `phase: Succeeded, health: Healthy`, it will only send a notification for the first state but not for the second.

Separating into these two conditions ensures the notification is sent in both cases. We can keep `Succeeded`, `Failed` and `Error` in the same condition because they are terminal states and it should not be possible to transition between them.

### Merge instructions

### DIFF
--- a/content/en/continuous_delivery/deployments/argocd.md
+++ b/content/en/continuous_delivery/deployments/argocd.md
@@ -94,7 +94,9 @@ metadata:
   name: argocd-notifications-cm
 data:
   trigger.cd-visibility-trigger: |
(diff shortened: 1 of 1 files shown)


andrea-mosk (approved):

GabrielAnca:
Thanks for the review @drichards-87! It's ready, feel free to merge at any time!

PR #25457, opened 2024-09-26
Fix out of date docs

nikolasleblanc:
Docs still suggest starting with require dd-trace/initialize when it should be dd-trace/init.

### Merge instructions


brett0000FF (approved):
@nikolasleblanc - Good catch. Thanks for making this contribution! I'll get the fix merged in for us shortly.

PR #25459, opened 2024-09-27
Add note about custom vital names

VivaLaPanda:

Docs currently don't note that custom vital names cannot have spaces or special characters. Ideally this would give something like a regex that matches all valid names, but I don't know that regex.

### Merge instructions


cy-moi (approved):

brett0000FF:
@VivaLaPanda - Thanks for making this contribution! I made a small tweak to make our Vale linter happy, but I'll get this merged for us shortly. 🚀

PR #25460, opened 2024-09-27
USM: add recommended values for Bottlerocket when using Helm

Yumasi:

Adding values to disable SELinux for the system-probe container when enabling Istio monitoring, when using the Bottlerocket distribution on Kubernetes clusters.
Istio monitoring will not work on Bottlerocket without these values.

### Merge instructions

### DIFF
--- a/content/en/universal_service_monitoring/setup.md
+++ b/content/en/universal_service_monitoring/setup.md
@@ -906,6 +906,21 @@ agents:
         - name: DD_SERVICE_MONITORING_CONFIG_TLS_ISTIO_ENABLED
           value: "true"
 ```
+
+If your cluster is using the Bottlerocket Linux distribution for its nodes, add the following to your values file:
+
+```
+agents:
+  containers:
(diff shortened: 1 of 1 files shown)


content/en/universal_service_monitoring/setup.md
```diff
@@ -906,6 +906,21 @@ agents:
...
+  containers:
+    systemProbe:
+      securityContext:
+        seLinuxOptions:
+          user: "system_u"
+          role: "system_r"
+          type: "spc_t"
+          level: "s0"
+```
+
```
guyarb, line 923: why is it under istio section?

guyarb: let's move it to a different section, as it can affect other components of system-probe and there is nothing unique for istio

Yumasi: Moved it below the instructions for GCE

neko-dd (approved):

PR #25466, opened 2024-09-27
Move Metrics section out of shortcode

drichards-87:

The Metrics section of the DBM MongoDB data collected shortcode appears in two articles. Each article requires the link in the Metrics section to point to a different URL. That's not possible, so as an alternative solution, we've decided to move the section outside of the shortcode.

### Merge instructions


pierreln-dd (approved):
Looks good to me!

brett0000FF (approved):

PR #25513, opened 2024-10-01
Add Info Alert to Correlating OpenTelemetry Traces and Logs

brett0000FF:

* The latest tracing library versions support 128-bit IDs.
* So, OTel traces and logs should be automatically correlated.
* Add note that the manual steps on this page are only required for older versions.

**Open questions**
* When did this change happen?

### Merge instructions


brett0000FF:
@dinooliva - Hey there. Could you please confirm that accuracy of this info banner? I have two question:
* Can we be more specific about "older versions"?
* Is the rest of this content even useful anymore? Our docs should generally cover the current state of the product, so I don't want to document outdated functionality.

Thanks!

brett0000FF:
Confirmed with Dino offline.

urseberry (approved):

PR #25533, opened 2024-10-02
(maint) Prevent workflow from failing on no-op

hestonhoffman:

The workflow that updates the synthetic-worker version was failing on the `commit` command when it didn't find any changes to the target file. Although the workflow wasn't broken (it just exited if no change was found), this behavior was confusing. This PR checks the existing version and writes a GitHub output variable that we can use to conditionally run the `commit` and `push` steps in the workflow. This should prevent the workflow from failing if there's nothing to commit.

test run: https://github.com/DataDog/documentation/actions/runs/11147491598/job/30981945184

### Merge instructions

### DIFF
--- a/local/bin/py/version_getter.py
+++ b/local/bin/py/version_getter.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 import re
+import os
 import json
 from io import StringIO
 import requests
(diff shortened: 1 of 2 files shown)


local/bin/py/version_getter.py
```diff
@@ -60,11 +61,24 @@ def get_versions(keys):
...
+        current_versions = json.load(open('data/synthetics_worker_versions.json'))
+    except:
+        current_versions = {}
     final_versions = get_versions(keys)
-    print(final_versions)
     
-    with open('data/synthetics_worker_versions.json', 'w') as f:
-        f.write(json.dumps(final_versions, indent=4, sort_keys=True))
+    if current_versions != final_versions:
+        print("New version detected!")
```
urseberry, line 75: Is it all right to leave the debugging print statements in the production version?

urseberry (approved):

local/bin/py/version_getter.py
hestonhoffman: Yep! I kept the echo step in as well, cos I think it helps whoever needs to debug this down the line

PR #25545, opened 2024-10-02
[ONBOARDING-68] Fix Signup Form Translation

bdhazman:

This PR fixes the translation on the signup modal by removing an extra `?` from the language parameter on the signup's URL. 

### Merge instructions

### DIFF
--- a/assets/scripts/components/global-modals.js
+++ b/assets/scripts/components/global-modals.js
@@ -17,7 +17,7 @@ const doOnLoad = () => {
         }
 
         if (lang === 'fr' || lang === 'ja' || lang === 'ko') {
-            langParam = `?lang=${lang}`;
+            langParam = `lang=${lang}`;
         } else {
             langParam = '';
         }


assets/scripts/components/global-modals.js
```diff
@@ -17,7 +17,7 @@ const doOnLoad = () => {
         }
 
         if (lang === 'fr' || lang === 'ja' || lang === 'ko') {
```
gverac, line 19: Huh, interesting that we limit to these languages considering we have others, like Portuguese and Spanish. I wonder if that's intended or not. Maybe the documentation page supports only these three?

gverac (approved):

assets/scripts/components/global-modals.js
bdhazman: Yeah it seems that is the case. There is a drop down on the bottom of the documentation page with these languages.

sseg-dd (approved):

stanleegoodspeed (approved):

PR #25549, opened 2024-10-02
Move Status page out one level

estherk15:

- Move the monitor status page out one level
- PM requested this change, no new content or updates in this change

### Merge instructions


urseberry (approved):
Would it make sense to set an alias on the moved page to the old URL?

estherk15:
> Would it make sense to set an alias on the moved page to the old URL?

I thinks this is what you meant, but if I misunderstood let me know! https://github.com/DataDog/documentation/blob/ce95ca78010bc8eb7d36520cdee28133469a80b7/content/en/monitors/status.md?plain=1#L6

PR #25575, opened 2024-10-04
Update Okta instructions to include forward slash for ACS URL (incident-31193)

gsaville12321:
Add / to company ID for okta instructions to account for issue in incident 31193

### Merge instructions

- [x ] Please merge after reviewing


sseg-dd (approved):

nobe4:
FWIW: without the prefix `/`, the redirect URL from Okta is just plainly wrong:

image

Notice how `assertion` and `id` are joined, when it should be `assertion/id`.

PR #25679, opened 2024-10-10
Setting MySQL 5.7 as the default tab for RDS

sadhbh-a:
The default tab on this this file is for MySQL 5.6, which is an older version of MySQL.
It makes more sense for MySQL 5.7 to be the initial tab, as this is more useful for customers & is consistent with with the Postgres documentation with latest version first: https://docs.datadoghq.com/database_monitoring/setup_postgres/selfhosted/?tab=postgres15#grant-the-agent-access

### Merge instructions


aldrickdev (approved):
Thanks for the update, all good 👍

PR #25686, opened 2024-10-10
OPA-2471: de-duplicate the multiple k8s config files

jdrouet:

When installing a worker in obs-pipelines, if you want to deploy to k8s, you need to select a cloud provider to pick the right `values.yaml` file. 3 are available but they are exactly the same. Therefore, I'm making another copy called `values.yaml` temporarily so that we can deploy the current fix. In a later PR, I will remove the 3 other files.

With this PR, the UI will be able to point where a file that, when downloaded, its name is `values.yaml` and we can update the installation script accordingly.

### Merge instructions


maycmlee (approved):
@jdrouet this looks good to me! Good to merge?

maycmlee:
@jdrouet docs repo doesn't have the merge queue workflow set up, so I'll go ahead and merge this.

PR #25743, opened 2024-10-15
Update Lambda Instrumentation Link in SFN doc

kimi-p:
The original link, which links to logs forwarder for Lambda instrumentation instructions, is incorrect. This PR point them to the Lambda instrumentation page for all runtimes.


buraizu (approved):
Confirmed in local build

sumedham (approved):

PR #25748, opened 2024-10-16
Fix the link to `glossary` page

y-yagi:
Currently, the link to the `glossary` is broken because the link name is wrong. This PR fixes it.

docs datadoghq com_ja_tracing_trace_collection_automatic_instrumentation_dd_libraries_ruby_

### Merge instructions


buraizu (approved):
Validated locally

buraizu:
Thanks for this update @y-yagi. I'll merge this for now, but there's a potential for this change to be overwritten later due to an automated update process. If this happens, it should be a transient change, as I've also coordinated with our translation service in this regard. Please let me know if you have any questions.

PR #25752, opened 2024-10-16
Mark roadrunner 3.x as supported

bwoebi:
First, roadrunner 3.x is supported.

However, some users have been confused because the SDK (https://github.com/roadrunner-php/http) is versioned differently from the web server (https://github.com/roadrunner-server/roadrunner).
These are independent, but people have been asking for whether we support version "2024.2", while the real thing they would need to know is if we support Roadrunner SDK 3.x (but they don't really grasp that).

Is there anything we can do here to make this clearer to viewers of that documentation page?

### Merge instructions


buraizu:
Thanks, created DOCS-9258 to investigate if there's anything else we can do here to make this clearer to viewers of this documentation page

brett0000FF (approved):
@bwoebi - Apologies for the delay on this! I think this is a good opportunity to use our new tooltip feature so we don't clutter the table, but we can include info about the version discrepancy.

What do you think of the change I pushed?

brett0000FF:
@PROFeNoM - What do you think of this change I pushed to clarify the customer confusion Bob highlighted? He's OOO so requested that I ping you. Thanks!

PROFeNoM (approved):
For me, this looks good! (+it's beautiful 😃)

PR #25755, opened 2024-10-16
PHP trace correlation identifiers - try to make it a little clearer

cecile75:
I found the current doc a little confusing.

##Preview
https://docs-staging.datadoghq.com/cecile/phplogscorrelation/tracing/other_telemetry/connect_logs_and_traces/php/

### DIFF
--- a/content/en/tracing/other_telemetry/connect_logs_and_traces/php.md
+++ b/content/en/tracing/other_telemetry/connect_logs_and_traces/php.md
@@ -82,7 +82,7 @@ use Illuminate\Support\Facades\Log;
 Log::info('Hello, World! [%dd.trace_id% %dd.span_id% %status%]');
 ```
 
-The PHP tracer replaces the placeholders with the corresponding values. For example, the logged message above will be transformed into:
+The PHP tracer replaces the placeholders with the corresponding values. For example, the logged message above is transformed into:
 
 ```
 [2022-12-09 16:02:42] production.INFO: Hello, World! [dd.trace_id="1234567890abcdef" dd.span_id="1234567890abcdef" status="info"]
--- a/content/en/tracing/other_telemetry/connect_logs_and_traces/php.md
+++ b/content/en/tracing/other_telemetry/connect_logs_and_traces/php.md
@@ -65,7 +65,7 @@ The PHP tracer adds the available trace correlation identifiers to the log conte
 
 #### Option 2: Use placeholders in your message {#use-placeholders-in-your-message}
 
-You can use placeholders in your message to select which trace correlation identifiers will be automatically injected into your logs. The PHP tracer supports the following placeholders:
+You can use placeholders in your message to select which trace correlation identifiers are automatically injected into your logs. The PHP tracer supports the following placeholders:
 - `%dd.trace_id%`: the trace ID
 - `%dd.span_id%`: the span ID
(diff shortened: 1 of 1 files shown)


content/en/tracing/other_telemetry/connect_logs_and_traces/php.md
```diff
@@ -55,17 +55,17 @@ use Illuminate\Support\Facades\Log;
...
 [2022-12-09 16:02:42] production.DEBUG: Hello, World! {"dd.trace_id":"1234567890abcdef","dd.span_id":"1234567890abcdef","dd.service":"laravel","dd.version":"8.0.0","dd.env":"production","status":"debug"}
 ```
 
 **Note**: If there is a placeholder in your message or if a trace ID is already present in the message, the PHP tracer does **not** add the trace correlation identifiers to the log context.
 
-#### Use placeholders in your message {#use-placeholders-in-your-message}
+#### Option 2: Use placeholders in your message {#use-placeholders-in-your-message}
 
-You can use placeholders in your message to automatically inject trace correlation identifiers into your logs. The PHP tracer supports the following placeholders:
+You can use placeholders in your message to select which trace correlation identifiers will be automatically injected into your logs. The PHP tracer supports the following placeholders:
```
buraizu, line 68:
```suggestion
You can use placeholders in your message to select which trace correlation identifiers are automatically injected into your logs. The PHP tracer supports the following placeholders:
```

content/en/tracing/other_telemetry/connect_logs_and_traces/php.md
```diff
@@ -82,7 +82,7 @@ use Illuminate\Support\Facades\Log;
 Log::info('Hello, World! [%dd.trace_id% %dd.span_id% %status%]');
 ```
 
-The PHP tracer replaces the placeholders with the corresponding values. For example, the logged message above could be transformed into:
+The PHP tracer replaces the placeholders with the corresponding values. For example, the logged message above will be transformed into:
```
buraizu, line 85:
```suggestion
The PHP tracer replaces the placeholders with the corresponding values. For example, the logged message above could be transformed into:
```

buraizu (changes requested):
Requesting some minor edits per the [documentation style guide](https://github.com/DataDog/documentation/blob/master/CONTRIBUTING.md#tense)

PROFeNoM (approved):
:smiley:

content/en/tracing/other_telemetry/connect_logs_and_traces/php.md
buraizu: Thanks for clarifying @cecile75, I've updated the recommendation. My point here is to avoid use of future tense, which is not recommended per the style guide.

Should this also be updated on line `58`, which likewise uses `could be`?

cecile75: line 58 is little different in my opinion.
With option 1, any trace correlation identifier with a value at the time of the log will be injected. So it's really more a "could" as the user doesn't chose explicitly which trace correlation identifiers will be injected.
While with option 2, as the coder explicitly choose to inject %dd.trace_id% %dd.span_id% %status% , I think we can assume that these items have a value at this time of the log and will be injected and present in the resulting log.

PR #25788, opened 2024-10-17
Modify spacing to separate nested list items

janine-c:

Small PR to add spaces after nested list items, so they don't all appear on the same line.

### Merge instructions


content/en/sensitive_data_scanner/library_rules.md
```diff
@@ -21,10 +21,10 @@ These rules can also be viewed in Datadog:
 
 1. Navigate to [Sensitive Data Scanner][1].
 1. Click **Scanning Rules Library** on the top right side of the page.
-1. To add rules from the library to a scanning group:
-  a. Select the rules you want to add.
-  b. Click **Add Rules to Scanning Group**.
-  c. Follow the steps in [Set Up Sensitive Data Scanner][2] to finish the setup
+1. To add rules from the library to a scanning group:   
+   a. Select the rules you want to add.   
+   b. Click **Add Rules to Scanning Group**.
```
neko-dd, line 26: Does this only work with the trailing spaces in place? I'm worried that they'll get deleted in the future, and they *shouldn't* be required in order for this to work.

Did you try changing the a. b. c. to 1. 1. 1.?

janine-c: Yeah, I did some playing around with it, and it looks like you either need trailing spaces, or empty lines between the nested list items for it to work. Do you think empty lines might be more appropriate here?

I just tried your 1. 1. 1. suggestion, and the nesting worked, but the numbering scheme was the same, which I think could be confusing: 
image

The only other thing I can think of here is just making it an HTML list instead, so we don't have to worry about the whitespace at all. Happy to do whichever approach you think is best!

neko-dd: I think this might be a CSS problem that can be corrected by the websites team. We can push out your current version but I can follow up to see if we can get a real fix for this.

neko-dd: Actually I think rather than trailing spaces, can you try adding <br> to the ends of the first two lines?

janine-c: That worked and is much more visible - updated, thank you!

neko-dd (approved):

PR #25811, opened 2024-10-18
Add details for GKE Autopilot compute class + cleanup

JacksonDavenport:

GKE Autopilot supports compute classes for dedicated hardware, and like their GKE Spot Pods, sort of silently manages these with Tainted nodes. Autopilot will automatically add Tolerations to the workloads you specify, however, you need to do this manually to the Agent. 

https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-compute-classes

Additionally does some minor cleanup and removes the Operator steps for Spot Pods, since the Operator does not support Autopilot anyway. 

From customer requests.

### Merge instructions

### DIFF
--- a/content/en/containers/kubernetes/distributions.md
+++ b/content/en/containers/kubernetes/distributions.md
@@ -294,9 +294,9 @@ providers:
 {{% /tab %}}
 {{< /tabs >}}
 
-### Spot pods and compute Classes
+### Spot pods and compute classes
 
-Using [Spot Pods][10] in GKE Autopilot clusters introduces [Taints][9] to the corresponding Spot GKE nodes. When using Spot Pods, additional configuration is required to provide the Agent DaemonSet with a matching Toleration.
+Using [Spot Pods][10] in GKE Autopilot clusters introduces [taints][9] to the corresponding Spot GKE nodes. When using Spot Pods, additional configuration is required to provide the Agent DaemonSet with a matching toleration.
 
 {{< tabs >}}
 {{% tab "Helm" %}}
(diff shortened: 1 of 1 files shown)


content/en/containers/kubernetes/distributions.md
```diff
@@ -330,7 +313,22 @@ agents:
 {{% /tab %}}
 {{< /tabs >}}
 
-**Note**: Network Performance Monitoring is not supported for GKE Autopilot.
+Similarly when using [GKE Autopilot Compute classes][11] to run workloads that have specific hardware requirements, take note of the [Taints][9] that GKE Autopilot is applying to these specific nodes and add matching Tolerations to the Agent DaemonSet. You can match the Tolerations on your corresponding pods. For example for the `Scale-Out` compute class use a Toleration like:
```
aliciascott, line 316:
```suggestion
Similarly when using [GKE Autopilot Compute classes][11] to run workloads that have specific hardware requirements, take note of the [Taints][9] that GKE Autopilot is applying to these specific nodes and add matching tolerations to the Agent DaemonSet. You can match the tolerations on your corresponding pods. For example for the `Scale-Out` compute class use a toleration like:
```

aliciascott: Is this generally industry accepted as capitalized? If not should probably use lower-case

content/en/containers/kubernetes/distributions.md
```diff
@@ -292,30 +294,11 @@ providers:
 {{% /tab %}}
 {{< /tabs >}}
 
-### Spot pods and instances
+### Spot pods and compute Classes
```
aliciascott, line 297:
```suggestion
### Spot pods and compute classes
```
I see "classes" is lower-case below, we should also lowercase here to adhere to sentence case for headings

aliciascott (approved):
Hi @JacksonDavenport went ahead and approved this as the changes requested are non-blocking. Note you can now add a comment to approved docs PR's with just a comment with `/merge` and go ahead and merge this when you are ready!

content/en/containers/kubernetes/distributions.md
JacksonDavenport: Oops I remembered to lower "compute" but forgot that.

content/en/containers/kubernetes/distributions.md
JacksonDavenport: Good call out, I thought K8s did so I mirrored, but I must have skimmed to quickly. Removed those. 
https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/

PR #25822, opened 2024-10-21
[SDCI] Jenkins and Custom providers running pipelines support

rodrigo-roca:

We added support for the running pipelines feature for both `Jenkins` and `custom` CI providers (data sent through the public API). 

### Merge instructions


content/en/continuous_integration/pipelines/_index.md
```diff
@@ -144,7 +144,7 @@ If your CI provider is not supported, you can try setting up Pipeline Visibility
 |  | Jenkins | GitLab | CircleCI | Buildkite | GitHub Actions | Azure Pipelines | Codefresh | TeamCity | AWS CodePipeline | Other CI Providers |
 |---|:---:|:---:|:---:|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
 | {{< ci-details title="Pipeline trace visualization" >}}Visualization of pipeline executions with associated tracing.{{< /ci-details >}} | {{< X >}} | {{< X >}} | {{< X >}} | {{< X >}} | {{< X >}} | {{< X >}} | {{< X >}} | {{< X >}} | {{< X >}} | {{< X >}} |
-| {{< ci-details title="Running pipelines" >}}Identification of pipelines executions that are running with associated tracing.{{< /ci-details >}} | | {{< X >}} | | | {{< X >}} | | | | {{< X >}} |
+| {{< ci-details title="Running pipelines" >}}Identification of pipelines executions that are running with associated tracing.{{< /ci-details >}} | {{< X >}}  | {{< X >}} | | | {{< X >}} | | | {{< X >}} | {{< X >}} |
```
rodrigo-roca, line 147: We forgot to add a checkmark here for AWS Codepipeline so I added it here

content/en/continuous_integration/pipelines/custom.md
```diff
@@ -43,7 +44,7 @@ To send pipeline events programmatically to Datadog, ensure that your [`DD_API_K
    - `DD-API-KEY`: Your [Datadog API key][14].
    - `Content-Type`: `application/json`.
 
-1. Prepare the payload body by entering information about the [pipeline execution][2] in a cURL command:
+2. Prepare the payload body by entering information about the [pipeline execution][2] in a cURL command:
```
rodrigo-roca, line 47: no related but we had `1.` on every item on the list

buraizu: If you use only `1.` for a list, Markdown will dynamically update each item to its correct number. With that said, I prefer this way, as it makes maintenance easier and avoids this kind of confusion.

content/en/continuous_integration/pipelines/custom.md
```diff
@@ -90,7 +91,19 @@ To send pipeline events programmatically to Datadog, ensure that your [`DD_API_K
    EOF
    {{< /code-block >}}
 
-1. After sending your pipeline event to Datadog, you can integrate additional event types such as `stage`, `job`, and `step`. For more information, see the [Send Pipeline Event endpoint][1].
+3. After sending your pipeline event to Datadog, you can integrate additional event types such as `stage`, `job`, and `step`. For more information, see the [Send Pipeline Event endpoint][1].
+
+## Running pipelines
+Pipelines sent with the `status` set to `running` should have the same `unique_id` as the final pipeline event. Additionally,
```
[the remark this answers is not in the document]
buraizu: Basing this suggestion on [RFC 2119](https://github.com/DataDog/documentation/blob/master/CONTRIBUTING.md#rfc-2119), but feel free to ignore this if there may exist valid reasons in particular circumstances for pipeline events with the `status` set to `running` to have a different `unique_id` as the final pipeline event.

content/en/continuous_integration/pipelines/custom.md
```diff
@@ -90,7 +91,19 @@ To send pipeline events programmatically to Datadog, ensure that your [`DD_API_K
...
+running pipelines can be updated by adding more information while they are still running. The flow is as follows:
+
+1. The initial running pipeline event with the `status` set to `running`.
+2. Optionally, `N` running pipelines with the same `unique_id` and the `status` set to `running` to update the pipeline with more information.
+3. The final pipeline event with a non `running` status and the same `unique_id`.
+
+**Note**: it is not guaranteed that the last value will be the one displayed in the UI when the same field is updated multiple times.
+For example: if the a tag called `my_tag` is set to `value1` in the first running pipeline and then we try to update it to `value2`,
+Datadog might sometimes show `value1` instead of `value2` in the UI.
```
buraizu, line 106:
```suggestion
**Note**: The most recent value may not always be the one displayed in the UI when a field is updated. For example, if the tag `my_tag` is set to `value1` in the first running pipeline, and then is updated to `value2`, you may see `value1` instead of `value2` in the UI.
```

buraizu: From my reading of this, it sounds like `updated multiple times` is not a requirement; the last value may not be shown if it is updated even once. Let me know if this is incorrect, or feel free to make adjustments to the words or phrasing I'm using here.

content/en/continuous_integration/pipelines/custom.md
```diff
@@ -90,7 +91,19 @@ To send pipeline events programmatically to Datadog, ensure that your [`DD_API_K
    EOF
    {{< /code-block >}}
 
-1. After sending your pipeline event to Datadog, you can integrate additional event types such as `stage`, `job`, and `step`. For more information, see the [Send Pipeline Event endpoint][1].
+3. After sending your pipeline event to Datadog, you can integrate additional event types such as `stage`, `job`, and `step`. For more information, see the [Send Pipeline Event endpoint][1].
+
+## Running pipelines
+Pipelines sent with the `status` set to `running` should have the same `unique_id` as the final pipeline event. Additionally,
+running pipelines can be updated by adding more information while they are still running. The flow is as follows:
```
buraizu, line 98:
```suggestion
Additionally, running pipelines can be updated by adding more information while they are still running. The flow is as follows:
```

buraizu: `The flow is as follows:` would seem to indicate that the following information is a series of steps. However, that doesn't seem to line up with the items, which don't direct a user to carry out any actions. Could this potentially be clarified?

buraizu (changes requested):

content/en/continuous_integration/pipelines/custom.md
```diff
@@ -90,7 +91,17 @@ To send pipeline events programmatically to Datadog, ensure that your [`DD_API_K
    EOF
    {{< /code-block >}}
 
-1. After sending your pipeline event to Datadog, you can integrate additional event types such as `stage`, `job`, and `step`. For more information, see the [Send Pipeline Event endpoint][1].
+3. After sending your pipeline event to Datadog, you can integrate additional event types such as `stage`, `job`, and `step`. For more information, see the [Send Pipeline Event endpoint][1].
+
+## Running pipelines
+Pipeline events sent with the `status` set to `running` have the same `unique_id` as the final pipeline event.
+Additionally, running pipelines can be updated by adding more information while they are still running. Essentially, a pipeline consists of the following events:
```
buraizu, line 98:
```suggestion
Pipeline events sent with the `status` set to `running` have the same `unique_id` as the final pipeline event. Running pipelines can be updated by adding more information while they are still running. A running pipeline consists of the following events:
```

content/en/continuous_integration/pipelines/custom.md
```diff
@@ -90,7 +91,17 @@ To send pipeline events programmatically to Datadog, ensure that your [`DD_API_K
...
+
+1. The initial running pipeline event with the `status` set to `running`.
+2. Optionally, `N` running pipelines with the same `unique_id` and the `status` set to `running` to update the pipeline with more information.
```
buraizu, line 101:
```suggestion
2. Optionally, `N` running pipelines with the same `unique_id` and the `status` set to `running`.
```

buraizu: I'm basing this suggestion on the assumption that this isn't directing the user to update the pipeline with more information, but rather describing a potential scenario with multiple running pipelines that have the same `unique_id`. Please let me know if this is inaccurate.

content/en/continuous_integration/pipelines/custom.md
```diff
@@ -90,7 +91,17 @@ To send pipeline events programmatically to Datadog, ensure that your [`DD_API_K
...
+3. The final pipeline event with a non `running` status and the same `unique_id`.
```
buraizu, line 102:
```suggestion
3. The final pipeline event **without** a `running` status and the same `unique_id`.
```

buraizu: If possible, can we tell users what the status will be, instead of what it will **not** be?

content/en/continuous_integration/pipelines/custom.md
buraizu: It sounds like this section is not about updating running pipelines, but rather describing them. Is this correct?

buraizu (changes requested):
Thanks for clarifying the **Running pipelines** section. I left a few additional suggestions and questions.

content/en/continuous_integration/pipelines/custom.md
rodrigo-roca: The end status may be `success`, `failed`, `canceled`, ... We do not know. We can either tell them it is not running or list every possible finished state

content/en/continuous_integration/pipelines/custom.md
rodrigo-roca: The user will be the one updating the pipeline in this case. The user needs to call the API in the way it is described in this section (they will be doing POST requests). The reader can go to the API endpoint for more details but the main goal of this section to describe running pipelines and letting them know, that, optionally, they can update them if they need to.

content/en/continuous_integration/pipelines/custom.md
rodrigo-roca: Correct, it is about describing them. They have the API documentation for the details

content/en/continuous_integration/pipelines/custom.md
rodrigo-roca: Let me know if something is not clear!

buraizu: Thanks again for clarifying. It sounds like `pipeline event` may be more accurate than `pipeline` in this case. I updated my suggestion accordingly.

buraizu (approved):

drodriguezhdez (approved):

PR #25870, opened 2024-10-23
enables typesense live syncing job

bgdeutsch:
- enables the typesense indexing job in live, so we can run typesense and algolia in parallel as we prepare to switch over.
- adds a `search.json` output for typesense specifically so no current Algolia processes are changed.

### Merge instructions
- [x] Please merge after websites reviews

### Additional notes
- `typesense_sync` job ran successfully in preview: https://gitlab.ddbuild.io/DataDog/documentation/-/jobs/680879329
- live typesense job is uncommented in gitlab config
- Current Algolia jobs are not affected by these changes

### DIFF
--- a/typesense.config.json
+++ b/typesense.config.json
@@ -1,7 +1,7 @@
 {
     "collections": [
         {
-            "name": "docs_main_v1",
+            "name": "docs_alias",
             "file_path": "./public/search.json",
             "synonyms": [
                 {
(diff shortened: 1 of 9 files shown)


typesense.config.json
```diff
@@ -2,7 +2,7 @@
     "collections": [
         {
             "name": "docs_main_v1",
```
bgdeutsch, line 4: we probably want to point this at the Alias name, but the docs API Key isn't configured for that at the moment.

rtrieu (approved):

davidejones (approved):

PR #25875, opened 2024-10-23
SDCD-437: Updating pagerduty service mapping logic in dora setup page

henrywalsh:

We are in the processes of changing the mapping rules for DORA failures when we get PagerDuty alerts. This PR provides details into those new mapping rules and modifies the rules that were already written for clarity.

### Merge instructions

- [ ] We are waiting for the customer-wide release of this before it is merged. I can update whomever reviews this PR when that happens and we are ready for merging.

### DIFF
--- a/content/en/dora_metrics/setup/failures.md
+++ b/content/en/dora_metrics/setup/failures.md
@@ -101,7 +101,7 @@ The matching algorithm works in the following steps:
    - If a single Datadog service matches, the incident metrics and events are emitted with the service and team.
    - If multiple Datadog services match, the incident metrics and events are emitted with the team.
 
-For more information about setting the PagerDuty service URL for a Datadog service, see [Use Integrations with Service Catalog][103].
+   For more information about setting the PagerDuty service URL for a Datadog service, see [Use Integrations with Service Catalog][103].
 
 3. If the PagerDuty service name of the incident matches a service name in the Service Catalog, the incident metrics and events are emitted with the service and team.
 4. If the PagerDuty team name of the incident matches a team name in the Service Catalog, the incident metrics and events are emitted with the team.
--- a/content/en/dora_metrics/setup/failures.md
+++ b/content/en/dora_metrics/setup/failures.md
@@ -87,12 +87,12 @@ The severity of the incident in the DORA Metrics product is based on the [incide
 
 ### Mapping PagerDuty services to Datadog services
 
-When an incident event is received for a specific [PagerDuty service][101], Datadog attempts to retrieve the related Datadog service and team from any relevant [Datadog monitors][108] and from the [Service Catalog][102].
+When an incident event is received for a specific [PagerDuty service][101], Datadog attempts to retrieve the related Datadog service and team from any triggering [Datadog monitors][107] and from the [Service Catalog][102].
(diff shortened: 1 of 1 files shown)


gturbat:
Also, should we use `Else if` instead of `If` to make it clear we stop going through the steps once we have a match?

henrywalsh:
> Also, should we use `Else if` instead of `If` to make it clear we stop going through the steps once we have a match?

We could, but it's not always `else if`. For example, we will look at static tags for `service` if the group by tags only include `env` and `team`. And if we can only get the `team` from the tags, we will still try to get the service from the PagerDuty URL matching. 

It's more like, "We will go through these steps until we have the clearest picture of who/what was effected."

content/en/dora_metrics/setup/failures.md
```diff
@@ -87,27 +87,36 @@ The severity of the incident in the DORA Metrics product is based on the [incide
...
-The matching algorithm works in the following scenarios:
+The matching algorithm works in the following steps:
+
+1. If the PagerDuty incident event was [triggered from a Datadog monitor][107]:
+   - If the monitor is in [Multi Alert mode][109], the incident metrics and events are emitted with the `env`, `service`, and `team` from the alerted group.
+   - If the monitor has [tags][110] for `env`, `service`, or `team`:
+     - `env`: If the monitor has a single `env` tag, the incident metrics and events are emitted with the environment.
+     - `service`: If the monitor has one or more `service` tags, the incident metrics and events are emitted with the provided services.
+     - `team`: If the monitor has a single `team` tag, the incident metrics and events are emitted with the team.
+2. If the service URL of the incident matches the PagerDuty service URL for any services in the Service Catalog:
```
calvarjorge, line 100:
```suggestion
     - `team`: If the monitor has a single `team` tag, the incident metrics and events are emitted with the team.
     
2. If the service URL of the incident matches the PagerDuty service URL for any services in the Service Catalog:
```

calvarjorge (approved):
LGTM

content/en/dora_metrics/setup/failures.md
```diff
@@ -87,27 +87,37 @@ The severity of the incident in the DORA Metrics product is based on the [incide
 
 ### Mapping PagerDuty services to Datadog services
 
-When an incident event is received for a specific [PagerDuty service][101], Datadog attempts to retrieve the related Datadog service and team from the [Service Catalog][102].
+When an incident event is received for a specific [PagerDuty service][101], Datadog attempts to retrieve the related Datadog service and team from any triggering [Datadog monitors][107] and from the [Service Catalog][102].
 
-The matching algorithm works in the following scenarios:
+The matching algorithm works in the following steps:
```
estherk15, line 92: Does the algorithm stop at the first matching condition or does it run through all the steps?

```suggestion
The matching algorithm works in the following order:
```

estherk15: This would make a really good diagram or decision tree. Consider submitting a request with [brand design]!(https://datadoghq.atlassian.net/jira/core/projects/BRAND/forms/form/direct/1/10002)!

estherk15 (approved):

PR #25968, opened 2024-10-29
Update monitor_aggregators.md

daniel-lee-datadoghq:
Correcting a description that I believe is incorrect.

It says `All values in the window must be above the threshold`, but in that 10 mins window, the minimum value is 8 at 3:03, which means the data is NOT above 10 at any point in the last 10 minutes. Therefore, I believe it is incorrect.


chinichilla (commented):
LGTM. Given the data provided, the monitor wouldn't go into Alert.

estherk15 (approved):
Thanks for the update!

PR #25999, opened 2024-10-30
[test visibility] Update EFD diagram for new tests

juan-fernandez:
Update EFD diagram to reflect new "new test" logic.

### Merge instructions

```
/merge
```


static/images/continuous_integration/early_flake_detection_commit_new_test_explanation.png
drichards-87, line 1: @juan-fernandez For the diagram, we'll need to keep the old file and instead add the new one and reference it with a different filename. This is due to cacheing issue with the docs site. If you replace an image and don't rename it, the site will often just use the old cached image and we then have to go and manually remove the file from cache.

drichards-87 (changes requested):

static/images/continuous_integration/early_flake_detection_commit_new_test_explanation.png
juan-fernandez: that makes sense. Done in https://github.com/DataDog/documentation/pull/25999/commits/9afe52ec3d496164111a17f52fe9cf93e63d2bbc

drichards-87 (approved):

PR #26197, opened 2024-11-08
Add allowlist for pull_config files

devindford:

- Adds the `pull_config_allowlist.yaml` file which will have a list of allowed repos we source from upstream
- Updates the python files to parse and check the allowlist 
- The allowlist will be applied for both the live and preview environments, the thought process here is that if a new repo is being added, during testing it will be in the preview config and once everything is ready will be in the live config. Since the repo will be trusted, there is no risk introducing it for both flows even if we're not pulling it into the live environment right away.
### Merge instructions

```
/merge
```

### Additional notes

image

Screenshot shows initial error output and then the cascade fail (job stop) from the `ValueError` that it throws (i commented out a repo to demonstrate this works)


local/bin/py/build/update_pre_build.py
```diff
@@ -150,6 +151,7 @@ def build_documentation(self):
 def main(disable_cache_on_retry=False):
     # Those hard-written variables should be set in the Makefile config later down the road.
     build_configuration_file_path = getenv("CONFIGURATION_FILE")
+    pull_config_allowlist_file_path = "./local/bin/py/build/configurations/pull_config_allowlist.yaml"
```
devindford, line 154: This pathing will change once it's in the ci repo since it will not be kept in the docs repo or on locals

bgdeutsch (approved):

PR #26235, opened 2024-11-12
Update is_match and is_exact_match description

SimoneT-DD:
The is_exact_match considers only the numeric values when used with numbers. Added a description that clarifies this scenario.

### DIFF
--- a/content/en/monitors/notify/variables.md
+++ b/content/en/monitors/notify/variables.md
@@ -31,9 +31,9 @@ The following conditional variables are available:
 |----------------------------|--------------------------------------------------------------------|
 | `{{#is_alert}}`            | The monitor alerts                                                 |
 | `{{^is_alert}}`            | The monitor does not alert                                         |
-| `{{#is_match}}`            | The context matches the provided substring. If a numeric value is used it will be converted to a string.|
+| `{{#is_match}}`            | The context matches the provided substring. If a numeric value is used, it is converted to a string.|
 | `{{^is_match}}`            | The context does not match the provided substring                  |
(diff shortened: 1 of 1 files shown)


content/en/monitors/notify/variables.md
```diff
@@ -31,9 +31,9 @@ The following conditional variables are available:
 |----------------------------|--------------------------------------------------------------------|
 | `{{#is_alert}}`            | The monitor alerts                                                 |
 | `{{^is_alert}}`            | The monitor does not alert                                         |
-| `{{#is_match}}`            | The context matches the provided substring                         |
+| `{{#is_match}}`            | The context matches the provided substring. If a numeric value is used it will be converted to a string.|
```
maycmlee, line 34:
```suggestion
| `{{#is_match}}`            | The context matches the provided substring. If a numeric value is used, it is converted to a string.|
```

content/en/monitors/notify/variables.md
```diff
@@ -31,9 +31,9 @@ The following conditional variables are available:
...
 | `{{^is_match}}`            | The context does not match the provided substring                  |
-| `{{#is_exact_match}}`      | The context exactly matches the provided string                    |
+| `{{#is_exact_match}}`      | The context exactly matches the provided string.<br> If a number is used, the numeric value will be considered, regardless of its type. This means that as long as two numbers have the same value, they will be considered equal by the function|
```
maycmlee, line 36:
```suggestion
| `{{#is_exact_match}}`      | The context exactly matches the provided string.<br> If a number is used, the numeric value is considered, regardless of its type. This means that as long as two numbers have the same value, they are considered equal by the function. |
```

maycmlee (approved):

PR #26457, opened 2024-11-22
Fix Step Function installation doc

lym953:

To enable tracing at forwarder level, I think we should set the `DD_STEP_FUNCTIONS_TRACE_ENABLED ` environment variable, not `DdStepFunctionsTraceEnabled`, because I see some forwarder is using the former and the former appears a lot in the codebase.
Need someone on Cloud Tracing team to confirm this. cc @avedmala 

### Merge instructions

```
/merge
```

### DIFF
--- a/content/en/serverless/step_functions/installation.md
+++ b/content/en/serverless/step_functions/installation.md
@@ -62,7 +62,9 @@ Then, to send your Step Functions logs to Datadog:
 
 5. To enable tracing, you have two options:
    - **Per Step Function**: Add the `DD_TRACE_ENABLED` tag to each Step Function and set the value to `true`.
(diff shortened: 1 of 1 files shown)


lym953:
Actually there's a `DdStepFunctionsTraceEnabled` parameter when we create the forwarder stack, but after creation, I think we can manually add the environment variable. Should we explain both options in the doc?

drichards-87 (approved):
Looks good from a Docs perspective. I'll let the Serverless team chime in on your other question. If you do end up updating the PR, reach out to the Docs Team and we'll give it another quick review. Thanks!

sumedham (approved):

PR #26535, opened 2024-11-28
Fix template variable example for Error Tracking monitors

GianlucaBortoli:

Fix template variable example for Error Tracking monitors


GianlucaBortoli:
> Hi @GianlucaBortoli, thanks for opening this. For the pipeline and merge-queue to function, the branch name must contain a `/`. Please reopen this change in a PR off a branch that follows that naming convention :D

Didn't notice the restriction, doing it now!

PR #26609, opened 2024-12-04
[hackathon] build english content only

bgdeutsch:
- Adds a job `build_preview:english` that will be the new default build/deploy job.
  - This will always run unless a translated file changes.
- Using Hugo's render segments feature, all languages other than English will be ignored in the build, will also be ignored during deploy (not deleting translated files when syncing)
- leaves a `build_preview:full` which would just build the full site.   this would run when there's a change to translated files (so we can still preview translation pipeline PRs).   in theory that would also run in the nightly scheduled build.  this is exactly what the current `build_preview` job does now.

### Merge instructions

```
/merge
```

### Additional notes

- https://gitlab.ddbuild.io/DataDog/documentation/-/jobs/727503587#L454-L472: english only job ran, hugo ignored other languages at render time.  the hugo build took ~30s.

- the first run of a new preview branch should always build everything because of gitsync it will look like every file is new.

- link checker is failing because its using the rendered version of the site hugo just built so it doesnt find any translated pages.  the pages are there in s3 though.  we likely need to find a way to have the link checker ignore anything under `/ja/, /fr/` etc


jhgilbert:

PR #26743, opened 2024-12-12
Update README.md with link to finding Slack username 

michaelcretzman:
the previous link went to page on finding the Profile ID, which is not the Slack username needed.

Merge readiness:
- [X] Ready for merge


aliciascott (approved):
much better URL thanks!

PR #26774, opened 2024-12-13
DOCS-9694: Remove outdated BigQuery integration doc

drichards-87:

Removes outdated BigQuery integration doc from the Japanese, Korean, and French docs. This is required in order for the redirect (which exists, but only works for the English docs) to properly work for the translated docs. i.e., a redirect won't work if the original doc still exists. In this particular case, the change hasn't made its way through Transifex yet.

### Merge instructions

Merge readiness:
- [] Ready for merge

```
/merge
```


aliciascott (approved):
looks good thanks for doing this!

PR #26993, opened 2025-01-07
Draft PR to add cross region support docs

FrancescoMarabotto:

This PR adds documentation for customers on how to configure cross region private links to send their data from a supported AWS region to US1 or AP1.

The idea is to add a tab containing the list of the supported regions and later on, add the proper instructions later (this requires a table similar to the existing one with an extra column).

While working on this I've notived that there is a JS file containing all the data and I would like to keep using it.
I've added a new variable to contain the list of the supported regions and I would like to have them printed, but I don't know how to tell Hugo to execute the code and get the return values. I've tried creating a shortcode but I couldn't make it work.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

```
/merge
```


hestonhoffman:
Closing in favor of #27717

PR #27110, opened 2025-01-14
[APM Tracing: Go] Update DD_TAGS documentation

mtoffl01:
Updates documentation on DD_TAGS. Default value of `[]` was incorrect because env var values are in string format. By default, no global tags are collected. Also updated the description to specify that key-value pairs must be string-convertible types.
Motivation: [Customer request](https://github.com/DataDog/dd-trace-go/pull/3067#issuecomment-2590424492)

### Merge instructions

Merge readiness:
- [ ] Ready for merge

```
/merge
```

### DIFF
--- a/content/en/tracing/trace_collection/library_config/go.md
+++ b/content/en/tracing/trace_collection/library_config/go.md
@@ -154,8 +154,8 @@ A JSON array of objects. Each object must have a `"sample_rate"`. The `"name"`,`
   - Set the sample rate to 100% for services that have a `priority` tag with the value `high`: `'[{"tags": {"priority":"high"}, "sample_rate": 1}]'`.
   
 `DD_TAGS`
-: **Default**: [] <br>
-A list of default tags to be added to every span and profile. Tags can be separated by commas or spaces, for example: `layer:api,team:intake,key:value` or `layer:api team:intake key:value`.
+: **Default**: none <br>
+A list of default tags to be added to every span and profile. Tags can be separated by commas or spaces, for example: `layer:api,team:intake,key:value` or `layer:api team:intake key:value`. Key-value pairs must be of string-convertible types.
 
 ### Agent  
 


KaibutsuX:
Assuming the file in this pr only applies to `go` code (go.md), then the description is still incorrect.
According to the parsing code:
```
func statsTags(c *config) []string {                                                                                              
    tags := globalconfig.StatsTags()                                                                                              
    if c.serviceName != "" {                                                                                                      
        tags = append(tags, "service:"+c.serviceName)                                                                             
    }                                                                                                                             
    // TODO: grab tracer config's env and hostname for globaltags                                                                 
    for k, v := range c.tags {                                                                                                    
        if vstr, ok := v.(string); ok {                                                                                           
            tags = append(tags, k+":"+vstr)                                                                                       
        }                                                                                                                         
    }                                                                                                                             
    return tags                                                                                                                   
}     
```
The doc states: 
```
Tags can be separated by commas or spaces, for example: `layer:api,team:intake,key:value` or `layer:api team:intake key:value`
```

Tags may not be in that format because they are parsed individually as a string:interface{} value. Multiple tags would not be supported.

maycmlee (commented):
Thanks @mtoffl01 for opening this PR! It looks good to me, but could you open a PR with the branch named in the format: `your-name/branch-name`?

darccio (approved):

mtoffl01:
Hi @KaibutsuX ,

This branch is closed in favor of another branch, but the information in the new PR will be the same.

Two sets of tags are referenced in the code snippet you provided: "global tags" and "local tags." 

`DD_TAGS`, or "global tags," accepts input like this: `DD_TAGS="key1:value2,key2:2,key3:true"` where comma and space separators are equally supported. Global tags are stored in a `map[string]interface{}` internally, but converted to a `[]string` that holds key-value pairs like so: `{"key1:value1","key2:value2"}` and is put onto our globalconfig ([Reference](https://github.com/DataDog/dd-trace-go/blob/main/ddtrace/tracer/option.go#L821-L836)). This array is later retrieved by contrib/database/sql ([Reference](https://github.com/DataDog/dd-trace-go/blob/main/contrib/database/sql/metrics.go#L54)), and merged with "local tags" set on contrib/database/sql (see [c.tags](https://github.com/DataDog/dd-trace-go/blob/main/contrib/database/sql/metrics.go#L59)), where "local tags" have been set via the [WithCustomTags](https://github.com/DataDog/dd-trace-go/blob/v1.70.0/contrib/database/sql/option.go#L256) option. The union of both sets of tags is used to populate the tags on our statsd client.

I'm not clear on what you mean here,
> Tags may not be in that format because they are parsed individually as a string:interface{} value. Multiple tags would not be supported.

But should you have any issue with DD_TAGS, let's re-route that discussion to the existing [dd-trace-go PR](https://github.com/DataDog/dd-trace-go/pull/3067).

Thanks!

PR #27162, opened 2025-01-16
(maint) Pin Vale to a specific version

hestonhoffman:
Pinning Vale to a slightly older version to get around a bug. Moving forward, we should probably stick to using a specific version of Vale instead of allowing the action to use the latest version. That way we can test the updates before we use them in CI.

### Merge instructions

Merge readiness:
- [x] Ready for merge

```
/merge
```


hestonhoffman:
Tested successfully

Screenshot 2025-01-16 at 9 25 08 AM

PR #27167, opened 2025-01-16
Fix broken pidMode link in ECS Fargate

mrdoggopat:
There was a broken link under the ECS Fargate tab, fixing this.


jhgilbert:
Hi @mrdoggopat, unfortunately the link is still broken -- I think the issue is that the link is not included inside the tab. It should be before the close tab, [like this link is](https://github.com/DataDog/documentation/pull/27167/files#diff-d239d38e542db1da81dd2aad4dc3672200cf3213277e4039c34d955532e5719bR283). Let me know if you have any questions!

estherk15 (approved):
Thanks for the fix!

PR #27218, opened 2025-01-21
init

retsguj:
### Merge instructions

Merge readiness:
- [ ] Ready for merge

```
/merge
```

### DIFF
--- a/layouts/partials/rbac-permissions-table.html
+++ b/layouts/partials/rbac-permissions-table.html
@@ -15,7 +15,7 @@
         {{- if ne .attributes.name nil -}}
         <tr>
           <td>
-            <code>{{- .attributes.name -}}</code>
+            {{- .attributes.display_name -}} (<code>{{- .attributes.name -}}</code>)
           </td>
           <td>
             {{- if ne .attributes.description_link nil -}}


layouts/partials/rbac-permissions-table.html
```diff
@@ -49,14 +49,16 @@
     <tbody>
       <tr>
         <td>
-          <a href="/logs/guide/logs-rbac-permissions/#logs_live_tail"><code>logs_live_tail</code></a>
+          <a href="/logs/guide/logs-rbac-permissions/#logs_live_tail">Logs Live Tail</a><br>
+          (<code>logs_live_tail</code>)
```
srosenthal-dd, line 53: worth noting: I never understood why this section is broken out separately and refers to these as "legacy" and "superseded" permissions. Out of scope for this particular task I think but I'd like to understand if the Logs team can help us clarify and simplify these.

layouts/partials/rbac-permissions-table.html
```diff
@@ -15,7 +15,7 @@
         {{- if ne .attributes.name nil -}}
         <tr>
           <td>
-            <code>{{- .attributes.name -}}</code>
+            {{- .attributes.display_name -}}<br>(<code>{{- .attributes.name -}}</code>)
```
srosenthal-dd, line 18: Tested on the [preview site](https://docs-staging.datadoghq.com/julia.gu/table-layout/account_management/rbac/permissions/#permissions-list) and it looks great! I believe @wangwillson1 had raised a concern that it might look crowded if we added another column but the parentheses solution is a nice simple one.

srosenthal-dd (approved):
It looks great to me, I'm OK with shipping!

In case you haven't shared with @aaronherrmann, he might like to know, but it doesn't seem controversial enough that he needs to weigh in. Screenshot for convenience from https://docs-staging.datadoghq.com/julia.gu/table-layout/account_management/rbac/permissions/#permissions-list: 
Screenshot 2025-01-21 at 5 05 49 PM

layouts/partials/rbac-permissions-table.html
retsguj: yeah i don't really get it either

nsollecito (approved):

PR #27424, opened 2025-02-03
Update manual_instrumentation.md adding .Net entry

gitstevenpham:
Adds `.Net` entry based on https://dd.slack.com/archives/C05C61J554L/p1738011042428139

### Merge instructions

Merge readiness:
- [x] Ready for merge

```
/merge
```

### DIFF
--- a/content/en/data_streams/manual_instrumentation.md
+++ b/content/en/data_streams/manual_instrumentation.md
@@ -141,8 +141,8 @@ var msg = consumer.Consume();
 var parentContext = new SpanContextExtractor().ExtractIncludingDsm(
     msg.Headers,
     GetHeader,
-    messageType: "<datastream-type>",
-    source: "<queue-or-topic-name>"
+    messageType: "<DATASTREAM-TYPE>",
+    source: "<QUEUE-OR-TOPIC-NAME>"
 );
 
 using (var scope = Tracer.Instance.StartActive("consume",
--- a/content/en/data_streams/manual_instrumentation.md
+++ b/content/en/data_streams/manual_instrumentation.md
@@ -96,7 +96,6 @@ set_consume_checkpoint(
 
 {{< /code-block >}}
 {{% /tab %}}
-{{% /tab %}}
 {{% tab ".Net" %}}
 <div class="alert alert-warning">
   <strong>Note:</strong> DSM instrumentation does not work in Async operations
--- a/content/en/data_streams/manual_instrumentation.md
+++ b/content/en/data_streams/manual_instrumentation.md
@@ -110,7 +110,7 @@ using Datadog.Trace;
 using (var scope = Tracer.Instance.StartActive("produce"))
 {
     var headers = new Headers();
-    var msg = new Message { Value = "any kind of value", Headers = headers};
+    var msg = new Message { Value = "<ANY-VALUE>", Headers = headers};
 
     new SpanContextInjector().InjectIncludingDsm(
         msg.Headers,
--- a/content/en/data_streams/manual_instrumentation.md
+++ b/content/en/data_streams/manual_instrumentation.md
@@ -97,7 +97,7 @@ set_consume_checkpoint(
 {{< /code-block >}}
 {{% /tab %}}
 {{% tab ".Net" %}}
-The following example propagates the trace context. See [Trace Context Propagation][1] for more information.
+The following example propagates the trace context. See [Trace Context Propagation][6] for more information.
 
 <div class="alert alert-warning">
   <strong>Note:</strong> In async operations, this may not work as expected because the context derived from the incoming message can be lost when producing a new message in different threads.
@@ -161,7 +161,7 @@ static IEnumerable<string?> GetHeader(Headers headers, string key)
 {
     yield return header.GetByKey(key);
 }
-{{< /code-block >}}[1]: https://docs.datadoghq.com/tracing/trace_collection/trace_context_propagation/?tab=net#additional-use-cases
+{{< /code-block >}}
 {{% /tab %}}
 {{< /tabs >}}
 ## Further Reading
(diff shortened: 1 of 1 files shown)


content/en/data_streams/manual_instrumentation.md
```diff
@@ -94,6 +94,69 @@ set_consume_checkpoint(
 "<datastream-type>", "<datastream-name>", headers.get
 )
 
+{{< /code-block >}}
+{{% /tab %}}
+{{% tab ".Net" %}}
+<div class="alert alert-warning">
+  <strong>Note:</strong> DSM instrumentation does not work in Async operations
```
vandonr, line 101: I wonder if this wording is not too strong... It will work, it's just that context is broken/lost on a `send` following a `receive`. I don't know how to convey that in a clear way.

content/en/data_streams/manual_instrumentation.md
```diff
@@ -94,6 +94,69 @@ set_consume_checkpoint(
...
+</div>
+
+In your producer
+{{< code-block lang="csharp" >}}
+using Datadog.Trace;
+
+using (var scope = Tracer.Instance.StartActive("produce"))
+{
+    var headers = new Headers();
+    var msg = new Message<Null, string> { Value = "any kind of value", Headers = headers};
```
vandonr, line 111:
```suggestion
    var msg = new Message { Value = "any kind of value", Headers = headers};
```
Any specific reason to have template parameters here ? I feel like they are distracting from what's actually necessary in this example

content/en/data_streams/manual_instrumentation.md
```diff
@@ -94,6 +94,69 @@ set_consume_checkpoint(
...
+using Datadog.Trace;
+
+var msg = consumer.Consume();
+var parentContext = new SpanContextExtractor().ExtractIncludingDsm(
+    msg.Headers,
+    GetHeader,
+    messageType: "<datastream-type>",
+    source: "<queue-or-topic-name>"
+);
+var startTime = DateTimeOffset.UtcNow;
```
vandonr, line 142:
```suggestion

var startTime = DateTimeOffset.UtcNow;
var msg = consumer.Consume();
var parentContext = new SpanContextExtractor().ExtractIncludingDsm(
    msg.Headers,
    GetHeader,
    messageType: "<datastream-type>",
    source: "<queue-or-topic-name>"
);
```
We force a different start time on the span because we want to capture the time it takes to execute the `Consume` operation, so we need to capture it before that.

vandonr (commented):
maybe we want to mention that doing that will _also_ propagate the trace context ?
We could also link to [the doc](https://docs.datadoghq.com/tracing/trace_collection/trace_context_propagation/?tab=net#additional-use-cases) explaining how to trace propagate _without_ DSM, which also has some nice examples of header getter/setter methods for some queues.

I'd even link in the other direction, from that doc linked above to here, like "hey, I see you want to inject trace context, how about injecting DSM at the same time ?"

content/en/data_streams/manual_instrumentation.md
gitstevenpham: How does:

`Note: In async operations, this may not work as expected because the context derived from the incoming message can be lost when producing a new message in different threads.`

Sound?

gitstevenpham:
> maybe we want to mention that doing that will _also_ propagate the trace context ? We could also link to [the doc](https://docs.datadoghq.com/tracing/trace_collection/trace_context_propagation/?tab=net#additional-use-cases) explaining how to trace propagate _without_ DSM, which also has some nice examples of header getter/setter methods for some queues.
> 
> I'd even link in the other direction, from that doc linked above to here, like "hey, I see you want to inject trace context, how about injecting DSM at the same time ?"

Added "The following example will propogate the trace context. See (Trace Context Propagation)[https://docs.datadoghq.com/tracing/trace_collection/trace_context_propagation/?tab=net#additional-use-cases] on how this is done in general."

content/en/data_streams/manual_instrumentation.md
```diff
@@ -94,6 +94,71 @@ set_consume_checkpoint(
...
+{{< /code-block >}}
+{{% /tab %}}
+{{% tab ".Net" %}}
+The following example propogates the trace context. See [Trace Context Propagation](https://docs.datadoghq.com/tracing/trace_collection/trace_context_propagation/?tab=net#additional-use-cases) on how this is done in general.
+
+<div class="alert alert-warning">
+  <strong>Note:</strong> In async operations, this may not work as expected because the context derived from the incoming message can be lost when producing a new message in different threads.
+</div>
+
+In your producer
```
buraizu, line 106:
```suggestion
#### Producer configuration

```

buraizu: Suggesting to add this is a sub-heading to make the distinction more visible, which also allows it to be linked to.

content/en/data_streams/manual_instrumentation.md
```diff
@@ -94,6 +94,71 @@ set_consume_checkpoint(
...
+}
+
+// Specific to how the header is modeled
+static void SetHeader(Headers headers, string key, string value)
+{
+    headers.Add(new Header(key, value));
+}
+{{< /code-block >}}
+
+In your consumer
```
buraizu, line 133:
```suggestion
#### Consumer configuration

```

content/en/data_streams/manual_instrumentation.md
```diff
@@ -94,6 +94,71 @@ set_consume_checkpoint(
...
+</div>
+
+In your producer
+{{< code-block lang="csharp" >}}
+using Datadog.Trace;
+
+using (var scope = Tracer.Instance.StartActive("produce"))
+{
+    var headers = new Headers();
+    var msg = new Message { Value = "any kind of value", Headers = headers};
```
buraizu, line 113:
```suggestion
    var msg = new Message { Value = "<ANY-VALUE>", Headers = headers};
```

buraizu (changes requested):
Requesting some minor edits

gitstevenpham:
@buraizu can you help with the linking. I can't seem to get it to work

buraizu (approved):

PR #27702, opened 2025-02-19
Clarify trace metric dependence on sampling

aw-dd:

Update serverless doc to clarify that overall trace metrics for a Lambda function are *only* calculated based on sampled invocations, not on 100% of traffic as would be the case with host-based tracing. 

Discussed during this escalation: https://datadoghq.atlassian.net/browse/APMS-13513

### Merge instructions

Merge readiness:
- [x] Ready for merge


buraizu (approved):

buraizu:
Verified in local build

PR #27797, opened 2025-02-25
Update aws.md to clarify containers can be billable

jhjacobson:
Clarifying that containers can be billable per customer's feedback

a customer pointed out that containers can actually be billable as [documented here](https://docs.datadoghq.com/account_management/billing/containers/). They asked if we could clarify that this is the case given that we call out custom metrics as billable.

### Merge instructions

I'm not sure if this will be more confusing since it isn't always the case that it's billable. But containers have an allotment similar to custom metrics so it seems fair to call it out here.
Merge readiness:
- [ ] Ready for merge

```
/merge
```


maycmlee (approved):

maycmlee:
Thanks @jhjacobson for opening this PR! I opened a new PR with the correct branch naming format (`<your_name>/<branch_name>`): https://github.com/DataDog/documentation/pull/28116, so will close this one.

PR #28013, opened 2025-03-06
added timeframe for change in unique tag values

ouracheal:
Added a note about the timeframe in which we grab the "change in unique tag values" column from the Custom Metrics Tags Cardinality Explorer.


jhgilbert:
Closing by request, we're not ready to have this info added yet.

PR #28066, opened 2025-03-10
GA Related Assets Facet - Update Docs

ouracheal:
Please approve, but only merge after feature is GA (PM will follow-up in thread)


buraizu (approved):

ouracheal:
Hi @buraizu! We're good to release these docs as the feature is now GA!

PR #28071, opened 2025-03-10
Update manage_data_collection.md

iadjivon:

This PR adds a clarification for what exclusions do and lets the user know that error events only match with the first rule. 

This update is motivated by a customer feedback from https://datadoghq.atlassian.net/browse/DOCS-10332

### Merge instructions

Merge readiness:
- [x] Ready for merge

```
/merge
```

### DIFF
--- a/content/en/error_tracking/manage_data_collection.md
+++ b/content/en/error_tracking/manage_data_collection.md
@@ -31,6 +31,8 @@ Each rule consists of:
 
 A given rule can be toggled on or off. An error event is included if it matches a query in one of the active inclusion filters _and_ it does not match any active nested exclusion queries.
 
+Each error event will be checked against the rules in order. The event is processed only by the first active rule it matches, and all subsequent rules are ignored. If the matched rule has an exclusion filter, the event will be excluded; otherwise, the event will be included.
+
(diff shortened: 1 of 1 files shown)


content/en/error_tracking/manage_data_collection.md
```diff
@@ -31,6 +31,8 @@ Each rule consists of:
 
 A given rule can be toggled on or off. An error event is included if it matches a query in one of the active inclusion filters _and_ it does not match any active nested exclusion queries.
 
+Each error event will be checked against the rules in order. The event is processed only by the first active rule it matches, and all subsequent rules are ignored. If the matched rule has an exclusion filter, the event will be excluded; otherwise, the event will be included.
```
maycmlee, line 34:
```suggestion
Each error event is checked against the rules in order. The event is processed only by the first active rule it matches, and all subsequent rules are ignored. If the matched rule has an exclusion filter, the event is excluded; otherwise, the event is included.
```

maycmlee (commented):
Just some small suggestions! We usually write to describe the present state of the product and avoid temporal words like "will".

maycmlee (approved):

PR #28254, opened 2025-03-19
add simple documentation for ddtrace-api package

emmettbutler:

This change adds basic documentation of the ddtrace-api Python package, recently released on PyPI.

### Merge instructions

Merge readiness:
- [x] Ready for merge

```
/merge
```


brett0000FF (commented):
@emmettbutler - Very cool! Thanks for adding this content.

I am going to be OOO for the rest of the week shortly, so I wanted to share a couple high-level thoughts before I can review in more detail next week.

It seems like we could simplify this quite a bit by integrating it into the existing [Datadog API](https://docs-staging.datadoghq.com/emmett.butler/dd-trace-api/tracing/trace_collection/custom_instrumentation/python/dd-api) page? From my quick scan, the main difference for users appears to be using `from dd_trace_api import tracer` instead of `from ddtrace import tracer`, plus installing an additional package.

The current tab structure (ddtrace-api, Datadog API, OpenTelemetry API) suggests these are parallel, alternative approaches, when ddtrace-api actually seems to be a complementary layer that works with the standard ddtrace implementation.

I might be missing certain technical context, but presenting this as a separate, duplicated documentation page might create confusion about the relationship between these packages and increase our maintenance burden when documentation needs to be updated.

emmettbutler:
@brett0000FF making this an extension of the Datadog API tab was the first approach I considered. I chose instead to make it a separate tab to make it as clear as possible that this is an early and unstable feature. Depending on how one looks at it, dd-trace-api can also be considered a parallel integration method to ddtrace and otel. The fact that it can be used as a drop-in replacement for the ddtrace API is important to the rollout plan, because we want to avoid ever requiring users to do a hard cutover from one method to the other.

If you think it's important enough, I could replace the new tab with a paragraph at the bottom of the Datadog API tab labelled "pre-release".

content/en/tracing/trace_collection/custom_instrumentation/python/ddtrace-api.md
```diff
@@ -0,0 +1,256 @@
...
+further_reading:
+    - link: 'tracing/other_telemetry/connect_logs_and_traces'
+      tag: 'Documentation'
+      text: 'Connect your Logs and Traces together'
+    - link: 'tracing/glossary/'
+      tag: 'Documentation'
+      text: 'Explore your services, resources, and traces'
+---
+
+*Note: the ddtrace-api Python package is in a pre-release state and may contain significant issues. If you'd rather use something more stable, use the [ddtrace API][9].*
```
wantsui, line 16: Worth wrapping this in a warning div, like DORA metrics: https://github.com/DataDog/documentation/blob/cb5309cc7a44fbe29207426501d66db9f04f5a8b/content/en/dora_metrics/_index.md?plain=1#L32, so this warning stands out more.

wantsui:
```suggestion
<div class="alert alert-warning">*Note: the ddtrace-api Python package is in a pre-release state and may contain significant issues. If you'd rather use something more stable, use the [ddtrace API][9].*</div> 

````

Comment (wantsui) [reply content/en/tracing/trace_collection/custom_instrumentation/python/ddtrace-api.md:16]: Also worth writing out some examples of issues to give an idea of what they could be.

Comment (wantsui) [inline content/en/tracing/trace_collection/custom_instrumentation/python/ddtrace-api.md:16]: Instead of `ddtrace API`, I think it's better to call this `Datadog API` to be consistent with the navigation bar.

Comment (bengl) [inline content/en/tracing/trace_collection/custom_instrumentation/python/ddtrace-api.md:16]: We're using "preview" as the terminology for these things, aren't we?

Comment (bengl) [reply content/en/tracing/trace_collection/custom_instrumentation/python/ddtrace-api.md:16]: Maybe, but it ought to be clear, maybe in parentheses, that this refers to the original `ddtrace` package.

File: content/en/tracing/trace_collection/custom_instrumentation/python/ddtrace-api.md [hunk 2]
```diff
@@ -0,0 +1,256 @@
...
+
+If you have not read the setup instructions for automatic instrumentation, start with the [Python Setup Instructions][6]
+
+If you aren't using supported library instrumentation (see [library compatibility][1]), you may want to manually instrument your code. You may also want to gain finer control
+over your application's instrumentation.
+
+## Additional Setup
+
+The stable public API for Datadog APM's custom Python instrumentation is implemented by the [ddtrace-api package][8]. This package doesn't implement any of the tracing functionality that creates and
```
bengl, line 25:
```suggestion
The stable public API for Datadog APM's custom Python instrumentation is implemented by the [ddtrace-api package][8]. This package doesn't implement any of the underlying functionality that creates and
```

At first glance I thought "wait, this does _does_ include tracing functionality, it's just implemented elsewhere". I think it may be clearer this way. Subjective/Non-blocking.

content/en/tracing/trace_collection/custom_instrumentation/python/ddtrace-api.md
```diff
@@ -0,0 +1,256 @@
...
+sends spans to Datadog; it *only* implements the API interface. This separation between interface in `ddtrace-api` and implementation in `ddtrace` allows users of cusom instrumentation to rely on
+an API that changes less frequently and more predictably, while allowing autoinstrumentation users to ignore semantic versioning entirely.
```
bengl, line 27:
```suggestion
an API that changes less frequently and more predictably, while allowing autoinstrumentation-only users to ignore API changes.
```

We (currently) use semver for purposes other than API management, such as deprecating support for things. Let's keep semver language out of this for now, so as to not lock us in to anything.

bengl (changes requested):

content/en/tracing/trace_collection/custom_instrumentation/python/ddtrace-api.md
emmettbutler: I reworded this to be less scary

brett0000FF:
> If you think it's important enough, I could replace the new tab with a paragraph at the bottom of the Datadog API tab labelled "pre-release".

@emmettbutler - I think I would be in favor of removing the separate tab for now and integrating this content into the existing DD API tab. I definitely see where you are coming from, but I think having this as a parallel tab, especially when in Preview, could create more confusion for users. Preview features are typically more hidden in the docs.

I'd be happy to take a crack at moving the content if that'd be helpful!

content/en/tracing/trace_collection/custom_instrumentation/python/dd-api.md
```diff
@@ -263,9 +263,45 @@ with tracer.trace("example") as span:
...
+<div class="alert alert-warning">The ddtrace-api Python package is in preview and may not include the API calls you need. If you'd rather use something more complete, use the API described above.
+The following steps are only necessary if you want to experiment with the preview `ddtrace-api` package.</div>
+
+A stable public API for Datadog APM's custom Python instrumentation is implemented by the [ddtrace-api package][8]. This package doesn't implement any of the underlying functionality that creates and
+sends spans to Datadog; it *only* implements the API interface. This separation between interface in `ddtrace-api` and implementation in `ddtrace` allows users of cusom instrumentation to rely on
+an API that changes less frequently and more predictably, while allowing autoinstrumentation-only users to ignore API changes. It also supports those users who integrate both single-step instrumentation
+and custom instrumentation by removing the need to depend on two copies of the `ddtrace` package.
```
brett0000FF, line 274:
```suggestion
The [ddtrace-api package][8] provides a stable public API for Datadog APM's custom Python instrumentation. This package implements only the API interface, not the underlying functionality that creates and sends spans to Datadog. 

This separation between interface (`ddtrace-api`) and implementation (`ddtrace`) offers several benefits:

- You can rely on an API that changes less frequently and more predictably for your custom instrumentation
- If you only use automatic instrumentation, you can ignore API changes entirely
- If you implement both single-step and custom instrumentation, you avoid depending on multiple copies of the `ddtrace` package
```
Just some light edits and breaking the content out.

content/en/tracing/trace_collection/custom_instrumentation/python/dd-api.md
```diff
@@ -263,9 +263,45 @@ with tracer.trace("example") as span:
...
+```python
+  from dd_trace_api import tracer
+
+  @tracer.wrap(service="my-sandwich-making-svc", resource="resource_name")
+  def get_ingredients():
+      # go to the pantry
+      # go to the fridge
+      # maybe go to the store
+      return
+```
```
brett0000FF, line 302:
`````suggestion
To use `ddtrace-api`:

1. Install both the `ddtrace` and `ddtrace-api` libraries:
   ```python
   pip install 'ddtrace>=3.1' ddtrace-api
   ```

2. Instrument your Python application using `ddtrace-run` by prefixing your Python entry-point command:
   ```shell
   ddtrace-run python app.py
   ```

3. After this is set up, you can write custom instrumentation exactly like the examples in the previous sections, but you import from `dd_trace_api` instead of `ddtrace`.

   For example:
   ```python
   from dd_trace_api import tracer
   
   @tracer.wrap(service="my-sandwich-making-svc", resource="resource_name")
   def get_ingredients():
       # go to the pantry
       # go to the fridge
       # maybe go to the store
       return
   ```
`````
Basically the same content, but this puts it into a numbered list.

brett0000FF (approved):
@emmettbutler - Looks great, thank you! I just left a few suggestions with minor edits for readability.

emmettbutler:
I'm ready to merge this. @brett0000FF looks like I don't have permission to merge, maybe you can help

PR #28277, opened 2025-03-20
Cdocs pilot prep

jhgilbert:
This is a PR from a feature branch to another feature branch, no review is needed aside from @hestonhoffman

- Add any usable scraped config
- Remove old test page / old test config
- Migrate the first pilot page
- Include a local tarball
- Build the author console at localhost:1313/cdocs/console
- Ignore the author console files in Git

LMK if you want me to undo any Makefile changes etc., happy to fix whatever is needed.


hestonhoffman (approved):
This looks great. I like the trait descriptions. Let's keep the hook in dependencies for now. If I figure out why adding it to start wasn't working, we can revisit, but I don't think it matters that much. I can ask someone in web about it too.

PR #28364, opened 2025-03-25
Update _index.md to include iOS entitlements for MAT

thang62590:
Include section for iOS entitlements

### DIFF
--- a/content/en/synthetics/troubleshooting/_index.md
+++ b/content/en/synthetics/troubleshooting/_index.md
@@ -147,7 +147,7 @@ Synthetic tests by default do not [renotify][12]. This means that if you add you
 
 If there are security checks during application startup, such as verifying if USB debugging is enabled, Datadog recommends uploading a version of the application that does not contain these checks. 
 
-### Ensuring Your App Functions Properly During Test Recording and Test Execution
+### Ensure proper app functioning during test recording and execution
 
 If certain features of your iOS app are not functioning as expected during test recording or test execution, it could be a result of the app resigning process. This resigning process is required so that mobile devices can trust the provided application, and so the resigning process could remove iOS entitlements that might be essential and cause issues (e.g. access to Contacts, Camera, Keychain, Photos, Health Kit, Home Kit, etc). To minimize the risk of entitlement-related issues, we recommend distributing your iOS app using Ad Hoc or Development provisioning profiles for improved compatibility. While using an Enterprise provisioning profile is an option, you might encounter challenges when using it.
 
--- a/content/en/synthetics/troubleshooting/_index.md
+++ b/content/en/synthetics/troubleshooting/_index.md
@@ -147,6 +147,10 @@ Synthetic tests by default do not [renotify][12]. This means that if you add you
 
 If there are security checks during application startup, such as verifying if USB debugging is enabled, Datadog recommends uploading a version of the application that does not contain these checks. 
 
+### Ensuring Your App Functions Properly During Test Recording and Test Execution
+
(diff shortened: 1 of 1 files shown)


content/en/synthetics/troubleshooting/_index.md
```diff
@@ -147,6 +147,10 @@ Synthetic tests by default do not [renotify][12]. This means that if you add you
 
 If there are security checks during application startup, such as verifying if USB debugging is enabled, Datadog recommends uploading a version of the application that does not contain these checks. 
 
+### Ensuring Your App Functions Properly During Test Recording and Test Execution
```
buraizu, line 150:
```suggestion
### Ensure proper app functioning during test recording and execution
```

content/en/synthetics/troubleshooting/_index.md
```diff
@@ -147,6 +147,10 @@ Synthetic tests by default do not [renotify][12]. This means that if you add you
...
+
+If certain features of your iOS app are not functioning as expected during test recording or test execution, it could be a result of the app resigning process. This resigning process is required so that mobile devices can trust the provided application, and so the resigning process could remove iOS entitlements that might be essential and cause issues (e.g. access to Contacts, Camera, Keychain, Photos, Health Kit, Home Kit, etc). To minimize the risk of entitlement-related issues, we recommend distributing your iOS app using Ad Hoc or Development provisioning profiles for improved compatibility. While using an Enterprise provisioning profile is an option, you might encounter challenges when using it.
```
buraizu, line 152:
```suggestion
If certain features of your iOS app are not functioning as expected during test recording or test execution, it could be a result of the app resigning process. This resigning process is required so that mobile devices can do the following:
- Trust the provided application
- Remove iOS entitlements that might be essential and cause issues (such as access to Contacts, Camera, Keychain, Photos, Health Kit, Home Kit, and so on)

To minimize the risk of entitlement-related issues, and for improved compatibility, Datadog recommends distributing your iOS app using Ad Hoc or Development provisioning profiles. While using an Enterprise provisioning profile is an option, you might encounter challenges when using it.
```

buraizu: In addition to the above edits, would it be possible to provide more detail about the challenges a user might face using the Enterprise provisioning profile? As written, this is likely to generate support tickets with customers asking about the challenges.

buraizu (changes requested):
Thanks for the PR; I've requested a few updates for clarity and per the guidelines outlined in the [documentation style guide](https://github.com/DataDog/documentation/blob/master/CONTRIBUTING.md). Let me know if you have any questions!

content/en/synthetics/troubleshooting/_index.md
thang62590: The suggested edits are mixing up the cause and effects.

Because the app resigning process is required, this could mean that
1. the original certificate is replaced by a new certificate or removed all together. If this new certificate isn’t from a trusted authority on the device, the mobile device may not trust the app which means the app will not be usable in Mobile Application Testing.
2. certain app entitlements, like access to device features (e.g., camera, microphone, or push notifications), can be stripped or invalidated during the resigning process. If the new signature lacks the proper entitlements, the app may not work as intended.

We don't know the exact challenges that a user might face when using the Enterprise provisioning profile--debugging these types of issues are difficult. And so, if customers open support tickets we'll have more information/data to work and as a result we can definitely update the docs to include more concrete examples.

buraizu: @thang62590 thanks for clarifying; I've updated the suggested edit.

Thanks also for clarifying about the Enterprise provisioning profile. Since this approach is untested, and Datadog recommends the Ad Hoc or Developer approach, my suggestion is to remove this content. Please feel free to follow up internally if you'd like to discuss further about the rationale for this.

buraizu (approved):

PR #28366, opened 2025-03-25
Fix customization menu behavior on small screens

jhgilbert:
This is a PR against a feature branch, it doesn't need a review aside from @hestonhoffman.

Most of the CSS updates are in the updated tarball, but I did nudge the documentation mobile TOC icon upward to avoid an awkward clash with the customization menu. If websites doesn't want to nudge it up on every page, we can nudge it up just on customized pages using JS.

Fixes:
- Dropdowns now stack nicely at small screen sizes, they were too tightly spaced before
- A gap appearing for smaller screen sizes at the very top of the page has been closed, scrolling text was peeking through before
- *.mdoc.md files should now be ignored by Hugo (they were publishing as path/to/page.mdoc, with weird looking output)

The TOC icon floats in front of the menu options at some screen sizes, but we seem to want it to float like that, since the other page content does not clear it either. I could update the customization menu to stop short of where the icon is, but the menu would look oddly narrow in other contexts. The user will likely make the majority of their selections before the menu sticks to the top of the page anyway.

Once the build runs, you should be able to squish the UI at https://docs-staging.datadoghq.com/jen.gilbert/cdocs-css-fixes/real_user_monitoring/guide/proxy-mobile-rum-data


buraizu:
Removed the review request from documentation for now, as it sounds like this doesn't need documentation team review currently. Please feel free to re-request a review if that's not the case.

hestonhoffman (approved):

hestonhoffman:
@jhgilbert Approved! I've noticed the TOC icon issue before - I think there might even be a ticket open for it somewhere. But I agree that it's out of scope for this

PR #28627, opened 2025-04-08
remove old ssi migration documentation

arbll:
Remove an old migration doc.

Motivation: we want to remove mention to the install script in the doc itself.


raphaelgavache (approved):

jhgilbert:
Hi @raphaelgavache, thanks for the update! I recommend removing just the link for now, not the doc itself. This is because translated versions of the doc trail behind the English version, and will still contain links to the file for a little while. Let me know if you have any questions on that. Thank you!

jhgilbert (approved):

PR #28643, opened 2025-04-08
Add documentation for new SQL functions available for Log Workspaces

vishalshah-dd:

A follow-up of #28591 where we're adding documentation for new functions that are now supported in Log Workspaces. This is in preparation for our GA launch. 

Adding documentation for the following functions: 
- `cardinality`
- `array_position`
- `strpos`
- `string_to_array`
- `split_part`
- `array_agg`
- `unnest`

Jira: https://datadoghq.atlassian.net/browse/LOGSQ-830

### Merge instructions

Merge readiness:
- [x] Ready for merge

**For Datadog employees**:

```
/merge
```

### DIFF
--- a/content/en/logs/workspaces/sql_reference.md
+++ b/content/en/logs/workspaces/sql_reference.md
@@ -97,7 +97,7 @@ The following SQL functions are supported. For Window function, see the separate
 | `to_char(timestamp t, string format)`            | string                                | Converts a timestamp to a string according to the given format.             |
 | `date_trunc(string unit, timestamp t)`           | timestamp                             | Truncates a timestamp to a specified precision based on the provided unit.  |
 | `regexp_like(string s, pattern p)`               | boolean                               | Evaluates whether a string matches a regular expression pattern.                 |
-
+| `cardinality(array a)`                           | integer                               | Returns the number of elements in the array, or 0 if the array is empty.    |
 
 {{% collapse-content title="Examples" level="h3" %}}
 
@@ -336,6 +336,14 @@ FROM
   emails
 WHERE
   regexp_like(email_address, '@example\.com$')
+{{< /code-block >}}
+
+### `CARDINALITY`
+{{< code-block lang="sql" >}}
+SELECT
+  CARDINALITY(recipients)
+FROM
+  emails
 {{< /code-block >}}
 
 {{% /collapse-content %}} 
--- a/content/en/logs/workspaces/sql_reference.md
+++ b/content/en/logs/workspaces/sql_reference.md
@@ -92,7 +92,7 @@ The following SQL functions are supported. For Window function, see the separate
 | `trim(string s)`                                 | string                                | Removes leading and trailing whitespace from the string.                    |
 | `replace(string s, string from, string to)`      | string                                | Replaces occurrences of a substring within a string with another substring. |
 | `substring(string s, int start, int length)`     | string                                | Extracts a substring from a string, starting at a given position and for a specified length. |
-| `strpos(string s, string substring)`             | integer                               | Returns the first index position of the substring in a given string, or 0 if there is no match |
+| `strpos(string s, string substring)`             | integer                               | Returns the first index position of the substring in a given string, or 0 if there is no match. |
 | `split_part(string s, string delimiter, integer index)` | string                         | Splits the string on the given delimiter and returns the string at the given position counting from one. |
 | `extract(unit from timestamp/interval)`          | numeric                               | Extracts a part of a date or time field (such as year or month) from a timestamp or interval. |
 | `to_timestamp(string timestamp, string format)`  | timestamp                             | Converts a string to a timestamp according to the given format.             |
@@ -103,7 +103,7 @@ The following SQL functions are supported. For Window function, see the separate
 | `array_position(array a, typeof_array value)`    | integer                               | Returns the index of the first occurrence of the value found in the array, or null if value is not found. |
 | `string_to_array(string s, string delimiter)`    | array of strings                      | Splits the given string into an array of strings using the given delimiter. |
 | `array_agg(expression e)`                        | array of input type                   | Creates an array by collecting all the input values.                        |
-| `unnest(array a [, array b...])`                 | type of array                         | Expands arrays into a set of rows. This form is only allowed in a FROM clause. |
+| `unnest(array a [, array b...])`                 | rows of a [, b...]                    | Expands arrays into a set of rows. This form is only allowed in a FROM clause. |
 
 {{% collapse-content title="Examples" level="h3" %}}
 
--- a/content/en/logs/workspaces/sql_reference.md
+++ b/content/en/logs/workspaces/sql_reference.md
@@ -98,6 +98,7 @@ The following SQL functions are supported. For Window function, see the separate
 | `date_trunc(string unit, timestamp t)`           | timestamp                             | Truncates a timestamp to a specified precision based on the provided unit.  |
 | `regexp_like(string s, pattern p)`               | boolean                               | Evaluates whether a string matches a regular expression pattern.                 |
 | `cardinality(array a)`                           | integer                               | Returns the number of elements in the array, or 0 if the array is empty.    |
+| `array_position(array a, typeof_array value)`    | integer                               | Returns the index of the first occurence of the value found in the array.   |
 
 {{% collapse-content title="Examples" level="h3" %}}
 
@@ -342,6 +343,14 @@ WHERE
 {{< code-block lang="sql" >}}
 SELECT
   CARDINALITY(recipients)
+FROM
+  emails
+{{< /code-block >}}
+
+### `ARRAY_POSITION`
+{{< code-block lang="sql" >}}
+SELECT
+  ARRAY_POSITION(recipients, 'hello@example.com')
 FROM
   emails
 {{< /code-block >}}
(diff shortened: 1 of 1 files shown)


content/en/logs/workspaces/sql_reference.md
```diff
@@ -92,12 +92,18 @@ The following SQL functions are supported. For Window function, see the separate
...
 | `to_timestamp(string timestamp, string format)`  | timestamp                             | Converts a string to a timestamp according to the given format.             |
 | `to_char(timestamp t, string format)`            | string                                | Converts a timestamp to a string according to the given format.             |
 | `date_trunc(string unit, timestamp t)`           | timestamp                             | Truncates a timestamp to a specified precision based on the provided unit.  |
 | `regexp_like(string s, pattern p)`               | boolean                               | Evaluates if a string matches a regular expression pattern.                 |
-
+| `cardinality(array a)`                           | integer                               | Returns the number of elements in the array, or 0 if the array is empty.    |
+| `array_position(array a, typeof_array value)`    | integer                               | Returns the index of the first occurence of the value found in the array.   |
```
petternasholm, line 103: - Spelling on "occurence"
- Consider mentioning what happens when you don't find the value.
- Also, just curious where you got the `typeof_array` from? I'm not necessarily against it, was just trying to figure out where it's from.

content/en/logs/workspaces/sql_reference.md
```diff
@@ -92,12 +92,18 @@ The following SQL functions are supported. For Window function, see the separate
...
+| `string_to_array(string s, string delimiter)`    | array of strings                      | Splits the given string into an array of strings using the given delimiter. |
+| `array_agg(expression e)`                        | array of input type                   | Create an array by concatenating the input values. Ordering is not deterministic |
```
petternasholm, line 105: "Concatenating" doesn't feel right here. High level, I'd look at the first variant mentioned at https://www.postgresql.org/docs/17/functions-aggregate.html.

I'm tempted to omit the fact about determinism just to keep things simple.

Also, should be "Creates" for consistency. End with a period.

content/en/logs/workspaces/sql_reference.md
```diff
@@ -92,12 +92,18 @@ The following SQL functions are supported. For Window function, see the separate
...
+| `unnest(array a)`                                | type of array                         | Expands an array into a set of rows.                                        |
```
petternasholm, line 106: Our impl is actually the 2nd variant from https://www.postgresql.org/docs/17/functions-array.html. Although it looks there like you need two args, which is not the case (even for Postgres). Worth noting is that it only works in a FROM clause.

content/en/logs/workspaces/sql_reference.md
```diff
@@ -231,6 +237,18 @@ FROM
   books
 {{< /code-block >}}
 
+### `STRPOS`
+{{< code-block lang="sql" >}}
+SELECT
+  STRPOS('high', 'ig')
```
petternasholm, line 243: Minor: Maybe we should use a different example than the Postgres docs? Just to not make it look like it was copied.

content/en/logs/workspaces/sql_reference.md
```diff
@@ -338,6 +356,53 @@ WHERE
...
+SELECT 
+  STRING_TO_ARRAY('a,b,c,d,e,f', ',')
+{{< /code-block >}}
+
+### `ARRAY_AGG`
+{{< code-block lang="sql" >}}
+SELECT 
+  ARRAY_AGG(x) arr1, 
+  ARRAY_AGG(all x) arr2, 
+  ARRAY_AGG(distinct x) arr3,
```
petternasholm, line 386: Upper case ALL and DISTINCT.

content/en/logs/workspaces/sql_reference.md
```diff
@@ -338,6 +356,53 @@ WHERE
...
+);
+{{< /code-block >}}
+
+### `UNNEST`
+{{< code-block lang="sql" >}}
+SELECT 
+  *, 
+  UNNEST(recipients)
+FROM 
+  emails;
```
petternasholm, line 403: Not supported (just give it a try in the UI). Needs to be used in FROM clause.

In fact, it would be good if you could sanity check all your added queries by issuing them in prod LW UI. You can still use a table like `emails` by just defining that in a separate cell and put in whatever data you want.

(If you really wanted to go the extra mile with that, you could create a workspace we'd dedicate for documentation examples (that we later can expand). In addition to the sanity check, this would help any documentation writer understand features better.)

content/en/logs/workspaces/sql_reference.md
```diff
@@ -338,6 +356,53 @@ WHERE
...
+{{< /code-block >}}
+
+### `ARRAY_AGG`
+{{< code-block lang="sql" >}}
+SELECT 
+  ARRAY_AGG(x) arr1, 
+  ARRAY_AGG(all x) arr2, 
+  ARRAY_AGG(distinct x) arr3, 
+  ARRAY_AGG(y) arr4
+FROM (
```
petternasholm, line 388: How about continuing to work with your emails theme here instead of these cryptic looking values (that I think I used somewhere)?

Also, I wonder if the UNION ALL stuff here just complicates things for users in terms of understanding.

Maybe you could just do something like aggregating email addresses from a user table. That can still show off DISTINCT (and ALL, which actually isn't useful as it doesn't do anything more than omitting a keyword here).

content/en/logs/workspaces/sql_reference.md
```diff
@@ -338,6 +356,53 @@ WHERE
...
+    SELECT 1 x, 4.1 y
+    UNION ALL
+    SELECT 2 x, 5.1 y
+    UNION ALL
+    SELECT 2 x, 6.1 y
+);
```
petternasholm, line 394: For consistency, end without semicolon. Ditto below.

content/en/logs/workspaces/sql_reference.md
vishalshah-dd: Thanks, fixing those. I didn't take `typeof_array` from anywhere, but needed to find a way to mention that. I didn't want to use a space as that might be confusing since we've been using a `type variable_name` format

content/en/logs/workspaces/sql_reference.md
vishalshah-dd: Thanks! That's actually what I was doing in [this workspace](https://app.datadoghq.com/logs/workspace/18740a38-ef3b-42b8-bed3-2b1818c5c0e7?from=1743794896553&to=1743795796553). Think I didn't copy-paste and made a mistake while writing it out. I'll double check the others

content/en/logs/workspaces/sql_reference.md
vishalshah-dd: Thanks I changed it to "collect" based on the docs, not sure if there's a better word though

content/en/logs/workspaces/sql_reference.md
petternasholm: Nice! Consider documenting this so it can be reused later.

content/en/logs/workspaces/sql_reference.md
```diff
@@ -92,18 +92,18 @@ The following SQL functions are supported. For Window function, see the separate
 | `trim(string s)`                                 | string                                | Removes leading and trailing whitespace from the string.                    |
 | `replace(string s, string from, string to)`      | string                                | Replaces occurrences of a substring within a string with another substring. |
 | `substring(string s, int start, int length)`     | string                                | Extracts a substring from a string, starting at a given position and for a specified length. |
-| `strpos(string s, string substring)`             | integer                               | Returns the index position of the substring in a given string.              |
-| `split_part(string s, string delimiter, integer index)` | string                         | Splits the string on the given delimiter and returns the string as the given position counting from 1. |
+| `strpos(string s, string substring)`             | integer                               | Returns the first index position of the substring in a given string, or 0 if there is no match |
```
petternasholm, line 95: end w/ period

content/en/logs/workspaces/sql_reference.md
```diff
@@ -92,18 +92,18 @@ The following SQL functions are supported. For Window function, see the separate
...
+| `array_position(array a, typeof_array value)`    | integer                               | Returns the index of the first occurrence of the value found in the array, or null if value is not found. |
 | `string_to_array(string s, string delimiter)`    | array of strings                      | Splits the given string into an array of strings using the given delimiter. |
-| `array_agg(expression e)`                        | array of input type                   | Create an array by concatenating the input values. Ordering is not deterministic |
-| `unnest(array a)`                                | type of array                         | Expands an array into a set of rows.                                        |
+| `array_agg(expression e)`                        | array of input type                   | Creates an array by collecting all the input values.                        |
+| `unnest(array a [, array b...])`                 | type of array                         | Expands arrays into a set of rows. This form is only allowed in a FROM clause. |
```
petternasholm, line 106: In the PG docs, they call the return type "setof anyelement, anyelement [, ... ]".

Instead of "type of array", maybe we could say "rows of a, b, ..." or something like that.

petternasholm (approved):

content/en/logs/workspaces/sql_reference.md
vishalshah-dd: Hmm thats true. `unnest` is a little special with the way it doesn't return specific values. I'll change it to `rows of a, b, ...`

PR #28684, opened 2025-04-09
remove unnecessary 'a' before 'Datadog' from Trace Context Propagation definition

dharapa:
I believe there is an unnecessary _a_ before _Datadog_ in the APM Terms and Concepts Overview table


buraizu:
Verified in local build

PR #28747, opened 2025-04-14
Standardize case sensitivity for SQL functions

vishalshah-dd:

This PR aims to standardize the case sensitivity of SQL functions in the documentation. We're taking an approach where: =
- lower case function names are used in tables when listing down functions 
- upper case function names are used in sample SQL queries where other key words are also in all upper case

This formatting behavior is also seen in docs like [Postgres](https://www.postgresql.org/docs/current/functions-window.html) and personally prefer it as lower case function names are easier on the eyes when reading a list, and upper case function names in SQL they're more widely used. 

Jira: https://datadoghq.atlassian.net/browse/LOGSQ-830

### Merge instructions

Merge readiness:
- [ ] Ready for merge

**For Datadog employees**:

```
/merge
```

### DIFF
--- a/content/en/logs/workspaces/sql_reference.md
+++ b/content/en/logs/workspaces/sql_reference.md
@@ -141,7 +141,7 @@ GROUP BY service_name
 {{< /code-block >}} 
 
 ### `CEIL`
-{{< code-block lang="sql" >}} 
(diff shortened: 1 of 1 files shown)


brett0000FF (approved):
Thanks for these updates!

PR #28983, opened 2025-04-25
[DSM] Add technology pages for DSM documentation

lkretvix:

https://datadoghq.atlassian.net/browse/DSMON-685

Customers can find it difficult to find what technologies are supported by DSM. In order to do this they need to go to each language and look at the tech libraries there. Let's create an Inversion where users can view the _technology_ and find the languages/libraries that are supported

Also some cleanup on duplicated sections of content, converting to shortcodes

### QA

https://docs-staging.datadoghq.com/lucas.kretvix/dsm-add-new-tech-pages/data_streams/technologies/rabbitmq/

### Merge instructions

Merge readiness:
- [x] Ready for merge

**For Datadog employees**:

```
/merge
```


content/en/data_streams/technologies/azure_service_bus.md
```diff
@@ -0,0 +1,32 @@
...
+  </thead>
+  <tbody>
+    <tr>
+      <td><a href="/data_streams/dotnet">.NET</a></td>
+      <td><a href="https://www.nuget.org/packages/Azure.Messaging.ServiceBus">Azure.Messaging.ServiceBus</a></td>
+      <td>2.53.0 </td>
+      <td>2.53.0 or later </td>
+    </tr>
+  </tbody>
+</table>
```
lkretvix, line 30: decided to use html table styling for these for consistency. Some of the tech tables have cells that span multiple columns (kafka's Go language has multiple libraries for ex.)

johannbotha:
image
Thanks for adding these, but it looks like it doesn't scroll to the right section. It takes you to the main page, but the setup matrix is still past the fold.

labbati (commented):
Thanks for the work @lkretvix, much clearer than before.

Just a few notes:
-  Any "Note" that prevents a mis-understanding, should be up front. For example, in `kafka.md` there is a not at the bottom related to kafka streams not fully supported. That should be more prominent, e.g. above the table or in any format that you find useful.
- I would try to avoid repetition of supported library names, versions, and tracer versions in two places (the tech page and the lang page). I know for experience that they _will_ go out of sync in weeks, not years. At the very least, I would extract them in importable fragments.
- In my opinion, it is not clear, from a page such as `kafka.md`, that they have to click the language name in the table's first column to proceed to per lang instructions. Should we add the language tiles all over again? This makes me question is we should remove the language tiles from the index page, but let's ignore that for a moment.

labbati:
> * The repetition is tough to get around. I can give myself a task to move them to shortcodes, but that means every version cell gets its own file. Not impossible to do, just takes time

That's be too much, so in this case feel free to ignore my comment. I was thinking more something like having the whole per-lang table to its own file and then maybe tabs, but we can do it once we see things get out of syncs.

> We do have a callout to go to the language page to see how to set up DSM on those technology pages, do you think that's not enough?

By callout you mean the language name on the first column, correct? It took me a while to get it as a _call to action_, but not a blocker for me. I am thinking if it would be worth to replace `Java` with `Setup Java` to make it more explicit, but probably worthless.

johannbotha:
> > Thanks for adding these, but it looks like it doesn't scroll to the right section. It takes you to the main page, but the setup matrix is still past the fold.
> 
> @johannbotha Not sure if there's a way to force it. I'm adding the anchor tag there but it doesn't seem to want to scroll 🤔

Can you reach out to the docs folks. The goal of the sidenav is to easily get the users to that matrix. I bet it's something with the sticky header or needing the id tag to match the sidenav.

johannbotha:
> > Thanks for the work @lkretvix, much clearer than before.
> > Just a few notes:
> > 
> > * Any "Note" that prevents a mis-understanding, should be up front. For example, in `kafka.md` there is a `Note` at the bottom related to kafka streams not fully supported. That should be more prominent, e.g. above the table or in any format that you find useful.
> > * I would try to avoid repetition of supported library names, versions, and tracer versions in two places (the tech page and the lang page). I know for experience that they _will_ go out of sync in weeks, not years. At the very least, I would extract them in importable fragments.
> > * In my opinion, it is not clear, from a page such as `kafka.md`, that they have to click the language name in the table's first column to proceed to per lang instructions. Should we add the language tiles all over again? This makes me question is we should remove the language tiles from the index page, but let's ignore that for a moment.
> 
> @labbati
> 
> * Happy to move the `Note` section higher up 👍
> * The repetition is tough to get around. I can give myself a task to move them to shortcodes, but that means every version cell gets its own file. Not impossible to do, just takes time
> * We do have a callout to go to the language page to see how to set up DSM on those technology pages, do you think that's not enough?
> 
> image

There should be a way to do this without "each cell getting it's own file." We're using Hugo and the markdown rendering engine and hugo supports the use of variables. See https://gohugo.io/templates/introduction/#variables

johannbotha (approved):
LGTM, thanks @lkretvix !

Would be great to schedule a swarm to add any small things we're missing now that we have a platform for tech specific things.

nsollecito (approved):

PR #29036, opened 2025-04-29
Bwest/community libraries restructure

brandonmwest:
Move from free-form community libraries list to yaml/shortcode like the other library types. We use this machine readable format to import library information into devhub.


hestonhoffman:
👋 A fair amount of these are 404ing on GitHub. Do you think we should remove them or make a best effort to find replacements?

brandonmwest:
> 👋 A fair amount of these are 404ing on GitHub. Do you think we should remove them or make a best effort to find replacements?

Thanks, should be sorted out now, I've removed the old libraries

hestonhoffman (approved):

davidejones (approved):

PR #29192, opened 2025-05-06
Update libraries.yaml

brandonmwest:
Add descriptions to libraries that were missing them


brandonmwest:
Now that the libraries have descriptions, the table doesn't look great and there's an h-scrollbar for some reason. 
image

A few solutions:
Truncate the descriptions
Update the templates to be card-like instead of tables
Deprecate this page and link out to devhub (which uses the same .yaml file)

fitzage (approved):

PR #29239, opened 2025-05-08
Remove incorrect DORA limitation about rebase

GabrielAnca:

Removes a limitation that was added incorrectly: for rebased branches and commits, we consider the timestamp at which the original commit was created. This is the desired logic, so no need to make any clarification about it.

### Merge instructions

Merge readiness:
- [x] Ready for merge

### DIFF
--- a/content/en/dora_metrics/setup/deployments.md
+++ b/content/en/dora_metrics/setup/deployments.md
@@ -299,7 +299,6 @@ If the two metadata entries are defined for a service, only `extensions[datadogh
 - Change lead time is not available for the first deployment of a service that includes Git information.
 - Change lead time is not available if the most recent deployment of a service was more than 60 days ago.
 - The Change Lead Time calculation includes a maximum of 5000 commits per deployment.
-- For rebased branches, *change lead time* calculations consider the new commits created during the rebase, not the original commits.
 - When using "Squash" to merge pull requests:
   - For GitHub and GitLab: Metrics are emitted for the original commits.
   - For other Git providers: Metrics are emitted for the new commit added to the target branch.


content/en/dora_metrics/setup/deployments.md
```diff
@@ -299,7 +299,6 @@ If the two metadata entries are defined for a service, only `extensions[datadogh
 - Change lead time is not available for the first deployment of a service that includes Git information.
 - Change lead time is not available if the most recent deployment of a service was more than 60 days ago.
 - The Change Lead Time calculation includes a maximum of 5000 commits per deployment.
-- For rebased branches, *change lead time* calculations consider the new commits created during the rebase, not the original commits.
```
GabrielAnca, line 302: I thought of rewording this sentence, but this is part of a section called 'limitations', and there are no limitations in the way rebased branches/commits are handled

henrywalsh: I wonder if it's worth calling out that it we will show the rebased commit sha?

Perhaps that's too in depth for what we need in this list and doesn't actually provide value to the customer. 🤷

henrywalsh (approved):

content/en/dora_metrics/setup/deployments.md
GabrielAnca: Great point! I was thinking about this and I think it is a bit tricky, because in this case it doesn't really matter whether we show the rebased commit or the original commit: they are identical to each other (except for the 'committer' timestamp). The rebased commit has both the original (author) timestamp and the new (committer) timesamp, so in a way it could be more useful than the original commit 😅

So I'm okay with not mentioning it in the docs, it feels like that detail may create confusion and not provide value. WDYT?

henrywalsh: I think I agree with that. It's a fine line of providing helpful information for customers and providing information that may be confusing or expose some detail that is irrelevant. 

Let's remove it from the docs altogether. If, similar to CLT calculations, we start seeing an increase in questions around this, we can reconsider at that time!

rtrieu (approved):

PR #29377, opened 2025-05-15
RUM SDK: add more details on view lifecycle and beforeSend

bcaudan:

Document view event specificity to help customer integration with beforeSend

### Merge instructions

Merge readiness:
- [ ] Ready for merge

**For Datadog employees**:

```
/merge
```

### DIFF
--- a/content/en/real_user_monitoring/browser/advanced_configuration.md
+++ b/content/en/real_user_monitoring/browser/advanced_configuration.md
@@ -552,6 +552,21 @@ You can update the following event properties:
 
 The RUM Browser SDK ignores modifications made to event properties not listed above. For more information about event properties, see the [RUM Browser SDK GitHub repository][15].
 
+**Note**: Unlike other events, view events are sent multiple times to Datadog to reflect the updates occurring during their lifecycle. An update on a previous view event can still be sent while a new view is active. So you should be mindful of this behavior when modifying the content of a view event.
+
+```javascript
+beforeSend: (event) => {
+    // discouraged, as the current view name could be applied the active view and on the previous views
+    event.view.name = getCurrentViewName()
+
+    // recommended
+    event.view.name = getViewNameForUrl(event.view.url)
+}
+```
+
+
+
+
 ### Discard a RUM event
 
 With the `beforeSend` API, discard a RUM event by returning `false`:


content/en/real_user_monitoring/browser/advanced_configuration.md
```diff
@@ -552,6 +552,21 @@ You can update the following event properties:
 
 The RUM Browser SDK ignores modifications made to event properties not listed above. For more information about event properties, see the [RUM Browser SDK GitHub repository][15].
 
+**Note**: Unlike other events, view events are sent multiple times to Datadog to reflect the updates occurring during their lifecycle. An update on a previous view event can still be sent while a new view is active. So you should be mindful of this behavior when modifying the content of a view event.
```
janine-c, line 555: I'm not clear about what behaviour the user would observe in the UI if they modified the view event content incorrectly (updating the view name?) or how this code sample is supposed to help users avoid that problem, or why the code sample is separate from the table that also contains information about `view.name`. Can we clarify this note to set expectations about what we'd like users to do more specifically? Happy to chat about how, if you'd find that helpful!
```suggestion
**Note**: Unlike other events, view events are sent multiple times to Datadog to reflect the updates occurring during their lifecycle. An update on a previous view event can still be sent while a new view is active. Datadog recommends being mindful of this behavior when modifying the content of a view event.
```

content/en/real_user_monitoring/browser/advanced_configuration.md
```diff
@@ -552,6 +552,21 @@ You can update the following event properties:
...
+
+```javascript
+beforeSend: (event) => {
+    // discouraged, as the current view name could be applied the active view and on the previous views
```
janine-c, line 559:
```suggestion
    // discouraged, as the current view name could be applied to both the active view and the previous views
```

content/en/real_user_monitoring/browser/advanced_configuration.md
```diff
@@ -552,6 +552,21 @@ You can update the following event properties:
...
+    event.view.name = getCurrentViewName()
+
+    // recommended
+    event.view.name = getViewNameForUrl(event.view.url)
+}
+```
+
+
+
+
```
janine-c, line 569:
```suggestion
```

janine-c (commented):
Hey Bastien! I'm not a RUM expert, so I had some questions about the functionality we're suggesting here. If you could help me clarify these, that would be really helpful!

content/en/real_user_monitoring/browser/advanced_configuration.md
bcaudan: > what behavior the user would observe in the UI if they modified the view event content incorrectly

with the view name discouraged example:

| time | action | behavior 
| ----- | ------| --------
| t0 | user defines current view name to 'first' | getCurrentViewName returns 'first'
| t1 | sdk sends view 1 | view 1 name is set to 'first'
| t2 | url changes | 
| t3 | user defines current view name to 'second' | getCurrentViewName returns 'second'
| t4 | sdk sends view 2 | view 2 name is set to 'second'
| t5 | sdk sends view 1 update | **view 1 name is incorrectly updated to 'second'**

so in the UI, the view 1 will have the name 'second' instead of 'first'

> why the code sample is separate from the table that also contains information about `view.name`

if you are referring to the `Override default RUM view names` section, it does not seem appropriate as:

- this section is not about modifying the event with the `beforeSend` API which is where a misconfiguration can occur if a user do not understand our behavior around view events
- all modifiable view attributes are concerned, not only the `view.name`

> Can we clarify this note to set expectations about what we'd like users to do more specifically? Happy to chat about how, if you'd find that helpful!

This API allows users to update our behavior with their own logic so a bit complicated to tell them exactly what they should do 😕️
My intent here is more about clarifying our behavior in order for them to achieve what they want to do.
Maybe the team will have other ideas about to make that clearer as well.
Happy to chat more about it!

janine-c: Thanks! I appreciate that incredibly thorough explanation 🙂 I was referring to the table directly preceding this new content (under `You can update the following event properties`), but it's helpful to know that it applies to all view attributes, so disregard that.

That all sounds good - I usually worry that advising users to be mindful of something isn't clear enough, but if it's open-ended, I don't want that to be a blocker.

janine-c (approved):

thomas-lebeau (approved):

PR #29510, opened 2025-05-22
(feat) Use params file for unsupported site banners

hestonhoffman:
### Merge instructions

Merge readiness:
- [ ] Ready for merge

**For Datadog employees**:

```
/merge
```

### DIFF
--- a/content/en/datadog_cloudcraft/_index.md
+++ b/content/en/datadog_cloudcraft/_index.md
@@ -4,11 +4,9 @@ further_reading:
 - link: "https://www.datadoghq.com/blog/cloud-architecture-diagrams-cost-compliance-cloudcraft-datadog/"
   tag: "Blog"
   text: "Plan new architectures and track your cloud footprint with Cloudcraft (Standalone)"
+cascade:
+  product_id: datadog_cloudcraft
 ---
-
-{{< site-region region="gov" >}}
-<div class="alert alert-warning">Cloudcraft in Datadog is not supported for your selected <a href="/getting_started/site">Datadog site</a> ({{< region-param key="dd_site_name" >}}).</div>
-{{< /site-region >}}
 
 ## Overview
 
--- a/layouts/partials/site_support_banner/site_support_banner.html
+++ b/layouts/partials/site_support_banner/site_support_banner.html
@@ -1,0 +1,23 @@
+{{/*  Inserts a banner at the top of the page if a site is not supported  */}}
+{{ $dot := . }}
+
+{{ $product_id := .Page.Params.product_id }}
+{{ $unsupported_sites := .Site.Params.unsupported_sites }}
+
+{{ $unsupported_product := index $unsupported_sites $product_id }}
+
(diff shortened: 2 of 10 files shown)


content/en/datadog_cloudcraft/_index.md
```diff
@@ -4,12 +4,10 @@ further_reading:
 - link: "https://www.datadoghq.com/blog/cloud-architecture-diagrams-cost-compliance-cloudcraft-datadog/"
   tag: "Blog"
   text: "Plan new architectures and track your cloud footprint with Cloudcraft (Standalone)"
+cascade:
+  product_id: datadog_cloudcraft
```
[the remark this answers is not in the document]
hestonhoffman: I thought about this, but there are some cases where a file covers a specific feature that isn't supported, even though the main product is supported. But maybe in those cases we just check for a param to override it 🤔

content/en/agent/troubleshooting/send_a_flare.md
```diff
@@ -32,6 +28,10 @@ When contacting Datadog Support with Remote Configuration enabled for an Agent,
 
 ## Send a flare from the Datadog site
 
+{{< site-region region="gov" >}}
```
jhgilbert, line 31: Should this one still be here? Assuming yes but flagging just in case!

layouts/partials/site_support_banner/site_support_banner.html
```diff
@@ -0,0 +1,42 @@
...
+{{$text := .Site.Params.site_support_banner | default "" | safeHTML }}
+
+
+{{/* "First checks if the file has a site_support_id param. Falls back to the file path. " */}}
+{{ $site_support_id := .Page.Params.site_support_id }}
+{{ $product_paths := split .Page.Path "/" }}
+{{ $unsupported_sites := .Site.Params.unsupported_sites }}
+
+{{ $unsupported_product := dict }}
+{{ if $site_support_id }}
```
jhgilbert, line 14: It seems like this should throw if the referenced site_support_id is not found anywhere in the YAML. This would catch misspelled or deprecated (and thus possibly confusing) site_support_ids. If the intention of `site_support_id` is just for support banners, and no banner is needed anymore, IMO it's cleaner to hard fail and force the writer to remove the `site_support_id` as soon as they have removed the corresponding YAML.

Catching misspelled folder names in the YAML would be more involved, and probably not worth the effort at the moment (because a cleanup script would be easy enough to write if the YAML is getting out of hand). But the above catch seems simple enough to be worth adding, even just in a follow-up PR if you need to get this one out the door.

jhgilbert (approved):
Approved with a couple of comments, thanks!

layouts/partials/site_support_banner/site_support_banner.html
hestonhoffman: Good point, I'm going to address this in a follow-up!

bgdeutsch (approved):

PR #29524, opened 2025-05-23
Rule creation guide

jacobotb:

Adds a guide for writing static analyzer rules. This is not a tutorial, but an advanced guide with full information, tips, and pitfalls.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

**For Datadog employees**:

```
/merge
```


iadjivon:
Created a Jira card for an editorial review: DOCS-11017

content/en/security/code_security/static_analysis/custom_rules/guide.md
```diff
@@ -0,0 +1,744 @@
+---
+description: A complete guide to writing custom rules for Datadog.
+title: Static Code Analysis Custom Rule Guide
+---
+
+This guide builds on the [Static Analysis Custom Rules Tutorial][1] to give a full description of how to write custom rules, along with tips and tricks and common pitfalls to avoid.
```
michaelcretzman: This guide builds on the [Static Analysis Custom Rules Tutorial][1], providing a full description of how to write custom rules, along with tips, tricks, and common pitfalls to avoid.

content/en/security/code_security/static_analysis/custom_rules/guide.md
```diff
@@ -0,0 +1,744 @@
...
+
+## Overview
+
+A static analyzer rule is composed of three parts: a Tree Sitter query to find interesting code constructs, a JavaScript function to analyze that code and generate findings, and tests that verify that the rule works correctly.
```
michaelcretzman:
```suggestion
A static analyzer rule consists of three parts: a Tree-sitter query to find relevant code constructs, a JavaScript function to analyze the code and generate findings, and tests that verify the rule works correctly.
```

content/en/security/code_security/static_analysis/custom_rules/guide.md
```diff
@@ -0,0 +1,744 @@
...
+(func-decl)
+```
+
+You can add patterns after the node type and before the closing parenthesis. The resulting pattern will match nodes with children that match those patterns.
+
+```scheme
+(func-decl (arg-list) (body))
+```
+
+This example matches nodes of type `func-decl` that contain a node of type `arg-list` followed by a node of type `body`, possibly with other nodes in between. In the following parse tree, this query would match the subtrees outlined in blue, but not the subtree outlined in orange.
```
michaelcretzman:
```suggestion
This example matches nodes of type `func-decl` containing a node of type `arg-list`, followed by a node of type `body`, possibly with other nodes in between. In the following parse tree, this query would match the subtrees outlined in blue, but not the subtree outlined in orange.
```

content/en/security/code_security/static_analysis/custom_rules/guide.md
```diff
@@ -0,0 +1,744 @@
...
+  const { cap4, cap5, cap6 } = query.capturesList;
+  /* check the captured nodes */
+  const err = /* generate a message for a finding */;
+  addError(err);
+}
+```
+
+### The `visit()` Function
+
+After executing the query, the static analyzer executes the `visit()` function for every match. This function receives three arguments:
```
michaelcretzman:
```suggestion
After executing the query, the static analyzer executes the `visit()` function for each match. This function receives three arguments:
```

content/en/security/code_security/static_analysis/custom_rules/guide.md
```diff
@@ -0,0 +1,744 @@
...
+  const parent = ddsa.getParent(funcDecl);
+  // Do something with the `funcDecl` node's parent
+  const children = ddsa.getChildren(funcDecl);
+  for (let child of children) {
+    // Do something with the `funcDecl` node's children
+  }
+}
+```
+
+You can also keep calling `ddsa.getParent(node)` and `ddsa.getChildren(node)` on the nodes returned by these functions to keep navigating the parse tree. Calling `ddsa.getParent()` on the root node returns `undefined`, while calling `ddsa.getChildren()` on a leaf tree returns an empty list.
```
michaelcretzman:
```suggestion
You can continue calling `ddsa.getParent(node)` and `ddsa.getChildren(node)` on the nodes returned by these functions to navigate the parse tree. Calling `ddsa.getParent()` on the root node returns `undefined`, while calling `ddsa.getChildren()` on a leaf tree returns an empty list.
```

content/en/security/code_security/static_analysis/custom_rules/guide.md
```diff
@@ -0,0 +1,744 @@
...
+}
+```
+
+You can then inspect and select the nodes you are interested in by checking their `cstType` and `text` properties.
+
+## Pitfalls
+
+### Adding predicates doesn't speed up the query
+
+If your Tree Sitter query is slow, you might think that you could speed it up by adding predicates to prune the search tree. However, the Tree Sitter query engine doesn't work that way: it ignores all predicates when it walks the tree in search of nodes that match the pattern, and it only applies those predicates at the end to filter the results list.
```
michaelcretzman:
```suggestion
If your Tree Sitter query is slow, you might try to speed it up by adding predicates to prune the search tree. However, this method doesn't work with the Tree Sitter query engine. The engine ignores all predicates when it walks the tree in search of nodes that match the pattern, and it only applies those predicates at the end to filter the results list.
```

content/en/security/code_security/static_analysis/custom_rules/guide.md
```diff
@@ -0,0 +1,744 @@
...
+```scheme
+(class_declaration
+  (method_declaration) @method1
+  (method_declaration) @method2
+)
+```
+
+For a class with only two methods, this query will only return one node; however, for a class with 10 methods, it will return 45. For a class with 100 methods, it will return 4950 nodes!
+
+To avoid this, you should use modifiers like `+` or `*` to capture the whole list of methods in one single query result, or use `.` to indicate that the child nodes must appear next to each other.
```
michaelcretzman:
```suggestion
To avoid this problem, use modifiers like `+` or `*` to capture the whole list of methods in one single query result. Alternatively, use `.` to indicate that the child nodes must appear next to each other.
```

content/en/security/code_security/static_analysis/custom_rules/guide.md
```diff
@@ -0,0 +1,744 @@
...
+  (#eq? @methodName "DoSomething")
+  (#eq? @varName @opName)
+)
+```
+
+The problem is that the Tree Sitter query engine will get every `var_declaration` node and every `call_expression` node, match them up pairwise, and then check the predicates for every pair. This results in an O(nm) operation.
+
+One solution is to write a query that finds one of the nodes and then use `ddsa.getParent()` and `ddsa.getChildren()` to find the other node.
+
+Another solution, which is not always possible, would be to gather all candidate nodes without trying to pair them up, and then process them in the JavaScript code.
```
michaelcretzman:
```suggestion
Another potential solution is to gather all candidate nodes without trying to pair them up, and then process them in the JavaScript code.
```

content/en/security/code_security/static_analysis/custom_rules/guide.md
```diff
@@ -0,0 +1,744 @@
...
+
+#### Trying to match a pattern at several nesting levels
+
+You can easily write a pattern that finds a node that contains a child node that matches a pattern. However, you cannot write a pattern that finds a node that contains a descendent at an arbitrary nesting level.
```
michaelcretzman:
```suggestion
You can write a pattern to find a node containing a child node that matches a pattern. However, you cannot write a pattern to find a node containing a descendent at an arbitrary nesting level.
```

content/en/security/code_security/static_analysis/custom_rules/guide.md
```diff
@@ -0,0 +1,744 @@
...
+
+```javascript
+// Code:
+function visit(query, filename, code) {
+  const { decl, fn } = query.captures;
+  // ... do something with decl and fn
+}
+```
+
+We've already mentioned the problems with this approach. First, the query engine goes down every branch of the parse tree to try and match the pattern, which could take a long time. Additionally, if there are two or more alternatives, the query engine will return one match for every set of nodes that matches one combination of choices.
```
michaelcretzman:
```suggestion
We've already mentioned the problems with this approach. First, the query engine goes down every branch of the parse tree to try and match the pattern, which might take a long time. Additionally, if there are two or more alternatives, the query engine will return one match for every set of nodes that matches one combination of choices.
```

content/en/security/code_security/static_analysis/custom_rules/guide.md
michaelcretzman:
```suggestion
This guide builds on the [Static Analysis Custom Rules Tutorial][1], providing a full description of how to write custom rules, along with tips, tricks, and common pitfalls to avoid.
```

hestonhoffman (approved):

PR #29559, opened 2025-05-27
[K9VULN-5345] Improve static sca setup documentation

piloulacdog:

Some customers are facing issues with the SBOM upload process and have reported that the documentation provided is inadequate. They have requested escalation and the booking of an engineer to support them with this critical installation for an important customer.

We should fix the documentation so that they can better understand how to install static SCA.

### Merge instructions

Merge readiness:
- [x] Ready for merge

### Preview
The preview can be found [here](https://docs-staging.datadoghq.com/pierrelouis.lacorte/K9VULN-5345-improve-sca-documentation/security/code_security/software_composition_analysis/setup_static/?tab=datadog)


michaelcretzman:
@piloulacdog have you reviewed the comments of @dastrong?

Let me know if you are going to make more changes and then I will review and merge the PR.

(DOCS editorial ticket is DOCS-11033)

piloulacdog:
@dastrong for your feedbacks:
> This makes it seem like we don't support GitLab for our product. This request access is to be able to see Code Snippets and get MR comments FWIR.

I think that I will likely need your help knowing exactly what it implies. I wasn't able to find a single org which had it enabled... so wasn't able to confirm that we offer...

> Why not put the supported language/package managers at the top? Afraid customers won't see this at all otherwise

That's because this list is only true if you scan using the Datadog/datadog-sbom-generator. It is not true if you use any other SBOM generator. The only true statement is the list of languages. But if you use the new `osv-scanner` (the google one, version v2) for example, you would have `uv` support. I've put it at the top of the `Run SCA scans in your CI Pipelines` section hoping that it helps with clarity!

> With the change above, we could remove the supported languages note. We could also probably shorten this whole section to be something like "Go to Security > Code Security to get started in app or following the documentation below."

Tried a rewording:
```

To get started:
1. Open [**Security** > **Code Security** > **Security Settings**][2].
2. Click **Manage Repositories** under "Activate scanning for your repositories".
3. Choose [where to run SCA scans](#select-where-to-run-static-sca-scans) (Datadog-hosted or CI Pipelines).
4. Follow the setup instructions for your source code provider.

The sections below cover the different ways to configure SCA for your repositories.

```

> I'd make this Other and says they need to use a customizable script and redirect. This Gitlab link doesn't lead to a script for SCA setup

✅ Updated!

piloulacdog:
Hi @michaelcretzman, I did ping @kassenq on Slack, I think she didn't get the chance to look at it just yet.
Please feel free to review! I don't expect push back from her. And she already gave us an initial blessing when I came up with the architecture of the re-write

PR #29592, opened 2025-05-28
[SAASINT-4057][openai][ccm] Documentation changes for admin key support

BoyangHuang:

An update to the admin key functionality has led to the need to update the documentation.


drichards-87 (changes requested):
Left a very small piece of feedback.

PR #29726, opened 2025-06-04
chore: warn on unsigned commits

avidal:
This patch configures merge gate to send warning Slack messages to authors of pull requests that contain unsigned commits.

There are built in exception filters for two known bots that create unsigned commits to this repository. We'll eventually remove these exceptions as bots begin signing commits.

### DIFF
--- a/repository.datadog.yml
+++ b/repository.datadog.yml
@@ -24,4 +24,5 @@ rules:
     excluded_emails:
       - 'robot-guacbot-guacamole@datadoghq.com'
       - 'guacbot@users.noreply.github.com'
+      - '41898282+github-actions[bot]@users.noreply.github.com'
     allow_unsigned_external: true


repository.datadog.yml
```diff
@@ -14,4 +14,15 @@ checks_to_ignore:
...
+schema-version: v1
+kind: mergegate
+rules:
+  - require: commit-signatures
+    enforcement: warn
+    excluded_emails:
+      - 'robot-guacbot-guacamole@datadoghq.com'
+      - 'guacbot@users.noreply.github.com'
+      - '41898282+github-actions[bot]@users.noreply.github.com'
+    allow_unsigned_external: true
```
avidal, line 28: This line allows unsigned commits from external contributors (ie, if that's the only source of an unsigned commit in a PR a warning will not be generated)

avidal:
cc @davidejones I can't actually perform the merge here even with the approvals

PR #29760, opened 2025-06-05
IPs should be separated by a space rather than a comma ,

aluiziod:
You can review this ticket here that shows a comma does not work for the customer.
https://datadog.zendesk.com/agent/tickets/2143260

Basically having different IPs separated by a comma have these recognised a single element of the no_proxy array, in reality we want each separate IP as a different element of the array.

This is under discussion in this thread: https://dd.slack.com/archives/C08PGQYM0SG/p1749131119035629


jszwedko (approved):
Approving for Agent engineering. The Agent expects spaces rather than commas.

PR #30092, opened 2025-06-24
[K9VULN-6032] Fix catch-all glob

dastrong:

We suggest that customers use `**/*` as a catch-all glob, but this will catch files in the root directory. This is especially problematic for the SCA product because dependency files are typically located at the root.

### Merge instructions

Merge readiness:
- [X] Ready for merge


piloulacdog (approved):

brett0000FF:
@dastrong - FYI the merge queue was disabled in this repo, so we have to manually click the Squash and merge button. Let me know if you don't have permissions for that, and I can help--thanks!

PR #30507, opened 2025-07-16
[CAP-2694] CRD filtering/sorting

christine-cay:

Here’s a quick summary of what’s new: CRD Filtering / Sorting: [before](https://app.datadoghq.com/orchestration/explorer/cr?cr-explorer-cols=name%2Ccluster%2C%40spec.maxReplicas%2C%40status.desiredReplicas&explorer-na-groups=false&ptfu=false)/[after](https://dd.datad0g.com/orchestration/explorer/cr?query=field%23status.agentList.available%3A1&cr-explorer-cols=name%2Cage%2Cfield%23apiVersion%2Cfield%23status.agentList.available%2Cfield%23status.agent.ready%2Cfield%23status.agentList.daemonsetName&crd_resource=datadogagent%2Fdatadoghq.com%2Fv2alpha1&explorer-na-groups=false)

- CR fields as columns using TagPopover: previously only partial data was shown, and CR field columns now support filtering, grouping, and column management — not just plain string values

- Filtering on CR fields is now fully supported: via the search bar (autocompletion and suggestions), facets, and TagPopover, etc.

- Sorting now works on CR field columns

> [!IMPORTANT]
>  Old CR columns don’t use the same field IDs, so features like sorting, TagPopover actions, etc. won’t work with them.
Since columns are stored in local storage, users may need to reset their columns and re-add CR fields in order to access the new functionality. Might be worth calling this out explicitly in the announcement as well.

Preview [here](https://docs-staging.datadoghq.com/christine.chun/update-crd-filtering-sorting/infrastructure/containers/configuration/?tab=datadogoperator#collect-custom-resources) and [here](https://docs-staging.datadoghq.com/christine.chun/update-crd-filtering-sorting/infrastructure/containers/orchestrator_explorer/?tab=datadogoperator#terms)

### Merge instructions

Merge readiness:
- [x] Ready for merge

### DIFF
--- a/content/en/infrastructure/containers/configuration.md
+++ b/content/en/infrastructure/containers/configuration.md
@@ -272,7 +272,7 @@ Follow these steps to collect the custom resources that these CRDs define:
 
 1.  Select **Enable Indexing** to save.
 
-After the fields are indexed, you can add them as columns in the explorer or as part of Saved Views.
+After the fields are indexed, you can add them as columns in the explorer and sort them, or include them in Saved Views. You can also filter on indexed fields using the prefix `field#`.
 
 ### Indexing complex types
 
--- a/content/en/infrastructure/containers/orchestrator_explorer.md
+++ b/content/en/infrastructure/containers/orchestrator_explorer.md
@@ -158,12 +158,12 @@ There are multiple types of terms available:
 
 | Type | Examples |
 |---|---|
-| **Tags**: Attached to resources by [the agent collecting them][20]. There are also additional tags that Datadog generates for Kubernetes resources. | `datacenter:staging`<br>`tag#datacenter:staging`<br>*(the `tag#` is optional)* |
+| **Tags**: Attached to resources by [the agent collecting them][20]. There are also additional tags that Datadog generates for Kubernetes resources. | `datacenter:staging`, `tag#datacenter:staging`<br>_(the `tag#` is optional)_ |
 | **Labels**: Extracted from [a resource's metadata][25]. They are typically used to organize your cluster and target specific resources with selectors. | `label#chart_version:2.1.0` |
 | **Annotations**: Extracted from [a resource's metadata][26]. They are generally used to support tooling that aid in cluster management. | `annotation#checksum/configmap:a1bc23d4` |
 | **Metrics**: Added to workload resources (pods, deployments, etc.). You can find resources based on their utilization. To see what metrics are supported, see [Resource Utilization Filters](#resource-utilization-filters). | `metric#cpu_usage_pct_limits_avg15:>80%` |
-| **String matching**: Supported by some specific resource attributes, see below.<br>*Note: string matching does not use the key-value format, and you cannot specify the attribute to match on.* | `"10.132.6.23"` (IP)<br>`"9cb4b43f-8dc1-4a0e"` (UID)<br>`web-api-3` (Name) |
-| **Fields**: Extracted from [a resource's metadata][27]. | `field#metadata.creationTimestamp:>=4wk`<br>`field#metadata.deletionTimestamp:<=1hr` |
+| **String matching**: Supported by some specific resource attributes, see below.<br>_Note: string matching does not use the key-value format, and you cannot specify the attribute to match on._ | `"10.132.6.23"` (IP),<br>`"9cb4b43f-8dc1-4a0e"` (UID),<br>`web-api-3` (Name) |
+| **Fields**: Extracted from [a resource's metadata][27] or from custom resources' indexed fields. | `field#metadata.creationTimestamp:>=4wk`, `field#metadata.deletionTimestamp:<=1hr`, `field#status.currentReplicas:3`, `field#status.conditions.Active.status:True` |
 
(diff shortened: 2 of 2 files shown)


content/en/infrastructure/containers/orchestrator_explorer.md
```diff
@@ -163,7 +163,8 @@ There are multiple types of terms available:
 | **Annotations**: Extracted from [a resource's metadata][26]. They are generally used to support tooling that aid in cluster management. | `annotation#checksum/configmap:a1bc23d4` |
 | **Metrics**: Added to workload resources (pods, deployments, etc.). You can find resources based on their utilization. To see what metrics are supported, see [Resource Utilization Filters](#resource-utilization-filters). | `metric#cpu_usage_pct_limits_avg15:>80%` |
 | **String matching**: Supported by some specific resource attributes, see below.<br>*Note: string matching does not use the key-value format, and you cannot specify the attribute to match on.* | `"10.132.6.23"` (IP)<br>`"9cb4b43f-8dc1-4a0e"` (UID)<br>`web-api-3` (Name) |
-| **Fields**: Extracted from [a resource's metadata][27]. | `field#metadata.creationTimestamp:>=4wk`<br>`field#metadata.deletionTimestamp:<=1hr` |
+| **Fields**: Extracted from [a resource's metadata][27] or from custom resources indexed fields. |  `field#metadata.creationTimestamp:>=4wk` <br> `field#metadata.deletionTimestamp:<=1hr` <br> `field#status.currentReplicas:3` <br> `field#status.conditions.Active.status:True` |
```
janine-c:
```suggestion
| **Fields**: Extracted from [a resource's metadata][27] or from custom resources' indexed fields. |  `field#metadata.creationTimestamp:>=4wk` <br> `field#metadata.deletionTimestamp:<=1hr` <br> `field#status.currentReplicas:3` <br> `field#status.conditions.Active.status:True` |
```
Not sure if this should be a possessive? I'll let you make the final call.

Also, when there are multiple examples in a cell, it could be useful to put additional line breaks or bullet points to differentiate them, but that's not limited to this PR, so it can just be a suggestion for now :)

christine-cay: Thanks for the suggestion! I updated here: 96749cf4acb39dd5ca28be905a4578206b6d8c59

| Before | After |
|--------|--------|
| Screenshot 2025-07-17 at 10 20 56 | Screenshot 2025-07-17 at 10 21 06 |

content/en/infrastructure/containers/configuration.md
```diff
@@ -272,7 +272,12 @@ Follow these steps to collect the custom resources that these CRDs define:
 
 1.  Select **Enable Indexing** to save.
 
-After the fields are indexed, you can add them as columns in the explorer or as part of Saved Views.
+After the fields are indexed, you can add them as columns in the explorer and sort them, or include them in Saved Views. You can also filter on indexed fields using the prefix `field#`.
+
+<div class="alert alert-info">
+If you see custom resource columns prefixed with <code>@</code>, they use an older format that doesn't support features like sorting or tag actions. To access the latest functionality, reset your columns and re-add the relevant custom resource fields.
+</div>
```
AntoineEsteve: ❓ Do we need this section? I feel like we should just treat those legacy columns as string field column and it should be transparent to the user

christine-cay: I think it could be confusing for users if we don't mention this. We're announcing features like:

* CR fields as columns now support filtering, grouping, sorting, and TagPopover — whereas previously they only showed plain string values.

If someone is still using legacy columns, none of that will work, and they won't know why. That was exactly the issue Kangyi had in [this thread](https://dd.slack.com/archives/C06TSC3PJKF/p1751977165141109). I'm worried that if we leave this out, we’ll get a lot of complaints from users saying sorting or TagPopover isn’t working — when in reality, they're just on the old column version
Screenshot 2025-07-17 at 10 25 57

AntoineEsteve: Isn't it very unlikely? They will only have an issue if the type of the field is not `string` (most are), and if they added fields as columns before the new implementation. I would like to avoid leaking implementation details in the documentation. If we think this case is not rare, then we should have some code to fix it automatically (by fetching the facet API to find the type and replacing the column, right?)

christine-cay: > Isn’t it very unlikely?

It’s hard to say — since column IDs are stored in local storage, we don’t have visibility into how many users still have legacy CR field columns. Also, those legacy columns use the `@` prefix, so they wouldn’t work as expected even for string columns.

I can look into whether it’s possible to strip the prefix, add the appropriate type suffix, and adjust the implementation to make it compatible with the new CR column IDs without the need to reset the columns

christine-cay: Removed this block since legacy columns will now be tackled automatically in the code: [PR](https://github.com/DataDog/web-ui/pull/216985)

christine-cay:
Hi @janine-c! We released the feature, would it be possible to merge this PR?

PR #30560, opened 2025-07-18
Update Java example with tags

michaelcretzman:
- adding tags to Java example at customer request
- fixed syntax error where there was a `.` with no following method

From @binitapoudel1 : 

> However, customer raised a case where they stated that if we could include tags also in our java example config that would be helpful:

### Merge instructions

Merge readiness:
- [X] Ready for merge


urseberry (approved):

binitapoudel1 (approved):
LGTM👍

PR #30562, opened 2025-07-18
APM .NET Library: update compatibility page

lucaspimentel:

**TL;DR:** Simplifies and improves accuracy of .NET APM compatibility documentation by removing redundant architecture sections, consolidating OS tables, and fixing various inconsistencies.

Updates the .NET APM library compatibility documentation for accuracy and consistency:

- Standardize date format from culture-specific `mm/dd/yyyy` to universal `Month dd, yyyy` format
- Remove entire "Supported processor architectures" section from both pages (architecture support is implicit in OS tables)
- Consolidate Nano Server into Windows Server entries with clarifying note
- Remove architecture labels (x64) from operating system names in tables
- Add note explaining Windows Server includes all editions and installation options (e.g. Nano Server)
- Update .NET Framework page to reference .NET Framework system requirements
- Standardize architecture capitalization (Arm64 → arm64) across all tables
- Fix alignment/formatting of Markdown tables (source code only, doesn't affect rendered HTML)
- Fix Process integration description for .NET Framework - change from quoted package to "built-in"
- Fix Process integration formatting for .NET Core - remove extra quotes
- Add explicit Windows OS support table to .NET Framework page (didn't exist before)
- Move .NET release notes reference to opening paragraph in .NET Core page
- Minor formatting fixes (periods, bold headers, etc.)

### Merge instructions

Merge readiness:
- [x] Ready for merge

**For Datadog employees**:

### DIFF
--- a/content/en/tracing/trace_collection/compatibility/dotnet-core.md
+++ b/content/en/tracing/trace_collection/compatibility/dotnet-core.md
@@ -74,14 +74,14 @@ The .NET Tracer supports Linux distributions as best-effort, based on minimum li
 | Operating System         | Version | Architectures | Support level                         | Package version                 |
 | -------------------------|---------|---------------|---------------------------------------|---------------------------------|
 | Alpine Linux (x64)       | 3.14+   |  x64          | [GA](#support-ga)                     | latest (.NET 5+ only, v1.27.0+) |
-| Alpine Linux (arm64)     | 3.18+   |  Arm64        | [GA](#support-ga)                     | latest (.NET 6+ only, v3.2.0+)  |
+| Alpine Linux (arm64)     | 3.18+   |  arm64        | [GA](#support-ga)                     | latest (.NET 6+ only, v3.2.0+)  |
 | CentOS Linux             | 7+      |  x64          | [Maintenance](#support-maintenance)   | latest (EOL in v4.0.0)          |
 | CentOS Stream Linux      | 8       |  x64          | [Maintenance](#support-maintenance)   | latest (EOL in v4.0.0)          |
-| Debian                   | 10+     |  x64, Arm64   | [GA](#support-ga)                     | latest                          |
+| Debian                   | 10+     |  x64, arm64   | [GA](#support-ga)                     | latest                          |
 | Fedora                   | 29+     |  x64          | [GA](#support-ga)                     | latest                          |
 | openSUSE                 | 15+     |  x64          | [GA](#support-ga)                     | latest                          |
 | Red Hat Enterprise Linux | 7+      |  x64          | [GA](#support-ga)                     | latest                          |
-| Ubuntu                   | 18.04+  |  x64, Arm64   | [GA](#support-ga)                     | latest                          |
+| Ubuntu                   | 18.04+  |  x64, arm64   | [GA](#support-ga)                     | latest                          |
 
 ### macOS
 
@@ -89,9 +89,9 @@ The .NET Tracer supports macOS for CI Test Optimization only.
 
 | Operating System         | Version | Architectures | Support level           | Package version            |
 | -------------------------|---------|---------------|-------------------------|----------------------------|
-| macOS                    | 12.0+   |  x64, Arm64   | [GA](#support-ga)       | latest                     |
+| macOS                    | 12.0+   |  x64, arm64   | [GA](#support-ga)       | latest                     |
 | macOS                    | 11.0    |  x64          | [EOL](#support-eol)     | < 3.0.0                    |
-| macOS                    | 11.0    |  Arm64        | [EOL](#support-eol)     | < 3.0.0 (Added in 2.20.0)  |
+| macOS                    | 11.0    |  arm64        | [EOL](#support-eol)     | < 3.0.0 (Added in 2.20.0)  |
 
 ## Integrations
 


content/en/tracing/trace_collection/compatibility/dotnet-core.md
```diff
@@ -30,44 +30,45 @@ The .NET Tracer supports automatic instrumentation on the following .NET and .NE
...
+| .NET Core 3.1        | December 13, 2022     | [GA](#support-ga)    | latest               |
+| .NET Core 3.0        | March 3, 2020         | [EOL](#support-eol)  | Not recommended      |
+| .NET Core 2.2        | December 23, 2019     | [EOL](#support-eol)  | Not recommended      |
+| .NET Core 2.1        | August 21, 2021       | [EOL](#support-eol)  | Not recommended      |
+| .NET Core 2.0        | October 1, 2018       | [EOL](#support-eol)  | Not recommended      |
 
 Additional information can be found in [Microsoft's .NET and .NET Core Lifecycle Policy][3], [End of life .NET runtime versions](#end-of-life-net-runtime-versions), and [.NET runtime support policy](#net-runtime-support-policy).
 
-## Supported processor architectures
+## Supported process architectures
```
bouwkast: Is `processor` -> `process` correct?

I don't think it is.

content/en/tracing/trace_collection/compatibility/dotnet-core.md
```diff
@@ -30,44 +30,45 @@ The .NET Tracer supports automatic instrumentation on the following .NET and .NE
...
 
 The .NET Tracer supports automatic instrumentation on the following architectures:
 
-| Processor architecture                    | Support level         | Package version                        |
+| Process architecture                      | Support level         | Package version                        |
 | ------------------------------------------|-----------------------|----------------------------------------|
 | Windows x64 (`win-x64`)                   | [GA](#support-ga)     | latest                                 |
-| Windows x86 (`win-x86`)                   | [GA](#support-ga)     | < 3.0.0 (e.g. 2.56.0)                  |
+| Windows x86 (`win-x86`)                   | [GA](#support-ga)     | latest                                 |
```
bouwkast: I think while we _can_ instrument 32-bit applications we don't support `win-x86` and our messaging throughout V3 has been that we don't _officially_ support it.

Unsure what the primary reason is/was but I thought it was mainly because we stopped shipping the 32-bit MSI.

I don't think we should change this here and if we do there are other locations that we'd need to update like here https://github.com/DataDog/documentation/blob/a69340731a615f68cc996f5e580708ca41c68ae0/content/en/tracing/trace_collection/automatic_instrumentation/dd_libraries/dotnet-framework.md?plain=1#L92

bouwkast (dismissed):
I don't think the `processor` -> `process` is correct nor is the support for `win-x86`, but the rest looks good to me thanks

content/en/tracing/trace_collection/compatibility/dotnet-core.md
lucaspimentel: My goal here is to clarify what we support and what we don't support. I think "processor" here is confusing, since we don't care what the _actual_ processor architecture is.

We care about:
- the OS architecture: for installation purposes, because our MSI only supports Windows `x64`, not to Windows `x86`
- the _process_ architecture: we support instrumenting both `x64` and `x86` processes on Windows

content/en/tracing/trace_collection/compatibility/dotnet-core.md
lucaspimentel: > we can instrument 32-bit applications we don't support `win-x86`

We _do_ support instrumenting `win-x86` processes (aka "32-bit Windows apps"), which is why I'm trying to clarify the ambiguity. 😅  What we don't support is using the MSI on Windows `x86` (which is the OS architecture, nor the processor).

Also, we should avoid the terms "32-bit" and "64-bit." They are ambiguous because Windows supports ARM architectures, of which there are also 32 and 64-bit versions.
- `win-x86` is 32-bit
- `win-x64` is 64-bit
- `win-arm` is 32-bit
- `win-arm64` is 64-bit

bouwkast: Yes but isn't this more ambiguous? We are saying `win-x86` as in the processor architecture not 32-bit applications

content/en/tracing/trace_collection/compatibility/dotnet-core.md
bouwkast: Hmm I don't think I follow as `win-x86` is a processor architecture and not a the process architecture?
Wouldn't that be 32-bit and 64-bit as the process architecture 🤷

bouwkast: I've reverted my request changes as maybe I don't fully understand the terminology

content/en/tracing/trace_collection/compatibility/dotnet-core.md
lucaspimentel: > isn't this more ambiguous?

Maybe? I'm glad we're having this conversation! 😄

Here's one example that I think is wrong:

image

I read that and to me it says, "to instrument a `win-x86` application, you need to use version < 3.0 of the tracer. That is not correct. Versions 3.0+ of the tracer support `win-x86` applications.

bouwkast: I think here we get into the semantics of "support" but also that was a `processor architecture` previously which I think is what we are saying that we don't "support" anymore, but you can definitely still instrument a 32-bit application.

I am unsure if you _could_ instrument a 32-bit application on `win-x86` as we said we didn't "support" that.

Maybe the tables in general should be reworked honestly 🤷

lucaspimentel: I think when you say `win-x86` you are talking about the Windows version and I'm talking about the executable.

> Maybe the tables in general should be reworked honestly

Yeah, maybe. I'll revert some of this architecture changes and we can deal with those separately.  Thanks for the feedback!

bouwkast: Okay maybe we rework it at some point to say https://learn.microsoft.com/en-us/dotnet/core/rid-catalog in a follow up as I would imagine that is what it is?

lucaspimentel: Maybe we could just remove the .NET RIDs (`{os}-{arch}`) and keep the more plain language ("Windows x86"). But I think we still need to make it clear that in this table, "Windows x86" means the process architecture (or the binary's target architecture), not the OS. Supported OSes are listed in a different table and they are not one-to-one.

lucaspimentel: Hmm, but it might be good to keep them because users need to know which one to use sometimes (for example, when using the `Datadog.Trace.Bundle` nuget package). Never mind.

content/en/tracing/trace_collection/compatibility/dotnet-core.md
```diff
@@ -50,24 +50,25 @@ The .NET Tracer supports automatic instrumentation on the following architecture
 | Processor architecture                    | Support level         | Package version                        |
 | ------------------------------------------|-----------------------|----------------------------------------|
 | Windows x64 (`win-x64`)                   | [GA](#support-ga)     | latest                                 |
-| Windows x86 (`win-x86`)                   | [GA](#support-ga)     | < 3.0.0 (e.g. 2.56.0)                  |
+| Windows x86 (`win-x86`)                   | [GA](#support-ga)     | latest                                 |
```
lucaspimentel: Do we or do we not support instrumenting `win-x86` processes? (regardless of the OS architecture)

Why did this table say `< 3.0.0`? I _think_ it was getting confused with the fact that the MSI installer no longer works on Windows x86. But OS architecture is not the same as process architecture, and afaik we still support `win-x86` processes (running on supported versions of Windows).

content/en/tracing/trace_collection/compatibility/dotnet-framework.md
```diff
@@ -48,12 +48,29 @@ Additional information can be found in [Microsoft's .NET Framework Lifecycle Pol
...
 The .NET Tracer supports automatic instrumentation on the following architectures:
 
-| Processor architectures | Support level       | Package version       |
-| ----------------------- | ------------------- | --------------------- |
-| Windows x64 (`win-x64`) | [GA](#support-ga)   | latest                |
-| Windows x86 (`win-x86`) | [EOL](#support-eol) | < 3.0.0 (e.g. 2.56.0) |
+| Processor architectures | Support level       | Package version |
+| ----------------------- | ------------------- | --------------- |
+| Windows x64 (`win-x64`) | [GA](#support-ga)   | latest          |
+| Windows x86 (`win-x86`) | [GA](#support-ga)   | latest          |
```
lucaspimentel: (same comment as the other file)

content/en/tracing/trace_collection/compatibility/dotnet-core.md
lucaspimentel: I reverted the change for now. We can discuss this one and make changes in a separate PR. Thanks!

content/en/tracing/trace_collection/compatibility/dotnet-core.md
lucaspimentel: Update: We _do_ support instrumenting `win-x86` processes in `dd-trace-dotnet` 3.x

bouwkast (approved):
Thanks!

PR #30747, opened 2025-07-28
Add a new page for Agent Tags

estherk15:

Added a new page for information on Agent tags that are automatically added to logs.

### DIFF
--- a/config/_default/menus/main.en.yaml
+++ b/config/_default/menus/main.en.yaml
@@ -425,7 +425,7 @@ menu:
       parent: agent
       weight: 5
     - name: Agent tags
-      url: agent/agent_tags/
+      url: agent/logs/agent_tags/
       parent: agent
       identifier: agent_tags
       weight: 501
--- a/config/_default/menus/main.en.yaml
+++ b/config/_default/menus/main.en.yaml
@@ -424,26 +424,31 @@ menu:
       identifier: agent_logs
       parent: agent
       weight: 5
+    - name: Agent tags
+      url: agent/agent_tags/
+      parent: agent
+      identifier: agent_tags
+      weight: 501
     - name: Advanced Configurations
       identifier: agent_logs_advanced_log_collection
       url: agent/logs/advanced_log_collection
       parent: agent_logs
-      weight: 501
+      weight: 502
     - name: Proxy
       identifier: agent_logs_proxy
       url: agent/logs/proxy
       parent: agent_logs
-      weight: 502
+      weight: 503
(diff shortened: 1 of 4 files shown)


config/_default/menus/main.en.yaml
```diff
@@ -424,26 +424,31 @@ menu:
       identifier: agent_logs
       parent: agent
       weight: 5
+    - name: Agent tags
+      url: agent/agent_tags/
```
DDuongNguyen: Hey! I'm not 100% confident but i think this maybe should be `agent/logs/agent_tags/` ? Since its specifically for the log agent and not the agent overall

estherk15: @DDuongNguyen Yup, you're right!

estherk15:
```suggestion
      url: agent/logs/agent_tags/
```

content/en/agent/logs/agent_tags.md
```diff
@@ -0,0 +1,52 @@
...
+| `version` | The Agent version | Always when available |
+| `filename` | Base name of the tailed file | File-based sources only |
+| `dirname` | Directory containing the tailed file | File-based sources only |
+| `source_host` | IP address of the socket source host | Socket sources (TCP/UDP) only |
+| `event_type` | Type of the Windows event | Windows events only |
+| `event_source` | Source of the Windows event | Windows events only |
+| `event_id` | Windows Event ID | Windows events only, if `tag_event_id: true` |
+| `sid` | Windows Security identifier | Windows events only, if `tag_sid: true` |
+| `truncated` | Source of truncation | If `logs_config.tag_truncated_logs: true` |
+| `multiline` | Source of multiline aggregation | If `logs_config.tag_multi_line_logs: true` |
```
estherk15:
```suggestion
| `multiline` | Source of multi-line aggregation | If `logs_config.tag_multi_line_logs: true` |
```

DDuongNguyen:
This makes sense to me! Its a bit hard to navigate to the exact page but I think the issue was the url thing i commented. I think if we can navigate to that page within the log collection section then it makes perfect sense to me. Let me know what you think! 🙇🏼

estherk15:
> This makes sense to me! Its a bit hard to navigate to the exact page but I think the issue was the url thing i commented. I think if we can navigate to that page within the log collection section then it makes perfect sense to me. Let me know what you think! 🙇🏼

Thanks for catching that! I'll merge both PRs!

PR #30770, opened 2025-07-29
Move API instructions inside of Generic script tab

GabrielAnca:
Alternative proposed by @andrea-mosk to integrate the API instructions inside of the generic script tab


GabrielAnca:
Option declined

PR #30891, opened 2025-08-04
Update App Builder identity information

OliviaShoup:
This PR updates information about running apps as author. 

### Merge instructions

Merge readiness:
- [x] Ready for merge

### DIFF
--- a/assets/styles/components/_tab-toggle.scss
+++ b/assets/styles/components/_tab-toggle.scss
@@ -193,6 +193,13 @@ $tab-active-shadow: 0 -4px 12px rgba(99, 44, 166, 0.08);
     .card-body img {
       border: 0px solid $ddgray;
     }
+
+    .tab-pane {
+      h3:before {
+        margin-top: 0;
+        height: 0;
+      }
+    }
   }
 
   // Horizontal tab navigation container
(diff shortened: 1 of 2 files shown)


barakshush:
Barak's feedback:

1. I see that the sidebar says: "Authentication" while the title says "Access and Authentication", can we make them aligned?
2. For this section: "Instead, actions use the identity of the app’s author" can we explain how to view who's the author of the app? see a screenshot below
image

barakshush (approved):

content/en/actions/app_builder/access_and_auth.md
```diff
@@ -5,11 +5,15 @@ aliases:
     - /service_management/app_builder/auth
 ---
 
-A few tools control access and authentication for apps and their components.
+[comment]: <> (This PR is for requesting changes to this file.)
+
```
janine-c:
```suggestion
```

janine-c (changes requested):
Looks good! Looks like there are still some links that go to the old filenames, so we want to make sure we catch those before merging. The internal page we have on [moving docs pages](https://datadoghq.atlassian.net/wiki/spaces/docs4docs/pages/2174550131/Moving+Docs+Pages) is also good for making sure we don't break links when we move pages, even though none of these pages have moved 🙂 

If you could do a search to update those links, and add aliases to both pages so their old URLs work, that would be great! Let me know if you'd like to show you how I do it.

janine-c (approved):
Made the changes I mentioned and tested them all 🙂 Feel free to take a look and merge whenever is convenient for you!

PR #30973, opened 2025-08-07
Rename Execution Replay -> Exception Replay

srosenthal-dd:
Tiny proposed fix: I just learned about this feature and it looks like this is a rebranding?

I see that https://docs.datadoghq.com/tracing/error_tracking/execution_replay automatically redirects to https://docs.datadoghq.com/error_tracking/backend/exception_replay/ already


janine-c (approved):
Great catch, thank you!!

PR #31175, opened 2025-08-19
[docseng-99] reading integrations and sources data from s3 instead of hugo module

bgdeutsch:
reading the integration markdown and `websites-sources` data from an s3 bucket rather than hugo module

motivation:

- hugo module is confusing for contributors, requires more security maintenance, and causes more issues like the go files being added to commits
- we are migrating away from hugo eventually so this will not be compatible with next gen docs
- previous system in websites-sources required a PR with manual review/merge which causes bottlenecks

### Merge instructions

Merge readiness:
- [ ] Ready for merge

**For Datadog employees**:


Makefile
```diff
@@ -81,20 +81,18 @@ build-llms-txt:
...
 	@make setup-build-scripts ## Build and run docs including external content.
 	@make dependencies
-	@make update_websites_sources_module
 	@make server
 
 # Skip downloading any dependencies and run the site (hugo needs at the least node)
 start-no-pre-build: node_modules  ## Build and run docs excluding external content.
 	@make setup-build-scripts
-	@make update_websites_sources_module
+	@make websites_sources_data
```
StefonSimmons: do we also want to get data from s3 on `make start` and `start-preserve-build` too?

bgdeutsch: @StefonSimmons this `websites_sources_data` call happens in `make dependencies` which happens in the commands you mentioned.   i don't think we really want to build all of the dependencies in `no-pre-build` but i figured we would want the integrations

StefonSimmons: ahh just seeing that. OK!

davidejones (commented):
So i did have one problem running this
```
Downloading data from websites-sources S3 bucket...
python3: can't open file '/Users/david.jones/_dev/projects/documentation/./local/bin/py/build/get_websites_sources_data.py': [Errno 2] No such file or directory
```

hestonhoffman (approved):
This is looking good to me!

Makefile
hestonhoffman: @evazorro brings up a really good point here. What if something in an integrations PR breaks the build? It would prevent us from previewing unrelated work. I think we should create a separate build command for local websites-sources previews and call this there instead. The make start stuff is fine, we just need to protect `make start-no-pre-build` from being blocked by unrelated work

Makefile
```diff
@@ -81,20 +81,18 @@ build-llms-txt:
...
 	@make build-cdocs
 	@make server
 
 # Leave build scripts as is for local testing
 # This is useful for testing changes to the build scripts locally
 start-preserve-build: dependencies
-	@make update_websites_sources_module
 	@make server
```
hestonhoffman, line 95:
```suggestion
	@make server
	
# Run the site with websites_sources_data (integrations previews)
start-integrations: node_modules
	@make setup-build-scripts
	@make websites_sources_data
	@make build-cdocs
	@make server
```

hestonhoffman (commented):
See comment below

PR #31181, opened 2025-08-19
[RFC, DO NOT MERGE] Spec for generating plaintext .md files for AI agents

jhgilbert:
# Spec for generating plaintext .md files for AI agents

## What is this?

This folder contains a listing of the components used on the docs site, and a description of their desired output format in the plaintext files generated for AI agents.

The HTML compiled from these files will be used as the test input HTML for the Markdown generator job.

## Where can I learn more?

1. Read [Datadog Docs for AI Agents](https://docs.google.com/document/d/1t48N0uDJElMP_228n2bIYi_cUIj_xMOKjdHuh1Z4rMg), which attempts to summarize the problem space and why this is important.
2. Ask Jen Gilbert.

## Disclaimer

Everything described here is *probably* possible based on the technical context and existing libraries we have so far, but it's not *guaranteed* to be possible until the docs are migrated to Markdoc.

If you'd like to follow along with changes in the spec, Jen will keep this PR up to date, with clear commit messages about what changed.

## Feedback is welcome!

Jen took her best shot at these, but she's not attached to any of it. Comment away!

### DIFF
--- a/content/en/md_conversion_test_files/common_components/alert.md
+++ b/content/en/md_conversion_test_files/common_components/alert.md
@@ -1,0 +1,25 @@
+---
+title: Alert
+---
+
+## Expected .md output
+
+Each alert is bordered with `[BEGIN <ALERT_LEVEL> ALERT]` and `[END <ALERT_LEVEL> ALERT]`. For example, `[BEGIN INFO ALERT]`.
+
+### Info
+
+<div class="alert alert-info">This is an info alert.</div>
+
+This is a random sentence.
+
+### Warning
+
+<div class="alert alert-warning">This is a warning alert containing an <a href="https://google.com">HTML link.</a></div>
+
+This is a random sentence.
+
+### Danger
+
+<div class="alert alert-danger">This is a danger alert.</div>
+
+This is a random sentence.
--- a/content/en/md_conversion_test_files/common_components/callouts.md
+++ b/content/en/md_conversion_test_files/common_components/callouts.md
@@ -1,0 +1,24 @@
+---
+title: Callouts
+---
+
+## Expected .md output
+
+- Each callout is marked by `[BEGIN CALLOUT]` and `[END CALLOUT]`.
+- Any attributes, such as the title or the URL, are included in the box contents:
+    - The title is a heading at the top of the box.
+    - The URL is included with verbiage at the bottom like "Visit this URL: <URL>".
+
+## Example inputs
+
+### Callout (preview / beta)
+
+{{< callout url="https://www.datadoghq.com/">}}
+  Datadog Apps are in Preview. Request access by using this form. Once approved, you can start getting creative and develop your App for you, your organization, or for publishing to the entire Datadog community alongside our other great Datadog Apps!
+{{< /callout >}} 
+
+### Learning center callout
+
+{{< learning-center-callout header="Try Foundations in the Learning Center" btn_title="Enroll Now" btn_url="https://learn.datadoghq.com/courses/datadog-foundation">}}
+  The Datadog Learning Center is full of hands-on courses to help you learn about this topic. Enroll today to learn more about Datadog Foundations.
+{{< /learning-center-callout >}}
--- a/content/en/md_conversion_test_files/common_components/code_blocks.md
+++ b/content/en/md_conversion_test_files/common_components/code_blocks.md
@@ -1,0 +1,60 @@
+---
+title: Code Blocks
+---
+
+## Scope
+
+This covers the `code-block` and `highlight` shortcodes. Plain fences are already covered in the "plain markdown" examples.
+
+## Expected .md output
+
+Each code block is wrapped in a fence, with the language included when available:
+
+```lang
+Code contents here.
+```
+
+### Filenames
+
+When a filename is present, it appears above the code block, like this:
+
+In the `block.java` file:
+
+```lang
+Code contents here.
+```
+
+### Highlighted lines
+
+When a highlight is present, it appears above the code block, like this:
+
+See lines 2-3 below:
+
+```lang
+Code contents here.
+```
+
+## Example inputs
+
+### Code block
+
+{{< code-block lang="java" filename="block.java" disable_copy="true" collapsible="true" >}}
+import com.datadoge.docs.SweetCodeBlock;
+import com.datadoge.docs.TryItOut;
+public class CodeBlocksForLife {
+    ...
+    }
+{{< /code-block >}}
+
+### Highlighted code block
+
+{{< highlight yaml "hl_lines=6-8" >}}
+    apiVersion: apps/v1
+    kind: Deployment
+    metadata:
+    name: springfront
+    labels:
+        tags.datadoghq.com/env: "dev"
+        tags.datadoghq.com/service: "springfront"
+        tags.datadoghq.com/version: "12"
+    {{< /highlight >}}
--- a/content/en/md_conversion_test_files/common_components/further_reading.md
+++ b/content/en/md_conversion_test_files/common_components/further_reading.md
@@ -1,0 +1,20 @@
+---
+title: Further Reading
+further_reading:
+  - link: "logs/processing/pipelines"
+    tag: "Documentation"
+    text: "Log processing pipelines"
+  - link: "some/blog/post"
+    tag: "Blog"
+    text: "Some blog post"
+---
+
+## Expected .md output
+
+A list of links, introduced by the text "Further reading:".
+
+## Example input
+
+## Further reading
+
+{{< partial name="whats-next/whats-next.html" >}}
--- a/content/en/md_conversion_test_files/common_components/images_and_videos.md
+++ b/content/en/md_conversion_test_files/common_components/images_and_videos.md
@@ -1,0 +1,17 @@
+---
+title: Images and Videos
+---
+
+## Expected .md output
+
+A placeholder including relevant information about the asset, such as `[IMAGE src="some/src/path" alt="some alt text"]`.
+
+## Example inputs
+
+## Image
+
+{{< img src="path/to/your/image-name-here.png" alt="Your image description" caption="Your image description" style="width:100%;" >}}
+
+## Video
+
+{{< img src="path/to/your/video-name-here.mp4" alt="Your image description" video="true" width="100%" >}}
--- a/content/en/md_conversion_test_files/common_components/plain_markdown.md
+++ b/content/en/md_conversion_test_files/common_components/plain_markdown.md
@@ -1,0 +1,80 @@
+---
+title: Plain Markdown Elements
+---
+
+## Expected .md output
+
+- Most of this should remains unchanged, or uses a format that does not meaningfully differ from its original form. (Some spacing changes can be difficult to avoid.)
+- In some cases, content may be slightly optimized for agents (for example, links will become inline).
+
+## Headings
+
+### Heading level 3
+
+Some text.
+
+#### Heading level 4 {#heading-level-4}
+
+Some text, with a custom heading ID.
+
+##### Heading level 5
+
+Some text.
+
+###### Heading level 6
+
+Some text.
+
+## Inline text formatting
+
+You can *italicize* text, or **bold** it.
+
+Text can also be formatted as a [plain link][1], **[bold link][1]**, or *[italic link][1]*.
+
+## Lists
+
+### Ordered
+
+Here's an ordered list of steps:
+
+1. Figure out how to migrate to [Astro][2].
+2. Make some *incredible docs*.
+3. **Profit.**
+
+### Unordered
+
+Here's an unordered version of the same list, which doesn't make as much sense, but that's okay.
+
+- Figure out how to migrate to [Astro][2].
+- Make some *incredible docs*.
+- **Profit.**
+
+## Code
+
+### Inline code
+
+Here's some `inline code`.
+
+### Code fence
+
+```javascript
+console.log("Hello, world!");
+```
+
+## Description list
+
+Service
+: Services are the building blocks of modern microservice architectures - broadly a service groups together endpoints, queries, or jobs for the purposes of building your application.
+
+Resource
+: Resources represent a particular domain of a customer application - they are typically an instrumented web endpoint, database query, or background job.
+
+`clusterChecksRunner.affinity.podAffinity.preferredDuringSchedulingIgnoredDuringExecution`
+: Required. A list of node selector terms. The terms are ORed.
+
+`site`
+: Set the site of the Datadog intake for Agent data:  {{< region-param key="dd_site" code="true" >}}. Defaults to `datadoghq.com`.
+
+
+[1]: https://www.google.com
+[2]: https://astro.build/
--- a/content/en/md_conversion_test_files/common_components/programming_lang_wrapper.md
+++ b/content/en/md_conversion_test_files/common_components/programming_lang_wrapper.md
@@ -1,0 +1,39 @@
+---
+title: Programming lang wrapper
+---
+
+## Expected .md output
+
+These are treated as tabs:
+
+- Each section starts with `[BEGIN TAB]` and end with `[END TAB]`.
+- Each section includes its label at the top (for example, "Python"), formatted as a heading.
+
+## Example input
+
+{{< programming-lang-wrapper langs="python,ruby,PHP" >}}
+
+{{< programming-lang lang="python" >}}
+Info about how to do this with Python.
+ ```python
+from datadog import initialize, statsd
+```
+{{< /programming-lang >}}
+
+{{< programming-lang lang="ruby" >}}
+Info about how to do this with Ruby.
+ ```ruby
+require 'datadog/statsd'
+```
+{{< /programming-lang >}}
+
+{{< programming-lang lang="PHP" >}}
+Info about how to do this with PHP.
+ ```php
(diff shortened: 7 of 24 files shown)


content/en/md_conversion_test_files/common_components/alert.md
```diff
@@ -0,0 +1,25 @@
+---
+title: Alert
```
onemec, line 2: **💬 suggestion:** ‏Maybe reusing [this syntax](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts) could be possible?

content/en/md_conversion_test_files/common_components/callouts.md
```diff
@@ -0,0 +1,24 @@
...
+- Each callout is marked by `[BEGIN CALLOUT]` and `[END CALLOUT]`.
+- Any attributes, such as the title or the URL, are included in the box contents:
+    - The title is a heading at the top of the box.
+    - The URL is included with verbiage at the bottom like "Visit this URL: <URL>".
+
+## Example inputs
+
+### Callout (preview / beta)
+
+{{< callout url="https://www.datadoghq.com/">}}
```
onemec, line 16: **❓ question:** ‏Do you have an example callout URL within the Datadog Docs page that I could view? I'm curious if there's a "standard" way we can render it in Markdown, but don't totally know what they look like

content/en/md_conversion_test_files/common_components/code_blocks.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Scope
+
+This covers the `code-block` and `highlight` shortcodes. Plain fences are already covered in the "plain markdown" examples.
+
+## Expected .md output
+
+Each code block is wrapped in a fence, with the language included when available:
+
+```lang
```
onemec, line 13: **🔨 warning:** ‏Not sure if this is an edge case you'll encounter or run into, but I've noticed occasionally code blocks contain triple backticks. A hacky workaround I've used (e.g. with LLMs) is to use quadruple backticks around code, though not sure if it's a case you'll run into and need to handle

content/en/md_conversion_test_files/common_components/further_reading.md
```diff
@@ -0,0 +1,20 @@
...
+    tag: "Documentation"
+    text: "Log processing pipelines"
+  - link: "some/blog/post"
+    tag: "Blog"
+    text: "Some blog post"
+---
+
+## Expected .md output
+
+A list of links, introduced by the text "Further reading:".
```
onemec: **💬 suggestion:** ‏I think you can use [custom anchors](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#custom-anchors) and [relative links](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#relative-links) for this functionality

content/en/md_conversion_test_files/common_components/alert.md
jhgilbert: I'm happy to try it! I don't always have as much control as I'd like, which is why the suggestions on output are fairly crude overall. But if there's a stretch goal like this, always happy to know about it and aim for it.

content/en/md_conversion_test_files/common_components/callouts.md
jhgilbert: Sure, here's [an example](https://docs.datadoghq.com/account_management/org_settings/ip_allowlist/#delete-ip-addresses-or-cidr-ranges), it's the box at the top of the page ("Get Started with IP Allowlist").

content/en/md_conversion_test_files/common_components/images_and_videos.md
```diff
@@ -0,0 +1,17 @@
+---
+title: Images and Videos
+---
+
+## Expected .md output
+
+A placeholder including relevant information about the asset, such as `[IMAGE src="some/src/path" alt="some alt text"]`.
```
onemec: **💬 suggestion:** ‏Markdown [supports](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#the-picture-element) the native `<picture>` HTML syntax

content/en/md_conversion_test_files/common_components/programming_lang_wrapper.md
```diff
@@ -0,0 +1,39 @@
+---
+title: Programming lang wrapper
+---
+
+## Expected .md output
+
+These are treated as tabs:
+
+- Each section starts with `[BEGIN TAB]` and end with `[END TAB]`.
```
onemec: **❓ question:** ‏What do you mean when you say `[BEGIN TAB]` and `[END TAB]`? Are these placeholder values that get populated? I saw them in some of the instructions, but didn't see them in examples

content/en/md_conversion_test_files/common_components/region_shortcodes.md
```diff
@@ -0,0 +1,36 @@
+---
+title: Region Shortcodes
+---
+
+## Site region shortcode
+
+### Expected .md output
+
+The contents are enclosed in `[BEGIN CONTENT FOR REGIONS "ap1", "gov"]` and `[END CONTENT FOR REGIONS "ap1", "gov"]`.
```
onemec: **💬 suggestion:** ‏I don't have a great suggestion for region-specific rendering, but maybe if you wanted to stick to using native Markdown AST processing, you could use a reserved namespace header like `# __reserved_docs.some_conditional_here` (which during artifact generation gets filtered out accordingly)

jhgilbert: I'm not worried about using a native Markdown AST, but it seems like that's a priority for you -- is that just to make sure the LLM will understand it, or is it because you want to be able to parse the files yourself for manipulation purposes? If it's the latter, I have some options around that I can show you, just to understand what you would prefer.

content/en/md_conversion_test_files/common_components/programming_lang_wrapper.md
jhgilbert: Sorry if that was confusing, the example is just the example input, not the output. In the output, a tab would look like this:

```
[BEGIN TAB]
# Python

Some Python-specific instructions.
[END TAB]
```

One issue I have is that it's sometimes hard to manipulate the content itself (for example, by indenting each line of it or something like that), so we may be stuck with start/stop delimiters in this quick "v0" version. If you need those delimiters to be machine-parsable, there are ways to do that -- for example, the Markdoc parser will happily parse Markdown that contains tags like `{% tab %}` and `{% /tab %}`. The `.md` file itself is just slightly less readable, but maybe that doesn't matter as long as the meaning is clear.

content/en/md_conversion_test_files/common_components/code_blocks.md
jhgilbert: Ooh, good to know, thank you!

content/en/md_conversion_test_files/common_components/further_reading.md
jhgilbert: I was just imagining plain Markdown links, since Further Reading links are to other pages (not to headers on the same page). But let me know if that wouldn't work for some reason, I'm not sure what the pros/cons are.

content/en/md_conversion_test_files/common_components/region_shortcodes.md
onemec: It's fully the latter (just for ease of parsing). Doesn't need to be now/soon, but as long as there's a non-browser-based way of parsing these, I'm open to your suggestions too (and don't feel strongly on this at all)

content/en/md_conversion_test_files/common_components/programming_lang_wrapper.md
onemec: ahh ok sure, I see what you mean. I would be in preference of adhering to any widely-used syntax (instead of "reinventing" ways to portray). to explain my thoughts a bit:

- The old non-native Markdown was problematic for LLMs since there was a lot of templated sections that resulted in missing context
- I can now see that "vanilla Markdown" doesn't cover all the necessary cases for your proposal, and agree it might not be the best fit (at least for the first set of changes)

in regards to how to do this, I'd veer in preference of whichever is the most widely adopted as long as:

- it's not causing crucial context to be missing (such as templated sections of missing content)
- it's able to be parsed (relatively) easily without needing a browser-based approach

(as an added benefit, the LLMs should also have more understanding of common syntax than adhoc syntax custom to us)

TLDR: `{% tab %}` and `{% /tab %}` seem good in my opinion unless there's another more common syntax that's feasible

content/en/md_conversion_test_files/common_components/further_reading.md
onemec: got it, sorry yep that sounds good to me! I misread that on first skim as `"Further reading:"` being some URL syntax 🤦‍♂️ sorry that's my bad

content/en/md_conversion_test_files/common_components/callouts.md
onemec: gotcha, thank you! I don't believe there's any native Markdown "banner", so I don't really have a concrete suggestion aside from my general preferences mentioned in this comment: https://github.com/DataDog/documentation/pull/31181#discussion_r2286260227

content/en/md_conversion_test_files/common_components/programming_lang_wrapper.md
jhgilbert: This is very helpful! So I can definitely switch to using easily readable tags, like `{% image src="something/or/other" %}` when I don't have a better "plain" option like "let's just convert this Further Reading section to a list of links".  It sounds like you would prefer that?

I'm limited in what "official Markdown" syntax I can use because this plain .md content is generated by Markdoc, which is intentionally pretty simple (it's designed for docs with lots of contributors who are not Markdown experts). 

It supports:

- the Commonmark Spec ([element list](https://markdoc.dev/docs/nodes))
- any tags you want (like `{% tab %}`, `{% box %}`, anything is fine). 
- And of course you can make plain text elements, like `[START TAB]` or whatever, since that's just text, but that's not parseable for you so best avoided, sounds like.

Any markup it generates is machine-parseable by Markdoc back into a very granular AST ([example](https://github.com/DataDog/corp-node-packages/pull/110/files#diff-cd4f1b299f1f326663ae61d9dc591305cadd45a776b0934e43d2eaf291ff822f)), which includes built-in functions for walking the AST, rendering it back to text, etc.

I should note that these limitations are just for v0. Markdoc is actually really powerful, and longer term we could write our own custom renderer that does whatever we want, but for this round, using its OOTB renderers at first means this takes weeks instead of months.

Do you think those tags would work okay for you? Most components wouldn't even need them, and we can name them whatever you want (so, `image`, etc., for max clarity).

onemec: Gotcha, perfect that makes total sense to me! thanks for all these details. It sounds like we're aligned on the path forward. I think OOTB parse-ability (and inferable context) are the main needs, which your plan seems to address 👍 Thanks for all the hard work on this

jhgilbert: Excellent, I will update all start/stop delimiters to tags instead. Thank you!

content/en/md_conversion_test_files/common_components/alert.md
jhgilbert: Updated to use tags, per other comment threads

content/en/md_conversion_test_files/common_components/region_shortcodes.md
jhgilbert: Resolved over DM, spec has been updated to be fully machine-parsable 👍

content/en/md_conversion_test_files/common_components/alert.md
```diff
@@ -0,0 +1,25 @@
...
+---
+
+## Expected .md output
+
+Each alert is enclosed in `{% alert %}` and `{% /alert %}` tags, with a `level` attribute.
```
jhgilbert, line 7: @onemec In cases like this, do you want me to also echo the attributes as plaintext content, like having "Info alert" as the bolded first line or something? Or does that get in the way? You can always add it yourself later in parsing as well, but happy to provide it up front if it's useful

content/en/md_conversion_test_files/common_components/plain_markdown.md
```diff
@@ -0,0 +1,79 @@
...
+
+Some text.
+
+###### Heading level 6
+
+Some text.
+
+## Inline text formatting
+
+You can *italicize* text, or **bold** it.
```
kayayarai: Do we need to account for using underscores to italicize and bold? This line could also be sourced as 
`You can _italicize_ text, or __bold__ it.`

jhgilbert: I believe Markdoc will pick it up either way, but I've added examples to make sure those instances are processed correctly, thank you!

content/en/md_conversion_test_files/common_components/tables.md
```diff
@@ -0,0 +1,72 @@
...
+title: Tables
+---
+
+## Plain Markdown tables
+
+### Expected .md output
+
+A key-value list for each table row:
+
+- Name: Jen
```
jhgilbert: @onemec I realized this is another instance of nonstandard formatting ... do you want this information expressed another way? I can just leave it as the plain Markdown table if you like, and you can break it up further on your end. I can also provide some kind of structured data if that would be better.

content/en/md_conversion_test_files/common_components/region_shortcodes.md
```diff
@@ -0,0 +1,37 @@
...
+{{< /site-region >}}
+
+## Region parameter shortcode
+
+### Expected .md output
+
+A placeholder: 
+
+- `<YOUR_DATADOG_SITE>` in code.
+- `{% user-datadog-site /%}` in copy.
```
jhgilbert, line 25: This one is tricky, we could also just say "your Datadog site" in plain text. But I wasn't sure if that would cause odd results sometimes in the content.

content/en/md_conversion_test_files/common_components/tables.md
```diff
@@ -0,0 +1,72 @@
+---
+title: Tables
+---
+
+## Plain Markdown tables
+
+### Expected .md output
```
brett0000FF, line 7: I think that this key-value pair reformat obscures the relationship between rows. In my opinion, it would be better to leave these as plain Markdown tables (maybe ensuring it follows GitHub flavored Markdown). These are already heavily represented in AI training data, so it should be well understood enough.

content/en/md_conversion_test_files/common_components/tables.md
```diff
@@ -0,0 +1,72 @@
...
+| Row 4, Col 1 | Row 4, Col 2 | Row 4, Col 3 |            |
+
+## HTML tables
+
+### Expected .md output
+
+Both of these should be included:
+
+- Whatever the generator can safely derive from the table, in "key-value list" form, with a disclaimer that the data may be incomplete.
+- The raw HTML contents, for further reference.
```
brett0000FF: For the fallback, we could wrap the raw HTML in a commented-out block or a generic `{% raw_html %}` tag. This would make it clear that the content is a less-reliable fallback and could be skipped by parsers if needed.

content/en/md_conversion_test_files/uncommon_components/kbd.md
```diff
@@ -0,0 +1,11 @@
+---
+title: kbd
+---
+
+## Expected .md output
+
+The `kbd` is replaced with **bold** markers.
```
brett0000FF: This definitely works, but I think it would be slightly better to represent keys with backticks rather than bold, since that is more common for keys and commands in Markdown.

content/en/md_conversion_test_files/uncommon_components/math.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Math
+---
+{{< jqmath-vanilla >}}
+
+## Expected .md output
+
+The literal contents are wrapped in a code fence labeled `jqmath`.
```
brett0000FF, line 8: Another stretch/long-term goal here could be to convert jqmath to LaTeX since that would be more universally recognized. Minor point though!

brett0000FF (commented):
This is so cool, Jen! Thank you for putting this together. I left a handful of small comments to consider, which don't necessarily have to be addressed in the initial iteration.

Outside of the specific transformations, I was curious if you have any thoughts on how to handle nested components? For example, alerts or code blocks inside of tabs, etc. It's probably straightforward in most cases, but it could be worth defining rules to disallow certain nesting behavior/defining fallbacks?

content/en/md_conversion_test_files/common_components/tables.md
jhgilbert: Thanks, Brett -- I was hearing that big tables were giving us some chunking issues, which made me think that third-party crawlers might encounter similar difficulties, but I'll check with Oliver to see if I'm overcorrecting. I'm completely fine with a plain Markdown table if there's no technical issue there!

PR #31347, opened 2025-08-28
Update supported monitor type by BitsAI

SimoneT-DD:
Update supported monitor type by BitsAI.
It replaces query_alert with APM Metrics, since query_alert is an internal label and not visible by customers


SimoneT-DD:
> Assuming APM Metrics is [commonly known](https://docs.datadoghq.com/monitors/types/apm/?tab=apmmetrics) as something different from "Trace Analytics Alert" (which we do not support)

This is how we label APM monitors in the UI and [documentation](https://docs.datadoghq.com/monitors/types/apm/?tab=apmmetrics#monitor-creation)

domalessi (approved):

PR #31393, opened 2025-09-01
Update notes about pytest-xdist support from ddtrace>=3.12.0

gnufede:
Updates the documentation regarding incompatible pytest plugins, now that we have worked on improving how our tracer interacts with them.

### Merge instructions

Merge readiness:
- [ ] Ready for merge


calvinbayer (approved):

content/en/tests/setup/python.md
```diff
@@ -452,11 +452,11 @@ Plugins for `pytest` that alter test execution may cause unexpected behavior.
...
 
-There are several issues when these plugins are used together with `ddtrace`. For example, a session, module, or suite may pass even when individual tests fail. Likewise, all the tests may pass and the suite/session/module fail. This happens because these plugins create worker subprocesses, and spans created in the parent process may not reflect the results from the child processes. For this reason, **the usage of `ddtrace` together with `pytest-xdist` and `pytest-forked` is not supported at the moment.**
+There are several issues when these plugins are used together with `ddtrace`, although all of them are resolved for `pytest-xdist` in the last versions of `dd-trace-py` (from 3.12.0 onwards). For example, a session, module, or suite may pass even when individual tests fail. Likewise, all the tests may pass and the suite/session/module fail. This happens because these plugins create worker subprocesses, and spans created in the parent process may not reflect the results from the child processes. For this reason, **the usage of `ddtrace` together with `pytest-forked` is not supported at the moment, while `ddtrace>=3.12.0` has support for `pytest-xdist`.**
```
joepeeples:
```suggestion
There are several issues when these plugins are used together with `ddtrace`, although they have been resolved for `pytest-xdist` in recent versions of `dd-trace-py` (3.12.0 and later). For example, a session, module, or suite may pass even when individual tests fail. Likewise, all the tests may pass and the suite/session/module fail. This happens because these plugins create worker subprocesses, and spans created in the parent process may not reflect the results from the child processes. For this reason, **the usage of `ddtrace` together with `pytest-forked` is not supported at the moment, while `pytest-xdist` only has support for `ddtrace>=3.12.0`.**
```

PR #31632, opened 2025-09-12
(maint) Replace shared workflow for version bumps

hestonhoffman:
I couldn't get Chainguard to work with a shared workflow, so I've replaced it with dedicated workflows and policies for each. A policy already exists for `bump-synthetic-worker-version.create-pr`. So I copied that one and modified it for `private-action-runner-version`.

### DIFF
--- a/.github/workflows/bump_synthetics_worker_version.yml
+++ b/.github/workflows/bump_synthetics_worker_version.yml
@@ -9,6 +9,7 @@ jobs:
     permissions:
       contents: write # for git push
       id-token: write # Needed to federate tokens.
+      pull-requests: write # Action creates a PR.
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/bump_synthetics_worker_version.yml
```diff
@@ -5,12 +5,73 @@ on:
   workflow_dispatch: # allows manual trigger
 
 jobs:
-  call-version-getter-shared-workflow:
+  synthetics_worker_version:
     permissions:
-      contents: write
-      pull-requests: write
```
devindford, line 11: are we sure we don't need this permissions since this does open a pull request?

hestonhoffman: Oh yes, thank you. I think I forgot to add this back in

PR #31868, opened 2025-09-26
add docs-blog-linker dir and contents to documentation repo

domalessi:
I wrote a script that can be used to add blog post links to the `further_reading` section of relevant documentation pages. This PR adds the script and supporting materials (all housed in a "docs-blog-linker" dir) to the documentation repo so the rest of the team can use it. 

### Merge instructions

Merge readiness:
- [x] Ready for merge

**For Datadog employees**:


local/bin/py/docs-blog-linker/excluded_docs.json
```diff
@@ -0,0 +1,5 @@
+{
```
jhgilbert: To integrate this tool into the build, it's helpful to get it down to one file (minus `requirements.txt`), even if the file is long. Until this list gets longer, could it just be a variable in the Python script itself?

If not, we could look into including this data in the Hugo config, but I wasn't sure how much you expect this list to grow over time.

local/bin/py/docs-blog-linker/rss_blog_to_docs.py
```diff
@@ -0,0 +1,663 @@
...
+yaml_rt.preserve_quotes = True
+yaml_rt.indent(mapping=2, sequence=4, offset=2)
+
+# ----------------------------
+# Config
+# ----------------------------
+
+BLOG_RSS = "https://www.datadoghq.com/blog/index.xml"
+DOCS_HOST = "docs.datadoghq.com"
+STATE_FILE = ".rss_state.json"
```
jhgilbert: I'd prefer that the design doesn't include an continually lengthening file that has to be checked into the repo, if we can avoid it. I noticed that your logic also includes a check on whether a given link is already included in a Further Reading section.

Considering the low number of posts published each week, and the fact that we know that two Tuesday Taskers will do this 2 weeks apart max, my recommendation is to default the script to looking 2 weeks back (or more if you want), and just let it process some links twice, since the worst that will happen is a no-op on a few links (if I'm understanding your code correctly). That way we can drop some of the logic around tracking processed links, giving us less to maintain and one less piece of perpetually expanding content.

local/bin/py/docs-blog-linker/rss_blog_to_docs.py
```diff
@@ -0,0 +1,663 @@
...
+GITHUB_API_BASE = os.getenv("GITHUB_API_BASE", "https://api.github.com")
+
+# Exclude list - documents that should not be modified
+def load_excluded_docs() -> dict:
```
jhgilbert: If excluded docs becomes a variable, this function would need to be updated.

local/bin/py/docs-blog-linker/rss_blog_to_docs.py
```diff
@@ -0,0 +1,663 @@
...
+        href = a["href"].split("#")[0].strip()
+        links.add(normalize_url(href))
+    return sorted(links)
+
+
+# ----------------------------
+# Frontmatter editing (ruamel.yaml, non-destructive)
+# ----------------------------
+
+def split_frontmatter(md_path: Path):
```
jhgilbert, line 174: I recommend using an existing module for loading frontmatter from a file, it will be less mistake-prone and more readable. Python isn't my first language, but something like `python-frontmatter` seems like it would work.

local/bin/py/docs-blog-linker/rss_blog_to_docs.py
```diff
@@ -0,0 +1,663 @@
...
+    Keeps all other keys (and their comments) in place.
+    """
+    if "title" in meta:
+        val = meta.pop("title")
+        # Insert at position 0. ruamel preserves comments on the map; if the title
+        # had inline comments, they typically aren't used; acceptable trade-off.
+        meta.insert(0, "title", val)
+
+
+def safe_update_further_reading(md_file: Path, blog_url: str, blog_title: str | None) -> tuple[bool, str]:
```
jhgilbert, line 343: This function is running pretty long, I recommend breaking it into smaller functions, and aiming for "code that just tells the story by invoking a bunch of clearly named functions" at the top level of the function.

local/bin/py/docs-blog-linker/rss_blog_to_docs.py
```diff
@@ -0,0 +1,663 @@
...
+        host = url.split("@", 1)[1].split(":", 1)[0]
+        owner_repo = url.split(":", 1)[1]
+    elif url.startswith("https://") or url.startswith("http://"):
+        parts = url.split("/")
+        host = parts[2]
+        owner_repo = "/".join(parts[3:])
+    return host, owner_repo
+
+
+def create_or_get_github_pr(repo_root: Path, branch: str, title: str, body: str, base: str = "master") -> str | None:
```
jhgilbert: GitHub token stuff tends to make the security folks nervous, and can be funky to maintain as the rules for repos change -- are you open to having the PR part be manual? The scope of this script could be "I added the blog links automatically, please push this branch and make a PR out of it."

local/bin/py/docs-blog-linker/RSS_BLOG_TO_DOCS_USAGE.md
```diff
@@ -0,0 +1,148 @@
...
+## Overview
+The `rss_blog_to_docs.py` script adds blog post links to the `further_reading` section of documentation pages that are referenced in Datadog blog posts. It does the following:
+1. Processes the [Datadog blog RSS feed](https://www.datadoghq.com/blog/index.xml)
+2. Finds documentation links in blog posts
+3. Updates the corresponding markdown files
+4. Commits the changes on a new branch, and pushes that branch to remote
+
+## Prerequisites
+
+1. [`ddtool`](https://datadoghq.atlassian.net/wiki/spaces/docs4docs/pages/3960766866/Build+setup+guide#Set-up-ddtool-for-token-handling) is installed.
```
jhgilbert: I'm thinking of this as a product for tech writers, and I want to make sure it's accessible to them. Env setup can be a major source of friction, especially for folks not comfortable with coding.

Ideally, we could roll this script into the docs build and make it usable with a simple `make` command, no setup required. Achieving that easily would require some simplifications to the tool, so I'm not sure how you feel about that. But I've made some suggestions to that end.

Ultimately, there are a lot of advantages in this problem space: there aren't a ton of blog posts shipping daily, we have a dedicated person to handle this with each Tuesday Task rotation, tech writers are used to Git and GitHub, etc. So I geared my feedback toward using those advantages and streamlining this tool. My hope is to ease its burden a little bit, in terms of how often it might break, how easy the code is to understand, how easy it is to "set up" (hopefully it's turnkey, otherwise people might be afraid to use it), and how often someone would need to maintain it. GitHub stuff especially just never stops changing, so what works one day can be permanently broken the next (or disallowed by new security guidelines).

Anyway, just wanted to share the philosophy I'm approaching this review with, because there's nothing wrong with what you have, I'm just aiming to be ... constructively cynical? ... about how much we want to maintain.

Happy to chat through it live too!

jhgilbert (commented):
I completed an initial pass, but there's a good amount to think about here, and it's pretty subjective. Happy to chat live!

local/bin/py/docs-blog-linker/rss_blog_to_docs.py
domalessi: @jhgilbert The reason I went with custom code is that when I tried `python-frontmatter` it would reorder the keys and change the indentation. The resulting format was correct, but all the changes led to messy diffs which I thought would make reviews harder. 

Thoughts? Definitely agreed that having custom code makes this whole thing more brittle and harder to read...

local/bin/py/docs-blog-linker/RSS_BLOG_TO_DOCS_USAGE.md
domalessi: Totally makes sense! Did some reworking to simplify things. Will post a more detailed comment shortly...

domalessi:
@jhgilbert Omg thank you SO MUCH for such a detailed review! Your feedback was so good 🥲 I tried to address everything and I think we're in a much better state, even if there's still work to be done! Here's what I've changed:
- **No more environment variable:** The script now auto-detects Git username for branch creation (worst case scenario, the user can change the branch name before pushing to remote)
- **Consolidated README and detailed guide**: No need for two! I can also delete the README and have a Confluence instruction guide if that's preferred? 
- **Simpler state handling:** Removed the state file. By default, the script reviews posts from last 14 days.
- **Removed "excluded docs" file**: Created a dict within the code that can be updated when/if needed.
- **Improved code readability**: Broke up the frontmatter parsing section for clarity. (Custom code remains -- see my comment in that section for my reasoning and my plea for help.)
- **Removed GH authentication**: The script no longer automatically push PRs; users can do that manually.
- **Automated dependency management**: Script checks and installs dependencies for you. 
- Simplified usage: Integrated with documentation repo's Makefile -- most people can just run `make update-blog-links`. README includes instructions for running the Python script directly for custom workflows.
- **Added blog publish dates**: Each output now includes the blog’s publish date for easier gut-checking.

local/bin/py/docs-blog-linker/rss_blog_to_docs.py
jhgilbert: Oh interesting, yeah I think it makes sense to keep your code if that's the case!

local/bin/py/docs-blog-linker/README.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+## How it works
+
+### Processing behavior
+- By default, the script processes blog posts from the last 14 days.
+- If a blog post link already exists in a doc's `further_reading` section, the script skips it (no-op).
+  - This design allows the script to be run multiple times safely without duplicating links.
+
+### Git workflow
```
jhgilbert, line 135: I don't recommend having your script interact with Git at all -- tech writers are used to making a new branch, running something, checking in the files themselves, etc., and any scripts that interact with Git tend to be treated as a security vulnerability. I would limit the script to just pulling the information it needs from RSS and updating the files. 

Beyond that, I would let tech writers manage their Git workflow however works best for them. It's still much easier than the existing workflow, but the domain of your script would be tightly scoped and straightforward for writers to understand. 

If you truly did want automatic branching/commits/etc., there doesn't need to be a tech writer involved at all beyond the oncall who approves the PR, since you could just run it in CI as a weekly job. But automating Git stuff is a more complex workflow that IMO is not worth the resources in this case, since your script is already offering such a drastic improvement!

As always, happy to chat live!

domalessi: Ah yes okay that makes a lot of sense! I just updated the script to remove all Git stuff. I think you're totally right -- removing this means users have to do a teeny bit more work, but it's part of their normal workflow anyway, and it simplifies the script greatly. Thank you for yet another round of v helpful feedback.

PR #31949, opened 2025-10-01
(fleet) drop uninstall instructions for remote agent management

arbll:
This PR removes the documentation entry suggesting users to run `datadog-installer purge` as remote agent management is now shipped as part of regular deb/rpm packages and no longer requires specific instructions to uninstall on linux.


buraizu (approved):
Verified in local build

PR #32395, opened 2025-10-24
Docs12110/metric discrepancy info

buraizu:
- Update metric discrepancy troubleshooting info
- Update style

### Merge instructions

Merge readiness:
- [ ] Ready for merge

**For Datadog employees**:

### DIFF
--- a/content/en/integrations/guide/aws-integration-troubleshooting.md
+++ b/content/en/integrations/guide/aws-integration-troubleshooting.md
@@ -55,7 +55,7 @@ See [Space aggregation][22] in the metric documentation for more information.
 
 #### Reconcile the discrepancy
 
-1. Go to **CloudWatch → All metrics**.
+1. Go to **CloudWatch**, and then **All metrics**.
 2. Search for and graph the metric. 
 3. Select the **Source** tab to show the full metric query.
 


content/en/integrations/guide/aws-integration-troubleshooting.md
```diff
@@ -20,49 +20,76 @@ The `sts:Assumerole` permission error indicates an issue with the trust policy a
...
 
-2. `min`, `max`, and `avg` have a different meaning within AWS than in Datadog. In AWS, average latency, minimum latency, and maximum latency are three distinct metrics that AWS collects. When Datadog pulls metrics from AWS CloudWatch, the average latency is received as a single timeseries per Elastic Load Balancer (ELB). Within Datadog, when you are selecting `min`, `max`, or `avg`, you are controlling how multiple timeseries are combined. For example, requesting `system.cpu.idle` without any filter returns one series for each host that reports that metric, and those series need to be combined to be graphed. If instead you requested `system.cpu.idle` from a single host, no aggregation is necessary and switching between average and max yields the same result.
+4. Confirm that the query in CloudWatch is scoped identically to the query in Datadog:
+   - Any [Dimensions][17] used in the CloudWatch metric query should match tags used in the Datadog metric query
+   - The [Statistic][18] used in the query should match the Datadog [space aggregator][19]
```
dwikle: This isn't really correct, but it's a bit too complicated for me to even try to explain it here.

I'm not sure how we can clearly explain this without explaining details which (today) we don't clearly document for customers.

On this specific point:
- We choose a specific statistic to ingest for each AWS Metric.  Often it's Average, but it may Sum, Min, or Max. 
- Sometimes we ingest multiple statistics for the same metric, and they are named differently in Datadog (with a suffix for additional stats)
- To compare with AWS, the statistic should be the one which we map to the Datadog metric in question.  The problem is we don't document that today.  We have the information (it's in our metrics specs) so we could explore adding that to the metric tables.

Example - https://github.com/DataDog/aws-metadata/blob/af858361e7bb49438391ac44ae447b33bf5d3734/metric-specs/aws/unified/amazon_lambda.yaml#L31-L75

--- 

None of the above has anything to do with the Datadog space aggregator.  The challenge in comparing AWS and Datadog is that we take very different approaches with space aggregation.  AWS models aggregation across multiple timeseries as separate, additional series, usually with the same metric name but a different dimension set.  On Datadog, we can aggregate over space at query time so we only ingest the most granular timeseries.

I think the recommendation for customers should be along the lines of "Datadog ingests the most granular time series, so the query on AWS should match"

dwikle: Another really important point that often trips up customers is the **Period** of the CloudWatch query.  We query metrics with a 1-minute granularity.  But the CloudWatch console often defaults to 5-minutes. This can cause mismatches.

So somewhere we should tell them to make sure the CloudWatch query is set to 1-minute precision/period.

dwikle: here's an internal doc I made which had some more details on the statistics. 

https://datadoghq.atlassian.net/wiki/spaces/AWS/pages/5434900614/CloudWatch+Metrics+Statistics

buraizu: @dwikle thanks for that feedback. I've pushed some updates in [bedf4a0](https://github.com/DataDog/documentation/pull/32395/commits/bedf4a0baf708174c45a1ce0713ea4894fe3a3f2). 

I've also opened DOCSENG-216 to explore listing the statistic in metric descriptions.

Thanks in advance for any additional feedback!

content/en/integrations/guide/aws-integration-troubleshooting.md
```diff
@@ -20,49 +20,75 @@ The `sts:Assumerole` permission error indicates an issue with the trust policy a
...
+
+Within Datadog, when you select `min`, `max`, or `avg`, you are controlling how multiple timeseries are combined. For example, requesting `system.cpu.idle` without any filter returns one series for each host that reports that metric, and those series need to be combined to be graphed. If instead you request `system.cpu.idle` from a single host, no aggregation is necessary and switching between average and max yields the same result.
+
+See [Space aggregation][22] in the metric documentation for more information.
+
+#### Reconcile the discrepancy
+
+1. Go to **CloudWatch → All metrics**.
+2. Search for and graph the metric. 
+3. Select the **Source** tab to show the full metric query.
```
michaelcretzman:
```suggestion
1. Go to **CloudWatch**, and then **All metrics**.
2. Search for and graph the metric. 
3. Select the **Source** tab to show the full metric query.
```

PR #32554, opened 2025-10-31
[AZINTS-3739] Reference LFO install template in integrations-management repo

gpalmz:
The Azure log forwarding install template is moving to a new location `https://raw.githubusercontent.com/DataDog/integrations-management/main/azure/logging_install/dist/azuredeploy.json`, so we should update the docs links. See linked ticket for reasons for moving this.

Looks like my editor deleted some trailing whitespace on save. I assume this is fine but lmk if not.


joepeeples (approved):
FYI, it's not necessary to update the Spanish version of the files, as they'll get updated as part of the translation process.

PR #32567, opened 2025-11-03
Delete content/en/data_streams/data_pipeline_lineage.md

johannbotha:
Removing lineage while we focus on other bets. We'll let DO be the entry point for this data.

### DIFF
--- a/content/en/data_streams/_index.md
+++ b/content/en/data_streams/_index.md
@@ -2,6 +2,7 @@
 title: Data Streams Monitoring
 aliases:
 - /data_streams/troubleshooting
+- /data_streams/data_pipeline_lineage
 further_reading:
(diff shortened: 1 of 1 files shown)


cswatt:
Pushed the following changes:
- Reset non-English menu files (we have a pipeline for all non-English content; you should never have to update a non-English file)
- Added an alias to the DSM index page so that the URL for the removed page will redirect to the index

johannbotha:
> * Reset non-English menu files (we have a pipeline for all non-English content; you should never have to update a non-English file)


Thanks I assumed there was a better way than manually editing them all 🙃

git-thuerk-done (approved):
looks good, please go ahead and hit the squash and merge button when this is ready!

PR #32835, opened 2025-11-14
Added DSM support for IBM MQ in Java

roisinlh:
### Merge instructions

Merge readiness:
- [ ] Ready for merge

**For Datadog employees**:

### DIFF
--- a/content/en/data_streams/setup/technologies/ibm_mq.md
+++ b/content/en/data_streams/setup/technologies/ibm_mq.md
@@ -12,11 +12,11 @@ title: Data Streams Monitoring for IBM MQ
 | [Java][4]  | [IBM MQ classes for Java and JMS][5]  | {{< dsm-tracer-version lang="java" lib="ibmmqjmsclient" type="minimal" >}}          | {{< dsm-tracer-version lang="java" lib="ibmmqjmsclient" type="recommended" >}}       |
 
 ### Limitations
-None of the IBM MQ tracers modify messages to add Data Streams Monitoring context information the way that tracers do for most other queue technologies. This is to avoid risk to customer services if unexpected additional fields appear in messages.
+For other queue technologies, Datadog's tracers modify messages to add Data Streams Monitoring context information. However, context propagation for IBM MQ is error-prone, as unexpected additional fields can appear in messages. To avoid risk to customer services, Datadog does not propagate context for IBM MQ traces.
 
-Due to this limitation, the Data Streams Monitoring pathways view will not be able to filter IBM MQ messages based on upstream pathway, and latency for pathways that entirely flow through IBM MQ will be approximated.
+Because of this limitation, the Data Streams Monitoring pathways view cannot filter IBM MQ messages based on upstream pathway.
 
-Message throughput and full presence on the Data Streams Topology map is fully supported in these tracers.
+Latency metrics for pathways that entirely flow through IBM MQ are available, though they are approximated. Message throughput and full presence on the Data Streams Topology map are fully supported.
 
 ### Setting up Data Streams Monitoring
 See setup instructions for [.NET][2] or [Java][4].


content/en/data_streams/setup/technologies/ibm_mq.md
```diff
@@ -9,10 +9,20 @@ title: Data Streams Monitoring for IBM MQ
 | Language     | Library                                                                                        | Minimal tracer version                                                            | Recommended tracer version                                                            |
 |--------------|------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------|---------------------------------------------------------------------------------------|
 | [.NET][2] | [IBMMQDotnetClient][3]           | {{< dsm-tracer-version lang="dotnet" lib="ibmmqdotnetclient" type="minimal" >}}          | {{< dsm-tracer-version lang="dotnet" lib="ibmmqdotnetclient" type="recommended" >}}          |
+| [Java][4]  | [IBM MQ classes for Java and JMS][5  | {{< dsm-tracer-version lang="java" lib="ibmmqjmsclient" type="minimal" >}}          | {{< dsm-tracer-version lang="java" lib="ibmmqjmsclient" type="recommended" >}}       |
```
shellymat-dd: This part doesn't show as link for me 
image

content/en/data_streams/setup/technologies/ibm_mq.md
```diff
@@ -9,10 +9,20 @@ title: Data Streams Monitoring for IBM MQ
...
+
+### Limitations
+None of the IBM MQ tracers modify messages to add Data Streams Monitoring context information the way that tracers do for most other queue technologies. This is to avoid risk to customer services if unexpected additional fields appear in messages.
```
shellymat-dd: Thank you for explaining the limitation, I think it's important for customers to understand what they'll get. 

Can we simplify? @ceciliawatt let us know if you prefer to do it. Basically what we are saying:
1. IBM MQ has a limitation preventing us from propagating context
2. This means that:
    a. You can still get understand data flows, visualize your pipeline automatically, and get throughput metrics 
    b. The latency metric would be an estimated average rather than exact distribution 
    
@roisinlh correct me if I got anything wrong

content/en/data_streams/setup/technologies/ibm_mq.md
roisinlh: I had a typo in the link, fixing

content/en/data_streams/setup/technologies/ibm_mq.md
roisinlh: For point 1, I'm not sure the best way to explain why we don't propagate context in IBM MQ for any language. It's not a limitation of IBM MQ itself, but rather that doing so would cause errors in production for many customers:

IBM MQ does have the technical capability to do so, but it was discovered when we launched for .Net that many customers would have production systems fail if there was any unexpected data included. Making matters worse, frequently this failure happens in a mainframe program written in something like COBOL that we can't provide direct support for and many customers don't have an ability to change.

We could in the future add context propagation for both Java and .Net, but it would have to be explicitly opt-in due to the significant risk involved.

---

For point 2, we could adjust the language to something like what you suggest if that is more appropriate.

cswatt: I think I might reword this segment to the following:

"For other queue technologies, Datadog's tracers modify messages to add Data Streams Monitoring context information. However, context propagation for IBM MQ is error-prone, as unexpected additional fields can appear in messages. To avoid risk to customer services, Datadog does not propagate context for IBM MQ traces.

Because of this limitation, the Data Streams Monitoring pathways view cannot filter IBM MQ messages based on upstream pathway. 

Latency metrics for pathways that entirely flow through IBM MQ are available, though they are approximated. Message throughput and full presence on the Data Streams Topology map are fully supported."

Any thoughts?

roisinlh: That sounds good, I've updated to include those changes.

cswatt (approved):

PR #32875, opened 2025-11-17
fix resource DD_TRACE_SAMPLING_RULES example

adrienbrault:
Hey, I only got the exclusion to work with a `/` before the resource/path.

I am using FastAPI


adrienbrault:
Nevermind, I thought DD_TRACE_SAMPLING_RULES was meant for the traced application, not the agent

PR #32932, opened 2025-11-19
[DOCS-12434] Clarify tag key overlap

evazorro:

Adds a paragraph clarifying that you can have the same tag key multiple times, and how to avoid it. I also cleaned up a few small things that I noticed while reading through the tagging docs.

### Merge instructions

Merge readiness:
- [x] Ready for merge

### DIFF
--- a/content/en/getting_started/tagging/_index.md
+++ b/content/en/getting_started/tagging/_index.md
@@ -99,6 +99,8 @@ As a best practice, Datadog recommends using unified service tagging when assign
 
 All metrics, logs, traces, and integrations go through a process of `host-tag` inheritance as data is ingested into Datadog. Since data is associated with a given hostname, those components inherit all the `host-level` tags associated with that host. These tags are visible in the [infrastructure list][12] for a given host, sourced from either the cloud provider or the Datadog Agent. See [missing `host-level` tags on new hosts or nodes][25] for more information.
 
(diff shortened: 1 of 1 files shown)


content/en/getting_started/tagging/_index.md
```diff
@@ -114,13 +113,15 @@ service:orders
 
 Downstream filters or dashboards should explicitly filter on the desired value if you expect only one.
 
+Also, given that tags can be inherited from multiple sources, you may want to take care with choosing key names that could overlap across sources. For example, if you've set a `service` key on a host (`service:my-host`) and a `service` key on a pod running on that host (`service:my-service`), your data inherits both tags. Opt for more differentiated key names (such as `infra_service`) to avoid duplicate tag keys.
```
evazorro: This is the main new content added.

All other changes in this PR are linter fixes, punctuation, formatting, etc.

JacksonDavenport: The added segment looks good. However, looking at the doc in full now I think it would read better to move this up one section to the "Tag Inheritance" section right above it. Especially since this is relative to exactly that that type of issue.

joepeeples: A few suggested edits to make the recommendation more direct. Also "overlap" is kinda metaphorical and might not be immediately understandable (and might be tricky for translations); maybe "duplicate" or something similar?
```suggestion
Also, because tags can be inherited from multiple sources, choose unique and specific key names to avoid duplicating them across sources. For example, if you've set a `service` key on a host (`service:my-host`) and a `service` key on a pod running on that host (`service:my-service`), your data inherits both tags. Opt for more differentiated key names (such as `infra_service`) to avoid duplicate tag keys.
```

evazorro: I like both these suggestions a lot, thank you!

PR #33112, opened 2025-12-01
[Cloud Run Jobs] Add instructions on how to disable APM and how to configure a retention filter

nhulston:

https://datadoghq.atlassian.net/browse/SVLS-8084

- In https://github.com/DataDog/datadog-agent/pull/43371, we now automatically create spans for the duration of a Cloud Run Job task
- We can now rely on the traces for task & execution duration instead of enhanced metrics. We can't rely on enhanced metrics because of high cardinality tags, which will soon be removed from the agent
- In https://github.com/DataDog/web-ui/pull/251927, I update the frontend to rely on traces instead of enhanced metrics with high cardinality. (We still use non-high-cardinality metrics in some places)
- Because the frontend now relies on traces, we must tell users how to configure a retention filter
- Additionally, since the agent now automatically creates spans, I give instructions on how to disable them

### Merge instructions

Merge readiness:
- [x] Ready for merge


DavidWeid:
@nhulston In preview, the cloud run jobs image doesn't load (404)

PR #33124, opened 2025-12-02
[DOCS-12786] Remove duplicate Further Reading sections

evazorro:
Someone reported that the [`rsyslog` integration](https://docs.datadoghq.com/integrations/rsyslog/) had two identical **Further Reading** sections one after the other. Upon investigation, I found that all the integrations in this part of the repo have the issue; the Further Reading partial was being applied twice during the build process. I've removed the in-line version of the partial so that it will only appear once ([see preview](https://docs-staging.datadoghq.com/eva/DOCS-12786-rsyslog/integrations/rsyslog/?tab=ubuntuanddebian#troubleshooting)).

_One thing to note is that the Further Reading heading that remains on the page doesn't function as a jump link._ Let me know if this is something we're concerned about.

### Merge instructions

Merge readiness:
- [x] Ready for merge

### DIFF
--- a/layouts/integrations/single.html
+++ b/layouts/integrations/single.html
@@ -24,7 +24,7 @@
     {{/* "We don't want to do this on integration guides/faqs for now" */}}
     {{- if and ($further_reading) (eq .Params.custom_kind "integration") -}}
 
-    <h2>Further Reading</h2>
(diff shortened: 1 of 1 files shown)


rtrieu:
Hi @evazorro nice fix! The Further Reading isn't that critical but it is nice to have. Do we know why it doesn't work as an anchor, and is there anything we can do about it?

PR #33571, opened 2025-12-24
fix: update feature flag node.js docs to properly set context

greghuels:
The Node.js docs for OpenFeature suggested using `OpenFeature.setContext` to set user-specified attributes. This is not a good practice for server SDKs since it leads to race conditions. Examples in the docs have been updated to evaluation-specific context.
### Merge instructions

Merge readiness:
- [x] Ready for merge

### DIFF
--- a/content/en/feature_flags/server/nodejs.md
+++ b/content/en/feature_flags/server/nodejs.md
@@ -74,7 +74,7 @@ app.get('/my-endpoint', async (req, res) => {
   await initializationPromise;
 
   const evaluationContext = {
-    userID: req.session?.userID,
+    targetingKey: req.session?.userID,
     companyID: req.session?.companyID
   };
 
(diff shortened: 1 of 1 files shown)


content/en/feature_flags/server/nodejs.md
```diff
@@ -73,12 +73,12 @@ const client = OpenFeature.getClient();
 app.get('/my-endpoint', async (req, res) => {
   await initializationPromise;
 
-  OpenFeature.setContext({
+  const evaluationContext = {
     userID: req.session?.userID,
     companyID: req.session?.companyID
-  });
+  };
```
dd-oleksii, line 79: shall we set `targetingKey` here?

greghuels: Yeah, we definitely should. I'll use userID as the targeting key

PR #33579, opened 2025-12-24
Add case custom status feature

estherk15:
### Merge instructions

Merge readiness:
> [!WARNING]
> Pending product GA


content/en/incident_response/case_management/customization.md
```diff
@@ -86,6 +87,29 @@ You can add attributes that reflect your team's specific needs, such as escalati
...
+- Standardize how work moves through each stage of a case lifecycle.
+- Reflect reviews, handoffs, and intermediate states.
+- Improve reporting and automation by aligning statuses with your processes.
+
+### Create a custom status
+
+1. Navigate to [**Settings > Shared Settings > Case Types**][2].
+2. Select the case type you want to update.
+3. Scroll to the **Statuses** section.
+4. Add a new status under one of the three existing status groups: **Open, In-Progress,** or **Closed**.
```
janine-c, line 104:
```suggestion
4. Add a new status under one of the three existing status groups: **Open, In Progress,** or **Closed**.
```

content/en/incident_response/case_management/customization.md
```diff
@@ -86,6 +87,29 @@ You can add attributes that reflect your team's specific needs, such as escalati
...
+5. (Optional) Set a new **default status** for each status group. Default Statuses are used in automations as the preferred statuses for the group when exact status names are not provided.
```
janine-c, line 105:
```suggestion
5. (Optional) Set a new **default status** for each status group. Default statuses are used in automations as the preferred statuses for the group when exact status names are not provided.
```

content/en/incident_response/case_management/customization.md
```diff
@@ -86,6 +87,29 @@ You can add attributes that reflect your team's specific needs, such as escalati
...
+
+<!-- \[Screenshot here\] -->
+
+* Each status group (Open, In-Progress, Closed) must contain at least one status.
```
janine-c, line 109:
```suggestion
* Each status group (Open, In Progress, Closed) must contain at least one status.
```

content/en/incident_response/case_management/customization.md
```diff
@@ -86,6 +87,29 @@ You can add attributes that reflect your team's specific needs, such as escalati
...
+* You can delete an existing status, but you must first migrate any cases currently using that status to another status in the same group.
+* Custom statuses behave exactly the same as Datadog's built-in statuses.
```
janine-c, line 111: These bullets are all good information, but I think they could use a little structural signposting to indicate what the bulleted list actually is. (I tried to think of an analogy, and maybe it's like if you opened an old recipe book and a sticky note fell out that said "temperature actually needs to be 375 degrees or else the muffins will explode." And you're like, this is good and likely correct information, but when do I apply it??)

Because it's more general information that doesn't directly pertain to creating custom statuses, I don't think it should be under the "Create a custom status" heading. Somewhere in the intro would work, and/or you could create an "Understand how custom statuses behave" header or something similar. That way, users will have a clearer idea of what the bulleted list is going to contain before they've even started reading it 🙂

content/en/incident_response/case_management/customization.md
```diff
@@ -86,6 +87,29 @@ You can add attributes that reflect your team's specific needs, such as escalati
      * Number
    * Choose whether to allow multiple values for this attribute.
 
+## Custom statuses
+
+Case Management supports customizable case statuses. By default, cases move through Open, In Progress, and Closed. You can add additional statuses to represent reviews, handoffs, or other workflow steps.
+
+Custom statuses help you:
+- Standardize how work moves through each stage of a case lifecycle.
```
janine-c, line 95:
```suggestion
- Standardize how work moves through each stage of a case life cycle.
```
Thank you Vale!

content/en/incident_response/case_management/customization.md
janine-c: I don't see the hyphen in the UI, and it's not a compound adjective, so I think this is okay?

janine-c (approved):
Fantastico! I have some small suggestions, but nothing showstoppy at all :)

PR #33620, opened 2025-12-30
Added the count object and its attributes

thang62590:
Added the count object and its attributes


git-thuerk-done (approved):
Looks good thank you!

PR #33687, opened 2026-01-06
[DOCS-13007] Add back removed links to SAML providers

evazorro:
Support flagged that there were two reference links showing up as `[7]`, `[8]` on this page. It appears that in my previous update to the doc, I accidentally removed the link references from the bottom of the file. This PR adds them back so the links display correctly.

### Merge instructions

Merge readiness:
- [x] Ready for merge


rtrieu (changes requested):
Hi @evazorro it looks like link 9 still needs to be added back in as well.

evazorro:
> Hi @evazorro it looks like link 9 still needs to be added back in as well.

Doh -- thank you! Just added it.

PR #33709, opened 2026-01-07
Fix a bunch of lettered lists and alignment

janine-c:

This PR looks large, but it should be easy to review! I did it because I noticed that there were instances of nested lists that were formatted like
```
1. Top-level list item
   a. Nested list item
```

Most commonly, using these lettered lists caused issues like the list items all appearing on a single line, or writers having to put in rogue `<br /` tags to make them appear as expected.

The formatting we need for lists to be rendered correctly looks like
```
1. Top-level list item
   1. Nested list item
```

So I went through and updated them. While I was at it, I also made a handful of minor typo/punctuation edits and whitespace edits to make code blocks align with those list items, which now had their own proper alignments! I know you're as thrilled as I am about this!

As always, I checked the changes I made in my local build while I did them. Your mission, should you choose to accept it, is to check to ensure I didn't delete something dumb or change what I shouldn't have, but overall it should be fairly easy to spot check. Let me know if you have any questions!

### Merge instructions

Merge readiness:
- [x] Ready for merge

**For Datadog employees**:

### DIFF
--- a/content/en/tracing/trace_collection/custom_instrumentation/android/dd-api.md
+++ b/content/en/tracing/trace_collection/custom_instrumentation/android/dd-api.md
@@ -28,684 +28,624 @@ Send [traces][1] to Datadog from your Android applications with [Datadog's
 ## Setup
 
 1. Add the Gradle dependency by declaring the library as a dependency in your `build.gradle` file:
-
-```groovy
-dependencies {
-  implementation "com.datadoghq:dd-sdk-android-trace:x.x.x"
-}
-```
-
+   ```groovy
+   dependencies {
+     implementation "com.datadoghq:dd-sdk-android-trace:x.x.x"
+   }
+   ```
 2. Initialize Datadog SDK with your application context, tracking consent, and
    the [Datadog client token][4]. For security reasons, you must use a client token: you cannot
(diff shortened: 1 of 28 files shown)


content/en/tracing/trace_collection/custom_instrumentation/android/dd-api.md
```diff
@@ -28,684 +28,624 @@ Send [traces][1] to Datadog from your Android applications with [Datadog's
...
-```kotlin
-class SampleApplication : Application() {
-   override fun onCreate() {
-       super.onCreate()
-       val configuration = Configuration.Builder(
+   ```
+   {{% /tab %}}
+   {{% tab "Java" %}}
+   ```java
+      public class SampleApplication extends Application {
```
fuzzybinary, line 97: I think sometimes these code blocks are getting extra indentation, and sometimes they're not. I'm not sure if it affects rendering but I thought I would point it out.

For example, this `java` block is indented one level above the "```java"  whereas the one below with `kotlin` isn't.

It looks like prior all code is indented to the same level as the opening block.

fuzzybinary (approved):
Super minor nit about the indentation within code blocks. It's possible it doesn't affect rendering though and if that's the case it can be ignored.

PR #33993, opened 2026-01-20
[DOCS-11387] Add MySQL source doc

maycmlee:
### Merge instructions

Merge readiness:
- [ ] Ready for merge

**For Datadog employees**:

### DIFF
--- a/content/en/observability_pipelines/sources/mysql.md
+++ b/content/en/observability_pipelines/sources/mysql.md
@@ -102,12 +102,15 @@ Datadog recommends using a unique incremental identifier for the incremental col
 
 ##### Checkpoint values
 
-Checkpoint values are updated every job run. To monitor the checkpoint value, there are Worker logs that contain the message `Checkpoint updated` and the latest value that was published. If a job fails or the Worker unexpectedly restarts or shuts down, the checkpoint value reverts to the start value. To determine the checkpoint value for a job that failed:
+Checkpoint values are updated every job run. To monitor the checkpoint value, there are Worker logs that contain the message `Checkpoint updated` and the latest value that the Worker queried. If a job fails or the Worker unexpectedly restarts or shuts down, the checkpoint value reverts to the start value set in the MySQL source. To determine the checkpoint value for a job that failed:
 
 1. Navigate to [Log Explorer][3] and search for Worker logs with the message `Checkpoint updated`.
-1. Check the value found in the latest Worker log to see what the Worker tracked.
-1. Check the log in your destination to which your logs were sent and determine the last value sent.
-1. Manually reset the checkpoint value in the MySQL source in the pipelines UI.
+1. Check the latest Worker log to see the last checkpoint value the Worker received from querying the database.
+1. Check the logs in your destination to see the last checkpoint value it received from the Worker. This value might not be the same as the checkpoint value seen in the latest Worker log. If you want to ensure that every record is sent to the destination at least once, reset the starting checkpoint value in the MySQL source with the destination's last checkpoint value.
+1. To reset the checkpoint value in the MySQL source:
+    1. Navigate to [Observability Pipelines][1].
+    1. Click on your pipeline and click.
+    1. In the MySQL source, enter the destination's last checkpoint value in the **Checkpoint value** field.
 
 #### External tools for validating queries
 


content/en/observability_pipelines/sources/database.md
```diff
@@ -0,0 +1,196 @@
...
+### External tools for testing
+
+Datadog recommends these tools for testing:
+
+- Third-party tools:
+  - DBeaver
+  - DataGrip
+  - TablePlus
+  - DbVisualizer
+- Native tools:
```
maycmlee: Only keep the relevant one when splitting to Postgres and MySQL docs.

content/en/observability_pipelines/sources/database.md
```diff
@@ -0,0 +1,196 @@
...
+  - For MySQL: MySQL Workbench
+  - For Postgres: psql or pgAdmin
+
+## Set up the source in the pipeline UI
+
+Set up the Database source and its environment variables when you [set up a pipeline](https://app.datadoghq.com/observability-pipelines). The information below is configured in the pipelines UI.
+
+1. Ensure you have completed the [prerequisite steps](#prerequisites).
+1. Select the type of database you want to collect data from:
```
maycmlee: Remove when splitting the docs.

content/en/observability_pipelines/sources/database.md
```diff
@@ -0,0 +1,196 @@
...
+    - **MySQL**
+      - Includes AWS Aurora
+    - **Postgres**
+1. Configure the connection between the Observability Pipelines Worker and the database.
+    1. Enter the connection string.
+        - **MySQL** connection string format: `mysql://${DB_USERNAME}:${DB_PASSWORD}@${DB_HOST}:${DB_PORT}/${DB_NAME}`
```
maycmlee: Keep the relevant one when splitting doc.

content/en/observability_pipelines/sources/database.md
```diff
@@ -0,0 +1,196 @@
...
+  - Edge device telemetry that gets stored in databases: Some smaller devices write events, such as diagnostics, maintenance, and errors, into SQL tables. For example, a pacemaker syncs periodically and saves the records in a database, which a DevOps team then uses in their logging tool.
+- Pull user or product information stored in databases to assist in troubleshooting issues or investigating threats.
+
+## Prerequisites
+
+Before you configure the Database source, complete the following prerequisites to help ensure that credentials, connectivity, and queries are validated before they are used in Observability Pipelines. These steps must be completed using [external tools](#external-tools-for-testing), such as in your database or with third-party tools.
+
+1. Create a database role for log collection, if you don't already have one. The role must:
+- Only have read-only access, so no permissions to modify or write data.
+- Have permission to execute the target queries that will be used for log collection.
```
maycmlee:
```suggestion
- Have permission to execute the target queries used for log collection.
```

content/en/observability_pipelines/sources/database.md
```diff
@@ -0,0 +1,196 @@
...
+
+- URI connection string
+  - THE URI that contains the necessary parameters, such as the database engine, host, port, and credentials, to connect to a database.
+  - Stored as environment variable: `DB_SOURCE_DATABASE_CONNECTION_STRING`
+
+## Incremental queries
+
+### Incremental query syntax
+
+For incremental querying, SQL queries must have the following to ensure consistent results.
```
maycmlee:
```suggestion
For incremental querying, SQL queries must have the following to help ensure consistent results.
```

content/en/observability_pipelines/sources/database.md
```diff
@@ -0,0 +1,201 @@
...
+- Use the correct host, port, database name, and authentication mechanism.
+- Be tested prior to configuring it in Observability Pipelines, to avoid runtime failures.
+
+1. Write, validate, and test SQL queries using either third-party or native database management tools.
+- All SQL queries must be validated and tested with an external tool prior to configuring it in Observability Pipelines.
+- Ensure that each query executes successfully using the read-only rule and returns the expected schema.
+
+### External tools for testing
+
+Datadog recommends these tools for testing:
```
maycmlee: Do we want to say `Datadog recommends..` or just these are `example external tools that you can use`?

content/en/observability_pipelines/sources/database.md
```diff
@@ -0,0 +1,201 @@
...
+### Queries
+
+#### Only read-only queries are executed
+
+- Only `SELECT` and `SELECT DISTINCT` statements are supported.
+- Queries that modify data (`INSERT`, `UPDATE`, `DELETE`, `DDL`) are all rejected.
+
+#### Validate and storing queries
+
+- SQL queries must be tested and validated outside of Observability Pipelines' UI. Datadog recommends using a third-party tool, such as DBeaver, or native SQL manager, such as MySQL Workbench, for testing.
```
maycmlee: For Postgres, mention psql or pgAdmin instead of MySQL Workbench.

content/en/observability_pipelines/sources/database.md
```diff
@@ -0,0 +1,201 @@
...
+
+- Database users, roles, and permission must be created and managed outside of Datadog.
+- Connection strings should reference environment variables for secrets.
+
+## Recommended monitors
+
+Datadog recommends creating these monitors:
+
+- Connection timeout monitor to alert you when the Worker cannot connect to the database.
+- Query failure monitor to alert you when the query is timing out and failing.
```
maycmlee:
```suggestion
- Query failure monitor to alert you when the query is times out and fails.
```

content/en/observability_pipelines/sources/database.md
```diff
@@ -0,0 +1,199 @@
...
+- `Varchar`
+- `Int`, `float`, `real`, `bigint`, `number`
+- `Timestamp`
+- `Datetime`
+
+Datadog recommends using an incremental ID for the incremental column . If you use an identifier that might not be unique, such as timestamps, it could result in data conflicts.
+
+**Notes**:
+- There could be data loss if you use strict operators, such as `>` or `<`, in your query.
+- There could be duplicated data if you use inclusive operators, such as `>=`.
```
tessneau: tioli: wonder if we can make it clear this is specific to using a non-unique identifier and not general?

content/en/observability_pipelines/sources/database.md
```diff
@@ -0,0 +1,199 @@
...
+### Incremental columns
+
+Incremental columns are used to track the progress of each new query. The following column types for incremental querying are supported:
+
+- `Varchar`
+- `Int`, `float`, `real`, `bigint`, `number`
+- `Timestamp`
+- `Datetime`
+
+Datadog recommends using an incremental ID for the incremental column . If you use an identifier that might not be unique, such as timestamps, it could result in data conflicts.
```
maycmlee:
```suggestion
Datadog recommends using an incremental ID for the incremental column . If you use an identifier that might not be unique, such as timestamps, it could result in data loss or duplicated data. For example:
```

content/en/observability_pipelines/sources/database.md
maycmlee: @tessneau ah okay, I hadn't realized it was for a specific situation. I'm thinking about adding it to the previous paragraph, let me know what you think of linking it to the previous paragraph:

```suggestion
- There could be data loss if you use strict operators, such as `>` or `<`, in your query.
- There could be duplicated data if you use inclusive operators, such as `>=`.
```

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,193 @@
...
+- There could be data loss if you use strict operators, such as `>` or `<`, in your query.
+- There could be duplicated data if you use inclusive operators, such as `>=`.
+
+### Checkpoint values
+
+Checkpoint values are updated every job run. To monitor the checkpoint value, there are Worker logs that contain the message `Checkpoint updated` and the latest value that was published. If a job fails or the Worker is restarted mid-job or it crashes, the checkpoint value reverts to the start value. To determine the checkpoint value for a job that failed:
+
+1. Navigate to [Log Explorer][3] and search for Worker logs with the message `Checkpoint updated`.
+1. Check the value found in the latest Worker log to see what the Worker tracked.
+1. Check the log in your destination to which your logs were sent and determine the last value sent.
```
maycmlee: Should the value match the one found in the Worker log?

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,193 @@
...
+1. Manually reset the checkpoint value in the Database source in the pipelines UI.
```
maycmlee: Are they resetting the value in the source to the value in the Worker log?

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,193 @@
...
+    - Be able to successfully connect to the database from the environment in which the Observability Pipelines Worker runs.
+    - Use the correct host, port, database name, and authentication mechanism.
+    - Be tested prior to configuring it in Observability Pipelines, to avoid runtime failures.
+1. Write, validate, and test SQL queries using either third-party or native database management tools.
+    - All SQL queries must be validated and tested with a tool external to Observability Pipelines and prior to configuring it in Observability Pipelines.
+    - Ensure that each query executes successfully using the read-only rule and returns the expected schema.
+
+### External tools for validating queries
+
+Datadog recommends these tools for validating and testing queries:
```
maycmlee: Do we want to say Datadog recommends these tools?

content/en/observability_pipelines/sources/database.md
tessneau: seems good to me 👍

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+- name: Logs
+  icon: logs
+  url: /observability_pipelines/configuration/?tab=logs#pipeline-types
+---
+
+{{< product-availability >}}
+
+## Overview
+
+Databases are often used to store many historical, audit, or operational records. For many legacy, enterprise resource planning (ERP), and IoT-based systems, these databases serve as storage layers for important information. Teams often depend on these records for monitoring, alerting, and creating dashboards in their preferred logging or security tool.
```
janine-c:
```suggestion
Databases often contain large numbers of historical, audit, or operational records. For many legacy, enterprise resource planning (ERP), and IoT-based systems, these databases serve as storage layers for important information. Teams often depend on these records for monitoring, alerting, and creating dashboards in their preferred logging or security tool.
```
Thought this might help with framing, but no pressure to actually accept it!

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+
+The Observability Pipelines' MySQL source (includes Amazon RDS and AWS Aurora) allows you to connect to your database so you can query and process the data in Observability Pipelines, and route your log events that are stored as database records.
```
janine-c:
```suggestion
The Observability Pipelines' MySQL source (includes Amazon RDS and AWS Aurora) allows you to connect to your database so you can query and process record data in Observability Pipelines, and route your log events that are stored as database records.
```
Not sure if this is right? Just thought we might want to make "the data" a little more specific if we can 🙂

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+
+**Note**: The Observability Pipelines Worker can only execute read-only SQL queries against supported databases.
+
+### When to use this source
+
+Common scenarios when you might use this source to:
```
janine-c:
```suggestion
You can use this source to:
```
This read a little awkwardly; I think we can keep it simple!

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+
+- Periodically extract transaction or access records stored in MySQL for audit and compliance reporting.
+- Analyze, alert, and build dashboards on event data stored in MySQL. For example:
+  - Application logs: Many legacy, regulated, or IoT devices write records to MySQL tables. These events often contain session activity, device information, and custom application logs. The Database source lets you extract the data you want and ingest it as logs for downstream alerting and investigation.
+  - Operational events and business-critical records: Many organizations store operational events in MySQL as a system of record. These databases contain data like ERP, billing, order, inventory, ticketing, and fulfillment info. Teams often query tables for dashboards, scheduled alerts, and investigations
```
janine-c:
```suggestion
  - Operational events and business-critical records: Many organizations store operational events in MySQL as a system of record. These databases contain data like ERP, billing, order, inventory, ticketing, and fulfillment info. Teams often query tables for dashboards, scheduled alerts, and investigations.
```

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+  - Edge device telemetry: Some smaller devices write events, such as diagnostics, maintenance, and errors, into SQL tables. For example, a pacemaker syncs periodically and saves the records in MySQL, which a DevOps team then uses in their logging tool.
+- Pull user or product information stored in MySQL to assist in troubleshooting issues or investigating threats.
+
+## Prerequisites
+
+Before you configure the MySQL source, complete the following prerequisites to help ensure that credentials, connectivity, and queries are validated before they are used in Observability Pipelines. These steps must be completed using [tools](#external-tools-for-testing) external to Observability Pipelines, such as MySQL Workbench or third-party tools.
```
janine-c:
```suggestion
Before you configure the MySQL source, complete the following prerequisites to ensure that Datadog can validate credentials, connectivity, and queries before using them in Observability Pipelines. Use a [tool](#external-tools-for-testing) external to Observability Pipelines, such as MySQL Workbench or third-party tools, to complete these steps.
```
Minor rephrasing to get rid of some passive voice and simplify the phrasing a bit 🙂

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+
+1. Create a database role for log collection, if you don't already have one. The role must:
```
janine-c: The top-level items in this list tell you what you need to do, but not how to do them, which makes me worry about a user being able to follow them. Can we link out to other docs, maybe, to help users feel a little less lost? I know it might be intentionally vague because it needs to happen on another tool, but maybe some idea of what users should see (for example, how do they know they've validated the connection string? How do they know they're using the correct host, port, etc.?) could be very helpful.

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+    - Only have read-only access, so no permissions to modify or write data.
```
janine-c:
```suggestion
    - Have read-only access, with no permissions to modify or write data.
```
I thought "only" was confusing, since the next bullet was about having more permissions 🙂

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+    - Have permission to execute the target queries used for log collection.
+    - Be scoped to only the databases, schemas, and tables required for your log collection.
+1. Validate the connection string. The connection string must:
+    - Be able to authenticate using the role from the previous step.
+    - Be able to successfully connect to the database from the environment in which the Observability Pipelines Worker runs.
+    - Use the correct host, port, database name, and authentication mechanism.
+    - Be tested prior to configuring it in Observability Pipelines, to avoid runtime failures.
+1. Write, validate, and test SQL queries using either third-party or native database management tools.
+    - All SQL queries must be validated and tested with a tool external to Observability Pipelines and prior to configuring it in Observability Pipelines.
```
janine-c:
```suggestion
    - Validate all SQL queries with a tool external to Observability Pipelines and prior to configuring it in Observability Pipelines.
```
I think this is stronger when we use the imperative instead of passive voice - "you," the user, have to do this.

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+    - Ensure that each query executes successfully using the read-only rule and returns the expected schema.
```
janine-c:
```suggestion
    - Ensure that each query executes successfully using the read-only role and returns the expected schema.
```
I'm not sure what "the expected schema" is?

content/en/observability_pipelines/sources/mysql.md
janine-c: It's one of several valid ways to frame it! I would either leave it as a recommendation, or if you don't want to commit to a full endorsement, you could frame it as a list of examples of third-party tools.

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+- Third-party tools:
+  - DBeaver
+  - DataGrip
+  - TablePlus
+  - DbVisualizer
+- Native MySQL tool: MySQL Workbench
+
+## Set up the source in the pipeline UI
+
+Set up the MySQL source and its environment variables when you [set up a pipeline][1]. The information below is configured in the pipelines UI.
```
janine-c:
```suggestion
When you're [setting up a pipeline][1], set up the MySQL source and its environment variables:
```
I think this gets across the same information? It took me an extra second to understand the UI part; I think most users are more focused on their objective than the UI, so I think it might be less confusing to leave it out.

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+Datadog recommends these tools for validating and testing queries:
+
+- Third-party tools:
+  - DBeaver
+  - DataGrip
+  - TablePlus
+  - DbVisualizer
+- Native MySQL tool: MySQL Workbench
+
+## Set up the source in the pipeline UI
```
janine-c:
```suggestion
## Set up the source while setting up a pipeline
```

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+Set up the MySQL source and its environment variables when you [set up a pipeline][1]. The information below is configured in the pipelines UI.
+
+1. Ensure you have completed the [prerequisite steps](#prerequisites).
+1. Configure the connection between the Observability Pipelines Worker and the database.
+    1. Enter the connection string.
+        - Connection string example: `mysql://${DB_USERNAME}:${DB_PASSWORD}@${DB_HOST}:${DB_PORT}/${DB_NAME}`
+        - **Note**: The connection string must be fully validated using a tool external to Observability Pipelines. See [Prerequisites](#prerequisites) for more information.
+1. Set the SQL query parameters.
+  1. Enter the name of your query.
+  1. Enter the path to local file containing the validated SQL query.
```
janine-c:
```suggestion
  1. Enter the path to the local file containing the validated SQL query.
```

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Set up the source in the pipeline UI
+
+Set up the MySQL source and its environment variables when you [set up a pipeline][1]. The information below is configured in the pipelines UI.
+
+1. Ensure you have completed the [prerequisite steps](#prerequisites).
+1. Configure the connection between the Observability Pipelines Worker and the database.
+    1. Enter the connection string.
+        - Connection string example: `mysql://${DB_USERNAME}:${DB_PASSWORD}@${DB_HOST}:${DB_PORT}/${DB_NAME}`
+        - **Note**: The connection string must be fully validated using a tool external to Observability Pipelines. See [Prerequisites](#prerequisites) for more information.
```
janine-c: I think we can maybe condense this a bit? If we have a section called "prerequisites," I think it's okay to assume that a user is responsible for their own issues if they skip that section 🙂 So I don't think we need to point back to it more than maybe once or twice for really big gotchas. I would also try to reduce the number of list items and levels; keeping the structure as simple as possible, without splitting information unnecessarily, helps the user stay on track.

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+1. Set the SQL query parameters.
+  1. Enter the name of your query.
+  1. Enter the path to local file containing the validated SQL query.
+      - The SQL query that the Worker executes must be stored in its own local file.
```
janine-c: Saving the query in its own file seems like a piece of information that belongs in the prerequisites to me.

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+      - See [Incremental query syntax](#incremental-query-syntax) for details about formatting and [Queries](#queries) for additional information.
```
janine-c: This also seems like it's oddly placed. If they've already validated the SQL query, they would have already used the syntax for it, right?

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+      - **Note**: All file paths are made relative to the configuration data directory, which is `/var/lib/observability-pipelines-worker/config/` by default.
+        - For example, if the SQL file path is `/DD_OP_DATA_DIR/config/db_queries/retrieve_incremental_with_start.sql`, enter the path `db_queries/retrieve_incremental_with_start.sql`.
+        - The file must be owned by the `observability-pipelines-worker group` and `observability-pipelines-worker` user, or at least readable by the group or user.
+        - See [Advanced Worker Configurations][2] for more information.
```
janine-c: Same note with all of this - if the SQL query information is included with the prerequisites, this seems too late for this level of detail.

content/en/observability_pipelines/sources/mysql.md
janine-c: I think instead, maybe we can set some more context about where in the UI this needs to happen. We've linked them to another procedure, but I'm not sure it's clear how the two procedures fit together, or when users should be looking at either page (we ideally wouldn't have them going back and forth).

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+    |----------------------------|-----------------|
+    | At minute 17 of every hour | `17 */1 * * *`  |
+    | At minute 45 of every hour | `45 * * * *`    |
+    | Every 1 min                | `*/1 * * * *`   |
+    | Once daily at 8 AM         | `0 8 * * *`     |
+
+## Set the environment variables
+
+- URI connection string
+  - THE URI that contains the necessary parameters, such as the database engine, host, port, and credentials, to connect to a database.
```
janine-c:
```suggestion
  - The URI that contains the necessary parameters, such as the database engine, host, port, and credentials, to connect to a database.
```

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+      1. Set the starting checkpoint value. This value tells the Worker to start from the specific row that contains the value. See [Checkpoint values](#checkpoint-values) for more information.
+6. Define your query scheduling using cron syntax. These are cron syntax examples for different query schedules:
+    | Query Schedule             | Cron Expression |
+    |----------------------------|-----------------|
+    | At minute 17 of every hour | `17 */1 * * *`  |
+    | At minute 45 of every hour | `45 * * * *`    |
+    | Every 1 min                | `*/1 * * * *`   |
+    | Once daily at 8 AM         | `0 8 * * *`     |
+
+## Set the environment variables
```
janine-c: I'm having trouble understanding this section. Because it's its own section and starts with a bulleted noun phrase rather than a full sentence, I don't understand where it fits in with other steps, or where the user is supposed to do this, or what it's supposed to do. Can we explain it a bit more?

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Incremental queries
+
+### Incremental query syntax
+
+For incremental querying, SQL queries must have the following to help ensure consistent results.
+
+- A `WHERE <incremental_column> <operation> <placeholder>` clause so that the source on each scheduled run can replace `placeholder` with the checkpoint value, and retrieve the latest results.
+  - `incremental_column`: A column with incremental values. This value must be the same as the column value set in the MySQL source configuration.
+  - `operation`: Enter `>` , `<`, `>=`, or `<=` base on your use case.
```
janine-c:
```suggestion
  - `operation`: Enter `>` , `<`, `>=`, or `<=` based on your use case.
```

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+  - `placeholder`: Enter `?`.
+- An `ORDER BY <incremental_column>` clause so that the database returns the rows in the expected order for the source to retrieve the latest values.
+- This is an example that uses all the options: `SELECT * FROM orders WHERE order_id > ? ORDER BY order_id LIMIT 500;`
```
janine-c:
```suggestion

This is an example that uses all the options: `SELECT * FROM orders WHERE order_id > ? ORDER BY order_id LIMIT 500;`
```
The sentence that introduced the list says that each query needs each of those options, so this doesn't really make sense in the list.

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Incremental query syntax
+
+For incremental querying, SQL queries must have the following to help ensure consistent results.
+
+- A `WHERE <incremental_column> <operation> <placeholder>` clause so that the source on each scheduled run can replace `placeholder` with the checkpoint value, and retrieve the latest results.
+  - `incremental_column`: A column with incremental values. This value must be the same as the column value set in the MySQL source configuration.
+  - `operation`: Enter `>` , `<`, `>=`, or `<=` base on your use case.
+  - `placeholder`: Enter `?`.
+- An `ORDER BY <incremental_column>` clause so that the database returns the rows in the expected order for the source to retrieve the latest values.
```
janine-c:
```suggestion
- An `ORDER BY <incremental_column>` clause so that the database returns the rows in the expected order for the source to retrieve the latest values:
```

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+- `Datetime`
+
+Datadog recommends using an incremental ID for the incremental column. If you use an identifier that might not be unique, such as timestamps, it could result in data loss or duplicated data. For example:
+
+- There could be data loss if you use strict operators, such as `>` or `<`, in your query.
+- There could be duplicated data if you use inclusive operators, such as `>=`.
+
+### Checkpoint values
+
+Checkpoint values are updated every job run. To monitor the checkpoint value, there are Worker logs that contain the message `Checkpoint updated` and the latest value that was published. If a job fails or the Worker is restarted mid-job or it crashes, the checkpoint value reverts to the start value. To determine the checkpoint value for a job that failed:
```
janine-c: "Checkpoint value" is defined earlier on the page, but if someone finds this section through search when they're trying to find a definition, it won't help them. Can we add a definition here too?

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+  - `incremental_column`: A column with incremental values. This value must be the same as the column value set in the MySQL source configuration.
+  - `operation`: Enter `>` , `<`, `>=`, or `<=` base on your use case.
+  - `placeholder`: Enter `?`.
+- An `ORDER BY <incremental_column>` clause so that the database returns the rows in the expected order for the source to retrieve the latest values.
+- This is an example that uses all the options: `SELECT * FROM orders WHERE order_id > ? ORDER BY order_id LIMIT 500;`
+  - If the last checkpoint value is `7`, this query retrieves all rows where the `order_id` column's value is greater than `7`.
+
+### Incremental columns
+
+Incremental columns are used to track the progress of each new query. The following column types for incremental querying are supported:
```
janine-c: Similar to above, can we add a definition here for users who arrive here from search, and to add some more context?

I might also consider putting this information in expanders closer to when it's relevant. It's not exactly in a bad spot here, but it might help it feel a little more relevant, rather than this part of the page feeling a bit more like an appendix.

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+
+- `Varchar`
+- `Int`, `float`, `real`, `bigint`, `number`
+- `Timestamp`
+- `Datetime`
+
+Datadog recommends using an incremental ID for the incremental column. If you use an identifier that might not be unique, such as timestamps, it could result in data loss or duplicated data. For example:
```
janine-c: I don't really understand how this would work, since the list of supported columns doesn't include IDs. Are you supposed to have an ID column in addition to the supported identifier column?

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+1. Check the log in your destination to which your logs were sent and determine the last value sent.
+1. Manually reset the checkpoint value in the Database source in the pipelines UI.
+
+## Limits and requirements
+
+### Worker and pipelines
+
+#### One database per pipeline
+
+- You can only have one Database source per pipeline.
```
janine-c:
```suggestion
- You can have only one database source per pipeline.
```

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+- Multiple instances of the same database type requires separate pipelines.
```
janine-c:
```suggestion
- Multiple instances of the same database type require separate pipelines.
```
"Require" here has to agree with the subject "instances," rather than "database type" 🙂

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+- One pipeline can run only one SQL query.
+
+#### Single-node execution requirements
+
+- Observability Pipelines Workers are deployed in a share-nothing architecture.
+- Datadog recommends deploying database pipelines using the Database source on one Worker node only. Otherwise, you run the risk of pulling duplicate data across multiple workers.
```
janine-c:
```suggestion
- Datadog recommends deploying database pipelines using the MySQL source on one Worker node only. Otherwise, you run the risk of pulling duplicate data across multiple Workers.
```
Not sure exactly what's supposed to be capitalized here, but I would try to stay consistent!

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Queries
```
janine-c: I don't really understand why this information would be separate from the other query-writing instructions, or why this is at the end. Can we put all of the query-related stuff in one place? I'm concerned that having the information broken up too much might cause users to miss things and get frustrated if they don't work.

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+
+#### Only read-only queries are executed
+
+- Only `SELECT` and `SELECT DISTINCT` statements are supported.
+- Queries that modify data (`INSERT`, `UPDATE`, `DELETE`, `DDL`) are all rejected.
+
+#### Validate and storing queries
```
janine-c:
```suggestion
#### Validate and store queries
```

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+
+- SQL queries must be tested and validated outside of Observability Pipelines' UI.See [External tools for validating queries](#external-tools-for-validating-queries).
```
janine-c:
```suggestion
- SQL queries must be tested and validated outside of Observability Pipelines' UI. See [External tools for validating queries](#external-tools-for-validating-queries).
```

content/en/observability_pipelines/sources/mysql.md
janine-c: This section seems to repeat instructions from earlier. I don't think this has to be stated again - I think maybe you can incorporate this into the prerequisites section, since these are things a user probably wants to know ahead of time.

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,174 @@
...
+- To execute a specific SQL query with Observability Pipelines, you must store the query in a local file for the Worker to read.
+
+#### SQL file requirements
+
+- The local file that contains the query for the Worker to run can only have one query.
```
janine-c:
```suggestion
- The local file that contains the query for the Worker to run can contain only one query.
```

content/en/observability_pipelines/sources/mysql.md
maycmlee: Ah should be `MySQL` source. `Workers` needs to be capitalized, like Agent.

content/en/observability_pipelines/sources/mysql.md
maycmlee: Link to MySQL create role docs: https://dev.mysql.com/doc/refman/9.2/en/create-role.html

content/en/observability_pipelines/sources/mysql.md
maycmlee: Makes sense, moving it up to prereqs, if it's not already there and removing from this section.

content/en/observability_pipelines/sources/mysql.md
maycmlee: Good point! Moving this info to prereqs.

content/en/observability_pipelines/sources/mysql.md
maycmlee: Yes, also good point. I need to consolidate this info. Might need to do it as a second iteration.

content/en/observability_pipelines/sources/mysql.md
maycmlee: Moved up

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,182 @@
...
+Before you configure the MySQL source, complete the following prerequisites to ensure that Observability Pipelines can validate credentials, connectivity, and queries before they are used in Observability Pipelines. Use a [tool](#external-tools-for-validating-queries) external to Observability Pipelines, such as MySQL Workbench or third-party tools, to complete these steps.
+
+1. [Create a database role](#1-create-a-database-role).
+1. [Validate the connection string](#2-validate-the-connection-string).
+1. [Write, validate, and test SQL queries](#3-write-validate-and-test-sql-queries).
+1. [Store the SQL query in a local file](#4-store-the-sql-query-in-a-local-file).
+
+### 1. Create a database role
+
+Create a database [role][4] for log collection, if you don't already have one. The role must:
```
maycmlee, line 45: Is it correct to link to this page for creating a role?
 https://dev.mysql.com/doc/refman/9.2/en/create-role.html

content/en/observability_pipelines/sources/mysql.md
```diff
@@ -0,0 +1,182 @@
...
+
+- Be able to authenticate using the role from the previous step.
+- Be able to successfully connect to the database from the environment in which the Observability Pipelines Worker runs.
+- Be in this format: `mysql://${DB_USERNAME}:${DB_PASSWORD}@${DB_HOST}:${DB_PORT}/${DB_NAME}`.
+- Use the correct host, port, database name, and authentication mechanism.
+- Be tested prior to configuring it in Observability Pipelines, to avoid runtime failures.
+
+### 3. Write, validate, and test SQL queries
+
+Write and test your SQL queries. Validate all SQL queries with a [tool](#external-tools-for-validating-queries) external to Observability Pipelines and prior to configuring it in Observability Pipelines. [Store the SQL query in a local file](#store-the-sql-query-in-a-local-file) for the Worker to execute.
```
tessneau, line 62: not sure why the "Store the SQL query in a local file" doesn't seem to work for me in the preview

content/en/observability_pipelines/sources/mysql.md
tessneau: good catch, not always 

the value in the worker logs is the last value the Worker _received_ from querying the Database but not necessarily the last value sent to the destination, for example if there's issues sending to the destination

tessneau: So I'd suggest leaning on the value in the destination if they want to ensure every record is sent at least once. If some record loss is acceptable for their use case, then the value in the OPW logs alone could suffice.

maycmlee: Thanks @tessneau! I updated the section with this info.

PR #34140, opened 2026-01-26
New guide for troubleshooting no data in monitors

estherk15:

- Convert internal troubleshooting guide to a public guide
-  [DOCS-11988](https://datadoghq.atlassian.net/browse/DOCS-11988)
### Merge instructions

Merge readiness:
- [ ] Ready for merge

[DOCS-11988]: https://datadoghq.atlassian.net/browse/DOCS-11988?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+
+When a monitor displays `No Data` or fails to evaluate as expected, there are several configuration settings and data characteristics to check. This guide helps you diagnose and resolve `No Data` issues in monitor evaluations, specifically:
+- [Troubleshooting the causes of No Data](#additional-causes-for-no-data).
+- [Troubleshooting why monitors are *not* alerting on No Data](#monitor-does-not-alert-on-no-data).
+
+Begin with [Key configuration settings](#key-configuration-settings) to check the three most common issues. To enable notifications, see [Enabling No Data notifications](#enabling-no-data-notifications).
+
+## Key configuration settings
+
+When a monitor shows `No Data` but you expect it to have data, check these three settings first:
```
janine-c, line 26: Can we give the user a link and maybe some additional context so they know where to find each of these settings? I did some digging in the UI and found it a little hard to guess! I wasn't sure if these settings applied to the explorer or to the monitors themselves.

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+- Metric monitors use the evaluation window (<code>last_1h</code>).<br>
+- Service checks use the default context expiration (24 hours).
+</div>
+
+
+## Further reading
+
+{{< partial name="whats-next/whats-next.html" >}}
+
+[1]: /monitors/guide/new-group-delay/
```
janine-c:
```suggestion
[1]: /monitors/configuration/#new-group-delay
```
This link 404ed when I clicked it. I think this is the right one?

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+  - **Do not require**: Monitor evaluates on partial data (recommended for sparse metrics).
+  - **Require**: Monitor waits for a full window before evaluating.
+
+## Additional causes for No Data
+
+{{% collapse-content title="No data actually exists" level="h3" %}}
+
+The monitor shows `No Data`, and when you graph the metric, no data appears.
+
+#### Methods to verify
```
janine-c:
```suggestion
#### Verify whether data exists
```
This doesn't have to be the exact wording! I just thought that this is really a single method, so the plural didn't make a ton of sense.

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+1. Identify the metric (such as AWS metric or custom metric).
+2. Graph it in a notebook or dashboard.
+3. Scope to the correct timeframe.
+4. Check if data points appear.
+
+#### Understanding (no groups found) vs. No Data
+While they look similar, `(no groups found)` is specific to multi alert monitors (grouped by a tag). It indicates that the monitor has no active entities to evaluate, whereas `No Data` refers to the metric itself being empty and is controlled by your monitor's [On Missing Data](#enabling-no-data-notifications) settings.
```
janine-c: The link confused me a little - when I clicked the link, nothing in that section said "On Missing Data," and then when I looked in the UI, I didn't see anything that matched. Can we tell users the exact setting to look for, and use standardized language for it, so we don't confuse them?

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+
+`(no groups found)` typically occurs if:
```
janine-c:
```suggestion
`(no groups found)` typically occurs in the following cases:
```
"If" makes it sound like there's a verb phrase coming, so the noun phrase sounds weird ("if your setup is new" sounds fine, but "if new setup" sounds weird, for example). This small wording edit helps it flow a little more nicely!

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+## Additional causes for No Data
+
+{{% collapse-content title="No data actually exists" level="h3" %}}
+
+The monitor shows `No Data`, and when you graph the metric, no data appears.
+
+#### Methods to verify
+1. Identify the metric (such as AWS metric or custom metric).
+2. Graph it in a notebook or dashboard.
+3. Scope to the correct timeframe.
```
janine-c: What's "the correct timeframe"? Is it the same timeframe as the monitor, so they can double check the data? Is it something else?

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+4. Check if data points appear.
+
+#### Understanding (no groups found) vs. No Data
+While they look similar, `(no groups found)` is specific to multi alert monitors (grouped by a tag). It indicates that the monitor has no active entities to evaluate, whereas `No Data` refers to the metric itself being empty and is controlled by your monitor's [On Missing Data](#enabling-no-data-notifications) settings.
+
+`(no groups found)` typically occurs if:
+
+- New Setup: The monitor has received no data for any groups since creation.
+- Aged Out: Previously active groups stopped reporting and have been cleared from the monitor's memory.
```
janine-c:
```suggestion
- Aged Out: Previously active groups stopped reporting and have been cleared from the monitor's memory. Different types of groups have different retention periods:
```
Not sure if this is the right terminology, but I think a helper sentence here could help smooth the transition into the bullets by providing context. These things are small, but I think they make an impact on the amount of mental load it takes for a user to keep up 🙂

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+  - **Host/Service Checks**: Groups age out after 48 hours.
+  - **Custom**: This can be extended up to 72 hours through the **On Missing Data** setting.
+
+{{% /collapse-content %}}
+
+{{% collapse-content title="Data is sparse or delayed" level="h3" %}}
+
+Data exists but doesn't appear consistently or arrives late.
+
+#### Methods to verify
```
janine-c: I'd like to make a suggestion like above to clarify what's being verified here...but I'm not really sure! Is this verifying that the sparse data is consistently sparse? In the last step, when we tell them to observe the data points, what are they actually looking for?

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+1. Graph the metric over 2-3 times the evaluation timeframe.
+   - Example: For a 5-minute evaluation window, graph the last 15 minutes.
+2. Observe the frequency and consistency of data points.
+
+#### Possible solutions
+- Enable **"Do not require a full window of data for evaluation"** in [Metric monitor Advanced Options][6].
```
janine-c:
```suggestion
- In [Metric monitor Advanced Options][6], enable the **"Do not require a full window of data for evaluation"** setting.
```
I like to name things in the order that the user will encounter them in, so we can save them a little mental bouncing around 🙂

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+- Increase the evaluation window to capture more data points.
+- Add an evaluation delay to account for data ingestion delays.
+- For aggregations like `avg`, `min`, and `max`, consider a larger timeframe.
+
+For more information, see [Adjusting No Data alerts for metric monitors][2].
```
janine-c:
```suggestion
To learn how to apply these changes, see [Adjusting No Data alerts for metric monitors][2].
```
I think this is right? I thought it might be worth making it explicit, in case someone sees something like "add an evaluation delay" and panics because they don't immediately know how.

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+
+{{% /collapse-content %}}
+
+{{% collapse-content title="Query contains a rollup" level="h3" %}}
+
+The monitor shows `No Data` even though data exists.
+
+#### Why this happens
+Rollups aggregate metric data into buckets. If the monitor doesn't have at least one complete bucket, it reports `No Data`. Datadog does not recommend defining rollups in monitor queries. Rollups make sense if you are monitoring one aggregated value over a long period or monitoring sparse metrics (typically over 24 hours).
```
janine-c: I'm not sure I understand how the last two sentences here relate to each other. Is the second sentence not something you would do with a monitor query? If not, how would you do it?

Maybe what's causing the confusion here for me is that they're not exactly parallel: "defining rollups in monitor queries" is a method of creating a query, but "if you are monitoring one aggregated value..." is a use case. If they were both methods or both use cases, I think this might be easier to follow, because they'd be easier to compare.

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+{{% /collapse-content %}}
+
+{{% collapse-content title="Tags with N/A values" level="h3" %}}
+
+The monitor shows `No Data`, but data exists in dashboards.
+
+#### How to troubleshoot
+1. Create a notebook or dashboard with your monitor query.
+2. Look for groups with N/A tag values.
+3. Adjust your grouping or scoping to exclude N/A tags.
```
janine-c: Can we clarify what we're troubleshooting or what the user should see when they try this? I think I'd use the "verification" language used above, rather than calling this "troubleshooting," because "troubleshooting" implies a solution to me. But I'm not 100% sure on that? The last step here is different from the solution below in this section, so I'm not sure if adjusting the grouping is supposed to fix the problem (is that just for verifying?) or if the real solution is to actually change the tags.

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+Require a full window of data
+: Found under **Advanced Options**, this setting determines whether the monitor requires a complete evaluation window of data before alerting.
+  - **Do not require**: Monitor evaluates on partial data (recommended for sparse metrics).
+  - **Require**: Monitor waits for a full window before evaluating.
+
+## Additional causes for No Data
+
+{{% collapse-content title="No data actually exists" level="h3" %}}
+
+The monitor shows `No Data`, and when you graph the metric, no data appears.
```
janine-c: I found the transitions between the drawer names and the first sentences a little jarring. They're describing pretty different things - the root cause, and the manifestation of what the user actually sees in the UI. I think we have a couple options to make this a little smoother:
- Swap them around and make the drawer names what the users sees, then put the cause inside the drawer. This would help us meet the user where they're at a little more effectively - they could be looking at a specific problem in front of them and we could connect our instructions with what they're seeing
- Add a little `#### Root cause` heading to try to smooth the transition between effect and cause; otherwise, the user has to fill in the gap themselves and that can make the reading process a little bumpier

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+Rollups aggregate metric data into buckets. If the monitor doesn't have at least one complete bucket, it reports `No Data`. Datadog does not recommend defining rollups in monitor queries. Rollups make sense if you are monitoring one aggregated value over a long period or monitoring sparse metrics (typically over 24 hours).
+
+#### Solution
+If you use rollups in your monitor query, add an evaluation delay at least equal to the rollup period. This ensures the monitor has at least one bucket of data to evaluate. For example, a 1-hour rollup requires a 3600-second evaluation delay.
+
+<div class="alert alert-tip">The <code>cumsum</code> function is a visual function designed for dashboards and notebooks. It doesn't work reliably in monitors because monitors lack context about which timeframe to use. Instead, use <a href="/monitors/configuration/?tab=thresholdalert#cumulative-time-windows">cumulative time windows</a> to monitor cumulative metrics.</div>
+
+{{% /collapse-content %}}
+
+{{% collapse-content title="Tags with N/A values" level="h3" %}}
```
janine-c:
```suggestion
{{% collapse-content title="Tags contain N/A values" level="h3" %}}
```
The other drawer names were verb phrases, but this was a noun phrase, so it felt a little out of place.

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+
+The monitor shows `No Data`, but data exists in dashboards.
+
+#### How to troubleshoot
+1. Create a notebook or dashboard with your monitor query.
+2. Look for groups with N/A tag values.
+3. Adjust your grouping or scoping to exclude N/A tags.
+
+#### Why this happens
+Monitors use AND logic for tags. If you group by multiple tags, the metric must have all those tags. Datadog automatically applies N/A values in dashboards when tags are missing, but in monitors, this results in a `No Data` status. For example, a query scoped by `env:production` won't match metrics where the `env` tag is N/A.
```
janine-c: This confused me a little - I think because I wasn't sure where "if you group by multiple tags" happens? In the monitor query? And then I wasn't sure where the "when tags are missing" part came from - like, which tags, the ones the user was grouping by?

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+
+#### Solution
+Ensure that all tags specified in your monitor are present on your metric data. If any tag is missing (shows as N/A), update your metric submission to include all required tags, or adjust your monitor query to group or scope only by tags that are always available.
```
janine-c: I'm still having trouble following this. Is the root cause grouping by multiple tags? Or is it missing tags...?

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+
+{{% /collapse-content %}}
+
+{{% collapse-content title="Insufficient historical data (Anomaly and Forecast monitors)" level="h3" %}}
+
+Anomaly or Forecast monitor shows `No Data`.
+
+#### Why this happens
+`No Data` in Anomaly or Forecast monitors is typically caused by either insufficient historical data or by the presence of rollups in your monitor query.
```
janine-c: This felt a little odd to me because there's a separate rollup section already. Does it apply differently for Anomaly or Forecast monitors?

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+
+| Cause                        | Details                                                                                                                                                                                                                    |
+|------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **Insufficient historical data** | - Anomaly monitors require at least 3 seasons of historical data to establish a baseline.<br>- Forecast monitors need varying amounts of historical data depending on the length of the forecast period.                |
```
janine-c: Is "season" a term users are likely to understand already? I hadn't seen it in other places timeframes came up, so just thought I'd double-check.

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+| **Rollups in the monitor query** | Adding a rollup in the monitor query overrides Datadog's automatic rollup, which is based on your alert timeframe. Datadog automatically selects rollups to provide 4-6 data points per evaluation. Manual rollups may result in too few points, increasing the risk of `No Data` or inaccurate evaluations.<br><br> Adjusting anomaly rollups can be used to address:<br>- **Sparse metrics**: Increase the rollup interval to capture more points<br>- **Noisy metrics**: Increase the rollup to smooth data by averaging points together|
```
janine-c:
```suggestion
| **Rollups in the monitor query** | Adding a rollup in the monitor query overrides Datadog's automatic rollup, which is based on your alert timeframe. Datadog automatically selects rollups to provide 4-6 data points per evaluation. Manual rollups may result in too few points, increasing the risk of `No Data` or inaccurate evaluations.<br><br> Address anomaly rollups based on the problem you're trying to solve:<br>- **Sparse metrics**: Increase the rollup interval to capture more points<br>- **Noisy metrics**: Increase the rollup to smooth data by averaging points together|
```
This is a little confusing; in the other drawer, it sounded like rollups are bad. But this one makes it sound like they're sometimes okay?

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+
+#### Solution
+Review both your data history and whether any manual rollups are present in your monitor query when troubleshooting `No Data` for these monitor types.
```
janine-c: This seems like a good guideline, but it feels vague for a solution. What is the user reviewing for? Maybe it would make sense to reshuffle the headings a bit if the "why this happens" section contains the information on how to fix the problem?

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+
+{{% /collapse-content %}}
+
+{{% collapse-content title="Arithmetic with NaN values" level="h3" %}}
+
+A monitor with arithmetic (`a / b`) shows `No Data`. **Note**: This behavior is specific to monitor evaluations. Graphs may display 0 values in the same scenarios.
```
janine-c:
```suggestion
A monitor evaluation with arithmetic (`a / b`) shows `No Data`, or a graph displays 0 values.
```
I think this might be easier to read?

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+
+#### Why this happens
+If any part of an arithmetic expression is NaN, the entire expression becomes NaN and won't evaluate. For example:
```
janine-c:
```suggestion
If any part of an arithmetic expression is NaN (Not a Number, which can happen in cases like trying to divide by zero), the entire expression becomes NaN and won't evaluate. For example:
```
I wasn't sure what NaN was, other than really making me want naan bread 😂 I wonder if a quick explainer like this might be useful?

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+- Expression: `a + b`.
+- If `a` is NaN, then `a + b` is NaN.
+- The monitor won't evaluate.
+
+<div class="alert alert-info">The exception is that queries using `as_count()` return 0 for missing values instead of NaN. For more information, see the <a href="/monitors/guide/as-count-in-monitor-evaluations/">as_count() documentation</a>.</div>
```
janine-c:
```suggestion
<div class="alert alert-info">The exception is that queries using <code>as_count()</code> return 0 for missing values instead of NaN. For more information, see the <a href="/monitors/guide/as-count-in-monitor-evaluations/">as_count() documentation</a>.</div>
```

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+
+#### How to troubleshoot
+1. Create a notebook or dashboard.
+2. Graph each part of the arithmetic expression separately.
+3. Identify which query returns NaN.
+4. If one query meets `No Data` criteria, the monitor behaved as expected.
```
janine-c: Might this be "at least one query"? This would happen if it were `a + b + c` and both `a` and `b` were NaN, right? Unless I'm misunderstanding?

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+
+{{% /collapse-content %}}
+
+
+## Monitor does not alert on No Data
+
+You expect a `No Data` alert but don't receive one. Make sure you have the required settings:
```
janine-c:
```suggestion
If you expect a `No Data` alert but don't receive one, make sure you have the required settings:
```

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+#### How to troubleshoot
+1. Create a notebook or dashboard.
+2. Graph each part of the arithmetic expression separately.
+3. Identify which query returns NaN.
+4. If one query meets `No Data` criteria, the monitor behaved as expected.
+
+{{% /collapse-content %}}
+
+
+## Monitor does not alert on No Data
```
janine-c:
```suggestion
## Troubleshoot No Data alerts
```
Might be nice to have a verb here that relates to the user's objective?

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+2. **Notification message** must:
+   - Apply to all state changes (without conditional blocks), or
+   - Include the `{{#is_no_data}}` template variable.
+
+### Common causes for missing alerts
+If your monitor has the correct settings but still isn't alerting as expected, check if the following causes could apply to your configuration:
+
+- `No Data` timeframe is shorter than the evaluation window.
+- Notification message uses conditional blocks that exclude `No Data`.
+- Monitor has `notify_no_data` disabled.
```
janine-c: After reading other stuff on this page about troubleshooting, I got a little confused about whether these are causes or solutions (i.e., things that should not or should be true of the user's configuration). Can we frame it explicitly in terms of solutions?

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Enabling No Data notifications
+
+1. **Notify on No Data** setting must be enabled.
+2. **No Data timeframe** should be at least 2x the evaluation window.
+3. **Notification message** must either:
+   - Apply to all state changes (no conditional blocks), or
+   - Use the `{{#is_no_data}}` template variable.
```
janine-c:
```suggestion
## Enable No Data notifications

1. Enable the **Notify on No Data** setting.
2. Set the **No Data timeframe** to at least 2x the evaluation window.
3. Set the **Notification message** to either:
   - Apply to all state changes (no conditional blocks), or
   - Use the `{{#is_no_data}}` template variable.
```
Small rephrasings to make this less of a checking procedure and more of a regular procedure 🙂

content/en/monitors/guide/troubleshooting-no-data.md
```diff
@@ -0,0 +1,193 @@
...
+
+<div class="alert alert-info">
+If <code>notify_no_data</code> is <code>true</code> but <code>no_data_timeframe</code> is missing:<br>
```
janine-c: Might be worth giving a little more context on where to find this?

content/en/monitors/guide/troubleshooting-no-data.md
estherk15: Yes, this needed to be clarified!

```suggestion
3. Scope the graph to the same timeframe as the monitor.
```

content/en/monitors/guide/troubleshooting-no-data.md
estherk15: Ended up removing this information, you're right, it's redundant!

content/en/monitors/guide/troubleshooting-no-data.md
estherk15: Yes! Added a screenshot and clarified where they can find these options.

PR #34263, opened 2026-02-02
update doc to recommend dd-trace-api usage

rochdev:

Update doc to recommend `dd-trace-api` usage. We released the public API package a while ago at this point, but have never documented to use it where it matters.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

**For Datadog employees**:

### DIFF
--- a/content/en/tracing/trace_collection/dd_libraries/nodejs.md
+++ b/content/en/tracing/trace_collection/dd_libraries/nodejs.md
@@ -88,11 +88,13 @@ Use the `--require` option to Node.js to load and initialize the tracer in one s
 node --require dd-trace/init app.js
 ```
 
-This approach requires using environment variables for all configuration of the tracer. If you need to use a programmatic configuration, initialize `dd-trace` in a dedicated file and require that instead:
+The above approach requires using environment variables for all configuration of the tracer. If you need to use a programmatic configuration, initialize `dd-trace` in a dedicated file and require that instead:
 
 ```sh
-node --require ./dd-trace app.js
+node --require ./dd-trace.js app.js
 ```
+
+The file should contain this:
 
 ```js
 // ./dd-trace.js


content/en/tracing/trace_collection/dd_libraries/nodejs.md
```diff
@@ -49,6 +49,21 @@ To install the Datadog tracing library (version 4.x of `dd-trace`) for end-of-li
 For more information on Datadog's distribution tags and Node.js runtime version support, see the [Compatibility Requirements][1] page.
 If you are upgrading from a previous major version of the library (0.x, 1.x, 2.x, 3.x or 4.x) to another major version, read the [Migration Guide][5] to assess any breaking changes.
 
+**Note**: In Serverless environments or when using Single-Step Instrumentation, the library is already pre-installed and shouldn't be added as a dependency. Instead it should be added as a dev dependency to get tracing locally.
+  ```shell
+  npm install dd-trace -D # instead of `npm install dd-trace`
+  ```
```
janine-c:
```suggestion
<div class="alert alert-info">In Serverless environments or when using Single-Step Instrumentation, the library is already pre-installed, so you don't need to add it as a dependency. Instead, add it as a dev dependency to get tracing locally:
  <div class="highlight code-snippet js-appended-copy-btn">
    <pre tabindex="0" class="chroma">
      <code class="language-shell" data-lang="shell"><span class="line"><span class="cl">npm install dd-trace -D <span class="c1"># instead of `npm install dd-trace`</span></span></span></code>
    </pre>
    <div class="code-button-wrapper position-absolute">
      <button class="btn text-primary js-copy-button">Copy</button>
    </div>
  </div>
</div>
```
This looks like a pile of spaghetti, but it looks good in my local build 😂 It renders the note more nicely and puts the code sample in the same div, so it's easier to spot them as one cohesive piece of information.

content/en/tracing/trace_collection/dd_libraries/nodejs.md
```diff
@@ -49,6 +49,21 @@ To install the Datadog tracing library (version 4.x of `dd-trace`) for end-of-li
...
+
+### Install the Datadog public API (optional)
+
+This step is only required when doing custom instrumentation in Serverless or with Single-Step Instrumentation. For other custom instrumentation use cases it's optional, but even in those cases there are still benefits as described in [Custom Instrumentation using the Datadog API][14].
```
janine-c:
```suggestion
This step is only required when doing custom instrumentation in Serverless or with Single-Step Instrumentation. For other custom instrumentation use cases, it's optional. For information on when to use the Datadog public API, see [Custom Instrumentation using the Datadog API][14].
```

content/en/tracing/trace_collection/dd_libraries/nodejs.md
```diff
@@ -273,3 +260,4 @@ Read [tracer settings][3] for a list of initialization options.
 [6]: /tracing/trace_collection/compatibility/nodejs/#complex-framework-usage
 [11]: /tracing/trace_collection/library_injection_local/
 [13]: /tracing/trace_collection/automatic_instrumentation/?tab=datadoglibraries#install-and-configure-the-agent
+[14]: /tracing/trace_collection/custom_instrumentation/nodejs/dd-api/
```
janine-c:
```suggestion
[14]: /tracing/trace_collection/custom_instrumentation/server-side/?api_type=dd_api&prog_lang=node_js
```
This page has some custom filters on it that we can deeplink to, which is a good idea because users might have made a different selection on another page and may default to other filter options. Does this look good to you? https://docs.datadoghq.com/tracing/trace_collection/custom_instrumentation/server-side/?api_type=dd_api&prog_lang=node_js

content/en/tracing/trace_collection/dd_libraries/nodejs.md
```diff
@@ -59,50 +74,32 @@ After you have completed setup, if you are not receiving complete traces, includ
...
+```sh
+node --require dd-trace/init app.js
 ```
 
-**Note**: `DD_TRACE_ENABLED` is `true` by default, which means some instrumentation occurs at import time, before initialization. To fully disable instrumentation, you can do one of the following:
-- import the module conditionally 
-- set `DD_TRACE_ENABLED=false` (if, for example, static or top-level ESM imports prevent conditional loading)
-
-##### TypeScript and bundlers
+This approach requires using environment variables for all configuration of the tracer. If programmatic configuration is needed, then initialize dd-trace in a dedicated file and require that instead.
```
janine-c:
```suggestion
This approach requires using environment variables for all configuration of the tracer. If you need to use a programmatic configuration, initialize `dd-trace` in a dedicated file and require that instead:
```

content/en/tracing/trace_collection/dd_libraries/nodejs.md
```diff
@@ -59,50 +74,32 @@ After you have completed setup, if you are not receiving complete traces, includ
...
-via environment variables, you can also use `dd-trace/init`, which loads and
-initializes in one step.
-
-```typescript
-import 'dd-trace/init';
+```js
+// ./dd-trace.js
+require('dd-trace').init({
+  // programmatic config
+})
```
janine-c, line 103: I found it confusing to see these two code samples directly beside each other. Can we maybe add a short line of context before the second one so a user can understand more immediately when to use either?

content/en/tracing/trace_collection/dd_libraries/nodejs.md
```diff
@@ -59,50 +74,32 @@ After you have completed setup, if you are not receiving complete traces, includ
...
 ```
 
-#### Option 2: Add the tracer with command line arguments
-
-Use the `--require` option to Node.js to load and initialize the tracer in one step.
+For cases where it's not possible to control the CLI arguments, an environment variable can be used instead: `NODE_OPTIONS='--require dd-trace/init'`.
```
janine-c:
````suggestion
For cases where it's not possible to control the CLI arguments, you can use an environment variable instead:
```
NODE_OPTIONS='--require dd-trace/init'
```
````
Easier to copy and paste as a standalone code block instead of inline 🙂

File: content/en/tracing/trace_collection/dd_libraries/nodejs.md [hunk 7]
```diff
@@ -59,50 +74,32 @@ After you have completed setup, if you are not receiving complete traces, includ
...
 
-```sh
-node --require dd-trace/init app.js
-```
-
-**Note:** This approach requires using environment variables for all configuration of the tracer.
+**Note**: `DD_TRACE_ENABLED` is `true` by default, which means some instrumentation occurs at import time, before initialization. To fully disable instrumentation, you can do one of the following:
+- import the module conditionally
+- set `DD_TRACE_ENABLED=false` (if, for example, static or top-level ESM imports prevent conditional loading)
```
janine-c:
```suggestion
<div class="alert alert-info"><code>DD_TRACE_ENABLED</code> is <code>true</code> by default, which means some instrumentation occurs at import time, before initialization. To fully disable instrumentation, you can do one of the following:
  <ul>
    <li>Import the module conditionally</li>
    <li>Set <code>DD_TRACE_ENABLED=false</code> (if, for example, static or top-level ESM imports prevent conditional loading)</li>
  <ul>
</div>
```
Same content, just formatted as the fancy note div again 🙂

content/en/tracing/trace_collection/dd_libraries/nodejs.md
rochdev: Not sure I understand which 2 samples you mean, can you clarify? Maybe it's just a formatting issue? (which as you could tell I'm not great at 😬)

janine-c: Sorry that wasn't clear! Maybe this screenshot will help. If I see some intro text and then a code sample, I go "oh cool, this is the situation I need to use that code sample in." But then I don't know what to do with the second one, because I'm not already familiar with this functionality 😅 
image

rochdev: They are supposed to be combined, that's why they are next to each other. Basically the first one is the new command that requires a local file instead of the module itself, and the second one is the content of that file. Let me try to maybe split the text in 2 🤔

janine-c: Ah yeah, that makes sense! Even just saying something like "the file should contain this:" would help a user understand why they're together but separate 🙂

rochdev: I pushed an update before seeing your comment. Let me know if you think it's better now, otherwise I'll revert it and add that small text between the 2 snippets instead.

janine-c: Sorry, I just see the one update with the rest of the review updates. I think it would help to have that small text, but happy to leave it to you for the final call!

rochdev: The push failed because of my previous merge. I will just add the small text and we can iterate if needed 🙂

PR #34419, opened 2026-02-06
Documents the Annotation Queue feature in LLMO

gsvigruha:

Documents the new Annotation Queue feature in LLMO.

### Merge instructions

Merge readiness:
- [ ] Ready for merge


cswatt:
Pushed changes: 
- Remove all marketing language
- Removed Key Concepts section — the only concept here that is unique to this context is annotation queues, which are explained in the first sentence on the page; "labels" and "schema" retain the same meaning that they have in general usage
- Moved optional/suggestion/recommendation-type content (When to use annotation queues, Best Practices) to bottom of page; critical information should come first
- Verified UI flow and made significant edits to match steps to what is currently present in the UI
- Removed excessive explanations of what a user is to see in the UI (we can tell them how to use the UI, but we don't need to tell them exactly what they'll see)
- Added screenshots
- Reformatting
- Various grammar/syntax edits

PR #34421, opened 2026-02-06
Add WP for spur threat intelligence

Raganof:

Spur threat intelligence tables are also used in Workload Protection.
Malwares are a primary case we want to cover with Workload Protection, so we want to highlight it in the documentation.

### Merge instructions

Merge readiness:
- [x] Ready for merge


datamarmot:
+1

PR #34432, opened 2026-02-06
[DOCS-11944] Add Node.js examples and non-JSON log correlation for OTel traces

brett0000FF:

Fixes DOCS-11944

Adds Node.js examples and non-JSON log format support to the OpenTelemetry trace-log correlation page:

- **New Node.js tab** in the "Inject trace context into your logs" section, showing how to use `@opentelemetry/instrumentation-winston` for automatic trace context injection.
- **New "Non-JSON logs" sub-section** under "Scrape logs from files", with a Node.js/Winston example for manual context injection into plain text logs and a corresponding `filelog` receiver config using `regex_parser`.

### References

- [OTel spec: Trace Context in Non-OTLP Log Formats](https://opentelemetry.io/docs/specs/otel/compatibility/logging_trace_context/)
- [@opentelemetry/instrumentation-winston](https://www.npmjs.com/package/@opentelemetry/instrumentation-winston)
- [@opentelemetry/instrumentation-pino](https://www.npmjs.com/package/@opentelemetry/instrumentation-pino)
- [Example: OTel Node.js trace context injection](https://github.com/billmeyer/otel-nodejs/blob/main/main.js#L109)

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes

Based on customer feedback requesting guidance for Node.js apps with non-JSON log formats.

### DIFF
--- a/content/en/tracing/other_telemetry/connect_logs_and_traces/opentelemetry.md
+++ b/content/en/tracing/other_telemetry/connect_logs_and_traces/opentelemetry.md
@@ -39,7 +39,7 @@ To correlate OpenTelemetry traces and logs in Datadog, you must:
 
 ### 1. Inject trace context into your logs
 
-The following examples for Go and Java use logging bridges. These bridges intercept logs from common logging libraries (such as `zap` and `Logback`), convert them into the OpenTelemetry log data model, and forward them to the OpenTelemetry SDK. This process automatically enriches the logs with the active trace context.
+The following examples use logging bridges or auto-instrumentation. These tools intercept logs from common logging libraries (such as `zap`, `Logback`, and `Winston`), convert them into the OpenTelemetry log data model, and forward them to the OpenTelemetry SDK. This process automatically enriches the logs with the active trace context.
 
 For complete, working applications, see the [Datadog OpenTelemetry Examples repository][2].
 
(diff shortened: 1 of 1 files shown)


billmeyer:
It's slightly confusing while reading through the doc- of the 3 language choices: Go, Java, Node.js that are presented, when the reader scrolls down to the [JSON logs](https://docs-staging.datadoghq.com/brett.blue/DOCS-11944-nodejs-non-json-trace-correlation/tracing/other_telemetry/connect_logs_and_traces/opentelemetry?tab=nodejs#json-logs) example, the language shifts to Python for that one example and then back to Node.js for the Non-JSON logs. It would be good to stay consistent with the examples throughout.

Also, the Python examples looks suspect to me. When logging as JSON log lines, I would expect to simply use an [add_fields](https://nhairs.github.io/python-json-logger/latest/reference/pythonjsonlogger/json/#pythonjsonlogger.json.JsonFormatter.add_fields) method so that the log line is written as JSON and no additional, manual parsing is needed.

brett0000FF:
@billmeyer - Thanks for the review! Great points about the page overall.
                                                                                              
The Python examples in the JSON logs and OTLP sections are pre-existing content, and this PR only added the Node.js tab at the top and the new Non-JSON logs subsection at the bottom.

The language inconsistency you're seeing was already there before this PR: Go/Java at the top, Python in the middle. My changes made it more noticeable by adding Node.js to the top tabs.

Looking at this more broadly, there's a bigger structural question here: OTel is treated as a "language" tab alongside Go, Java, Node.js, etc. on the parent page. This needs to be totally restructured with language first, then Datadog vs. OTel. That would eliminate the mixed-language examples problem entirely.

But that's a significant restructure, so I'll create another ticket to tackle that. 🫡

OliviaShoup (approved):

PR #34437, opened 2026-02-06
[DOCS-13285] Add Integration column to NDM Supported Devices table

git-thuerk-done:

Fixes DOCS-13285

This PR enhances the NDM Supported Devices page by adding a dedicated "Integration" column to the table. Previously, some vendors had integration links embedded in their vendor names, making it unclear which vendors have additional integration support beyond SNMP profiles.

Changes include:
- Added new "Integration" column to the supported devices table
- Cleaned up vendor names by removing embedded integration links
- Added 17 vendors with dedicated integration links to the new column
- Added explanatory note differentiating between Config files (SNMP profiles) and Integration pages (dedicated integration support with dashboards and enhanced capabilities)

This makes it easier for users to discover which vendors have both SNMP profiles and dedicated integrations available.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes

The Integration column is only populated for vendors that have dedicated integration pages. All vendors listed have SNMP profile support via the Config Files column.

### DIFF
--- a/content/en/network_monitoring/devices/supported_devices.md
+++ b/content/en/network_monitoring/devices/supported_devices.md
@@ -13,12 +13,10 @@ further_reading:
 
 multifiltersearch:
   headers:
-    - name: Vendor
+    - name: Vendor/Integration
       id: vendor
     - name: Config Files
       id: config_files
-    - name: Integration
-      id: integration
   data:
     - vendor: 3com
       config_files: '[3com.yaml](https://github.com/DataDog/integrations-core/blob/master/snmp/datadog_checks/snmp/data/default_profiles/3com.yaml) <br> [3com-huawei.yaml](https://github.com/DataDog/integrations-core/blob/master/snmp/datadog_checks/snmp/data/default_profiles/3com-huawei.yaml)'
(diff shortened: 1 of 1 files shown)


content/en/network_monitoring/devices/supported_devices.md
```diff
@@ -199,6 +218,10 @@ The [generic profile][114] collects metrics for all devices not supported by a v
 The following vendor devices are supported with dedicated profiles.
```
janine-c:
```suggestion
The following vendor devices are supported with dedicated profiles. Some vendors provide dedicated integration pages with additional setup instructions, dashboards, and enhanced monitoring capabilities. All vendors provide SNMP profile configuration files that define which metrics to collect from devices.
```

content/en/network_monitoring/devices/supported_devices.md
```diff
@@ -199,6 +218,10 @@ The [generic profile][114] collects metrics for all devices not supported by a v
...
 If a vendor or device type isn't listed, see the [Troubleshooting](#troubleshooting) section below.
 
+**Note**: The table below shows two types of support for network devices:
+- **Config files**: SNMP profile configuration files that define which metrics to collect from devices. All listed vendors are supported through SNMP.
+- **Integration**: Dedicated integration pages with additional setup instructions, dashboards, and enhanced monitoring capabilities. Some vendors have both SNMP profiles and dedicated integrations available.
```
janine-c:
```suggestion
```
I actually liked the integrations in the first column rather than being in their own, mostly empty column. I might consider calling the first column something like "Name/integration (if applicable)" and seeing how that works?

I also thought that putting this information in the intro paragraph would reduce the amount of mental load that goes into reading this table. It's kind of like the meme about how good docs shouldn't need a "how to read these docs" section - I think that the big benefit of tables is that you can dump a lot of information in one place, and let the spatial relationships do the heavy lifting.

Lastly, I might consider using the products' full names? For example, I noticed that because "Meraki" lacks "Cisco" in the name, it doesn't show up if I search for "Cisco." But the integration page calls it "Cisco Meraki," and if I were only familiar with that, maybe I'd find it confusing.

janine-c (approved):
Looks good! I have some suggestions, but nothing that's strictly necessary :)

content/en/network_monitoring/devices/supported_devices.md
git-thuerk-done: Hi @janine-c thanks for the feedback! The PM wanted to have a separate column to differentiate that those particular integrations had a separate integration support in addition to the SNMP supported yaml profile, but I agree it makes that column pretty empty.
I do like how you suggested to combine the differences in the intro paragraph ! 
Also good tip on changing Meraki to Cisco Meraki !

git-thuerk-done: But on that note I do think I'll fiddle with re-naming the first column, we also don't want customers to think we are lacking integration support for so many vendors :)

PR #34440, opened 2026-02-06
[DOCS-13311] Update US1-FED support portal link

domalessi:
### What does this PR do? What is the motivation?

Fixes DOCS-13311

Updated the support portal link for US1-FED from `http://help.ddog-gov.com/` to `https://govsupport.ddog-gov.com/` in two locations on the [Access Your Support Ticket](https://docs.datadoghq.com/account_management/guide/access-your-support-ticket/) page.

The old link is no longer valid. The new link matches what's documented on the [Datadog GovCloud Support Terms](https://www.datadoghq.com/legal/datadog-govcloud-support-terms/) page.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes

Changes made:
- Line 19: Updated support page link in "Creating a support ticket" section
- Line 43: Updated support page link in "Accessing existing tickets" section

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #34441, opened 2026-02-06
[DOCS-13292] Clarify default HTTP header collection in APM

domalessi:

Fixes DOCS-13292

Updates the APM standard attributes documentation to accurately reflect default HTTP header collection behavior:

- Clarifies that the `Host` header is always collected as `http.request.headers.host` by default
- Clarifies that the `User-Agent` header is collected by default as `http.useragent`

Previously, the documentation incorrectly stated that "None are collected by default" for HTTP request headers. Engineering confirmed that both Host and User-Agent headers are collected by default in tracing libraries (confirmed for .NET, behavior in other libraries unconfirmed but expected to be consistent).

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes

Based on engineering feedback from APMS-18505 escalation. The `User-Agent` header is stored in `http.useragent` tag (not `http.request.headers.*`) for historical reasons.


janine-c (approved):

janine-c:
Seems to include changes that should be in #34440 instead?

domalessi:
> Seems to include changes that should be in #34440 instead?

Yes, you are right. Apologies! I just merged that other PR so then when I update this with master, only the diff should remain.

PR #34449, opened 2026-02-06
[DOCS-13346] Update CCM Tag Pipelines docs for replace tag feature

rtrieu:

Updates the Tag Pipelines documentation to reflect the new dropdown feature that allows users to choose whether to replace, append, or skip applying a tag when it already exists.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes

Includes a warning that the "Replace" option can overwrite existing data.


content/en/cloud_cost_management/allocation/tag_pipelines.md
ashleyyip: this should be updated with the new options, it works exactly the same as the "Add key" section

ashleyyip: same suggestion as above

PR #34506, opened 2026-02-10
(Actions) update Events page in App Builder

OliviaShoup:
This PR updates the Events page in App Builder to clarify the difference between reactions and state functions. 

Resolves [this ticket](https://datadoghq.atlassian.net/browse/DOCS-13374). 

### Merge instructions

Merge readiness:
- [ ] Ready for merge


content/en/actions/app_builder/events.md
aananiadis:
```suggestion
UI components can trigger reactions on an **Event**. Event triggers differ according to the component. For example, a button component can trigger a reaction on a click event, and a table component can trigger a reaction on a page change or table row click event. To see what event triggers are available for a given component, see [Components][1].
```

aananiadis, line 24: This can be separate, but we should take a pass through all the docs to change `query` --> `action`

aananiadis: extra word in 
> a table component event can trigger a reaction

instead
> a table component can trigger a reaction

to mirror the previous clause

aananiadis (approved):
All changes look good, just one comment elsewhere on the page

content/en/actions/app_builder/events.md
OliviaShoup: thank you! i've made a Jira ticket for this change: https://datadoghq.atlassian.net/browse/DOCS-13382

content/en/actions/app_builder/components/_index.md
```diff
@@ -22,7 +22,11 @@ further_reading:
 
 This page provides a list of UI components that you can use when creating apps in App Builder.
 
-Many component properties allow you to select from provided values. If you want to use an expression for a property's value, click **&lt;/&gt;** next to the property to use the code editor. For more information on using JavaScript in App Builder, see [JavaScript Expressions][7]. For more information about saving your components as a template, see [Reusable Modules][12].
+Many component properties allow you to select from provided values. If you want to use an expression for a property's value, click **&lt;/&gt;** next to the property to use the code editor.
```
janine-c, line 25: You don't have to address this now, because I can see this isn't new for this PR. But I'm a little concerned that someone using a screen reader wouldn't know what to look for in the UI if they came across this.

content/en/actions/app_builder/components/_index.md
```diff
@@ -282,9 +272,6 @@ Is Visible
 Event
 : **Value**: change
 
-Reaction
-: **Values**: custom, close modal, download file, open modal, open side panel, close side panel, set component state, set state variable value, toast notification, trigger action
-
 State Functions
 : setValue<br>
 **Example**: See [events][9].
```
janine-c: There are several of these that were left in; is this intentional?

content/en/actions/app_builder/events.md
```diff
@@ -27,14 +27,36 @@ In addition to working with events manually, you can use Bits AI to set up event
...
    1. Enter a custom prompt for an event, or try the prompt `How can you help me with events?`.
 
-### State functions
 
-App Builder provides functions for some types of app state changes. These functions are available in custom event reactions for specific components and in post-query hooks.
+### Events and reactions
+
+In App Builder, events and reactions work together to determine component behavior. 
+
+Events are triggered both by components (such as a button click or a form submit) and actions (execution finished). Events differ depending on which component you use. For component-specific event details, see [Components][1].
```
janine-c:
```suggestion
Components (such as button clicks or form submissions) and actions (such as execution finished) can both trigger events. Events differ depending on which component you use. For component-specific event details, see [Components][1].
```
"Execution finished" is the name of an action? It sounds like a notification that an action is complete? This sounds a little awkward as a result.

content/en/actions/app_builder/components/_index.md
```diff
@@ -65,13 +69,6 @@ Is Visible
 Event
 : **Value**: click
 
-Reaction
```
janine-c, line 68: It's helpful to provide your reviewer with some context about what you're doing here. I'm having trouble understanding why what looks like helpful information is being removed from this page. If you could describe this in your PR description next time (along with a more descriptive title!), it can really help someone understand what you're doing and figure out how they can help you do it.

content/en/actions/app_builder/events.md
```diff
@@ -27,14 +27,36 @@ In addition to working with events manually, you can use Bits AI to set up event
...
+
+Reactions specify what happens when an event is triggered. Any component that can trigger an event has the following available reactions: custom, close modal, download file, open modal, open side panel, close side panel, set component state, set state variable value, toast notification, and trigger action.
```
janine-c: I usually try to stay away from these lists because they're hard to parse, and when there are so many items, it's hard to retain all of them. I'd see if I could maybe format them with bullets.

Is the reason they were taken out of the other file that they're now all the same? I'd be concerned about findability because this information is on a completely different page now.

content/en/actions/app_builder/events.md
```diff
@@ -27,14 +27,36 @@ In addition to working with events manually, you can use Bits AI to set up event
    1. Click the **Build with AI** icon (**<i class="icon-bits-ai"></i>**).
    1. Enter a custom prompt for an event, or try the prompt `How can you help me with events?`.
 
-### State functions
 
-App Builder provides functions for some types of app state changes. These functions are available in custom event reactions for specific components and in post-query hooks.
+### Events and reactions
```
janine-c:
```suggestion
## Events and reactions
```
There was no H2 before this, so we should make sure we don't skip it! That's why it's missing from the heading navigation on the right. And sequential headings are helpful for people using screen readers 🙂

content/en/actions/app_builder/events.md
```diff
@@ -27,14 +27,36 @@ In addition to working with events manually, you can use Bits AI to set up event
...
+
+In App Builder, events and reactions work together to determine component behavior. 
+
+Events are triggered both by components (such as a button click or a form submit) and actions (execution finished). Events differ depending on which component you use. For component-specific event details, see [Components][1].
+
+Reactions specify what happens when an event is triggered. Any component that can trigger an event has the following available reactions: custom, close modal, download file, open modal, open side panel, close side panel, set component state, set state variable value, toast notification, and trigger action.
 
+#### Custom reactions
```
janine-c:
```suggestion
### Custom reactions
```

content/en/actions/app_builder/events.md
```diff
@@ -27,14 +27,36 @@ In addition to working with events manually, you can use Bits AI to set up event
...
+        makeRequest0?.fetch()
+    } else if (teams.some(t => t?.name === "team2")) {
+        makeRequest1?.fetch()
+    } else {
+        makeRequest2?.fetch()
+    }
+}}
+```
+
+### State functions
```
janine-c:
```suggestion
## State functions
```

content/en/actions/app_builder/events.md
```diff
@@ -27,14 +27,36 @@ In addition to working with events manually, you can use Bits AI to set up event
...
 
-fetch
-: Executes the query that you run this function on. This function is available for all components that offer the custom event reaction type.<br>
-**Example**: `query0.fetch()` executes the query named `query0`.
```
janine-c, line 37: ~Was this removed intentionally?~ Saw the linked ticket that said it was, but if you can give that context upfront, it helps your reviewer figure out what they should be seeing!

content/en/actions/app_builder/components/_index.md
OliviaShoup: this is a great point! making note of that for another PR. thank you :)

content/en/actions/app_builder/components/_index.md
OliviaShoup: thanks for asking! yes, the `function` bits that i took out were the only thing wrong with the State Functions sections

content/en/actions/app_builder/components/_index.md
OliviaShoup: thank you for the advice ❤️ i will def be better about that in the future

content/en/actions/app_builder/events.md
OliviaShoup: thank you! yes, i was trying to reduce work for future writers so that they wouldn't have to update this list every time a new reaction is added/removed. but you're right that the list is now separate from the most relevant page. do you think the solution is a shortcode or something like that? and def agree about the bullet points

janine-c: Oh interesting, I don't really understand why moving it to another page would reduce maintenance work? But yeah, reusing content is always a great idea when you want to have the same content in multiple places without increasing maintenance overhead!

OliviaShoup: ah sorry, it's because when it was on the previous page, the list was inside each individual component's text, so the list was in there like 10 times. do you think it's better to put it back on the original page with some kind of short code? i just want the best experience for the user :')

OliviaShoup:
@janine-c thank you for all the helpful comments! i talked to Ursula about the issue of findability with Reactions and we came to a compromise: the Components page now has some example reactions and links to the full list on the Events page. i think that'll be much better for users :)

content/en/actions/app_builder/events.md
```diff
@@ -28,15 +28,25 @@ In addition to working with events manually, you can use Bits AI to set up event
...
+- custom
+- close modal
+- download file
+- open modal
+- open side panel
+- close side panel
+- set component state
+- set state variable value
+- toast notification
+- trigger action
```
janine-c: Can we make sure this list matches the UI?
- The first two are switched around
- The lowercase is weird? We don't really do that unless it matches the UI
- This list seems to be missing "Open URL," which I can see in demo

content/en/actions/app_builder/components/_index.md
```diff
@@ -69,7 +69,9 @@ Is Visible
 Event
 : **Value**: click
 
-For more information on events, see [Events][1].
+Reaction
+: **Values**: examples include open modal, trigger action, and set component state<br>
```
janine-c, line 73: Sorry I missed your last message on the thread we had going about this! If you chatted with Ursula, I don't think we need to belabour this :) My recommendation would usually be a shortcode - it isn't a lot of information, so my instinct is that it isn't worth making a user clicking through to another page to see it. A shortcode helps you satisfy a few tech writing principles:
- Providing information as soon as the user needs it (not making them wonder and hunt, which takes cognitive load)
- Treating "every page as page one" (accounting for the fact that users might not read pages in the order you intend)

It's up to you about whether you want to implement them; if the examples are the most common, for example, or if you and Ursula had a good reason not to use shortcodes, that's fine. I wanted to explain where I was coming from!

content/en/actions/app_builder/events.md
OliviaShoup: thank you for catching these! updated 🫡

content/en/actions/app_builder/components/_index.md
OliviaShoup: it's okay! i know you're busy! thank you for explaining :) i've never made a shortcode before so i think it would take some tooling around on my part. i'm going to experiment with it when i get a chance and open a separate PR!

PR #34653, opened 2026-02-17
(DOCS-13380) Move/rename dev section

hestonhoffman:
This renames the "Developers" section to "Extend Datadog." Things are in a bit of a weird state at the moment because I haven't moved the IDE plugins section. The section still exists in the Extend side nav, but when you click on  one of the pages, you'll notice it still points to `/developers/`. This is because @joepeeples is going to move that section in a separate PR.


dd-dominic:
@hestonhoffman will existing links continue to function? We have integration tiles that link to `https://docs.datadoghq.com/developers/*`

hestonhoffman:
> @hestonhoffman will existing links continue to function? We have integration tiles that link to `https://docs.datadoghq.com/developers/*`

Yes, these should redirect

joepeeples:
Opened [DOCS-13441](https://datadoghq.atlassian.net/browse/DOCS-13441) to follow up with in-depth review. 

[DOCS-13441]: https://datadoghq.atlassian.net/browse/DOCS-13441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

content/en/extend/faq/_index.md
```diff
@@ -0,0 +1,34 @@
...
+### Libraries and community contributions
+
+* [OmniOS Install From Source][5]
+* [Is it possible to integrate with ThousandEyes?][6]
+* [Deploying the Agent on RaspberryPI][7]
+* [Custom Legacy OpenMetrics Check][9]
+
+[1]: /extend/faq/use-our-webhook-integration-to-create-a-trello-card/
+[2]: /extend/faq/how-to-post-appdynamics-events-to-datadog/
+[3]: /extend/faq/data-aggregation-with-dogstatsd-threadstats/
```
joepeeples: I very randomly discovered this link 404's. I think there might be a redirect missing in the target file `extend/dogstatsd/data_aggregation.md`. But we should also update the hardcoded link here:
```suggestion
[3]: /extend/dogstatsd/data_aggregation/
```

content/en/extend/dogstatsd/data_aggregation.md
```diff
@@ -3,11 +3,12 @@ title: DogStatsD Data Aggregation
 description: Learn how the DogStatsD server aggregates your data before sending it to Datadog
 aliases:
     - /developers/faq/data-aggregation-with-dogstatsd-threadstats
```
joepeeples, line 5: I think we need another redirect here. I got a 404 when I followed a link pointing to this page from [/extend/faq](https://docs-staging.datadoghq.com/heston/migrate-developers/extend/faq/).
```suggestion
    - /developers/faq/data-aggregation-with-dogstatsd-threadstats
    - /extend/faq/data-aggregation-with-dogstatsd-threadstats
```

content/en/extend/_index.md
```diff
@@ -1,7 +1,8 @@
 ---
-title: Developers
-description: Learn how to develop an integration on Datadog.
+title: Extend Datadog
```
joepeeples, line 2: The table on this page has a few links that still include `/developers/`. Redirects will probably catch them, but we should still update the source URLs, too.

Looks like the table is generated from the partial `requests.html` which pulls data from `data/partials/requests.yaml`. [That YAML file](https://github.com/DataDog/documentation/blob/d5c6c2fe0c0032a670f5f109e715f967d1a252ee/data/partials/requests.yaml#L11) is where the links are hardcoded (I see three when I search for `/developers`).

content/en/extend/faq/_index.md
hestonhoffman: I think this one actually predates our change (ie, it's an old dev file that doesn't exist)!

content/en/extend/_index.md
hestonhoffman: Good catch, I pushed a change to fix these as well as other links from the shortcodes directory that hadn't been caught yet. Also updated the Claude skill

PR #34713, opened 2026-02-19
(AI) Add llm page hints

hestonhoffman:
We publish `.md` versions of most pages on the docs site to help LLMs read our documentation. But agents seem to have trouble finding these pages when a user asks one to fetch and read a page. 

This PR adds adds an "invisible" div to the top of the page with directions to the `.md` version (if one exists).

### How it works

1. Once per build, `pages-json-cache.html` uses partialCached to download pages.json from the docs site and store the data in a cached Hugo dict.
2. Once per page, `page-agent-hint.html` checks if a key exists in the cached dict and returns the `.md` URL. This also uses partialCached to prevent multiple calls for the same page.
3. If a key exists, the div is added to the page.

Note:
- I borrowed a lot of this functionality from something @jhgilbert was working on.
- There is a bit of futzing with URL strings to get the correct .md path to the page.
- I needed to add this higher in the `body` element to get it to work

Screenshot 2026-02-19 at 11 51 01 AM

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -28,10 +28,6 @@
   {{- if and (ne $.Params.disable_opengraph_meta_tags true) .File -}}
     {{- partial "meta.html" . -}}
   {{- end -}}
-  {{- $mdURL := partialCached "page-agent-hint.html" (dict "ctx" .) .RelPermalink -}}
-  {{- with $mdURL -}}
-  <meta name="plaintext-source" content="{{ . }}">
-  {{- end -}}
 </head>
 {{- $bodyClass := $.Scratch.Get "bodyClass" -}}
 {{- $customClass := $.Params.customclass -}}


layouts/_default/baseof.html
```diff
@@ -28,13 +28,20 @@
...
+  {{- end -}}
 </head>
 {{- $bodyClass := $.Scratch.Get "bodyClass" -}}
 {{- $customClass := $.Params.customclass -}}
 {{ $ctx := . }}
 
 <body
   class="{{ .Site.Language.Lang }} {{ if .Kind }}kind-{{.Kind}}{{ end }} {{ if .IsPage }} {{ replace $.Type "/" "-" }} {{ else }} {{ .Section }} {{ end }} {{ $bodyClass }} {{ $customClass }} {{ if .Site.Params.announcement_banner }}announcement{{ end }}">
+  {{- with partialCached "page-agent-hint.html" (dict "ctx" .) .RelPermalink -}}
+  <div style="display:none" class="exclude-from-md">For a plain-text version of this page, see: {{ . }}</div>
```
jhgilbert:
```suggestion
  <div style="display:none" class="exclude-from-md">For a plaintext version of this page, see: {{ . }}</div>
```
A nitpick but I feel like this is the convention these days

PR #34844, opened 2026-02-25
Add missing quotation mark to Directory check description

evazorro:

Removes a stray quotation mark at the end of the Directory check description in the commonly requested technologies table on the Developers page.

### Merge instructions

Merge readiness:
- [x] Ready for merge

### Additional notes

### DIFF
--- a/data/partials/requests.yaml
+++ b/data/partials/requests.yaml
@@ -24,7 +24,7 @@ requests:
     link: "/infrastructure/process"
     image: "int_processes.png"
   - name: "Files and Directories"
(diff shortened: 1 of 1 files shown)


janine-c:
Judging by the structure of the rest of the file, I wonder if we should be _adding_ a quotation mark to the beginning of this string?

evazorro:
> Judging by the structure of the rest of the file, I wonder if we should be _adding_ a quotation mark to the beginning of this string?

Yeesh good catch!! I think you're right. Updating.

PR #35045, opened 2026-03-05
add ECS MI daemon instructions

stzou:
Update ECS MI documentation to include instructions for setting up the Datadog Agent as a Daemon. This is now the preferred method of setting up the Datadog Agent(as opposed to sidecar deployment). The sidecar instructions are still included in the documentation so existing users with this setup can reference it, but it is now discouraged. 

### Merge instructions

**Do not merge until 4/02**

Merge readiness:
- [ ] Ready for merge


iadjivon:
Added an editorial review card: DOCS-13654

PR #35126, opened 2026-03-09
[DOCS-13596] Clarify entity creation workflow and API definition upload

drichards-87:

Fixes DOCS-13596

Clarifies the "Through the Datadog UI" workflow in Create Entities by adding an intro paragraph that describes the Kind dropdown and how form fields vary by entity type, including API specification upload for the API kind.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes


margheritadonnici:
Looks good to me, wondering if it's worth talking about the "+ API" button in the API view of Software Catalog?

drichards-87:
> Looks good to me, wondering if it's worth talking about the "+ API" button in the API view of Software Catalog?

That’s a good point. I thought about that as well. I ended up choosing this approach to reduce the maintenance burden, updating existing docs instead of trying to add Software Catalog specific instructions to the App and API Protection docs. Also, Datadog offers multiple ways to complete the same task in the UI. Documenting each and every option isn’t feasible, and I’m not sure it would be a good idea either. So I think going with this approach makes the most sense right now. We can always revisit this if it continues to cause confusion with customers.

margheritadonnici:
> > Looks good to me, wondering if it's worth talking about the "+ API" button in the API view of Software Catalog?
> 
> That’s a good point. I thought about that as well. I ended up choosing this approach to reduce the maintenance burden, updating existing docs instead of trying to add Software Catalog specific instructions to the App and API Protection docs. Also, Datadog offers multiple ways to complete the same task in the UI. Documenting each and every option isn’t feasible, and I’m not sure it would be a good idea either. So I think going with this approach makes the most sense right now. We can always revisit this if it continues to cause confusion with customers.

Alright, makes sense!

content/en/internal_developer_portal/software_catalog/set_up/create_entities.md
```diff
@@ -37,18 +37,21 @@ To add [entity definitions][13] to Software Catalog, you can:
...
+Software Catalog provides a guided workflow for creating entity definitions. After you select an entity type from the **Kind** dropdown (Service, API, System, and others), the form presents metadata fields and options specific to that type. For example, selecting **API** surfaces fields for uploading an OpenAPI or gRPC specification, while selecting **Service** surfaces fields for defining the service type and lifecycle.
+
+To create an entity definition:
 
 1. Navigate to the [Software Catalog Setup & Config][3] page.
 1. Click **Create a New Entry**.
-1. Specify your service details, including metadata such as ownership and documentation links.
-1. (Optional) Switch to **YAML** or **JSON** to see the generated code and cURL command. In the code editors, Datadog automatically flags invalid data. 
+1. Select the entity type from the **Kind** dropdown.
+1. Fill in the metadata fields, such as ownership and documentation links.
```
iadjivon:
```suggestion
1. Fill in the metadata fields, such as ownership and reference links.
```
I'd suggest calling it reference since the link can be documentation or code repos etc.

content/en/internal_developer_portal/software_catalog/set_up/create_entities.md
```diff
@@ -37,18 +37,21 @@ To add [entity definitions][13] to Software Catalog, you can:
...
+1. (Optional) Switch to **YAML** or **JSON** to see the generated code and cURL command. In the code editors, Datadog automatically flags invalid data.
```
iadjivon, line 48: I'd recommend modifying the image since the new UI has **YAML** or **JSON** but the current image does not.

iadjivon (approved):
Hi DeForest! Approved with some comments :)

PR #35343, opened 2026-03-17
Clarifying tag key parameters

clreaume:
Based on some internal confusion here: https://dd.slack.com/archives/C8JDGHUCE/p1773760043260989

Docs team discussion here: https://dd.slack.com/archives/GGHR55PNV/p1773761413651209

Clarifies the limitations around tag keys—i.e., which types of characters are and are not allowed.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### AI assistance

Did not use for drafting (did use Gemini conversationally to learn more about Unicode)

### Additional notes
I have not yet set up my local docs build, and so am opening this PR as a draft!


content/en/getting_started/tagging/_index.md
```diff
@@ -171,4 +172,4 @@ For more information, see [Using Tags][1].
 [22]: /getting_started/tagging/using_tags/#developers
 [23]: /account_management/billing/usage_attribution/
 [24]: /getting_started/tagging/using_tags/#ci-visibility
-[25]: /containers/troubleshooting/log-collection?tab=datadogoperator#missing-host-level-tags-on-new-hosts-or-nodes
\ No newline at end of file
+[25]: /containers/troubleshooting/log-collection?tab=datadogoperator#missing-host-level-tags-on-new-hosts-or-nodes
```
clreaume, line 183: mrrr I didn't add this back in (intentionally)

maycmlee: just lurking, and thought I'd respond :) This happens regularly...I'm not sure why though...I've seen it my own PR and other's.

content/en/getting_started/tagging/_index.md
```diff
@@ -58,7 +58,8 @@ Below are Datadog's tagging requirements:
 
 1. Tags must **start with a letter** and after that may contain the characters listed below:
 
-    - Alphanumerics
+    - Letters (all Unicode letters are supported, e.g., a, ó, 気, 녕, ك, and ดี)
```
tibo-mllr: Correct me if I'm wrong, but aren't emojis a subset of Unicode? Does that mean we allow emojis in tags?

content/en/getting_started/tagging/_index.md
clreaume: thanks May! it's helpful to know you've seen this often. I was just looking closer and realized git is saying, "this content was removed from line 174 and added on line 175." oddly, that pattern is true of all of the content on this page below my change (since my change added a line), but it's only recognized on this (last) line

do you always see this kind of thing with the last line of a file, or does the location vary?

maycmlee: Ohh interesting, I never looked into it, but that kind of makes sense. I do always see it with the last line of a file.

clreaume: ok, noted! thank you!

content/en/getting_started/tagging/_index.md
clreaume: something to note here is that we allow Unicode _letters_, not all Unicode characters. I just chatted with Claude, and he told me we do not support emojis in tags. I've made a change to state that explicitly, and am going to seek human validation before merging!

clreaume:
I made some changes to this page after chatting with Claude, and am seeking human validation before publishing. namely that:

- tags do not support commas (our docs [currently state explicitly that we do](https://docs.datadoghq.com/getting_started/tagging/#define-tags)) 
- tags support ASCII letters (a-z) as well as all Unicode letters (for example, ó, 気, 녕, ك, and ดี)
- with regard to slashes, tags only support _forward_ slashes, not back slashes
- tags don't support emoji
- in tags, leading/trailing underscores are stripped, and contiguous underscores are collapsed into one

@nhamlani101 (hi Noman, nice to meet ya!), you are the PM who looked most closely aligned to this work from looking at [this spreadsheet](https://docs.google.com/spreadsheets/d/1myTg0Z8bcyTARiJjYU6mRPo-wHEgxJAaRACwpeE602w/edit?pli=1&gid=347181619#gid=347181619). could you please validate my checks in this comment, or point me to someone who could? thank you so much!

I'll also add below a screenshot of my convo with Claude, if anyone is curious:
[update: I removed my screenshot of my convo with sir Claude, but message me if you'd like to see it!]

clreaume:
just noting that I continued to augment these docs (the Overview and Define tags sections of Getting Started with Tags) today after I learned about some related-but-slightly-new feedback @maycmlee told me about in this ticket: https://datadoghq.atlassian.net/browse/DOCS-11311. today's changes are in this commit: https://github.com/DataDog/documentation/pull/35343/changes/4e9a3d10cd7896965289a8a74c2d57f6f41c1765

content/en/getting_started/tagging/_index.md
```diff
@@ -32,10 +32,10 @@ algolia:
 
 Tags are a way of adding dimensions to Datadog telemetries so they can be filtered, aggregated, and compared in Datadog visualizations. [Using tags][1] enables you to observe aggregate performance across several hosts and (optionally) narrow the set further based on specific elements. In summary, tagging is a method to observe aggregate datapoints.
 
-Tags are `key:value` pairs containing two parts:
+A tag can be formatted as `<value>` or `<key>:<value>`. When using a `<key>:<value>` pair:
```
jszwedko: Tags that are `key:value` are strongly encouraged vs. tags that are just `value`s (e.g. they are easier to query). That might be why we don't document their existence 🤔 

I'd suggest we de-emphasize `value` tags, at least, and push people to `key:value`.

jszwedko: Ah, nvm, I do see we document the existence of `value` tags below.

clreaume: ah, this is valuable context! updated, thank you

content/en/getting_started/tagging/_index.md
```diff
@@ -56,29 +56,30 @@ Because containers and cloud environments regularly churn through hosts, using t
...
+    - Forward slashes
 
-    Other special characters are converted to underscores.
+    All other characters (including, but not limited to, commas, emoji, and spaces) are converted to underscores.
 
-2. Tags can be **up to 200 characters** long and support Unicode letters (which includes most character sets, including languages such as Japanese).
-3. Tags are converted to lowercase. Therefore, `CamelCase` tags are not recommended. Authentication (crawler) based integrations convert camel case tags to underscores, for example `TestTag` --> `test_tag`.
-4. A tag can be in the format `value` or `<KEY>:<VALUE>`. Commonly used tag keys are `env`, `instance`, and `name`. The key always precedes the first colon of the global tag definition, for example:
+- Tags can be **up to 200 characters** long. If the tag has the format `<key>:<value>`, the `<key>`, `:`, and `<value>` all count toward the character limit.
+- For log attributes and span tags, tags are case-sensitive. For metric tags, tags are converted to lowercase.
```
Reallnas: Note: There are Span tags and span attributes. Span tags and attributes are collected automatically by the instrumentation, but span attributes can also be defined in the code with `span.Tag(...)` (yes...). The former are normalized like metrics (lowercase, normalized), attributes are mostly left untouched. 

There is this public [page](https://docs.datadoghq.com/tracing/trace_collection/tracing_naming_convention#span-tags) that explains the difference between both. But since the public documentation on the tracer are explicitly referring to the `attributes` as `tags` (e.g. [java](https://docs.datadoghq.com/tracing/trace_collection/custom_instrumentation/server-side/?api_type=dd_api&prog_lang=java#adding-tags-java)), the wording can be confusing for spans. 

Since tags are defined on the tracer as `Tag(key, value)` and are never in the `key:value` form (except when querying), maybe it could be worth to change the wording to explain that attributes are not normalized but tags are ? What do you think we could do to make this clearer ?

(Yes, this is confusing 😓 )

clreaume: hey, thanks for catching this! I just pushed an update that hopefully clarifies that span tags are normalized to lowercase, while span attributes are not. would you mind taking another look?

Reallnas (approved):

content/en/getting_started/tagging/_index.md
```diff
@@ -56,29 +56,30 @@ Because containers and cloud environments regularly churn through hosts, using t
...
-    - Underscores
+    - Letters (all Unicode letters are supported---for example, a, ó, 気, 녕, ك, and ดี)
+    - Numbers
+    - Underscores (leading/trailing underscores are stripped, and contiguous underscores are collapsed into one)
     - Minuses
     - Colons
-    - Commas
     - Periods
-    - Slashes
+    - Forward slashes
```
rayram23, line 70: Is this to say we do not allow back slashes? `\` ?

rayram23: Would be good to add clarity there.

clreaume: hi Ray, that's what Claude told me! can you please verify? (I'd also love if you could verify that commas aren't allowed.) if so, I can add it to the parenthetical in the line below ("All other characters..."). good call!

rayram23 (approved):

content/en/getting_started/tagging/_index.md
rayram23: Commas in tag keys are not allowed and are sanitized to underscores (_) — no error is returned.

clreaume: thanks! you called out tag _keys_—are commas allowed in tag values? my understanding was that these constraints applied to the entire tag string (i.e., all of `<key>:<value>` or `<value>`)

rayram23: The allowed characters in tag values are: alphanumeric (a-z, 0-9), _, -, ., /, :, @. Commas are not in the valid set.

content/en/getting_started/tagging/_index.md
```diff
@@ -54,31 +54,32 @@ Because containers and cloud environments regularly churn through hosts, using t
...
 
-    Other special characters are converted to underscores.
+    All other characters (including, but not limited to, commas, emoji, and spaces) are converted to underscores.
```
clreaume:
```suggestion
    All other characters (including, but not limited to, commas, emoji, back slashes, and spaces) are converted to underscores.
```

content/en/getting_started/tagging/_index.md
clreaume: cool cool, thanks for confirming that commas are not allowed in tag keys nor values. but! you just introduced a new allowed character that's not in the existing list: @

before I add it to the allowlist, just double-checking that they are indeed allowed, because Claude told me otherwise—that @s aren't. it cited [this test](https://github.com/DataDog/dd-trace-rb/blob/master/spec/datadog/core/tag_normalizer_spec.rb#L35), where `fun:ky@@@tag/3` is normalized to `fun:ky_tag/3`

as for your original comment that it "Would be good to add clarity [on back slashes not being allowed]," I've pushed an update that explicitly says that

clreaume: hi @nhinsch — thanks for [volunteering](https://dd.slack.com/archives/C03A4P0207Q/p1775164129480529?thread_ts=1774618563.282709&cid=C03A4P0207Q) to jump in here in Ray's stead! I'm hoping to move this change along, so please provide any additional feedback by Monday, 4/13. as of now, I'm planning to leave @ out of the allowlist, per my Claude-assisted research. as I mentioned [here](https://dd.slack.com/archives/C03A4P0207Q/p1775224159711669?thread_ts=1774618563.282709&cid=C03A4P0207Q), if there are additional important, log-specific distinctions, feel free to share them! thank you!

nhinsch: To circle back on this, it turns out that in logs, `@` is allowed by our backend, but is removed by the Datadog agent. Nonetheless I will approve the PR and let you decide how you want to proceed with this docs change 👍

nhinsch (approved):

content/en/getting_started/tagging/_index.md
clreaume: thanks! as I said [here](https://dd.slack.com/archives/C07FZ0UG4CC/p1776866784092949?thread_ts=1776146071.343129&cid=C07FZ0UG4CC), I think it makes sense to specify in the docs that @s in tags for logs ingested via HTTP are allowed/not normalized to _s. I just pushed a commit that makes that change

with that, I think we're ready to publish this!

content/en/getting_started/tagging/_index.md
```diff
@@ -32,10 +32,10 @@ algolia:
 
 Tags are a way of adding dimensions to Datadog telemetries so they can be filtered, aggregated, and compared in Datadog visualizations. [Using tags][1] enables you to observe aggregate performance across several hosts and (optionally) narrow the set further based on specific elements. In summary, tagging is a method to observe aggregate datapoints.
 
-Tags are `key:value` pairs containing two parts:
+A tag can be formatted as `<key>:<value>` or `<value>`. It is recommended to use the `<key>:<value>` format, as it is often semantically clearer and allows for richer querying capabilities (for example, grouping by key). When using a `<key>:<value>` pair:
```
buraizu:
```suggestion
A tag can be formatted as `<key>:<value>` or `<value>`. Datadog recommends using the `<key>:<value>` format, as it is often semantically clearer and allows for richer querying capabilities (for example, grouping by key). When using a `<key>:<value>` pair:
```

content/en/getting_started/tagging/_index.md
```diff
@@ -54,35 +54,37 @@ Because containers and cloud environments regularly churn through hosts, using t
...
 
-Below are Datadog's tagging requirements:
+Tag strings (that is, the entire content of `<key>:<value>` or `<value>`) must meet the following requirements:
 
-1. Tags must **start with a letter** and after that may contain the characters listed below:
+- Tag strings must **start with a letter** (this applies regardless of whether the tag uses the format `<key>:<value>` or `<value>`). After the leading letter, the tag string may contain the characters listed below:
 
-    - Alphanumerics
-    - Underscores
+    - Letters (all Unicode letters are supported---for example, a, ó, 気, 녕, ك, and ดี)
```
buraizu:
```suggestion
    - Letters (all Unicode letters are supported—for example, a, ó, 気, 녕, ك, and ดี)
```

buraizu: Docs typically use the Unicode em-dash character — rather than ---. Example from a sibling file (content/en/getting_started/tagging/using_tags.md): 
`"narrow down the scope even further—for example…"`

content/en/getting_started/tagging/_index.md
```diff
@@ -54,35 +54,37 @@ Because containers and cloud environments regularly churn through hosts, using t
...
     - Colons
-    - Commas
     - Periods
-    - Slashes
-
-    Other special characters are converted to underscores.
+    - Forward slashes
+    - (For tags on logs [ingested via HTTP][28] only) at signs (@)
 
+    All other characters (including, but not limited to, commas, emoji, back slashes, and spaces) are converted to underscores.
```
buraizu:
```suggestion
    All other characters (including commas, emoji, backslashes, and spaces) are converted to underscores.
```

content/en/getting_started/tagging/_index.md
```diff
@@ -54,35 +54,37 @@ Because containers and cloud environments regularly churn through hosts, using t
...
 
-2. Tags can be **up to 200 characters** long and support Unicode letters (which includes most character sets, including languages such as Japanese).
-3. Tags are converted to lowercase. Therefore, `CamelCase` tags are not recommended. Authentication (crawler) based integrations convert camel case tags to underscores, for example `TestTag` --> `test_tag`.
-4. A tag can be in the format `value` or `<KEY>:<VALUE>`. Commonly used tag keys are `env`, `instance`, and `name`. The key always precedes the first colon of the global tag definition, for example:
-
+- Tags can be **up to 200 characters** long. If the tag has the format `<key>:<value>`, the `<key>`, `:`, and `<value>` all count toward the character limit.
+- [Span tags][26] and metric tags are normalized to lowercase. Because of this, using `CamelCase` is not recommended. Authentication (crawler) based integrations also convert camel case tags to underscores before lowercasing, for example `TestTag` --> `test_tag`.
```
buraizu:
```suggestion
- [Span tags][26] and metric tags are normalized to lowercase, so avoid camelCase. Authentication (crawler) based integrations also convert camel-case tags to underscores before lowercasing. For example, `testTag` becomes `test_tag`.
```

buraizu: These examples aren't actually camel case, and it's worth being precise here because these are identifier-convention terms with established technical meanings.

Additionally, `-->` is an HTML comment sentinel and renders inconsistently, and using prose might be more understandable for screen readers.

content/en/getting_started/tagging/_index.md
```diff
@@ -54,35 +54,37 @@ Because containers and cloud environments regularly churn through hosts, using t
...
+    - Underscores (leading/trailing underscores are stripped, and contiguous underscores are collapsed into one)
     - Minuses
     - Colons
-    - Commas
     - Periods
-    - Slashes
-
-    Other special characters are converted to underscores.
+    - Forward slashes
+    - (For tags on logs [ingested via HTTP][28] only) at signs (@)
```
buraizu:
```suggestion
    - At signs (`@`)—only for tags on logs [ingested through HTTP][28]
```

buraizu: Moving the parenthetical to the end makes this a bit easier to scan

content/en/getting_started/tagging/_index.md
```diff
@@ -54,35 +54,37 @@ Because containers and cloud environments regularly churn through hosts, using t
...
 
-2. Tags can be **up to 200 characters** long and support Unicode letters (which includes most character sets, including languages such as Japanese).
-3. Tags are converted to lowercase. Therefore, `CamelCase` tags are not recommended. Authentication (crawler) based integrations convert camel case tags to underscores, for example `TestTag` --> `test_tag`.
-4. A tag can be in the format `value` or `<KEY>:<VALUE>`. Commonly used tag keys are `env`, `instance`, and `name`. The key always precedes the first colon of the global tag definition, for example:
-
+- Tags can be **up to 200 characters** long. If the tag has the format `<key>:<value>`, the `<key>`, `:`, and `<value>` all count toward the character limit.
+- [Span tags][26] and metric tags are normalized to lowercase. Because of this, using `CamelCase` is not recommended. Authentication (crawler) based integrations also convert camel case tags to underscores before lowercasing, for example `TestTag` --> `test_tag`.
+    - Separate from tags, [span attributes][27] and log attributes are case-sensitive, and are not normalized.
```
buraizu:
```suggestion
    - Unlike tags, [span attributes][27] and log attributes are case-sensitive, and are not normalized.
```

content/en/getting_started/tagging/_index.md
```diff
@@ -54,35 +54,37 @@ Because containers and cloud environments regularly churn through hosts, using t
...
+Tag strings (that is, the entire content of `<key>:<value>` or `<value>`) must meet the following requirements:
 
-1. Tags must **start with a letter** and after that may contain the characters listed below:
+- Tag strings must **start with a letter** (this applies regardless of whether the tag uses the format `<key>:<value>` or `<value>`). After the leading letter, the tag string may contain the characters listed below:
 
-    - Alphanumerics
-    - Underscores
+    - Letters (all Unicode letters are supported---for example, a, ó, 気, 녕, ك, and ดี)
+    - Numbers
+    - Underscores (leading/trailing underscores are stripped, and contiguous underscores are collapsed into one)
```
buraizu:
```suggestion
    - Underscores (leading and trailing underscores are removed, and contiguous underscores are collapsed into one)
```

content/en/getting_started/tagging/_index.md
buraizu: My perception is that `including` already implies `but not limited to`, but open to feedback if you have a different perspective

content/en/getting_started/tagging/_index.md
```diff
@@ -54,35 +54,37 @@ Because containers and cloud environments regularly churn through hosts, using t
...
     - The `DD_TAGS` environment variable uses whitespace as a separator between tags. Whitespace in `DD_TAGS` values is **not** converted to underscores. For example, `DD_TAGS="test:this is a test"` produces four separate tags: `test:this`, `is`, `a`, and `test`. To set a tag value that contains spaces, use a YAML configuration file or integration annotations, where whitespace is converted to underscores.
 
-2. Tags can be **up to 200 characters** long and support Unicode letters (which includes most character sets, including languages such as Japanese).
-3. Tags are converted to lowercase. Therefore, `CamelCase` tags are not recommended. Authentication (crawler) based integrations convert camel case tags to underscores, for example `TestTag` --> `test_tag`.
-4. A tag can be in the format `value` or `<KEY>:<VALUE>`. Commonly used tag keys are `env`, `instance`, and `name`. The key always precedes the first colon of the global tag definition, for example:
-
+- Tags can be **up to 200 characters** long. If the tag has the format `<key>:<value>`, the `<key>`, `:`, and `<value>` all count toward the character limit.
```
buraizu:
```suggestion
- Tags can be **up to 200 characters** long. If the tag has the format `<key>:<value>`, the key, `:`, and value all count toward the character limit.
```

buraizu: Minor update to untangle conceptually different uses of "key" and "value". Alternatively, since the key/value concept is already defined earlier in the page, this could be shortened to:
`Tags can be up to 200 characters long, including the key, the :, and the value.`

content/en/getting_started/tagging/_index.md
```diff
@@ -54,35 +54,37 @@ Because containers and cloud environments regularly churn through hosts, using t
...
+- When using the format `<key>:<value>`, the key always precedes the first colon of the global tag definition. For example:
+    
     | Tag                | Key           | Value          |
     | ------------------ | ------------- | -------------- |
     | `env:staging:east` | `env`         | `staging:east` |
     | `env_staging:east` | `env_staging` | `east`         |
 
-5. Tags should not originate from unbounded sources, such as epoch timestamps, user IDs, or request IDs. Doing so may infinitely [increase the number of metrics][2] for your organization and impact your billing.
-6. Limitations (such as downcasing) only apply to metric tags, not log attributes or span tags.
+- Tags should not originate from unbounded sources, such as epoch timestamps, user IDs, or request IDs. Doing so may infinitely [increase the number of metrics][2] for your organization and impact your billing.
```
buraizu:
```suggestion
- Tags should not originate from unbounded sources, such as epoch timestamps, user IDs, or request IDs. Doing so can cause unbounded growth in the number of metrics.
```

buraizu: Optional suggestion since this is not actually part of your changes

buraizu (commented):
Left some suggestions, but happy to keep iterating on this if you have any feedback

content/en/getting_started/tagging/_index.md
clreaume: thanks! I can see how switching the order makes this list item easier to scan, but I wonder if it carries a cost—like people skimming past the parenthetical

in general I try to put conditional clauses before the instruction or information they modify (see this concept further explained on Google's [Sentence Structure](https://developers.google.com/style/sentence-structure) page)

content/en/getting_started/tagging/_index.md
clreaume: I use `but not limited to` when I want to stress that the list is non-exhaustive, but I don't think it's super important to do so in this case. applying your suggestion, thanks!

content/en/getting_started/tagging/_index.md
clreaume: thanks, I'm down to apply this suggestion, but also just curious—did you remove the clause about billing being impacted because that's implied?

content/en/getting_started/tagging/_index.md
clreaume: thanks! I made an update to use "camel case," since it seems like that's what we've used elsewhere in the docs. (I like the idea of styling it autologically, but didn't feel it was imperative)

content/en/getting_started/tagging/_index.md
clreaume: nice catch! I took your suggestion

content/en/getting_started/tagging/_index.md
clreaume: I took most of your suggestion, but left in the link (but modified it to only apply to "metrics")

content/en/getting_started/tagging/_index.md
clreaume: I applied your suggestions to switch `via` to `through` and to code format the `@`, but left the parenthetical at the beginning of the text. let me know whatcha think @buraizu !

content/en/getting_started/tagging/_index.md
```diff
@@ -57,35 +57,37 @@ Because containers and cloud environments regularly churn through hosts, using t
...
 
-2. Tags can be **up to 200 characters** long and support Unicode letters (which includes most character sets, including languages such as Japanese).
-3. Tags are converted to lowercase. Therefore, `CamelCase` tags are not recommended. Authentication (crawler) based integrations convert camel case tags to underscores, for example `TestTag` --> `test_tag`.
-4. A tag can be in the format `value` or `<KEY>:<VALUE>`. Commonly used tag keys are `env`, `instance`, and `name`. The key always precedes the first colon of the global tag definition, for example:
-
+- Tags can be **up to 200 characters** long. If the tag has the format `<key>:<value>`, the key, `:`, and value all count toward the character limit.
+- [Span tags][26] and metric tags are normalized to lowercase, so avoid using camel case. Authentication (crawler) based integrations also convert camel case tags to underscores before lowercasing. For example `testTag` becomes `test_tag`.
+    - Unlike tags, [span attributes][27] and log attributes are case-sensitive, and are not normalized.
```
buraizu, line 81:
```suggestion
- [Span tags][26] and metric tags are normalized to lowercase, so avoid using camel case. Authentication (crawler) based integrations also convert camel case tags to underscores before lowercasing. For example, `testTag` becomes `test_tag`.
    - Unlike tags, [span attributes][27] and log attributes are case-sensitive and are not normalized.
```

content/en/getting_started/tagging/_index.md
buraizu: Thanks for the link — that page is a useful reference, but I'd push back a bit on whether it applies here.
                                                                                                                                                                  
The Google guidance is explicitly about instructions: "If you want to tell the reader to do something, try to mention the circumstance...before you provide the instruction" (their example is "Click Delete if you want to delete the entire document" → "To delete the entire document, click Delete"). The bullet at line 68 isn't an instruction, though — it's an item in a noun list enumerating allowed characters. The circumstance-before instruction rule is about procedural sentences, not list items.                                                                                                                                      
                  
The trade-off I see: every other bullet in this list leads with the character class as the subject ("Letters", "Numbers", "Underscores", and so on). The @  bullet is the only one that inverts that, so a reader scanning to answer "is @ allowed?" has to read the parenthetical before discovering which bullet they're on. The skim-past risk applies in either ordering — readers can skim past a trailing qualifier or a leading parenthetical equally — but the parallelism cost is paid on every scan.

If the concern is keeping the conditional prominent without breaking parallelism, one option is to promote it out of the bullet:                                
   

> - At signs (@)                                                                                                                                                
>       
> **Note**: At signs are allowed only in tags on logs [ingested through HTTP][28].

Up to you.

buraizu (approved):
Thanks for applying that feedback. I'm approving with one additional minor punctuation suggestion, and also a reply to your comment about the `@` symbols with a potential middle-ground solution. I'll leave its implementation up to you.

content/en/getting_started/tagging/_index.md
clreaume: thanks! the Google page does focus on instructions, but the third example shows how the principle can be applied to non-instructional information:

- bad: Using custom domains might add noticeable latency to responses if your app is located in one of the following regions:
- good: If your app is located in one of the following regions, using custom domains might add noticeable latency to responses:

> The skim-past risk applies in either ordering

this is true, but I think the negative impact of using an `@` sign in a non-HTTP-ingested log tag is greater than someone missing that they actually _can_ use `@`s in a HTTP-ingested-log tag. in the former, they'll generate an error in Datadog, and have to go back and rename the tag

as such, I'm going to leave it as is. thank you for this discussion, @buraizu—I enjoy digging into these things, especially with someone with a slightly different perspective!

PR #35429, opened 2026-03-20
[APPSEC-61834] Fix Python IAST OS compatibility table

avara1986:

Fixes APPSEC-61834: https://datadoghq.atlassian.net/browse/APPSEC-61834

The Python IAST OS compatibility table:

- The macOS row for ddtrace v3 (`>=3, <4`) was missing — macOS is supported for IAST in both v3 and v4, as IAST native extensions are compiled on all non-Windows platforms.
- Windows was correctly removed from both v3 and v4 rows since IAST native extensions (`_stacktrace`, `iastpatch`, `_native`) are not compiled on Windows (confirmed in `setup.py`).

### DIFF
--- a/content/en/security/code_security/iast/setup/_index.md
+++ b/content/en/security/code_security/iast/setup/_index.md
@@ -834,7 +834,12 @@ Two release branches are supported:
 
 And the library supports the following runtimes:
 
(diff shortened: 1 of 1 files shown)


avara1986:
@codex review

PR #35926, opened 2026-04-10
[WEB-8784] Enforce max image width via Fastly IO params in img shortcode

nsollecito:

Fixes WEB-8784: https://datadoghq.atlassian.net/browse/WEB-8784

The `img.html` shortcode computed a max display width (`$img_w`: 850px standard, 1170px wide) but never applied it to the Fastly IO image URL. Every image was served at full resolution via `?auto=format` only, regardless of actual display size.

This change appends `&w=$img_w` to the Fastly IO URL in all three `$e` assignment branches, capping image downloads to the content column constraints. Also renames `$imgix_w` → `$img_w` for CDN-agnostic naming.

Noticeable difference in larger-asset sizes. Validate downloaded file sizes on the /logs page compared with live: https://docs-staging.datadoghq.com/nsollecito/img-max-width/logs/

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes

Content column is `col-lg-7` of a 1170px container (~682px rendered width). The 850px cap for standard images provides ~1.25x retina headroom; 1170px for wide images matches the full container width.

### DIFF
--- a/config/_default/params.yaml
+++ b/config/_default/params.yaml
@@ -2,10 +2,10 @@ environment: default
 # do not delete repo_name, used by websites-modules as context
 repo_name: "documentation"
 integrations_img_url: "https://static.datadoghq.com/static/images/logos/"
-imgix:
-  header_jpg: "?ch=Width&fit=max&fm=jpg&auto=format&lossless=true"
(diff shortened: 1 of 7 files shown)


devindford (approved):
Image sizes being returned on the preview are smaller but still the same quality

PR #36044, opened 2026-04-16
[SDTEST-3046] Document 100k hard limit for non-paginated known tests endpoint

borja-gomez:
## Summary
- Documents the new 100,000 test hard limit for non-paginated requests to the known tests endpoint
- Replaces the old 500k limit reference with the new 100k limit
- Notes that the existing Compatibility minimums cover EFD's launch, not pagination support

## Context
The known tests endpoint (`POST /api/v2/ci/libraries/tests`) now enforces a hard limit of 100k tests for non-paginated requests. All tracer libraries have pagination support, so clients on recent versions are unaffected. Clients on older versions with services larger than 100k known tests need to upgrade.

### Why drop the 500k mention?

The original doc said:

> Older tracer versions limit the number of known tests fetched to 500k. If you need to fetch more than 500k known tests, update to the latest tracer version.

That mention is now obsolete from the customer-visible perspective:

- With the new 100k hard limit added — also only on non-paginated — **the 100k always fires first**. There is no realistic scenario where a customer reaches the 500k threshold without hitting 100k first.
- **Paginated** requests have no upper bound at all; the per-page cap is `MaxPageSize` (100k) but the client can request as many pages as it wants.

So mentioning 500k alongside 100k would just be noise. The 100k limit fully supersedes it in user-visible behavior.


juliomalegria-dd (approved):

content/en/tests/flaky_tests/early_flake_detection.md
```diff
@@ -103,8 +103,8 @@ The test framework compatibility is the same as [Test Optimization Compatibility
 
 {{< /tabs >}}
 
-<div class="alert alert-danger">
-Older tracer versions limit the number of known tests fetched to 500k. If you need to fetch more than 500k known tests, update to the latest tracer version.
+<div class="alert alert-warning">
+Test services with more than 100,000 known tests require a Datadog library version that supports pagination on the known tests endpoint. The minimum versions above cover EFD's initial release; pagination support arrived later. Without it, the library cannot fetch the known tests baseline and Early Flake Detection does not run for that session.
```
juan-fernandez, line 107: can we add what library versions added support for pagination?

borja-gomez: Sure. Do you have the list?

borja-gomez: Added the list

hestonhoffman (approved):

PR #36067, opened 2026-04-16
Add agent-only shortcode

jhgilbert:

Adds a [requested agent-only shortcode](https://datadoghq.atlassian.net/jira/people/5fd13b2734847e0069ea3ec5/boards/18030?assignee=712020%3Abd78b98e-f427-4e98-820c-3c772f1d4ab4&selectedIssue=WEB-8815) to Hugo. I made a test page in a [separate PR](https://github.com/DataDog/documentation/pull/36098).

### Merge instructions

Merge readiness:
- [ ] Ready for merge

**For Datadog employees**:

### DIFF
--- a/layouts/shortcodes/agent-only.html
+++ b/layouts/shortcodes/agent-only.html
@@ -1,0 +1,1 @@
+<div class="plaintext-section" title="Instructions for AI agents" style="display:none;">{{ .Inner }}</div>
--- a/layouts/shortcodes/agent-only.html
+++ b/layouts/shortcodes/agent-only.html
@@ -1,1 +1,1 @@
-<div class="plaintext-section agent-only" title="Instructions for AI agents">{{ .Inner }}</div>
+<div class="plaintext-section agent-only" title="Instructions for AI agents">{{ .Inner | markdownify }}</div>
--- a/layouts/shortcodes/agent-only.html
+++ b/layouts/shortcodes/agent-only.html
@@ -1,1 +1,1 @@
-<div class="plaintext-section" title="Instructions for AI agents" style="display:none;">{{ .Inner }}</div>
+<div class="plaintext-section agent-only" title="Instructions for AI agents">{{ .Inner }}</div>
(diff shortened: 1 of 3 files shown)


layouts/shortcodes/agent-only.html
```diff
@@ -0,0 +1 @@
+<div class="plaintext-section" title="Instructions for AI agents" style="display:none;">{{ .Inner }}</div>
```
hestonhoffman, line 1: Curious what you think about the styles we're using in other places for the same purpose - right now, I think that's just [here](https://github.com/DataDog/documentation/blob/cf682bf6fd7a269a5009096e27252804a0852087/layouts/_default/baseof.html#L39). Is the styling used there overkill? I adopted it after reading [this blog post](https://dacharycarey.com/2026/03/01/make-hugo-site-agent-friendly/#:~:text=5.%20Add%20an%20Agent%20Discovery%20Directive) and I think we did see a small bump in LLM success after making the change. If we do end up using that, maybe we should apply CSS on the class universally?

jhgilbert: I don't think the styling there is overkill for what you were doing, but my primary concern when I'm working inside the content area is making sure screen readers skip this content. I think `display: none;` is considered the most common workhorse for that particular case, since the content won't be read aloud but `display: none;` is still expected to be indexed by everything else, including AI crawlers. 

But honestly, I don't really know how much of an advantage one has over the other, beyond the screen reader consideration.

jhgilbert: If we want to play around with it in the future, I'd be in favor of setting up an `agent-only` class and styling it with `display: none;` but we could experiment with it later if we wanted to. I could invoke `agent-only` in Cdocs, so we could change the behavior in both Hugo and Cdocs without requiring multiple PRs.

hestonhoffman (approved):

bgdeutsch (approved):

PR #36077, opened 2026-04-17
Fix Go SDK code sample errors

RomainMuller:

Fixes two errors in the Go SDK code sample:
- Corrects `r.Query()` to `r.URL.Query()` (`http.Request` has no `.Query()` method).
- Replaces an out-of-scope `metadata` variable reference in the login failure
  tracking call with `nil`.

### Merge instructions

Merge readiness:
- [x] Ready for merge

### Additional notes

### DIFF
--- a/content/en/security/application_security/setup/go/sdk.md
+++ b/content/en/security/application_security/setup/go/sdk.md
@@ -132,7 +132,7 @@ func loginHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Track failed login
-	appsec.TrackUserLoginFailure(r.Context(), req.Username, true, metadata)
+	appsec.TrackUserLoginFailure(r.Context(), req.Username, true, nil)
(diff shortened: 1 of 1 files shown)


content/en/security/application_security/setup/go/sdk.md
```diff
@@ -132,7 +132,7 @@ func loginHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Track failed login
-	appsec.TrackUserLoginFailure(r.Context(), req.Username, true, metadata)
+	appsec.TrackUserLoginFailure(r.Context(), req.Username, true, nil)
```
RomainMuller, line 135: There is no `metadata` in scope here... Using `nil` here is valid... We could also have introduced a `metadata` variable there alternatively. I think both approaches are valid for the sake of the example.

brett0000FF (approved):

PR #36212, opened 2026-04-22
[DOCS-14138] Migrate data warehouse connection guides to Cdocs

iadjivon:

Fixes DOCS-14138

Consolidates the four per-warehouse connection guides under `/experiments/guide/` (BigQuery, Databricks, Redshift, Snowflake) into a single filterable Cdocs page at `/experiments/guide/connecting_a_data_warehouse/`. Readers select their warehouse from the filter at the top of the page, and the previous URLs redirect via aliases so existing links continue to work.

Changes:
- New consolidated `.mdoc.md` page with a per-warehouse `if` block for each warehouse.
- Shared top-level "Step 3: Configure experiment settings" heading with per-warehouse content, to ensure the in-page anchor resolves correctly across all filters.
- New option group `experiments_warehouse_options` in `customization_config/en/option_groups/product_analytics.yaml` (BigQuery is the default).
- Four new options (`bigquery`, `databricks`, `redshift`, `snowflake`) added to `customization_config/en/options/general.yaml`.
- The four old `connecting_*.md` files deleted; their URLs added as aliases on the new page.
- Prerequisites section in `defining_metrics.md` rewritten to a single sentence linking to the connection guide. Readers pick their warehouse from the Cdocs filter on the destination page.
- `_index.md` simplified to a single "Connect a Data Warehouse" entry.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes

**Why the prerequisites links don't preselect a warehouse.** An earlier iteration linked from `defining_metrics.md` directly to each warehouse filter (`?database=bigquery`, `?database=databricks`, etc.). Cdocs option persistence (cookies/local storage) caused those links to land on whichever variant the reader had previously selected, not the one in the link. Even after clearing site data, the intended option did not load reliably. Switched to a single link to the connection guide so readers pick their warehouse from the filter on arrival.

**S3/CloudFront drops query strings without a trailing slash.** While the filter-link approach was in place, links written as `/path?query` were redirected to `/path/` with the query string stripped. Hugo's dev server does not reproduce this, so it only surfaced on staging. Not load-bearing now that the filter links are gone, but worth noting for future Cdocs filter linking.


iadjivon:
Editorial review card: DOCS-14168

content/en/experiments/guide/_index.md
```diff
@@ -9,14 +9,8 @@ cascade:
...
-{{< whatsnext desc="Connect warehouse data to Datadog Experiments:" >}}
-    {{< nextlink href="experiments/guide/connecting_bigquery" >}}Connect BigQuery Data Warehouse
-{{< /nextlink >}}
-    {{< nextlink href="experiments/guide/connecting_databricks" >}}Connect Databricks Data Warehouse
-{{< /nextlink >}}
-    {{< nextlink href="experiments/guide/connecting_redshift" >}}Connect Redshift Data Warehouse
-{{< /nextlink >}}
-    {{< nextlink href="experiments/guide/connecting_snowflake" >}}Connect Snowflake Data Warehouse
+{{< whatsnext desc="Warehouse Connections:" >}}
+    {{< nextlink href="experiments/guide/connecting_a_data_warehouse" >}}Connect your data warehouse for warehouse-native experiment analysis
```
jeff-morgan-dd, line 13: Minor non-blocking note but the page is called Connect **a** data warehouse (vs. "**your** data warehouse" here) - do you want these to match?

jeff-morgan-dd (commented):
The pages look good! One thing we should verify before publishing - the links from the defining_metrics page look correctly formed, but aren't going to the corresponding option.

Can you check with one of the cdocs folks whether there's a way to force links to go to a certain Cdocs option? I know there's persistence of option selection, which probably impacts linking behavior, but even when i cleared out cookies/data for the staging site, i couldn't get the links to go to the intended option. 

It might be that it's not possible, in which case we should probably link differently from the defining_metrics page.

content/en/experiments/guide/_index.md
iadjivon: I intended for it to **not** match the title exactly. But fair point to make the "a" vs "your" consistent since that's technically the only change.

iadjivon: > One thing we should verify before publishing - the links from the defining_metrics page look correctly formed, but aren't going to the corresponding option.

Oh odd! It worked while in local preview. I will loop the Cdocs folks in as you suggested!

Thanks!

content/en/experiments/guide/connecting_a_data_warehouse.mdoc.md
```diff
@@ -0,0 +1,618 @@
+---
+title: Connect a Data Warehouse for Warehouse-Native Experiment Analysis
```
iadjivon:
```suggestion
title: Connect your Data Warehouse for Warehouse-Native Experiment Analysis
```

content/en/experiments/guide/_index.md
jeff-morgan-dd: > I intended for it to **not** match the title exactly. But fair point to make the "a" vs "your" consistent since that's technically the only change.

It's fine to leave it if you had a strong intention to have it different on the landing page. What was the reason, out of curiosity?

iadjivon: Not a strong enough intention to keep them different. My thinking was that the landing page should feel more direct (your) since the reader has not yet seen the list of warehouse sources, so it's addressing them personally. The destination page then uses "a" because it introduces the list neutrally. But, "your" works great on both pages, and matches the tone of the page.

jeff-morgan-dd (approved):

content/en/experiments/guide/connecting_a_data_warehouse.mdoc.md
```diff
@@ -0,0 +1,618 @@
...
+1. In the left navigation, hover over **Settings**, then click **Experiments**.
+1. Select the **Warehouse Connections** tab.
+1. Click **Connect a data warehouse**. If you already have a warehouse connected, click **Edit** instead.
+1. Select the **Snowflake** tile.
+1. Enter the **Account**, **Role**, **Warehouse**, **Database**, and **Schema** you configured in [Step 1](#step-1-prepare-the-snowflake-service-account). If your database and schema do not appear in the dropdown, enter them manually to add them to the list.
+1. Click **Save**.
+
+{% img src="/product_analytics/experiment/guide/snowflake_experiment_setup.png" alt="The Edit Data Warehouse modal with Snowflake selected, showing two sections: Select Snowflake Account with fields for Account, Role, and Warehouse, and Select Database and Schema with fields for Database and Schema." style="width:90%;" /%}
+
+After you save your warehouse connection, create experiment metrics using your Snowflake data. See [Create Experiment Metrics][1].
```
rtrieu: suggesting this last sentence be the same as the other warehouse options
```suggestion
After you save your warehouse connection, [create experiment metrics][1] using your Snowflake data.
```

rtrieu (approved):
one minor suggestion for consistency, but otherwise looks good!

PR #36317, opened 2026-04-27
[DOCS-13132] Add Google Cloud collection scope decision support

buraizu:

Fixes DOCS-13132

Adds decision support for choosing between Google Cloud's two collection approaches (automatic project discovery vs. manual per-project setup), which are currently buried inside two setup tabs without an upfront comparison. This addresses the cluster of escalated multi-project / org-level support tickets analyzed under DOCS-13132 — customers were picking the wrong scope because the trade-offs weren't surfaced before the setup steps.

Two changes:

1. **New shortcode `layouts/shortcodes/google-cloud-collection-scope.md`** — five-line "Choose your scope" intro covering automatic project discovery (recommended) and manual per-project setup, framed by mechanism so folder-level customers find their term explicitly. Covers both metric and resource collection. Referenced from `getting_started/integrations/google_cloud.md` between `### Metric collection` and the existing VPC Service Controls alert.

2. **Heading flatten on the per-project cost and API quota attribution section** — removes the misleading plural `#### Best practices for monitoring multiple projects` H4 wrapper and promotes the inner H5 ("Enable per-project cost and API quota attribution") to H4. The existing `#enable-per-project-cost-and-api-quota-attribution` anchor is preserved (text-based slug, level-independent), so the two inbound callers (`google_cloud.md:107` and `_vendor/.../google-cloud-platform.md:231`) keep working unchanged.

A companion change in `integrations-internal-core` will apply both edits to the integration tile page (`google_cloud_platform/README.md`); that PR will be opened after this one merges so the new shortcode is available before the upstream reference goes live.

Follow-up work tracked in DOCS-14219: decision rubric (org vs. folder vs. project tier choice), project-discovery mechanism (re-scan cadence, triggers, removal behavior), and architecture diagram. SME-blocked.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes

### DIFF
--- a/layouts/shortcodes/google-cloud-metric-collection-scope.md
+++ b/layouts/shortcodes/google-cloud-metric-collection-scope.md
@@ -1,0 +1,4 @@
+Datadog supports two metric collection approaches:
+
+- **Automatic project discovery (recommended)**: Assign IAM roles at the organization or folder level. Datadog automatically discovers and monitors all current and future projects in scope. Requires the corresponding Admin role at the chosen scope.
+- **Manual per-project setup**: Assign IAM roles to specific projects. Use this approach when org- or folder-level admin access is not available, or to limit Datadog's monitoring scope.
--- a/content/en/getting_started/integrations/google_cloud.md
+++ b/content/en/getting_started/integrations/google_cloud.md
@@ -74,7 +74,7 @@ Use this guide to get started with monitoring your Google Cloud environment. Thi
 
 ### Metric collection
 
-{{% google-cloud-metric-collection-scope %}}
+{{% google-cloud-collection-scope %}}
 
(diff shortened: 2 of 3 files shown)


layouts/shortcodes/google-cloud-metric-collection-scope.md
```diff
@@ -0,0 +1,4 @@
+Datadog supports two metric collection approaches:
```
ChrisHansenDatadog: This also applies to resource collection. Might make sense to make that clear?

buraizu: Applied — broadened the lead-in to cover both metrics and resource collection, and renamed the shortcode to `google-cloud-collection-scope` (call site updated) in fcff604.

ChrisHansenDatadog (approved):

hestonhoffman (approved):

PR #36342, opened 2026-04-29
[DBM] Add work_mem troubleshooting section for Postgres

pierreln-dd:

Adds a new subsection under "Queries from Agent are slow and/or have a high impact on the database" in the Postgres DBM troubleshooting doc.

Covers the case where DBM collection queries spill to temporary disk files because the `work_mem` for the `datadog` user is too low. This manifests as elevated write IOPS on the database host, gradually worsening after DBM is enabled and temporarily resolving after a database restart.

The new section provides:
- A Metrics Explorer query to confirm the issue (`postgresql.queries.temp_blks_written` filtered by `user:datadog`)
- A SQL estimation query to size the correct `work_mem` value based on actual schema and `pg_stat_statements` state
- The `ALTER ROLE` command to apply the fix without a database restart

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes

New section is inserted after the existing `#### High value for pg_stat_statements.max` subsection.


pierreln-dd:
Consensus from PR discussion in Slack:
`work_mem` more of a mitigation that hides the issue than a real fix, and we should overall be conservative
Slow queries are preferable over OOM or cache eviction from memory pressure

PR #36385, opened 2026-04-30
Fix incorrect RUM attribute names in standard attributes page

ncreated:

Fixes 4 incorrect RUM attribute entries in the standard attributes page. Each was verified against the [`rum-events-format`](https://github.com/DataDog/rum-events-format) JSON schemas and SDK source code:

- **`user.id` removed, `usr.id` expanded to all platforms** — All SDKs use `usr.id`; the `user.id` entry was incorrect. The existing `usr.id` entry was updated to include Android and Roku.
- **`error.issue_id` removed, `error.id` added** — `error.issue_id` doesn't exist in any SDK or schema. The correct attribute is `error.id` (consistent with `resource.id` and `action.id`), which was missing from the page.
- **`action.name` removed, `action.target.name` description updated** — `action.name` doesn't exist in the schema or any SDK event payload. The description of `action.target.name` was updated to cover both automatically collected and custom actions.
- **`action.frustration.type:rage_click/dead_click/error_click` replaced with a single `action.frustration.type` entry** — The `attribute:value` naming convention used here is inconsistent with the rest of the page. Replaced with one entry listing the possible values in the description, following the same pattern as other enum attributes (e.g. `resource.provider.type`).

### Merge instructions

Merge readiness:
- [x] Ready for merge

### AI assistance

Used Claude Code to cross-reference attribute names against `rum-events-format` schemas and SDK source code (browser, iOS, Android).

### DIFF
--- a/content/en/standard-attributes/_index.md
+++ b/content/en/standard-attributes/_index.md
@@ -412,6 +412,15 @@ attributes:
     type: string
     domain: RUM core attributes
 
+  - name: application.name
+    description: The Datadog application name.
+    product_source:
+      - icon-rum
+      - android
+      - browser
+      - ios
+    type: string
+    domain: RUM core attributes
 
   - name: device.type
     description: The device type as reported by the device (System User-Agent).
--- a/content/en/standard-attributes/_index.md
+++ b/content/en/standard-attributes/_index.md
@@ -412,15 +412,6 @@ attributes:
     type: string
     domain: RUM core attributes
 
-  - name: application.name
-    description: The Datadog application name.
-    product_source:
-      - icon-rum
-      - android
-      - browser
-      - ios
-    type: string
-    domain: RUM core attributes
 
   - name: device.type
     description: The device type as reported by the device (System User-Agent).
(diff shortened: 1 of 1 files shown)


content/en/standard-attributes/_index.md
```diff
@@ -412,15 +412,6 @@ attributes:
...
 
-  - name: application.name
-    description: The Datadog application name.
-    product_source:
-      - icon-rum
-      - android
-      - browser
-      - ios
-    type: string
-    domain: RUM core attributes
```
bcaudan, line 423: **💬 suggestion:** ‏This one seems to be added by the backend though.
As we are listing other computed fields like device, os, geo maybe we could keep it 🤷‍♂️

bcaudan (approved):

content/en/standard-attributes/_index.md
ncreated: Fair point 👌 - changed, keeping `application.name` 👍.

content/en/standard-attributes/_index.md
```diff
@@ -963,18 +957,8 @@ attributes:
...
-      - icon-rum
-      - android
-      - browser
-      - ios
-      - roku
-    type: string
-    domain: Action (Browser events, Android events, iOS events, Roku events)
   - name: action.target.name
-    description: Element that the user interacted with. Only for automatically collected actions.
+    description: The name of the element the user interacted with (for automatically collected actions) or the name provided in the API call (for custom actions, for example `Click on checkout`).
```
estherk15:
```suggestion
    description: For automatically collected actions, the name of the element the user interacted with. For custom actions, the name provided in the API call (for example, `Click on checkout`).
```

estherk15 (approved):
Thanks for the update, left one suggestion on wording, but this is approved!

plousada (approved):

PR #36386, opened 2026-04-30
[DBM][Oracle] Add Multi-tenant/Non-CDB tabs to RAC setup page

HassData:
## What

Added Multi-tenant, Non-CDB, and Oracle 11 tab structure to the Oracle RAC setup page, matching the pattern already used on the self-hosted Oracle setup page. The `Create the Datadog user` section now shows the correct `CREATE USER` statement per deployment type, and the `Configure the Agent` section now shows the correct `username` value (`c##datadog` for CDB, `datadog` for Non-CDB/Oracle 11) in each tab. A note ties the two sections together: "the `username` value must match the user created above."

## Why

The RAC page previously used a generic shortcode for user creation while hardcoding `username: 'c##datadog'` in the YAML example. Users on non-CDB or Oracle 11 RAC deployments would create a `datadog` user but copy a config with `c##datadog`, causing authentication failures. Making the username choice explicit per deployment type prevents this mismatch.

## Docs preview

https://docs.datadoghq.com/database_monitoring/setup_oracle/rac/

### DIFF
--- a/content/en/database_monitoring/setup_oracle/rac.md
+++ b/content/en/database_monitoring/setup_oracle/rac.md
@@ -40,7 +40,27 @@ Complete the following to enable Database Monitoring with your Oracle database:
 
 ### Create the Datadog user
 
-{{% dbm-create-oracle-user %}}
+{{< tabs >}}
+{{% tab "Multi-tenant" %}}
+```SQL
+CREATE USER c##datadog IDENTIFIED BY &password CONTAINER = ALL ;
+
+ALTER USER c##datadog SET CONTAINER_DATA=ALL CONTAINER=CURRENT;
+```
+{{% /tab %}}
+
+{{% tab "Non-CDB" %}}
+```SQL
+CREATE USER datadog IDENTIFIED BY &password ;
+```
+{{% /tab %}}
+
+{{% tab "Oracle 11" %}}
+```SQL
+CREATE USER datadog IDENTIFIED BY &password ;
+```
+{{% /tab %}}
+{{< /tabs >}}
 
 ### Securely store your password
 {{% dbm-secret %}}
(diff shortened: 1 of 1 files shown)


content/en/database_monitoring/setup_oracle/rac.md
```diff
@@ -53,10 +73,12 @@ For installation steps, see the [Agent installation instructions][9].
 
 ### Configure the Agent
 
-Configure the Agent for each RAC node by following the instructions for [self-hosted Oracle databases][3].
+Configure the Agent for each RAC node. Use the tab that matches your Oracle deployment type (the `username` value must match the user created above).
```
estherk15: Removing directional language to match our docs style guide:

```suggestion
Configure the Agent for each RAC node. Use the tab that matches your Oracle deployment type. The `username` value must match the user you created previously.
```

content/en/database_monitoring/setup_oracle/rac.md
```diff
@@ -40,7 +40,27 @@ Complete the following to enable Database Monitoring with your Oracle database:
 
 ### Create the Datadog user
 
-{{% dbm-create-oracle-user %}}
```
estherk15, line 43: To confirm you want to drop this text:

> If you already have the legacy Oracle integration installed, the user already exists, and you can skip this step. 
>
> Create a read-only login to connect to your server and grant the required permissions:

content/en/database_monitoring/setup_oracle/rac.md
estherk15: This was flagged during review, there isn't a section on this page for this anchor link.

```suggestion
1. Install or verify the Oracle integration
```

estherk15 (commented):
@HassData Thanks for the update, I left a question on the removal of the shortcode text and suggestions on wording. Let me know if you have any questions!

HassData:
The `#install-or-verify-the-oracle-integration` anchor link in the setup checklist was already present in master before this PR — neither introduced nor removed by this change. The same broken anchor exists on the self-hosted page as well. Leaving it out of scope for this PR to keep the change focused.

content/en/database_monitoring/setup_oracle/rac.md
HassData: we can mark this as resolved but if you'd like to fix, or suggest a fix I am happy to do it

HassData:
@estherk15 Thank you so much for your detailed review here and help! Please let me know if anything else is needed :)

content/en/database_monitoring/setup_oracle/rac.md
pierreln-dd: I think we're far enough along that most customers don't have the legacy integration anymore, and even if they do, re-creating the user will not be a failed command, so removing that first line is probably not a problem:

> If you already have the legacy Oracle integration installed, the user already exists, and you can skip this step.

HassData: Good call — agreed. Removed in fb3f95ed.

HassData:
I rebased so the new preview is shown, seems to worked as expected thankfully

estherk15 (approved):

PR #36400, opened 2026-04-30
Add Hex integration doc to Data Observability BI section

heron--:
The flag for this integration was just recently enabled for all orgs, so we're ready to move this into Data Observability. This PR creates a copy of the Hex integration page from the integrations config.

### Merge instructions
No blockers for merge

Merge readiness:
- [x] Ready for merge

### AI assistance
Claude was used to copy the files and for Git operations, but was not used for writing (at least in this PR. I can't speak to its source)

Documentation Editorial Review: https://datadoghq.atlassian.net/browse/DOCS-14278


jeff-morgan-dd:
Hi Devin @heron-- ! Thanks very much for the PR. Following up on the Slack conversation - I'm fine to approve this PR, but i think overall the better solution going forward is to reference/link to the integration in the app and treat that as the source of truth (Which i think is where your Slack conversation landed or was trending towards).

From a consistency standpoint, it's OK to break the cycle now and just link to the Hex integration. User impact is minimal since they're still landing on setup instructions and probably won't care that much exactly where they land. We could then non-urgently update the previous integration links in a separate PR. But if you want to go forward with this PR as-is, let me know.

Thoughts?

jeff-morgan-dd (approved):

PR #36476, opened 2026-05-05
Fix Java feature flags doc: API key belongs on Agent, not application

aarsilv:

Two fixes to `content/en/feature_flags/server/java.md`:

**1. `DD_API_KEY` does not belong on the application.**

The page lists `DD_API_KEY` as a required application-level environment variable. It is not — the Java SDK reaches Datadog only through the local Agent for feature flagging:
- Flag config flows over the standard Agent-mediated Remote Configuration poller.
- Exposures route through the Agent's EVP proxy. `EVENT_PLATFORM` in [`Intake.java`](https://github.com/DataDog/dd-trace-java/blob/master/internal-api/src/main/java/datadog/trace/api/intake/Intake.java) is constructed with agentless mode hardcoded to `false`, so the API-key code path in `BackendApiFactory` is unreachable.
- Evaluation metrics export to the Agent's OTLP receiver at `http://localhost:4318` by default ([dd-openfeature README](https://github.com/DataDog/dd-trace-java/blob/master/products/feature-flagging/feature-flagging-api/README.md)).

Removed `DD_API_KEY` / `-Ddd.api.key` from the application env vars and system properties tabs, clarified the compatibility requirement that the API key goes on the Agent, and added an Agent-side prerequisite to the troubleshooting checklist (so the key still appears, in the right place).

**2. The provider is no longer experimental — only the env var name is.**

The page reads as if Java Feature Flags is still an experimental feature. It isn't; the `EXPERIMENTAL_` prefix on `DD_EXPERIMENTAL_FLAGGING_PROVIDER_ENABLED` is kept for backwards compatibility. Updated the top-of-page banner from a yellow warning to an info note that explains the prefix is historical, and dropped "experimental" from descriptive prose throughout (env-var comments, the auto-start sentence, the red alert, and the troubleshooting prereq). The actual variable and system property names are unchanged.

Translated content (`content/es/...`) is not edited per the contributing guidelines; it will re-sync from the English source.

### Merge instructions

Merge readiness:
- [x] Ready for merge

### Additional notes

### DIFF
--- a/content/en/feature_flags/server/java.md
+++ b/content/en/feature_flags/server/java.md
@@ -10,7 +10,7 @@ further_reading:
   text: "Java APM and Distributed Tracing"
 ---
 
-<div class="alert alert-info">The Java Feature Flags provider is enabled by setting <code>DD_EXPERIMENTAL_FLAGGING_PROVIDER_ENABLED=true</code>. The <code>EXPERIMENTAL_</code> prefix in the variable name is preserved for backwards compatibility — the provider itself is no longer experimental. See the <a href="#configuration">Configuration section</a> for details.</div>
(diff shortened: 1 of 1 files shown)


typotter (approved):

brett0000FF (approved):
Thanks, looks good!

PR #36579, opened 2026-05-11
Fix grace period docs: scope Disabled condition and add fix-present case

TovRudyy:
## Summary

- Adds the missing `fix not present` condition to the Disabled bullet in the grace period section (was incorrectly implying the test stays Disabled even when the fix is present in the commit)
- Adds an explicit bullet clarifying that when the fix IS present, Datadog treats the test as healthy regardless of previous status

This matches the actual behavior in `processFixedTests` (`libraries.go`): when `IsAncestor` returns `true`, the test is skipped and not returned to the client (treated as healthy) for all previous statuses.

## Test plan

- [ ] Review rendered markdown on the docs preview


TovRudyy:
Closing — redoing as a clean branch off latest master

PR #36591, opened 2026-05-11
[DOCS-10268] Document when RUM Browser SDK collects each Core Web Vital

rtrieu:

Addresses user feedback by adding a "When each core web vital is collected" reference and a troubleshooting subsection to the RUM browser performance page, so users have a place to land for "why is INP/LCP/CLS/FCP missing from my view?" questions.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes

### DIFF
--- a/content/en/real_user_monitoring/application_monitoring/browser/monitoring_page_performance.md
+++ b/content/en/real_user_monitoring/application_monitoring/browser/monitoring_page_performance.md
@@ -51,7 +51,7 @@ You can access performance telemetry for your views in:
 
 {{< img src="real_user_monitoring/browser/core-web-vitals-1.png" alt="Core Web Vitals summary visualization" >}}
 
-- Interaction to Next Paint and Largest Contentful Paint are not collected for pages opened in the background (for example, in a new tab or a window without focus).
+- LCP and FCP are not collected for pages that start hidden (for example, opened in a background tab). For more on when each vital is reported and how to troubleshoot a missing value, see [When each Core Web Vital is collected](#when-each-core-web-vital-is-collected).
 - Telemetry collected from your real users' pageviews may differ from those calculated for pages loaded in a fixed, controlled environment such as a [Synthetic browser test][6]. Synthetic Monitoring displays Largest Contentful Paint and Cumulative Layout Shift as lab telemetry, not real telemetry.
 
 | Datapoint                   | Focus            | Description                                                                                           | Target value |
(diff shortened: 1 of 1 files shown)


hestonhoffman (approved):

content/en/real_user_monitoring/application_monitoring/browser/monitoring_page_performance.md
```diff
@@ -70,6 +70,34 @@ RUM reports the element that is associated with each Core Web Vital instance:
 - For First Input Delay, RUM reports the CSS selector of the first element the user interacted with.
 - For Cumulative Layout Shift, RUM reports the CSS selector of the most shifted element contributing to the CLS.
 
+### When each core web vital is collected
+
+The RUM Browser SDK reports each Core Web Vital under different conditions. If a vital is missing from a view event, the SDK most likely did not capture a value for it.
+
+- **Largest Contentful Paint**: Reported on the initial view only. Captured when the `largest-contentful-paint` PerformanceObserver entry fires before the page is hidden and before the user's first interaction. Not reported if the page starts hidden, if the user interacts before any LCP entry arrives, or after a 10-minute cap.
```
allspain: Both caveats here, start-hidden and first-interaction, come straight from the LCP spec rather than anything Datadog-specific. Worth pointing at the upstream source so anyone comparing our numbers to Lighthouse can verify we're doing the same thing.

```suggestion
- **Largest Contentful Paint**: Reported on the initial view only. Captured when the `largest-contentful-paint` PerformanceObserver entry fires before the page is hidden and before the user's first interaction. Not reported if the page starts hidden, if the user interacts before any LCP entry arrives, or after a 10-minute cap. See [LCP: Differences between the metric and the API](https://web.dev/articles/lcp#differences-metric-api) for the upstream definition.
```

content/en/real_user_monitoring/application_monitoring/browser/monitoring_page_performance.md
```diff
@@ -70,6 +70,34 @@ RUM reports the element that is associated with each Core Web Vital instance:
...
+- **Interaction to Next Paint**: Reported on every view, including SPA route changes. Requires at least one user interaction during the view's lifetime. If the user never interacts, no INP value is emitted. Requires browser support for the `event` PerformanceObserver entry type and `PerformanceEventTiming.interactionId`. If unsupported, no INP value is emitted.
```
allspain: The interaction requirement is the INP definition itself, and the `interactionId` part lives in the web-vitals support table. A link helps if anyone wonders whether "no interaction → no INP" is our choice or the metric's.

```suggestion
- **Interaction to Next Paint**: Reported on every view, including SPA route changes. Requires at least one user interaction during the view's lifetime (see [INP on web.dev](https://web.dev/articles/inp)). If the user never interacts, no INP value is emitted. Requires browser support for the `event` PerformanceObserver entry type and `PerformanceEventTiming.interactionId`. If unsupported, no INP value is emitted.
```

content/en/real_user_monitoring/application_monitoring/browser/monitoring_page_performance.md
```diff
@@ -70,6 +70,34 @@ RUM reports the element that is associated with each Core Web Vital instance:
...
+- **Cumulative Layout Shift**: Reported on every view, including SPA route changes. Requires browser support for the `layout-shift` PerformanceObserver entry type and `WeakRef`. If unsupported, no CLS value is emitted.
```
allspain: Small thing worth keeping separate: `layout-shift` is the spec API, but `WeakRef` is an SDK implementation detail. As written they read as equivalent requirements, which could leave a reader thinking Google's CLS definition involves `WeakRef`. Splitting them keeps the spec/Datadog line clean.

```suggestion
- **Cumulative Layout Shift**: Reported on every view, including SPA route changes. Requires browser support for the [`layout-shift` PerformanceObserver entry type](https://web.dev/articles/cls). The SDK also relies on `WeakRef` internally; if either is missing, no CLS value is emitted.
```

allspain (commented):
Doc reads accurately already, so these are suggestions rather than fixes. The thread I'd pull on: since Core Web Vitals are Google's spec, linking each caveat back to the canonical web.dev definition shifts the framing from "here's what our SDK does" to "here's the spec, here's where we conform." That's a stronger answer for anyone comparing our numbers to Lighthouse or PageSpeed Insights and trying to figure out whether a difference is a Datadog quirk or expected spec behavior. Most of the inline notes are just spec-link nudges; a couple separate spec requirements from SDK implementation details (`WeakRef`, the 10-minute cap) so readers don't attribute those to Google.

content/en/real_user_monitoring/application_monitoring/browser/monitoring_page_performance.md
```diff
@@ -70,6 +70,34 @@ RUM reports the element that is associated with each Core Web Vital instance:
...
+- **First Contentful Paint**: Reported on the initial view only. Captured when the entry fires before the page is hidden, within 10 minutes of view start.
```
allspain: Same split as the CLS note: initial-view-only is FCP spec behavior, the 10-minute cap is ours. Reads cleaner if those are attributed separately.

```suggestion
- **First Contentful Paint**: Reported on the initial view only (see [FCP on web.dev](https://web.dev/articles/fcp)). Captured when the entry fires before the page is hidden. The SDK applies a 10-minute ceiling per view.
```

content/en/real_user_monitoring/application_monitoring/browser/monitoring_page_performance.md
```diff
@@ -70,6 +70,34 @@ RUM reports the element that is associated with each Core Web Vital instance:
 - For First Input Delay, RUM reports the CSS selector of the first element the user interacted with.
 - For Cumulative Layout Shift, RUM reports the CSS selector of the most shifted element contributing to the CLS.
 
+### When each core web vital is collected
+
+The RUM Browser SDK reports each Core Web Vital under different conditions. If a vital is missing from a view event, the SDK most likely did not capture a value for it.
```
allspain, line 75: One thing this section could call out more directly: the per-view reset on CLS and INP for SPA route changes is ahead of the stable spec. It tracks the direction of Chrome's [Soft Navigations API experiment](https://developer.chrome.com/docs/web-platform/soft-navigations-experiment), while LCP and FCP stay on the stable definition (initial view only). As written, a reader could miss that the SPA behavior is intentional. A sentence here would make it land as a deliberate feature.

allspain (approved):

content/en/real_user_monitoring/application_monitoring/browser/monitoring_page_performance.md
rtrieu, line 75:
```suggestion
The RUM Browser SDK reports each Core Web Vital under different conditions. If a vital is missing from a view event, the SDK most likely did not capture a value for it.

**Note**: Note: Re-reporting CLS and INP on each view, including SPA route changes, is intentional. It follows Chrome's [Soft Navigations API experiment][#], which extends these metrics beyond the initial page load. LCP and FCP follow the stable definition and are reported on the initial view only.
```

rtrieu: what do you think of the suggestion below?

rtrieu: would need to add link https://developer.chrome.com/docs/web-platform/soft-navigations at the bottom

PR #36627, opened 2026-05-12
Add static defaults to region-param shortcode for OTel LLM Obs docs

ZStriker19:

The `region-param` shortcode renders an empty `<code>` (or `<span>`) element that is populated client-side by JavaScript based on the visitor's region selection. Readers that don't execute JavaScript — LLMs, AI assistants, search bots — see an empty string, which makes code samples like the OTLP trace endpoint configuration look broken or ambiguous.

This PR:

1. Extends `layouts/partials/region-param.html` so the existing `text=` parameter is used as the element's inner content for the `code` and `span` variants (it was already used for the `link` variant). The change is backward-compatible — usages without `text=` continue to render empty elements.
2. Adds `text="https://otlp.datadoghq.com/v1/traces"` to every `region-param` call in `content/en/llm_observability/instrumentation/otel_instrumentation.md` so the OTLP endpoint renders with a sensible US default in static HTML. JavaScript still swaps in the correct region-specific value for users with JS enabled.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes

Claude session: \`e51e8def-07e3-46ff-a673-98ef4988878b\`
Resume: \`claude --resume e51e8def-07e3-46ff-a673-98ef4988878b\`

### DIFF
--- a/content/en/llm_observability/instrumentation/otel_instrumentation.md
+++ b/content/en/llm_observability/instrumentation/otel_instrumentation.md
@@ -28,7 +28,7 @@ Set the following environment variables in your application:
 
 ```
 OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=http/protobuf
-OTEL_EXPORTER_OTLP_TRACES_ENDPOINT={{< region-param key="otlp_trace_endpoint" code="true" >}}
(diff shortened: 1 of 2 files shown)


jeff-morgan-dd (approved):
LGTM, thank you!

PR #36657, opened 2026-05-13
Live Debugger Go: support log templates and expressions

tylfin:

The Go SDK now supports log templates and condition expressions for Live Debugger / Dynamic Instrumentation. This PR moves that capability from the Unsupported to the Supported features list on the Go enablement page.

### Merge instructions

Merge readiness:
- [x] Ready for merge

### Additional notes


grantseltzer (approved):
You're absolutely right - good catch!

maycmlee (approved):

PR #36667, opened 2026-05-13
Add Monitors and Automation page to DSM Kafka docs

shellymat-dd:

Adds a new page, **Monitors and Automation**, under `data_streams/kafka/` so customers landing on the Kafka docs can find:

- The cluster- and topic-level monitor templates that Data Streams Monitoring ships out of the box (with the exact metrics, scopes, and default thresholds the product uses).
- Concrete examples of using webhooks to automate a response when a Kafka monitor triggers — consumer lag rising, lag approaching retention (data-loss risk), broker disk exhaustion, and offline/under-replicated partitions.

The monitor template details (titles, descriptions, metrics, default thresholds) are taken from the in-product templates defined in `web-ui` (`getClusterRecommendedMonitors` / `getTopicRecommendedMonitors`), so the docs match what users see in the {{< ui >}}Monitors{{< /ui >}} side panel on a cluster or topic page.

Also adds a `weight` field to `setup.md` and the new page so the left nav reads **Setup → Monitors and Automation**, and adds a pointer to the new page from the Kafka section's `_index.md`.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### AI assistance

Drafted with Claude Code. The monitor templates section was rewritten against the source of truth in `web-ui/packages/apps/data-streams/private/runtime/kafka/components/KafkaMonitorsSidePanel/kafka-monitors-side-panel.utils.ts` to make sure metric names, scopes, and the 80% retention threshold match what's actually shipped. Webhook examples were authored with Shelly's input on which "when" conditions to cover.

### Additional notes

- The **Consumer lag is approaching bytes retention limit** template is conditional in-product (only shown when `hasKafkaBrokerMetrics` is true). The doc notes this requirement inline rather than omitting the template.
- No screenshots in this first pass — happy to add a screenshot of the Monitor Templates side panel in a follow-up if reviewers want one.


maycmlee:
Created https://datadoghq.atlassian.net/browse/DOCS-14396 for docs review.

content/en/data_streams/kafka/monitors_and_automation.md
```diff
@@ -0,0 +1,94 @@
...
+
+Signals that unread messages are approaching the topic's retention window. If lag exceeds retention, those messages get deleted before the consumer can read them.
+
+**Potential action:** Trigger an emergency runbook that can temporarily extend retention on the affected topic, pause the upstream producer, or scale the consumer group ahead of the threshold.
+
+### Broker disk filling up
+
+Signals that a broker host is running low on disk space. If the disk fills up, the broker goes offline and its partitions become unavailable.
+
+**Potential action:** Trigger a capacity workflow to add storage, expand the cluster, or compact a candidate topic.
```
piochelepiotr: instead of compact a candidate topic, reduce retention on a candidate topic

content/en/data_streams/kafka/monitors_and_automation.md
```diff
@@ -0,0 +1,94 @@
...
+| Consumer lag is high for topic                             | `kafka.estimated_consumer_lag`                                          | Consumer lag exceeds a threshold for a topic and consumer group. |
+| Incoming message rate has dropped                          | `kafka.topic.message_rate`                                              | Produce rate to the topic drops below a threshold. Catches silent producer failures. |
+| Offline partitions on topic                                | `kafka.partition.offline`                                               | Any partition for this specific topic goes offline, indicating data unavailability for that topic. |
+| Consumer lag is approaching time retention limit           | `kafka.estimated_consumer_lag` / `kafka.topic.config.retention_ms`      | Estimated lag approaches the topic's time-based retention. Beyond the retention limit, the consumer cannot recover lost data. |
+| Consumer lag is approaching bytes retention limit          | `kafka.consumer_lag` × throughput / `kafka.topic.config.retention_bytes` | Estimated lag approaches the topic's bytes-based retention. Requires Kafka broker metrics to be available. |
```
piochelepiotr: Let's link to the broker integration? https://docs.datadoghq.com/opentelemetry/integrations/kafka_metrics/?tab=host

piochelepiotr (approved):
Looks good to me. I see we are pretty in depth on technologies not owned by the DSM team (like webhooks, workflows). The downside is that these docs will run out of date if the webhook / workflow docs get updated. Also, my guess is that most users only want to get paged when an issue happens, and not take automatic actions.

content/en/data_streams/kafka/monitors_and_automation.md
```diff
@@ -0,0 +1,94 @@
...
+| Template                                | Metric                              | Condition                                  |
+|-----------------------------------------|-------------------------------------|--------------------------------------------|
+| Offline partitions detected             | `kafka.partition.offline`           | Any partition in the cluster is offline.   |
+| Under-replicated partitions detected    | `kafka.partition.under_replicated`  | Any partition in the cluster is under-replicated, which puts data at risk if a broker fails. |
+
+Both monitors are grouped by `kafka_cluster_id` so each cluster alerts its own owner.
+
+### Topic-level templates
+
+| Template                                                   | Metric                                                                  | Condition |
```
johannbotha: It feels weird to have "recommended templates" without having a section that clearly lays out what additional metrics you get with DSM's KA. 

For example, what is the difference between underreplicated partitions here https://docs.datadoghq.com/opentelemetry/integrations/kafka_metrics/?tab=host vs what we have?

content/en/data_streams/kafka/monitors_and_automation.md
```diff
@@ -0,0 +1,94 @@
...
+|------------------------------------------------------------|-------------------------------------------------------------------------|-----------|
+| Consumer lag is high for topic                             | `kafka.estimated_consumer_lag`                                          | Consumer lag exceeds a threshold for a topic and consumer group. |
```
johannbotha: The important part of this metric is it's estimating in seconds and not offsets.

johannbotha (commented):
I like that we're suggesting what monitors to have at each level. As a user I'd still be confused to understand what metrics I get with KA and what is free with DD. I suggest adding a metrics section under `kafka` that lets you know all the metrics you get in addition to the free metrics. 

Then we can easily link to those in our recommended monitors section.

content/en/data_streams/kafka/monitors_and_automation.md
shellymat-dd: There may not be a difference other than how it's being collected - often the challenge is setting up the JMX integration. We can add a metrics section under Kafka in a separate PR, I don't think it should be blocking

shellymat-dd: The goal is to show what is possible with our product, not to focus on the nuances or differentiators.

content/en/data_streams/kafka/monitors_and_automation.md
shellymat-dd: Added a "description" column to explain the impact so we can highlight it's in seconds. Want to take a look at all descriptions? Preview is still WIP but you should be able to see the code change easily

content/en/data_streams/kafka/monitors_and_automation.md
johannbotha: I do see that it's helpful to reflect he monitors that we suggest. As a follow on, we should make it explicit what additional metrics you get from KA.

content/en/data_streams/kafka/monitors_and_automation.md
johannbotha: Yes, that makes it clearer that this one is in seconds.

johannbotha (approved):

cswatt:
Pushed some small style-guide edits! I actually removed the `weight` field on the pages — the `weight`s are already being declared in `main.en.yaml`, so having them in the frontmatter is redundant.

cswatt (approved):

PR #36672, opened 2026-05-13
[DOCS-14086] Centralize site-support banners for 9 integrations

hestonhoffman:
Moves site-support banners from inline integration README blocks to centralized `unsupported_sites` entries in `params.yaml`. The docs build system renders the banner automatically at build time, so inline blocks in integration source repos are redundant.

## Companion PRs (remove inline banners from source repos)

- DataDog/integrations-internal-core#3218
- DataDog/integrations-extras#2992
- DataDog/integrations-core#23604

## Merge instructions

This should be merged just before we merge the integrations PRs to make sure there isn't a large lag where the pages have double banners.


janine-c (approved):
:shipit:

PR #36685, opened 2026-05-14
Add query optimization guidance to Advanced Queries Analysis

vitalii-dd:
## Summary

Adds a new **Writing efficient queries** section to the Notebooks Advanced Analysis docs page. The section walks customers through eight common query antipatterns we observe in Trino failures, each with a short Before / After SQL rewrite.

Editorial choices:

- Tone is suggestive, not prescriptive (no absolutes like _"never longer than 7 days"_).
- No internal implementation details — CPU/memory caps, scan limits, and error codes are not exposed.
- Examples use generic vocabulary (`service:checkout-api`, `logs`, `traces`, `rum`); no real customer query is recognizable.
- Section fits between **Viewing and exporting data** and **Further reading**.

## Why

The Notebooks Trino query engine has been absorbing a steady stream of failing customer queries — most of them avoidable by following a small set of query-authoring patterns. The same handful of antipatterns recurs across every investigation we have done.

Recent telemetry:

- **May 14 extract**: 43,569 failed Trino queries — 41,667 CPU-limit, 1,902 memory-limit — across 77 customer orgs. **36 queries caused 80%** of the total failed CPU.
- **Apr 6 – Apr 9 spike window**: four separate hourly spike investigations recorded ~1,000–1,900 failures each. The same notebooks reappeared in 3 of 4 (and 2 of them in all 4) consecutive spikes, indicating customers have no clear self-service guidance for fixing the failures.

A docs reference with concrete Before / After examples should cut the recurring failure rate by giving customers something to consult when a query is slow or canceled.

## Investigations behind this draft

- [Trino Performance Investigation — Notebooks Query Source (Jan–Apr 2026)](https://app.datadoghq.com/notebook/14325701)
- [Spike investigation — Apr 6, 2026](https://app.datadoghq.com/notebook/14335217)
- [Spike investigation — Apr 7, 2026](https://app.datadoghq.com/notebook/14335405)
- [Spike investigation — Apr 8, 2026](https://app.datadoghq.com/notebook/14335479)
- [Spike investigation — Apr 9, 2026](https://app.datadoghq.com/notebook/14325748)

## Test plan

- [ ] `hugo serve` — preview page locally and confirm rendering
- [ ] Verify all internal `[N]` link references still resolve
- [ ] Confirm no internal limits, error codes, or customer fingerprints appear in the rendered output
- [ ] Notebooks PM / TPM sign-off
- [ ] Docs writer pass for house-style consistency


buraizu:
Thanks for the PR, created DOCS-14413 for documentation team review

content/en/notebooks/advanced_analysis/_index.md
```diff
@@ -206,6 +206,217 @@ You can view, search, edit, and delete existing report schedules from the **Repo
 
 **Note**: To schedule reports and view other users' schedules, users need the **CSV Report Schedules Write** permission. To edit other users' schedules, users need the **CSV Report Schedules Manage** permission. These permissions can be granted by a user with the **Org Management** permission.
 
+## Writing efficient queries
```
shivrams: Since this advice applies to all advanced analysis queries and products, is there a better home for it than the notebooks -->  advanced analysis feature page? Curious for documentation team input here

content/en/notebooks/advanced_analysis/_index.md
```diff
@@ -206,6 +206,217 @@ You can view, search, edit, and delete existing report schedules from the **Repo
...
+SELECT *
+FROM logs
+WHERE service = 'checkout-api'
+  AND env     = 'prod'
+  AND timestamp >= NOW() - INTERVAL '1 day';
+```
+
+### Choose a time range that fits the question
+
+Scan cost grows with the time range. Consider starting with a window that covers the question (often a few hours or a day) and widening only when the question demands more. For long-term trends, a pre-aggregated metric or a daily-summary cell is usually cheaper than re-scanning raw events on every run.
```
shivrams: The challenge here is notebooks time ranges are baked into the time picker, and authors/readers change that generously. 

Maybe we encourage the author to set a default narrow time range for computational notebooks? It avoids users trying different large time frames to understand a narrative

content/en/notebooks/advanced_analysis/_index.md
```diff
@@ -206,6 +206,217 @@ You can view, search, edit, and delete existing report schedules from the **Repo
...
+```
+
+**After**
+
+```sql
+SELECT *
+FROM logs
+WHERE service = 'checkout-api'
+  AND env     = 'prod'
+  AND timestamp >= NOW() - INTERVAL '1 day';
```
shivrams: Let's leave this as 7 day to illustrate to the customer they can achieve their intended result with better filters (and not necessarily shorten time range)

content/en/notebooks/advanced_analysis/_index.md
```diff
@@ -206,6 +206,217 @@ You can view, search, edit, and delete existing report schedules from the **Repo
...
+```
+
+### Choose a time range that fits the question
+
+Scan cost grows with the time range. Consider starting with a window that covers the question (often a few hours or a day) and widening only when the question demands more. For long-term trends, a pre-aggregated metric or a daily-summary cell is usually cheaper than re-scanning raw events on every run.
+
+**Before**
```
shivrams: To help land the before vs after better, could we add one line about what the author is trying to understand. eg: "I want to understand error log volume in my services over the last 30 days", and show an example of an inefficient and efficient query for the same.

content/en/notebooks/advanced_analysis/_index.md
```diff
@@ -206,6 +206,217 @@ You can view, search, edit, and delete existing report schedules from the **Repo
...
+
+Asking for millions of raw rows isn't free—the engine has to buffer them in memory, transmit them over the network, and render them in your notebook, even if the underlying scan is bounded. If your goal is to understand the data (top-N, counts per category, distributions), a `GROUP BY` returns a small summary that costs far less and is easier to read.
+
+**Before**
+
+```sql
+-- Returns up to 5,000,000 raw rows
+SELECT *
+FROM logs WHERE service = 'orders-api'
+LIMIT 5000000;
```
shivrams: A customer might have this cell without the limit, and have a top list visualization cell that aggregates by status_code below it. Is doing the aggregation in the SQL query more efficient than doing it in the visualization?

content/en/notebooks/advanced_analysis/_index.md
```diff
@@ -206,6 +206,217 @@ You can view, search, edit, and delete existing report schedules from the **Repo
...
+WHERE timestamp >= NOW() - INTERVAL '7 days';    -- no service filter, unbounded set
+```
+
+**After**
+
+```sql
+SELECT DISTINCT user_email
+FROM logs
+WHERE service = 'checkout-api'
+  AND timestamp >= NOW() - INTERVAL '1 day';     -- same DISTINCT, scoped scan
```
shivrams: Instead of changing the interval here, it might be more useful to keep the time window the same and do a pre-aggregation for per day user-email, and then do a distinct across days?

The author starts with a goal of "what are the distinct emails in this service's logs over the last 7 days" - we cannot tell them to change the query to 1 day to make it more efficient.

content/en/notebooks/advanced_analysis/_index.md
```diff
@@ -206,6 +206,217 @@ You can view, search, edit, and delete existing report schedules from the **Repo
...
+```
+
+### Checklist
+
+Walk through this before re-running a heavy query:
```
shivrams: Let's move this up top?

content/en/notebooks/advanced_analysis/_index.md
```diff
@@ -206,6 +206,217 @@ You can view, search, edit, and delete existing report schedules from the **Repo
...
+SELECT status_code, count(*) AS hits
+FROM logs WHERE service = 'orders-api'
+GROUP BY status_code
+ORDER BY hits DESC
+LIMIT 100;
+```
+
+### Combine self-joins into a single scan
+
+Joining one source to itself many times to correlate different events is one of the most common causes of slow notebook queries. Most self-joins can be rewritten as a single scan with `CASE` expressions, window functions, or `GROUP BY ... HAVING`. When you do need a `JOIN`, prefer a high-cardinality key (`user_id`, `request_id`, `trace_id`)—coarse keys like `service` or `status` can expand a moderate input into billions of intermediate rows. The rewrite reduces structural overhead; for a large dataset, also narrow the data source filter and time range.
```
shivrams: I think the high cardinality join recommendation could become its own heading

content/en/notebooks/advanced_analysis/_index.md
```diff
@@ -206,6 +206,217 @@ You can view, search, edit, and delete existing report schedules from the **Repo
 
 **Note**: To schedule reports and view other users' schedules, users need the **CSV Report Schedules Write** permission. To edit other users' schedules, users need the **CSV Report Schedules Manage** permission. These permissions can be granted by a user with the **Org Management** permission.
 
+## Writing efficient queries
+
+Notebook queries that read large amounts of data or run heavy computations can be slow or return resource errors. The patterns below are the most common causes—each with a rewrite that usually resolves the issue.
+
+The shortest summary: **filter early, aggregate early.** Work pushed into the query filter or a `GROUP BY` summary is computed against the index; work left to a `JOIN` or a wide `LIMIT` is held in memory.
+
+### Filter on the data source
```
shivrams: I feel like the recommendations can be grouped under top anti-patterns causing inefficient queries, and then we can easily iterate on the doc over time.

Maybe we group them like

* Aggregations
* Filtering
* Joins
* ?

vitalii-dd: grouped by 
- Filtering
- Column selection
- Aggregations
- Joins
- Expressions

content/en/notebooks/advanced_analysis/_index.md
vitalii-dd: done

content/en/notebooks/advanced_analysis/_index.md
vitalii-dd: added:
>In Notebooks, you can lock a cell's time range independently of the global time picker. Setting a conservative default prevents readers from accidentally widening the window to a much larger scan.

content/en/notebooks/advanced_analysis/_index.md
vitalii-dd: added
> Goal: understand log volume from the billing service over the last month.

content/en/notebooks/advanced_analysis/_index.md
vitalii-dd: It could be the same. If we don't have limit for the query, it uses also time range values.
I don't expect that this optimization will have effect for the 100%, at least for ddsql.
I believe, for the visualization cell we should optimize it ourselves.

content/en/notebooks/advanced_analysis/_index.md
vitalii-dd: Added a new section "Use a high-cardinality join key"

content/en/notebooks/advanced_analysis/_index.md
vitalii-dd: updated

content/en/notebooks/advanced_analysis/_index.md
vitalii-dd: moved checklist to the top

adamwalzer-datadog (approved):

mhintz (commented):
Ideally we'd include examples of using the "indexes" and "filter" arguments to the dd.logs table function, to improve the filtering performance for the underlying logs queries

content/en/ddsql_reference/_index.md
```diff
@@ -1199,6 +1199,241 @@ FROM aws.ec2_instance
...
+Scan time grows with the time range. Consider starting with a window that covers the question (often a few hours or a day) and widening only when the question demands more. For long-term trends, a pre-aggregated metric or a daily-summary query runs faster than re-scanning raw events on every run.
+
+*Goal: understand log volume from the billing service over the last month.*
+
+**Before**
+
+```sql
+-- A single query scanning 31 days of raw events
+SELECT *
+FROM logs
```
petitcl: Are we sure we want to use `select * from logs` as an example for this? given the way to query logs in DDSQL Editor is through a PTF. So users reading this might be misled in thinking that `select * from logs` is a valid way of fetching logs

content/en/ddsql_reference/_index.md
```diff
@@ -1199,6 +1199,241 @@ FROM aws.ec2_instance
 | akeys(hstore tags)                            | Array of text | Gets the keys of an HSTORE as an array                                                            |
 | avals(hstore tags)                            | Array of text | Gets the values of an HSTORE as an array                                                          |
 
+## Writing efficient queries
+
+Queries that read large amounts of data or run heavy computations can be slow or return resource errors. The patterns below are the most common causes—each with a rewrite that usually resolves the issue.
```
OliviaShoup:
```suggestion
Queries that read large amounts of data or run heavy computations can be slow or return resource errors. The patterns below are the most common causes, each with a rewrite that usually resolves the issue.
```

content/en/ddsql_reference/_index.md
```diff
@@ -1199,6 +1199,241 @@ FROM aws.ec2_instance
...
+
+Walk through this before re-running a slow query:
+
+- Does the query filter include a selective token (not a wildcard like `service:*` or `env:*`)?
+- Is the time range as small as it can be for the question?
+- Are you selecting only the columns you actually use?
+- If you have a large `LIMIT`, could a `GROUP BY` summary work instead?
+- If you have many JOINs, could it be rewritten as a single scan?
+- When joining across sources, is each side filtered?
+- Is the JOIN key high-cardinality (user / request / trace ID)?
```
OliviaShoup:
```suggestion
- If you have many `JOIN` clauses, could the query be rewritten as a single scan?
- When joining across sources, is each side filtered?
- Is the `JOIN` key high-cardinality (user/ request/trace ID)?
```

content/en/ddsql_reference/_index.md
```diff
@@ -1199,6 +1199,241 @@ FROM aws.ec2_instance
...
+- Is the same regex being run more than once per row?
+
+### Filtering
+
+#### Filter on the data source
+
+Always include a selective token in your data source query—`service:`, `host:`, `env:`, or any `@attribute:value`. Wildcard filters such as `service:*` or `env:*` match every event, so they don't actually narrow the data—treat them as equivalent to leaving the filter blank.
```
OliviaShoup:
```suggestion
Always include a selective token in your data source query, such as `service:`, `host:`, `env:`, or any `@attribute:value`. 

Wildcard filters such as `service:*` or `env:*` match every event, so they don't actually narrow the data. Treat these filters as equivalent to leaving the filter blank.
```

content/en/ddsql_reference/_index.md
```diff
@@ -1199,6 +1199,241 @@ FROM aws.ec2_instance
...
 
+## Writing efficient queries
+
+Queries that read large amounts of data or run heavy computations can be slow or return resource errors. The patterns below are the most common causes—each with a rewrite that usually resolves the issue.
+
+The shortest summary: **filter early, aggregate early.** Work pushed into the query filter or a `GROUP BY` summary is computed against the index; work left to a `JOIN` or a wide `LIMIT` is held in memory.
+
+### Checklist
+
+Walk through this before re-running a slow query:
```
OliviaShoup:
```suggestion
Before re-running a slow query, ask the following questions:
```

content/en/ddsql_reference/_index.md
```diff
@@ -1199,6 +1199,241 @@ FROM aws.ec2_instance
 | akeys(hstore tags)                            | Array of text | Gets the keys of an HSTORE as an array                                                            |
 | avals(hstore tags)                            | Array of text | Gets the values of an HSTORE as an array                                                          |
 
+## Writing efficient queries
+
+Queries that read large amounts of data or run heavy computations can be slow or return resource errors. The patterns below are the most common causes—each with a rewrite that usually resolves the issue.
+
+The shortest summary: **filter early, aggregate early.** Work pushed into the query filter or a `GROUP BY` summary is computed against the index; work left to a `JOIN` or a wide `LIMIT` is held in memory.
```
OliviaShoup:
```suggestion
**Filter early and aggregate early**. Computations in the query filter or a `GROUP BY` summary run against the index. Computations in a `JOIN` or a wide `LIMIT` are held in memory.
```

content/en/ddsql_reference/_index.md
jinatdatadog: agreed - lets replace it with a logs PTF? To make it valid DDSQL that is agnostic to DDSQL Editor or Notebooks

content/en/ddsql_reference/_index.md
```diff
@@ -1199,6 +1199,241 @@ FROM aws.ec2_instance
...
+  AND timestamp >= NOW() - INTERVAL '7 days'
+GROUP BY 1
+ORDER BY 1;
+```
+
+### Column selection
+
+#### Project only the columns you use
+
+Each column in your query is fetched from storage. Trim the column list to what you use downstream—wide attributes like the raw `message` or full HTTP headers can slow down the query significantly.
```
OliviaShoup:
```suggestion
Each column in your query is fetched from storage. Trim the column list to what you use downstream. Wide attributes like the raw `message` or full HTTP headers can slow down the query significantly.
```

content/en/ddsql_reference/_index.md
```diff
@@ -1199,6 +1199,241 @@ FROM aws.ec2_instance
...
+```sql
+-- Returns ~10 rows (one per status_code), answers the actual question
+SELECT status_code, count(*) AS hits
+FROM logs WHERE service = 'orders-api'
+GROUP BY status_code
+ORDER BY hits DESC
+LIMIT 100;
+```
+
+**Note:** Aggregating in SQL with `GROUP BY` is more efficient than fetching raw rows and aggregating in a downstream step—the engine filters and summarizes data at the source.
```
OliviaShoup:
```suggestion
**Note:** Aggregating in SQL with `GROUP BY` is more efficient than fetching raw rows and aggregating in a downstream step. The engine filters and summarizes data at the source.
```

content/en/ddsql_reference/_index.md
```diff
@@ -1199,6 +1199,241 @@ FROM aws.ec2_instance
...
+SELECT timestamp, service, host                     -- only what the analysis uses
+FROM logs
+WHERE service = 'checkout-api';
+```
+
+### Aggregations
+
+#### Return summaries, not raw rows
+
+When your goal is to understand the data—top-N, counts per category, distributions—a `GROUP BY` returns a focused result that is faster to compute and easier to work with than scanning millions of raw rows.
```
OliviaShoup:
```suggestion
When your goal is to understand the data (such as top-N, counts per category, or distributions), a `GROUP BY` returns a focused result that is faster to compute and easier to work with than scanning millions of raw rows.
```

content/en/ddsql_reference/_index.md
```diff
@@ -1199,6 +1199,241 @@ FROM aws.ec2_instance
...
+GROUP BY status_code
+ORDER BY hits DESC
+LIMIT 100;
+```
+
+**Note:** Aggregating in SQL with `GROUP BY` is more efficient than fetching raw rows and aggregating in a downstream step—the engine filters and summarizes data at the source.
+
+#### Narrow the scan before aggregating on a high-cardinality column
+
+`SELECT DISTINCT` or `GROUP BY` on a high-cardinality column (an email, an IP, a request ID) keeps one entry per distinct value across workers, which grows unboundedly without a tight filter. Narrow the data source filter first so the aggregation runs over fewer rows. For wide cardinalities, pre-aggregate to one row per time bucket and key, then count distinct across buckets. `approx_distinct(...)` is faster than exact `count(distinct ...)`.
```
OliviaShoup:
```suggestion
When working with high-cardinality columns (such as emails, IPs, or request IDs):
-  Use `SELECT DISTINCT` or `GROUP BY` to return one entry per distinct value across workers. Without a tight filter, the result set grows without bound.
- Narrow the data source filter first so the aggregation runs over fewer rows.
- For wide cardinalities, pre-aggregate to one row per time bucket and key, then count distinct across buckets.
- Use `approx_distinct(...)` instead of exact `count(distinct ...)` for better performance.
```

content/en/ddsql_reference/_index.md
```diff
@@ -1199,6 +1199,241 @@ FROM aws.ec2_instance
...
+) daily;
+```
+
+### Joins
+
+#### Combine self-joins into a single scan
+
+Joining one source to itself many times to correlate different events is one of the most common causes of slow queries. Most self-joins can be rewritten as a single scan with `CASE` expressions, window functions, or `GROUP BY ... HAVING`.
+
+**Before—4 self-joins**
```
OliviaShoup:
```suggestion
**Before: 4 self-joins**
```

content/en/ddsql_reference/_index.md
```diff
@@ -1199,6 +1199,241 @@ FROM aws.ec2_instance
...
+     b AS (SELECT user_id FROM logs WHERE service='payment-api'),
+     c AS (SELECT user_id FROM logs WHERE service='shipping-api'),
+     d AS (SELECT user_id FROM logs WHERE service='orders-api')
+SELECT user_id
+FROM a JOIN b USING (user_id)
+       JOIN c USING (user_id)
+       JOIN d USING (user_id);
+```
+
+**After—single scan**
```
OliviaShoup:
```suggestion
**After: single scan**
```

content/en/ddsql_reference/_index.md
```diff
@@ -1199,6 +1199,241 @@ FROM aws.ec2_instance
...
+SELECT user_id
+FROM logs
+WHERE service IN ('checkout-api','payment-api','shipping-api','orders-api')
+GROUP BY user_id
+HAVING count(DISTINCT service) = 4;
+```
+
+#### Use a high-cardinality join key
+
+Prefer a high-cardinality key (`user_id`, `request_id`, `trace_id`) when joining. Coarse keys like `service` or `status` can expand a moderate input into billions of intermediate rows. For large datasets, also narrow the data source filter and time range on each side of the join.
```
OliviaShoup:
```suggestion
When joining, prefer a high-cardinality key like `user_id`, `request_id`, or `trace_id`. Coarse keys like `service` or `status` can expand a moderate input into billions of intermediate rows. For large datasets, also narrow the data source filter and time range on each side of the join.
```

content/en/ddsql_reference/_index.md
```diff
@@ -1199,6 +1199,241 @@ FROM aws.ec2_instance
...
+
+#### Filter both sides of a cross-source join
+
+When you `JOIN` across two data sources (logs + RUM, logs + traces, feed + logs), apply a selective filter on each side. An unfiltered side becomes a full scan that has to be held in memory for the join. Where possible, pre-aggregate each source separately and join the summaries.
```
OliviaShoup:
```suggestion
When you `JOIN` across two data sources, apply a selective filter on each side. Examples include logs + RUM, logs + traces, and feed + logs. An unfiltered side becomes a full scan that has to be held in memory for the join. Where possible, pre-aggregate each source separately and join the summaries.
```

OliviaShoup (changes requested):
hey, thank you for the PR! it looks great :) i left some comments to make the text align with our style guide (removing em dashes, or shortening/rearranging sentences to make the page easier to scan quickly). please let me know if you have questions about anything!

OliviaShoup (approved):
thanks for the updates 🚀

vitalii-dd:
> Ideally we'd include examples of using the "indexes" and "filter" arguments to the dd.logs table function, to improve the filtering performance for the underlying logs queries

@mhintz , yes, we decided to move it to the ddsql section and changed to the dd.logs table function. 
All examples are mirrored in the https://app.datadoghq.com/notebook/14596226/writing-efficient-ddsql-queries.

vitalii-dd:
Fixed issues after [OliviaShoup](https://github.com/OliviaShoup) review.

PR #36741, opened 2026-05-15
(maint) Add site support checker

hestonhoffman:

Sends an alert to a PR when an author tries to add a site support banner directly to the top of the page. We'd prefer people use the existing site-support automation for this.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### AI assistance

Generated this workflow with Claude.


devindford (commented):
## Security Review

The workflow is **safe for use in a public repository**:
- Uses `pull_request` (not `pull_request_target`) — fork PRs run in an isolated context without access to secrets. This is the critical distinction; `pull_request_target` would grant elevated permissions to fork-contributed code and is the common source of "pwn request" vulnerabilities in public repos.
- Actions pinned to commit SHAs with version comments ✓
- Minimal permissions (`contents: read`, `pull-requests: write`) ✓
- `persist-credentials: false` ✓
- File paths from `git diff` are passed to the shell via an `env:` block (`CHANGED_FILES: ${{ steps.changed_files.outputs.files }}`), not via direct `${{ }}` interpolation in the script body — this is the correct pattern and avoids shell injection ✓
- `flagged_files.txt` content (file paths) is wrapped in backticks in the comment template, limiting markdown injection risk ✓

---

## Issues

### Bug: `github.ref` won't filter the translation branch

```yaml
if: github.ref != 'refs/heads/guacbot/translation-pipeline'
```

On `pull_request` events, `github.ref` is `refs/pull/{number}/merge`, never `refs/heads/...`, so this condition is always `true` and the translation pipeline branch is never actually skipped. Should be:

```yaml
if: github.head_ref != 'guacbot/translation-pipeline'
```

### Missing: No comment cleanup when the issue is resolved

When a PR author fixes their files (removes the flagged shortcode) and pushes a new commit, the warning comment will remain since the `Post warning comment` step only runs when `has_flagged == 'true'`. The workflow should handle the resolved case by deleting or editing the existing comment. Example addition:

```yaml
- name: Resolve warning comment
  if: steps.check.outputs.has_flagged == 'false'
  uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
  with:
    github-token: ${{ secrets.GITHUB_TOKEN }}
    script: |
      const { data: comments } = await github.rest.issues.listComments({
        owner: context.repo.owner,
        repo: context.repo.repo,
        issue_number: context.issue.number,
      });
      const marker = 'Possible site support note detected';
      const existing = comments.find(c => c.user.type === 'Bot' && c.body.includes(marker));
      if (existing) {
        await github.rest.issues.deleteComment({
          owner: context.repo.owner,
          repo: context.repo.repo,
          comment_id: existing.id,
        });
      }
```

### Minor: Fragile indentation stripping

```js
const body = `...`.replace(/^            /gm, '');
```

This strips exactly 12 leading spaces to un-indent the template literal. If the indentation ever shifts (editor formatting, refactor), the regex silently stops working and comments gain unexpected leading whitespace. A more robust approach is to not indent the template literal content, or use a small dedent helper.

devindford (commented):
## Re-review

All three issues from the previous review are addressed:

- **`github.head_ref` fix** ✓ — branch filter now correctly evaluates on PR events
- **Cleanup step** ✓ — `Resolve warning comment` step deletes the existing bot comment when `has_flagged == 'false'`
- **Fragile indentation stripping** ✓ — replaced the `.replace(/^ +/gm, '')` approach with an array `.join('\n')`, which is clean and won't silently break

One nit: `--diff-filter=AMD` now includes deleted files (`D`), but the original `AM` filter was correct for this use case. Deleted files are harmlessly skipped by the `[ ! -f "$file" ]` check, but there's no reason to include them in the list at all. Worth reverting that filter back to `AM` to keep the intent clear.

Otherwise this looks good to me.

devindford (approved):

PR #36892, opened 2026-05-21
Add experiments analysis methods docs

emielver:
## Summary

- Add a new Experiments `Analysis Methods` documentation page covering fixed-sample frequentist, sequential frequentist, sequential hybrid, and Bayesian analysis methods.
- Link the new page from the Experiments overview, Plan and Launch Experiments, and Reading Experiment Results pages.
- Add the page to the Experiments left navigation.

## Validation

- `git diff --check`
- Front matter and menu YAML parsed locally with Ruby
- Added link targets checked locally
- `git commit` hooks passed

Note: a full local `yarn run build:hugo` reached rendering but failed on a missing generated `layouts/partials/markdoc-assets.html` partial in this checkout, which appears to require the repo's local build-script setup and is unrelated to this docs change.


Dpananos (approved):

content/en/experiments/analysis_methods.md
```diff
@@ -0,0 +1,102 @@
...
+| Sequential frequentist | Monitor results while the experiment runs and make a decision when you are ready. | Supports flexible decision-making while controlling the false positive rate. | Has less power than fixed-sample analysis, so it can require more samples to detect the same effect. |
+| Sequential hybrid | Use sequential analysis while the experiment runs, then switch to a fixed-sample interval at the planned end date. | Combines early stopping with stronger power at the end of a planned experiment. | Requires an end date, and intervals are wider than fixed-sample intervals. |
+| Bayesian | Combine experiment data with a prior belief about plausible lifts, then make decisions from the posterior distribution. | Supports nuanced decisions, especially when sample sizes are small or business tradeoffs are asymmetric. | Requires trust in the prior and alignment on how to interpret probabilities. |
```
janine-c:
```suggestion
| [**Fixed-sample frequentist**](#fixed-sample-frequentist-analysis) | Choose a sample size or duration before launching the experiment, wait until that point, then make a decision. | Provides the most power for a fixed sample size. | Requires an upfront plan and can lose its statistical guarantees if you stop early or extend the experiment based on observed results. |
| [**Sequential frequentist**](#sequential-frequentist-analysis) | Monitor results while the experiment runs and make a decision when you are ready. | Supports flexible decision-making while controlling the false positive rate. | Has less power than fixed-sample analysis, so it can require more samples to detect the same effect. |
| [**Sequential hybrid**](#sequential-hybrid-analysis) | Use sequential analysis while the experiment runs, then switch to a fixed-sample interval at the planned end date. | Combines early stopping with stronger power at the end of a planned experiment. | Requires an end date, and intervals are wider than fixed-sample intervals. |
| [**Bayesian**](#bayesian-analysis) | Combine experiment data with a prior belief about plausible lifts, then make decisions from the posterior distribution. | Supports nuanced decisions, especially when sample sizes are small or business tradeoffs are asymmetric. | Requires trust in the prior and alignment on how to interpret probabilities. |
```
Suggesting some bold links in the first column to bring some more visual attention to the names of the methods, and to make it easier for users to click through to the relevant parts of the page 🙂

content/en/experiments/analysis_methods.md
```diff
@@ -0,0 +1,102 @@
...
+
+Sequential frequentist analysis is the default because it lets you monitor results and make ship or rollback decisions without invalidating the false positive rate. Fixed-sample analysis can be more powerful when everything goes according to plan, but it requires a stricter decision process. Sequential hybrid and Bayesian methods support more specialized decision-making workflows.
+
+Configure the analysis method in the experiment's [statistical analysis plan][1].
+
+## Fixed-sample frequentist analysis
+
+Fixed-sample frequentist analysis is the most direct way to analyze experiment results. Before launching the experiment, choose when you will evaluate the results. This decision point can be a fixed duration or a target sample size. When the experiment reaches that point, compare each treatment variant to control and decide whether to ship, roll back, or continue with a new experiment.
```
janine-c:
```suggestion
Fixed-sample frequentist analysis is the most direct way to analyze experiment results. Before launching the experiment, choose when you will evaluate the results. This decision point can be a fixed duration or a target sample size. When the experiment reaches that point, compare each treatment variant to the control and decide whether to ship, roll back, or continue with a new experiment.
```
Not sure if this just reads awkwardly to me, but if it's common to drop the article in this context, ignore me 🙂

content/en/experiments/analysis_methods.md
```diff
@@ -0,0 +1,102 @@
...
+The main challenge is choosing the decision point. If you evaluate too early, the experiment may not have enough power to detect a real effect. If you evaluate too late, you may expose users to an inferior experience longer than needed.
+
+<div class="alert alert-warning">For fixed-sample analysis, avoid changing the sample size or duration based on interim results. Stopping early because results look unusually good or bad, or extending the experiment because results are close to significant, can bias the lift estimate and increase the false positive rate.</div>
+
+There are two common ways to choose the decision point:
+
+- **Choose a fixed duration.** This works when product, business, or operational constraints determine how long the experiment can run. Make sure the duration is long enough to detect a meaningful effect.
+- **Run a power analysis.** Use historical metric behavior to estimate how many subjects you need to detect a given [minimum detectable effect][2].
+
+Use fixed-sample analysis when sample size is scarce and your team can commit to the decision criteria before the experiment starts.
```
janine-c:
```suggestion
Use fixed-sample analysis when sample size is scarce and your team can commit to the decision criteria before the experiment starts.

The main challenge is choosing the decision point. If you evaluate too early, the experiment may not have enough power to detect a real effect. If you evaluate too late, you may expose users to an inferior experience longer than needed. There are two common ways to choose the decision point:

- **Choose a fixed duration.** This works when product, business, or operational constraints determine how long the experiment can run. Make sure the duration is long enough to detect a meaningful effect.
- **Run a power analysis.** Use historical metric behavior to estimate how many subjects you need to detect a given [minimum detectable effect][2].

<div class="alert alert-warning">For fixed-sample analysis, avoid changing the duration or sample size based on interim results. Stopping early because results look unusually good or bad, or extending the experiment because results are close to significant, can bias the lift estimate and increase the false positive rate.</div>
```
Lightly rearranging the content to provide explanatory information first (when to use it; methods to choose the decision point) and then caveats second (not changing your decision for your method). I think this sort of progressive disclosure helps users take in the information a little more easily!

content/en/experiments/analysis_methods.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Sequential frequentist analysis
+
+Sequential frequentist analysis lets you monitor experiment results continuously and make a decision without preselecting a final sample size. This is useful when you need to react to strong positive or negative results, or when the assumptions behind a fixed-sample plan may change while the experiment is running.
+
+Sequential analysis controls the false positive rate while allowing repeated checks of the results. The tradeoff is power: for the same number of subjects, sequential analysis is less likely than fixed-sample analysis to detect a true effect. To reach the same power, the experiment may need to run longer.
+
+Use sequential analysis when flexibility matters more than maximizing power for a predetermined sample size. It is a good default for many experiments because it lets you:
```
janine-c, line 47:
```suggestion
Use sequential analysis when flexibility matters more than maximizing power for a predetermined sample size. It is the default for experiments because it lets you:
```
I thought it was a little easy to miss that this is the default below the table, so I thought it was worth restating somewhere in here again.

content/en/experiments/analysis_methods.md
```diff
@@ -0,0 +1,102 @@
...
+- Continue collecting data without invalidating the analysis.
+- Avoid restarting the experiment when the original sample size assumptions were wrong.
+
+## Sequential hybrid analysis
+
+Sequential hybrid analysis combines sequential and fixed-sample approaches. While the experiment is running, Datadog uses a sequential method so you can monitor results and stop early when needed. At the planned end date, Datadog switches to a fixed-sample interval, which gives the final analysis more power than a purely sequential method.
+
+The tradeoff is that sequential hybrid analysis requires a planned end date, and the intervals are slightly wider than fixed-sample intervals. This protects the statistical guarantees across both phases of the analysis.
+
+Sequential hybrid analysis is useful when you want to stop early for clear regressions but prefer to wait until the planned end date before declaring a winning variant. In that workflow, teams often use the sequential phase to protect users from harmful variants and the fixed-sample phase to make final shipping decisions with more power.
```
janine-c:
```suggestion
Use sequential hybrid analysis when you want to stop early for clear regressions, but prefer to wait until the planned end date before declaring a winning variant. In that workflow, teams often use the sequential phase to protect users from harmful variants and the fixed-sample phase to make final shipping decisions with more power.
```
To match the structure of the above methods

content/en/experiments/analysis_methods.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Bayesian analysis
+
+Bayesian analysis uses experiment data to update a prior belief about plausible lift values. The result is a posterior distribution that describes which lift values are most compatible with the prior and the observed data.
+
+Frequentist methods ask how likely the observed data would be if the treatment and control had no true difference. Bayesian methods instead take the observed data and prior as given, then estimate the probability of different lift values. This can make results easier to discuss with stakeholders because statements such as "the treatment is likely to be better than control" map more directly to the output of the analysis.
+
+Bayesian analysis is useful when:
```
janine-c:
```suggestion
Use Bayesian analysis when:
```

content/en/experiments/analysis_methods.md
```diff
@@ -0,0 +1,102 @@
...
+
+Sequential hybrid analysis is useful when you want to stop early for clear regressions but prefer to wait until the planned end date before declaring a winning variant. In that workflow, teams often use the sequential phase to protect users from harmful variants and the fixed-sample phase to make final shipping decisions with more power.
+
+## Bayesian analysis
+
+Bayesian analysis uses experiment data to update a prior belief about plausible lift values. The result is a posterior distribution that describes which lift values are most compatible with the prior and the observed data.
+
+Frequentist methods ask how likely the observed data would be if the treatment and control had no true difference. Bayesian methods instead take the observed data and prior as given, then estimate the probability of different lift values. This can make results easier to discuss with stakeholders because statements such as "the treatment is likely to be better than control" map more directly to the output of the analysis.
```
janine-c:
```suggestion
Frequentist methods ask how likely the observed data would be if the treatment and control had no true difference. Bayesian methods instead take the observed data and prior as given, then estimate the probability of different lift values. This can make results easier to discuss with stakeholders because statements such as "the treatment is likely to be better than the control" map more directly to the output of the analysis.
```

content/en/experiments/analysis_methods.md
```diff
@@ -0,0 +1,102 @@
...
+
+Bayesian analysis is useful when:
+
+- You need to make a decision with limited data.
+- The decision depends on the probability of a lift exceeding a business threshold, not just whether the interval excludes zero.
+- Different metrics have asymmetric risks or rewards.
+- Stakeholders are prepared to interpret probabilities rather than binary significant or not significant results.
```
janine-c:
```suggestion
- Stakeholders are prepared to interpret probabilities rather than binary significant or non-significant results.
```
I think having a compound adjective here makes this a little easier to read

content/en/experiments/analysis_methods.md
```diff
@@ -0,0 +1,102 @@
...
+
+The prior matters most when sample sizes are small. With enough data, the posterior is driven mostly by observed experiment behavior. With little data, an incorrect prior can influence the lift estimate and interval enough to change the decision.
+
+<div class="alert alert-info">Bayesian intervals are technically credible intervals, though Datadog may present them alongside confidence intervals in the experiment results UI. Unlike frequentist intervals, Bayesian intervals do not provide the same false positive rate guarantee.</div>
+
+## Related settings
+
+Analysis methods are only one part of the statistical analysis plan. Datadog Experiments also supports:
```
janine-c:
```suggestion
Analysis methods are only one part of the statistical analysis plan. Datadog Experiments also supports modifying the following [settings][1]:
```
I thought adding a link would be helpful, so if a user's like "oh cool, yes!" they can click through to get more information on implementing it 🙂

janine-c (approved):
This looks great! I made some really minor writing suggestions for clarity 🙂

content/en/experiments/analysis_methods.md
```diff
@@ -0,0 +1,102 @@
...
+
+| Method | Description | Strengths | Tradeoffs |
+| --- | --- | --- | --- |
+| Fixed-sample frequentist | Choose a sample size or duration before launching the experiment, wait until that point, then make a decision. | Provides the most power for a fixed sample size. | Requires an upfront plan and can lose its statistical guarantees if you stop early or extend the experiment based on observed results. |
+| Sequential frequentist | Monitor results while the experiment runs and make a decision when you are ready. | Supports flexible decision-making while controlling the false positive rate. | Has less power than fixed-sample analysis, so it can require more samples to detect the same effect. |
+| Sequential hybrid | Use sequential analysis while the experiment runs, then switch to a fixed-sample interval at the planned end date. | Combines early stopping with stronger power at the end of a planned experiment. | Requires an end date, and intervals are wider than fixed-sample intervals. |
```
tbuffington7: > switch to a fixed-sample interval at the planned end date. 

This is only if the experiment is not stopped before then. If the experiment was ended early, it stays sequential forever.

I'd say this instead:

> Use sequential analysis before a pre-planned end date, then switch to a fixed sample interval on the end date if the experiment was not stopped early.

content/en/experiments/analysis_methods.md
tbuffington7: I'd remove this part: 

> or business tradeoffs are asymmetric.

Many people use the standard frequentist analysis with alpha=0.05 precisely because of asymmetric tradeoffs - they think that a false positive is worse than a false negative

content/en/experiments/analysis_methods.md
```diff
@@ -0,0 +1,102 @@
...
+| Bayesian | Combine experiment data with a prior belief about plausible lifts, then make decisions from the posterior distribution. | Supports nuanced decisions, especially when sample sizes are small or business tradeoffs are asymmetric. | Requires trust in the prior and alignment on how to interpret probabilities. |
+
+Sequential frequentist analysis is the default because it lets you monitor results and make ship or rollback decisions without invalidating the false positive rate. Fixed-sample analysis can be more powerful when everything goes according to plan, but it requires a stricter decision process. Sequential hybrid and Bayesian methods support more specialized decision-making workflows.
```
tbuffington7: I'd say "without inflating the false positive rate"

lukasgoetzweiss (approved):
lgtm!

content/en/experiments/analysis_methods.md
```diff
@@ -0,0 +1,102 @@
...
+
+The main challenge is choosing the decision point. If you evaluate too early, the experiment may not have enough power to detect a real effect. If you evaluate too late, you may expose users to an inferior experience longer than needed.
+
+<div class="alert alert-warning">For fixed-sample analysis, avoid changing the sample size or duration based on interim results. Stopping early because results look unusually good or bad, or extending the experiment because results are close to significant, can bias the lift estimate and increase the false positive rate.</div>
+
+There are two common ways to choose the decision point:
+
+- **Choose a fixed duration.** This works when product, business, or operational constraints determine how long the experiment can run. Make sure the duration is long enough to detect a meaningful effect.
```
tbuffington7: This is a little confusing because it also sounds like running a power analysis (how do you know if you can detect a meaningful effect?). Also, the takeaway from running a power analysis per the second bullet point is to choose a fixed duration. 

Rather than saying there are two common ways, I would consolidate it to something like this:

The decision point should be informed by a power analysis, which helps experimentation teams choose a duration that gives the experiment enough power to detect meaningful effects while adhering to product, business, or operational constraints.

content/en/experiments/analysis_methods.md
tbuffington7: I'd remove the part about "the sample size is scare"

Maybe

"Use fixed-sample analysis for additional statistical power when your team can commit to a fixed duration before the experiment starts."

content/en/experiments/analysis_methods.md
```diff
@@ -0,0 +1,102 @@
...
+
+Bayesian analysis uses experiment data to update a prior belief about plausible lift values. The result is a posterior distribution that describes which lift values are most compatible with the prior and the observed data.
+
+Frequentist methods ask how likely the observed data would be if the treatment and control had no true difference. Bayesian methods instead take the observed data and prior as given, then estimate the probability of different lift values. This can make results easier to discuss with stakeholders because statements such as "the treatment is likely to be better than control" map more directly to the output of the analysis.
+
+Bayesian analysis is useful when:
+
+- You need to make a decision with limited data.
+- The decision depends on the probability of a lift exceeding a business threshold, not just whether the interval excludes zero.
+- Different metrics have asymmetric risks or rewards.
```
tbuffington7: I'd remove the part about asymmetric risks or rewards. I can't think of any way we codify that in the product, so it's almost certain to confuse customers

content/en/experiments/analysis_methods.md
tbuffington7: Stakeholders are prepared to interpret probabilities rather than binary significant or not significant results.

I don't really agree with this point - you have probabilities (p-values) in the frequentist approaches too. I would say something like this instead:

Stakeholders prefer to think in terms of probabilities of success rather than in terms of frequentist p-values.

content/en/experiments/analysis_methods.md
```diff
@@ -0,0 +1,102 @@
...
+- Stakeholders are prepared to interpret probabilities rather than binary significant or not significant results.
+
+The prior matters most when sample sizes are small. With enough data, the posterior is driven mostly by observed experiment behavior. With little data, an incorrect prior can influence the lift estimate and interval enough to change the decision.
```
tbuffington7: There is somewhat of a myth that A/B testing sample sizes are large enough that they dominate Bayesian priors. 

I'd maybe say this:

The prior matters most when sample sizes are small. With enough data, the posterior is driven mostly by observed experiment behavior, but empirically grounded priors are often strong enough to affect results even at large sample sizes. A poorly-specified prior can influence the lift estimate and interval enough to change the decision.

content/en/experiments/analysis_methods.md
```diff
@@ -0,0 +1,102 @@
...
+
+<div class="alert alert-info">Bayesian intervals are technically credible intervals, though Datadog may present them alongside confidence intervals in the experiment results UI. Unlike frequentist intervals, Bayesian intervals do not provide the same false positive rate guarantee.</div>
+
+## Related settings
+
+Analysis methods are only one part of the statistical analysis plan. Datadog Experiments also supports:
+
+CUPED
+: Uses pre-experiment data from each subject to reduce metric variance and improve experiment sensitivity. With CUPED enabled, displayed lift may differ from the naive treatment-minus-control calculation.
```
tbuffington7: > With CUPED enabled, displayed lift may differ from the naive treatment-minus-control calculation.

This isn't quite right - we have CUPED treatment and control values, so if you apply the "naive treatment-minus-control calculation" on _those_ values, it won't differ. 

I would say something like:

With CUPED enabled, displayed lift and metric values may differ from the naive estimates calculated from the raw data.

tbuffington7 (commented):
Looks good overall. Just a few points that to update/add from a stats perspective

content/en/experiments/analysis_methods.md
emielver: Adjusted, thank you!

content/en/experiments/analysis_methods.md
emielver: Removed, thanks!

content/en/experiments/analysis_methods.md
emielver: Gotcha

content/en/experiments/analysis_methods.md
emielver: Nice, consolidated, much clearer!

content/en/experiments/analysis_methods.md
emielver: Done

content/en/experiments/analysis_methods.md
emielver: 👍

content/en/experiments/analysis_methods.md
emielver: Removed!

content/en/experiments/analysis_methods.md
emielver: Done, thanks for clearing that up!

content/en/experiments/analysis_methods.md
emielver: Done!

PR #36939, opened 2026-05-22
Add .NET logo and fix C++ logo in SCA language support section

gorkavicente:

Updates the SCA language support cards in the `sca-lang-support.html` partial:

- Adds the `.NET` logo (using `dotnet_avatar.svg`) after the Java logo, where the Go logo previously appeared
- Fixes the C++ logo entry (replaces the incorrectly used `c_sharp.png` with `cpp.png`) and corrects the alt text to `c++`

Both `.NET` and `C++` are now properly represented in the language support grid.

### Merge instructions

Merge readiness:
- [x] Ready for merge

### Additional notes


drichards-87:
@gorkavicente Just FYI, there’s a merge conflict in the partial file. I took a quick look, but I wasn’t sure how it should be resolved and didn’t want to mess up the changes you were making. Could you take a look?

StefonSimmons (approved):

StefonSimmons (commented):
The change looks good. One thing - every image card links to the same anchor, so the links aren't actually navigating users anywhere distinct. Should we drop the <a> wrappers, or are there plans to link each language to its own section?
CC: @drichards-87

drichards-87:
> The change looks good. One thing - every image card links to the same anchor, so the links aren't actually navigating users anywhere distinct. Should we drop the wrappers, or are there plans to link each language to its own section? CC: @drichards-87

@StefonSimmons I think we did have separate pages at one time, but then we removed them later. I agree though, if we don't have separate language-specific setup pages that we can link to, we should probably convert this to a list, instead of using a partial.

PR #37125, opened 2026-05-29
Bump cdocs and add cdocs card-grid e2e tests

hestonhoffman:

- Adds render hooks to handle tooltips in card-grids.
- Bumps `cdocs-hugo-integration` to v4.4.0 and adds Playwright e2e tests for the `card-grid` and `image-card` Markdoc tags.
- Updates some stale snapshots

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes

Package URL should be updated to the `master` build once DataDog/corp-node-packages#282 merges.


OliviaShoup (approved):
🚀

domalessi (approved):

assets/scripts/datadog-docs.js
```diff
@@ -128,6 +129,10 @@ const doOnLoad = () => {
         initCodeTabs();
     }
 
+    if (document.querySelector('.card-grid-card[data-bs-toggle="tooltip"]')) {
```
jhgilbert: I think I would just run initCardGrid no matter what. Even if it's not on the page at first, it might be later, right? Worst case, the init script won't match on anything, and `forEach` will be on an empty array, which is probably comparable performance-wise to the guard you have here, and then only needs to run once no matter what, instead of twice if your guard comes back affirmative.

jhgilbert (approved):
One minor suggestion, thanks!

PR #37195, opened 2026-06-02
Clarify AND logic for multiple source filters in Custom Allocation Rules

AustinLeaming:
## What does this PR do?

Adds a note to the Custom Allocation Rules documentation clarifying that when multiple source filters (`costs_to_allocate`) are specified, they are combined with **AND** logic. A cost line must match all filter conditions to be selected for allocation.

## Why is this important?

This behavior is not documented anywhere — not in the public docs, API reference, or Terraform provider docs. A Tier 1 customer (CLOUDS-8179) raised this question, and we confirmed the AND behavior by reviewing:

- Backend source code (`dd-analytics/DailyProportionalAggregator.scala`) — uses `.forall()` (Scala AND)
- Anthropic OOTB seed data (`k8s-resources`) — three separate rules with AND'd filters
- Internal Terraform configs (`dd-source`) — real rules with multiple `costs_to_allocate` blocks
- Protobuf definition (`dd-source`) — `repeated ArbitraryAllocationFilter costs_to_allocate`

## Preview

The note appears in **Step 1 - Define the source**, right after the example, and reads:

> **Note:** When you define multiple source filters, they are combined with **AND** logic. A cost line must match **all** filter conditions to be selected for allocation.

## Ref
- CLOUDS-8179


content/en/cloud_cost_management/allocation/custom_allocation_rules.md
```diff
@@ -38,6 +38,9 @@ You can manage custom allocation rules using the [API][4], [Terraform][5], or di
 
    _Example: Untagged support costs, shared database costs._
 
+**Note:** When you define multiple source filters, they are combined with **AND** logic. A cost line must match **all** filter conditions to be selected for allocation. For example, if you filter by `aws_product is ec2` and `region is us-east-1`, only costs matching **both** conditions are allocated. To apply OR logic (matching costs that satisfy either condition independently), create separate allocation rules.
```
rtrieu:
```suggestion
**Note**: When you define multiple source filters, they are combined with **AND** logic. A cost line must match **all** filter conditions to appear in allocation. For example, if you filter by `aws_product is ec2` and `region is us-east-1`, only costs matching **both** conditions are allocated. To apply **OR** logic, create separate separate allocation rules.
```

rtrieu (changes requested):
minor suggestion to use active voice and remove redundancy

safipatel (commented):
looks good from an accuracy standpoint

safipatel (approved):

rtrieu (approved):

PR #37251, opened 2026-06-04
[SOURCE-323] Add review signals to claude_review action

xopham:
## Summary

- Add start and finish signal jobs that manage requester reactions and the `Claude review` commit status.
- Harden reaction cleanup so stale bot eyes/thumb reactions are removed before setting the current result.
- Update the Claude review artifact upload/download action pins from the sandbox work.

## Validation

- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/claude_review.yml"); puts "YAML OK"'`
- `git diff --cached --check`

### DIFF
--- a/.github/workflows/claude_review.yml
+++ b/.github/workflows/claude_review.yml
@@ -3,19 +3,25 @@
 #
 # Security model — read before editing:
 #
-# 1. The pipeline has three jobs: `gate` (verify the /review caller
-#    has repo write access and resolve the PR head SHA), `review`
-#    (run Claude with read-only tools and produce a JSON review
-#    artifact), and `post` (sanitize and submit the review via the
-#    GitHub API). Treat the checked-out tree and PR/comment
-#    metadata as hostile input subject to prompt injection.
+# 1. The pipeline has five jobs: `gate` (verify the /review caller
+#    has repo write access and resolve the PR head SHA), `start_signal`
+#    (acknowledge the request and create the PR status), `review` (run
+#    Claude with read-only tools and produce a JSON review artifact),
+#    `post` (sanitize and submit the review via the GitHub API), and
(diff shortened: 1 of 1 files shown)


.github/workflows/claude_review.yml
```diff
@@ -165,6 +171,78 @@ jobs:
         echo "head_sha=$head_sha" >> "$GITHUB_OUTPUT"
         echo "base_sha=$base_sha" >> "$GITHUB_OUTPUT"
 
+  start_signal:
+    name: Acknowledge request
+    needs: gate
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+      statuses: write
```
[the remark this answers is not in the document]
xopham: this is incorrect as we are reacting to PR comments

hestonhoffman (approved):

.github/workflows/claude_review.yml
```diff
@@ -165,6 +171,78 @@ jobs:
...
+          }
+
+          await github.rest.repos.createCommitStatus({
+            owner, repo,
+            sha: headSha,
+            state: "pending",
+            context: statusContext,
+            target_url: runUrl,
+            description: `Claude is reviewing PR #${issueNumber}.`,
+          });
```
KSerrania, line 244: I understand the part about comment reactions, but why do we need a commit status on top of it? Is the status from the GitHub Action workflow run itself not enough?

KSerrania (approved):

.github/workflows/claude_review.yml
xopham: We are running on an issue comment event (which are also comments on PRs) and consequently the check run is not attached to the PR. So, we do that manually.

xopham: or in other words: without this, there is no indication of the check run on the PR

PR #37334, opened 2026-06-08
Remove deprecated APM Investigator experience

brett0000FF:

Remove APM Investigator latency experience.

Changes:
- **Delete** the APM Investigator guide page (`/tracing/guide/latency_investigator/`) and its images.
- **Redirect** the old URL to the Bits Chat page with an alias, so existing links and bookmarks land on the replacement experience instead of a 404. (Hugo aliases redirect to the page, not a specific anchor, so it lands on the Bits Chat page rather than scrolling directly to the Latency investigations section.)
- **User and account tracing guide**: remove the APM Investigator mentions and screenshot; user/account impact now points to Error Tracking issues only.
- **Bits Investigation page**: drop the retired preview callout and the point-plot step from the APM latency entry point, and link to Bits Chat latency investigations for details.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes


buraizu (commented):
Looks good overall — but could you confirm if it's intentional to delete all the screenshots? Some translated docs will still reference these screenshots until the change has worked its way through the translation platform. For example, here:
https://docs.datadoghq.com/ja/tracing/guide/latency_investigator/

brett0000FF:
> Looks good overall — but could you confirm if it's intentional to delete all the screenshots? Some translated docs will still reference these screenshots until the change has worked its way through the translation platform. For example, here: https://docs.datadoghq.com/ja/tracing/guide/latency_investigator/

@buraizu - Good catch!

buraizu (approved):

PR #37381, opened 2026-06-09
Fix link within tab (again)

evazorro:
The same link is used in both normal body text _and_ content within a tab, and when I noticed and fixed the link within the tab, I accidentally broke the body text version. The link reference needs to be within the tab itself, so we can't actually use the same link reference for both. This PR adds back the original reference (`[13]`) and updates the reference within the tab to use `[100]` (because presumably we won't have enough links on the page ever to cause a collision).

### Merge instructions

Merge readiness:
- [x] Ready for merge

### AI assistance

None, all natural

### DIFF
--- a/content/en/agent/guide/datadog-disaster-recovery.md
+++ b/content/en/agent/guide/datadog-disaster-recovery.md
@@ -160,7 +160,7 @@ Then, follow the prompt to scope the hosts and telemetry (metrics, logs, traces)
 
 <div class="alert alert-danger">Cloud Integrations can only run in either your primary or DDR Datadog site, but not both at the same time, so failing them over ceases Cloud Integration data in your primary site. <strong>During an integration failover, integrations run only in the DDR data center.</strong> When no longer in failover, disable the failover policy to return integration data collection to the primary org.</div>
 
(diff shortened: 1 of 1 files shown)


content/en/agent/guide/datadog-disaster-recovery.md
janine-c: Hello! Does this mean this one should be [100] as well? Otherwise, I'm not sure what's referring to the in-tab link!

```suggestion
From the [Fleet Automation][100] page in your failover org, on the {{< ui >}}Configure Agents{{< /ui >}} tab, you can create a failover policy or reuse an existing one, and apply it to your fleet of Agents. Soon after the policy is enabled, Agents begin dual-shipping telemetry to both the primary and DDR (failover) observability sites.
```

janine-c (approved):
Quick question, but approving so you can merge once it's settled!

content/en/agent/guide/datadog-disaster-recovery.md
evazorro: You are totally correct! Good catch thank you!!

janine-c: GO TEAM 🚀

PR #37398, opened 2026-06-09
(DOCS-14612): Adjust Fleet Automation site-support

hestonhoffman:
Fleet View is in preview on gov sites, which means we can no longer use a blanket support parameter for Fleet Automation. This PR adds a site-support-id to most Fleet Automation pages and a banner to Fleet View scoped to gov 1 and 2.

### DIFF
--- a/content/en/agent/fleet_automation/fleet_view.md
+++ b/content/en/agent/fleet_automation/fleet_view.md
@@ -10,13 +10,12 @@ further_reading:
   text: "Send a flare"
 ---
 
-<div class="alert alert-info">
 {{< site-region region="gov,gov2" >}}
+<div class="alert alert-info">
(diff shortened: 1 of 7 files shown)


content/en/agent/fleet_automation/fleet_view.md
```diff
@@ -10,6 +10,14 @@ further_reading:
   text: "Send a flare"
 ---
 
+<div class="alert alert-info">
```
evazorro: Should the div be inside the site-region partial?

hestonhoffman: Oh yeah, probably better if it is. I'll fix that

evazorro (approved):

PR #37484, opened 2026-06-12
Fix Developer Platform link in OAuth2 in Datadog doc

holdenmc:

Updates the \"Developer Platform\" link in the \"Implement the OAuth protocol\" section from `https://app.datadoghq.com/apps` to `https://app.datadoghq.com/publishing-platform`.

**Note:** There is a second occurrence of `https://app.datadoghq.com/apps` on this page — it appears as the anchor text for \"SHA-256\" in the Code Challenge Method table of the \"Use the PKCE extension\" section. That link appears to be a dead end / incorrect target, but I'm not sure what it should point to instead. Leaving it for a follow-up.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes

### DIFF
--- a/content/en/extend/authorization/oauth2_in_datadog.md
+++ b/content/en/extend/authorization/oauth2_in_datadog.md
@@ -114,5 +114,5 @@ The [PKCE protocol][11] integrates with the authorization code grant flow by com
 [13]: https://datatracker.ietf.org/doc/html/rfc7636#section-4.2
 [14]: https://datatracker.ietf.org/doc/html/rfc7636#section-4.3
 [15]: https://datatracker.ietf.org/doc/html/rfc6234#section-4.1
-[16]: https://app.datadoghq.com/apps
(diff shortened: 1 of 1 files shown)


evazorro:
> **Note:** There is a second occurrence of `https://app.datadoghq.com/apps` on this page — it appears as the anchor text for "SHA-256" in the Code Challenge Method table of the "Use the PKCE extension" section. That link appears to be a dead end / incorrect target, but I'm not sure what it should point to instead. Leaving it for a follow-up.

@holdenmc yeah, it looks like that link was introduced when the doc was first created, but I agree that is probably not what it should point to. I'm just going to remove the link, because SHA-256 is an easy enough thing to google; I don't think we need to link to an article about it or anything.

holdenmc:
@evazorro agreed, that makes sense!

evazorro (approved):
Thank you for fixing this!

PR #37623, opened 2026-06-18
Clarify Go DSM context propagation and add goroutine/channel guidance

ericfirth:

Improves the Go Data Streams Monitoring manual instrumentation page in two ways:

- **Clarifies the Confluent Kafka steps 3–4**: expands the intro sentence to explain what `ExtractFromBase64Carrier` returns and why, and explicitly calls out that passing `context.Background()` at the produce site instead creates a new pathway root and breaks end-to-end visibility. This gives the code steps meaning beyond "copy this."
- **Adds a "Goroutines and channels" subsection**: a common failure mode where the DSM pathway restarts at the produce step because the consume `ctx` didn't survive a goroutine/channel handoff. The new section explains the mechanism in one sentence, shows how to carry the ctx through a job struct (with `context.WithoutCancel` for async workers), and adds a brief note on fan-out and fan-in.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes

No Jira ticket for this one.

### DIFF
--- a/content/en/data_streams/setup/language/go.md
+++ b/content/en/data_streams/setup/language/go.md
@@ -92,7 +92,7 @@ producer, err := ddkafka.NewProducer(&kafka.ConfigMap{
 }, ddkafka.WithDataStreams())
 ```
 
-If a service consumes data from one point and produces to another point, propagate context between the two places using the Go context structure. The `ctx` returned by `ExtractFromBase64Carrier` carries the upstream DSM pathway. Pass it to `SetDataStreamsCheckpointWithParams` when you produce, then inject it into the outbound message with `InjectToBase64Carrier`. Passing `context.Background()` at the produce site instead — for example, in a worker goroutine that has lost the consume ctx — creates a new pathway root and breaks end-to-end visibility.
+If a service consumes data from one point and produces to another point, propagate context between the two places using the Go context structure. The `ctx` returned by `ExtractFromBase64Carrier` carries the upstream DSM pathway. Pass it to `SetDataStreamsCheckpointWithParams` when you produce, then inject it into the outbound message with `InjectToBase64Carrier`. Passing `context.Background()` at the produce site creates a new pathway root and breaks end-to-end visibility. This happens, for example, in a worker goroutine that has lost the consume `ctx`.
 
 3. Extract the context from headers
 
@@ -107,7 +107,7 @@ If a service consumes data from one point and produces to another point, propaga
 
 ##### Goroutines and channels
 
-Go channels and goroutines do not carry `context.Context` automatically. If your service fans consumed messages out to worker goroutines before producing, thread the consume ctx to the produce site by including it in the work item you send over the channel:
+Go channels and goroutines do not carry `context.Context` automatically. If your service fans consumed messages out to worker goroutines before producing, pass the consume `ctx` to the produce site by including it in the work item you send over the channel:
 
 ```go
 type job struct {
@@ -137,7 +137,8 @@ for j := range jobs {
 }
 ```
 
-When one consumed message fans out to multiple produce calls, pass the same consume ctx to each produce checkpoint — each call creates its own child node in the pathway. For the reverse case (many consumed messages merging into one produce), combine the inbound contexts with `datastreams.MergeContexts(ctxs...)` before producing.
+- **Fan-out**: When one consumed message fans out to multiple produce calls, pass the same consume `ctx` to each produce checkpoint. Each call creates its own child node in the pathway.
+- **Fan-in**: When many consumed messages merge into one produce call, combine the inbound contexts with `datastreams.MergeContexts(ctxs...)` before producing.
 
 #### Other queuing technologies or protocols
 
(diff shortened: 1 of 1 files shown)


content/en/data_streams/setup/language/go.md
```diff
@@ -92,7 +92,7 @@ producer, err := ddkafka.NewProducer(&kafka.ConfigMap{
 }, ddkafka.WithDataStreams())
 ```
 
-If a service consumes data from one point and produces to another point, propagate context between the two places using the Go context structure:
+If a service consumes data from one point and produces to another point, propagate context between the two places using the Go context structure. The `ctx` returned by `ExtractFromBase64Carrier` carries the upstream DSM pathway. Pass it to `SetDataStreamsCheckpointWithParams` when you produce, then inject it into the outbound message with `InjectToBase64Carrier`. Passing `context.Background()` at the produce site instead — for example, in a worker goroutine that has lost the consume ctx — creates a new pathway root and breaks end-to-end visibility.
```
git-thuerk-done: `ctx` is a code identifier and should use backtick formatting when it appears in prose. Also, the last sentence is complex with nested em dashes — splitting it aids readability.

```suggestion
If a service consumes data from one point and produces to another point, propagate context between the two places using the Go context structure. The `ctx` returned by `ExtractFromBase64Carrier` carries the upstream DSM pathway. Pass it to `SetDataStreamsCheckpointWithParams` when you produce, then inject it into the outbound message with `InjectToBase64Carrier`. Passing `context.Background()` at the produce site creates a new pathway root and breaks end-to-end visibility. This happens, for example, in a worker goroutine that has lost the consume `ctx`.
```

content/en/data_streams/setup/language/go.md
```diff
@@ -105,6 +105,40 @@ If a service consumes data from one point and produces to another point, propaga
     datastreams.InjectToBase64Carrier(ctx, ddsarama.NewProducerMessageCarrier(message))
     ```
 
+##### Goroutines and channels
```
git-thuerk-done, line 108: This `#####` heading is at the same level as `##### Confluent Kafka client` and `##### Sarama Kafka client`, making it a sibling section rather than a subsection. If this is intended to be a subsection of the Confluent client section, change it to `######`. See the higher-level comment for options.

content/en/data_streams/setup/language/go.md
```diff
@@ -105,6 +105,40 @@ If a service consumes data from one point and produces to another point, propaga
...
+
+Go channels and goroutines do not carry `context.Context` automatically. If your service fans consumed messages out to worker goroutines before producing, thread the consume ctx to the produce site by including it in the work item you send over the channel:
```
git-thuerk-done: "Thread" as a verb meaning "pass through" is uncommon and may not read clearly to all readers. Also, `ctx` needs backtick formatting in prose.

```suggestion
Go channels and goroutines do not carry `context.Context` automatically. If your service fans consumed messages out to worker goroutines before producing, pass the consume `ctx` to the produce site by including it in the work item you send over the channel:
```

content/en/data_streams/setup/language/go.md
```diff
@@ -105,6 +105,40 @@ If a service consumes data from one point and produces to another point, propaga
...
+        options.CheckpointParams{PayloadSize: int64(out.Value.Length())},
+        "direction:out", "type:kafka", "topic:"+outTopic)
+    if ok {
+        datastreams.InjectToBase64Carrier(ctx, ddsarama.NewProducerMessageCarrier(out))
+    }
+    producer.SendMessage(out)
+}
+```
+
+When one consumed message fans out to multiple produce calls, pass the same consume ctx to each produce checkpoint — each call creates its own child node in the pathway. For the reverse case (many consumed messages merging into one produce), combine the inbound contexts with `datastreams.MergeContexts(ctxs...)` before producing.
```
git-thuerk-done: `ctx` needs backtick formatting in prose (two instances). The fan-out and fan-in patterns are distinct enough to merit separate bullets for scannability.

```suggestion
- **Fan-out**: When one consumed message fans out to multiple produce calls, pass the same consume `ctx` to each produce checkpoint. Each call creates its own child node in the pathway.
- **Fan-in**: When many consumed messages merge into one produce call, combine the inbound contexts with `datastreams.MergeContexts(ctxs...)` before producing.
```

git-thuerk-done (changes requested):
Two structural issues need resolving before merge: (1) the heading level of the new "Goroutines and channels" section makes it a sibling of the Kafka client sections rather than a subsection, and the code example uses Sarama APIs despite the section's placement in/near the Confluent section — see the inline comment on line 108 for options; (2) `ctx` needs backtick formatting throughout the new prose. Comments on the specific lines below.

content/en/data_streams/setup/language/go.md
ericfirth: Keeping this as `#####` intentionally — the goroutines/channels guidance applies to both Sarama and Confluent clients, so it belongs as a sibling section under `#### Manual instrumentation` rather than as a subsection of either client. The Sarama code example is there because Sarama is the more fully covered client in the page; the same ctx-threading pattern applies to Confluent.

content/en/data_streams/setup/language/go.md
ericfirth: fixed

git-thuerk-done (approved):
looks good thank you!

PR #37639, opened 2026-06-19
Update AAP verify setup shortcode

e-n-0:

The Verify section of AAP pages where not loading the templated code associated to the shortcode.

This PR updates App and API Protection setup pages to use the nested AAP verify setup shortcode.

Keeps the legacy verify setup shortcode available for translated pages that still reference it.

### Merge instructions

Merge readiness:
- [x] Ready for merge


janine-c (approved):
Great catch on the duplicate shortcode!

I think we'd be okay to remove the legacy shortcode to prevent people from using it. I think the way our build works, the new shortcode would propagate to the localized pages. But we can do that in a separate PR 🙂

PR #36629, opened 2026-05-13
Replace card grid HTML partials with Hugo shortcodes

hestonhoffman:
### What does this PR do? What is the motivation?

Replaces ~100 one-off Hugo partials that render link card grids with a single reusable `card-grid`/`image-card` shortcode pair. Authors can now define card grids directly in content files instead of creating a new HTML partial for each one.

**Before:** Every card grid required a bespoke partial in `layouts/partials/` with hardcoded HTML.

**After:**
```
{{< card-grid min_width="150px" >}}
  {{< image-card href="/path/" src="integrations_logos/logo.svg" alt="Logo" >}}
{{< /card-grid >}}
```

I also replaced instances in the translated docs to prevent the build from failing.

### Params

**card-grid**

| Param | Default | Description |
|-------|---------|-------------|
| `card_width` | `150px` | Card width. Controls columns per row. Accepts any CSS unit. |
| `image_width` | `150` | Default image width (px) for all child cards. |

**image-card**

| Param | Default | Description |
|-------|---------|-------------|
| `href` | — | Link target (required) |
| `src` | — | Image path. Omit for text-only cards. |
| `alt` | `""` | Image alt text |
| `title` | — | Label below image (or main text for text-only cards) |
| `subtitle` | — | Secondary text below the title |
| `image_width` | Inherits from parent | Overrides the parent's `image_width` for this card |
| `tooltip` | — | Bootstrap tooltip text on hover |

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes

- Design spec: `docs/superpowers/specs/2026-05-12-card-grid-shortcode-design.md`
- 7 deferred partials (conditional logic): `actions/expressions`, `apm/apm-compatibility`, `code_analysis/sca-getting-started`, `continuous_delivery/cd-getting-started`, `continuous_testing/ct-cicd-integrations`, `platforms/platforms`, `support/support`

PR #37654, opened 2026-06-22
Remove migrated AAP compatibility pages and repoint links

eliottness:

The per-language App and API Protection (AAP) compatibility pages and the cross-language compatibility index under `security/application_security/setup/` are now generated from the compatibility matrix and injected at build time. This PR removes the leftover in-repo copies and repoints all internal links to the consolidated `/security/application_security/setup/compatibility/<lang>` pages, matching the existing Go setup.

- Removed `setup/compatibility/_index.md` and `setup/{dotnet,java,nodejs,php,python,ruby}/compatibility.md`
- Repointed 47 references (41 content links + 6 shortcodes) from `setup/<lang>/compatibility` to `setup/compatibility/<lang>`
- Corrected a copy-paste link on the nginx setup page that pointed to the PHP compatibility page

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes

Translated content (`content/es`, `content/ko`) is managed externally and is not modified here.


rtrieu:
hi @eliottness thanks for this PR! it mostly looks good, but you will need to add aliases for the deleted pages so that users navigating to those pages won't 404. for example, `/security/application_security/_index.md` should have  `/security/application_security/threats/setup/compatibility/`. `content/en/security/application_security/setup/dotnet/docker.md` should have `/security/application_security/setup/dotnet/compatibility`, etc. can you add the missing aliases and then request a re-review when ready? thanks!

rtrieu (changes requested):
missing aliases

eliottness:
Thanks @rtrieu! Addressed:

**Per-language redirects** — added an alias for each removed page so the old URLs redirect instead of 404. `/security/application_security/setup/{dotnet,java,nodejs,php,python,ruby}/compatibility` are now aliases on each language's setup landing page (`setup/<lang>/_index.md`).

**Compatibility index** — I kept `setup/compatibility/_index.md` in the repo rather than re-homing its alias. That page's content is provided by the shared `websites-modules` Hugo module, but the module's current version still references the `security-platform/appsec-languages.html` partial that was removed in #36629 — so the repo file has to stay for now (it overrides the module copy; removing it fails the Hugo build). Keeping it preserves `/security/application_security/setup/compatibility/` and its `threats/setup/compatibility/` alias with no 404. Removing it will be a follow-up once the module is updated to emit the card-grid shortcode.

Re-requesting review — thanks!

rtrieu:
hi @eliottness thank you for making the changes! one thing to flag: the deleted `compatbility.md` files contained aliases for older `/threats/setup/compatibility/{lang}` URLs (dotnet, java, nodejs, ruby). those aliases are now gone, and the auto-generated `setup/compatibility/(lang)` pages would need to include them to avoid 404s for anyone using those old paths. this might be out of scope for this PR but worth tracking. do you know if the auto-generation system can be updated to carry those aliases over?

eliottness:
Good news — the auto-generated pages already carry those aliases, so nothing is lost. The module-generated `setup/compatibility/<lang>` pages include the old `/threats/setup/compatibility/<lang>` aliases (plus `/enabling/compatibility/<lang>` for dotnet/nodejs/ruby).

Verified on this PR's preview (where the repo `compatibility.md` files are deleted) — all resolve, no 404:
- `/security/application_security/threats/setup/compatibility/{dotnet,java,nodejs,ruby,php,python,go}` → 200
- `/security/application_security/enabling/compatibility/{dotnet,nodejs,ruby}` → 200

So this is already handled by the auto-generation system — no follow-up needed. (Before this PR those aliases lived on *both* the repo pages and the generated pages, i.e. duplicated; now they live only on the generated pages.)

Re-requesting review — thanks!

eliottness:
To close the loop on the tracking point, @rtrieu:

- **`/threats/setup/compatibility/<lang>`** — these are emitted by the generator for every language (verified 200 on the preview), so they carry over automatically; nothing is lost when the repo `compatibility.md` files are removed.
- **`/setup/<lang>/compatibility`** (the canonical URLs of the removed pages) — these aren't on the generator. This PR redirects them via aliases on each language's setup landing page (also 200 on the preview).

So there are no 404s either way. As an optional cleanup we can move the `/setup/<lang>/compatibility` redirects into the generator too, so they point straight at the new compat page and survive regeneration — I've written that up as a follow-up. It's not required for this PR.

rtrieu (approved):
thanks for making the changes!

PR #37778, opened 2026-06-25
Consolidate Azure Container Apps instrumentation hub

rohanagarwaldd:

Consolidates the flow to reach setup guide for Azure Container Apps instrumentation into a single landing page with an interactive method/language picker, updates the overview copy, and adds a "Best for" row to the comparison table. The goal of this change is to remove redundant unnecessary/redundant info and provide users with a faster time-to-instrumentation.

### DIFF
--- a/content/en/serverless/azure_container_apps/in_container/_index.md
+++ b/content/en/serverless/azure_container_apps/in_container/_index.md
@@ -1,14 +1,6 @@
 ---
 title: In-Container Instrumentation
-type: multi-code-lang
+external_redirect: /serverless/azure_container_apps/
 aliases:
   - /serverless/guide/aca_serverless_init/
 ---
-
-First, set up the **[Azure Integration][1]** to collect metrics and logs from Azure Container Apps.
-
-Then, instrument your application using one of the following guides:
-
-{{% container-languages path="azure_container_apps/in_container" %}}
-
-[1]: /integrations/azure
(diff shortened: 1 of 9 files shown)


content/en/serverless/azure_container_apps/in_container/_index.md
```diff
@@ -1,14 +1,6 @@
 ---
 title: In-Container Instrumentation
-type: multi-code-lang
+external_redirect: /serverless/azure_container_apps/
```
janine-c, line 3: I noticed that when I click the new "Go to [method] setup for [language]" button, none of the pages render, and this seems to be why. Without the `multi-code-lang` param and the `container-languages` shortcode in this and the sidecar index file, the setup docs have nowhere to actually live, and thus don't render. (I also noticed that the links to the `in-container` and `sidecar` pages are still in the left nav, which we should remove if we're going to make these pages redirect, but I'm not sure that's what we want to do!)

For next steps, I think we need to figure out where to make this content live, so we can link to it nicely. This could also be a good candidate for [Cdocs](https://datadoghq.atlassian.net/wiki/spaces/docs4docs/pages/5673910318/Cdocs+Customizable+docs), which gives us better filtering abilities for situations like this that have multiple options, and the benefit there would be that it wouldn't require designing any new components.

Give it a think, and let us know how we can help. If you have any questions, reach out anytime in #documentation!

janine-c (changes requested):
Hey Rohan, thanks for submitting this! I tested this on the staging links, and it looks like the setup pages all broke. I provided some ideas on next steps below; let me know if you'd like to chat further 🙂

PR #37790, opened 2026-06-26
reading results update ready for review

lukasgoetzweiss:
### AI assistance

Gave Claude a draft, edited output manually


content/en/experiments/reading_results.md
```diff
@@ -12,38 +12,96 @@ further_reading:
...
+<div class="alert alert-info"><strong>How metrics are calculated</strong><br><br>
+Datadog analyzes experiments at the <strong>subject</strong> level—the unit you configured when you set up the experiment, typically a user. Datadog computes a metric value for each enrolled subject (for example, revenue per user or whether the user completed a signup). These per-subject values form a distribution for each variant. Datadog's statistical engine then compares these distributions between control and treatment.<br><br>
+<strong>Relative lift</strong> measures how much the treatment shifted the average per-subject metric value compared to the control:<br><br>
+<pre><code>Relative lift = (Treatment − Control) / Control</code></pre>
+A relative lift of 10% means the treatment group's average per-subject value is 10% higher than the control group's average. Negative lift means the treatment performed worse on average.
+</div>
+
+### Confidence intervals
+
+The confidence interval is a range of lift values that the observed data supports. The true lift could fall outside this range, but values inside the interval are more consistent with what the experiment measured.
```
tbuffington7: I'd prefer to say "that are consistent with the observed data." The way frequentist hypothesis testing works is that it doesn't directly "support" specific values/null hypotheses, but rather it "rules out" hypotheses/values that (if true) would make the observed data very unlikely. I think it's technically more precise to say "not inconsistent with the observed data" but it reads like a double negative so I'm willing to compromise there haha.

content/en/experiments/reading_results.md
```diff
@@ -12,38 +12,96 @@ further_reading:
...
+
+- If the **entire interval is above zero**, the result is statistically significant in the positive direction. The improvement to the metric is unlikely to be attributable to random variation.
```
tbuffington7: Stats snobs would push back on these because they sound like Bayesian interpretations (the wording sounds like it's referring to P(lift is real | data) rather than P(observed lift at least this large | true effect = 0). I'd say something like "An improvement at least this large is unlikely to occur if there is no true effect" or even just "The observed lift is inconsistent with a no true effect."

content/en/experiments/reading_results.md
```diff
@@ -12,38 +12,96 @@ further_reading:
...
+- If the **entire interval is below zero**, the result is statistically significant in the negative direction. The treatment likely reduced the metric.
+- If the **interval crosses zero**, the result is not statistically significant. The observed lift may have occurred by chance.
```
tbuffington7: I'd say "the result is consistent with a true effect of zero" instead of "The observed lift may have occurred by chance." 

Lift may have occurrerd by chance sounds like it's making a statement about P(H0) not P(Data | H0)

content/en/experiments/reading_results.md
```diff
@@ -12,38 +12,96 @@ further_reading:
...
+### Global lift
+
+Experiments typically enroll only a subset of eligible users. Switch to the {{< ui >}}Global lift{{< /ui >}} tab on the metric scorecard to estimate how rolling out the treatment to all eligible users would affect your overall metric totals. See [Global Lift][7] for the full methodology.
+
+{{< img src="/product_analytics/experiment/exp_reading_global_lift.png" alt="The Global lift tab of the experiment scorecard showing control and treatment average metric values, coverage, and global lift for each decision metric." style="width:90%;" >}}
+
+For each metric, the {{< ui >}}Global lift{{< /ui >}} tab displays:
+
+- **Control and treatment values**: The average per-subject metric value in each variant—the same values shown on the main scorecard tab.
+- **Coverage**: The estimated proportion of your global metric total that would come from the eligible population under a control-only rollout.
```
tbuffington7: I think that the "control-only rollout" description sounds a little jargon-ey and obscures the nice intuition of coverage. Maybe something like:

"The estimated proportion of your global metric total associated with the experiment's eligible population (excluding the effect of the experiment)"

I really want people to grok coverage as e.g., the % of revenue exposed to the change being tested rather than some technical causal inference concept. I think the "control only" correction is an afterthought/implementation detail

static/images/product_analytics/experiment/exp_reading_exps_overview.png
brett0000FF, line 1: Can you please add this file back? We have a job that deletes outdated images. We have to leave this hear temporarily so that the image doesn't break on non-English pages. Thanks!

content/en/experiments/reading_results.md
```diff
@@ -12,38 +12,96 @@ further_reading:
...
+- **Confidence interval**: A range of lift values consistent with the observed data, shown as a bar centered on the relative lift estimate.
+
+The width and interpretation of the confidence interval depend on the experiment's configured [analysis method][2].
+
+<div class="alert alert-info"><strong>How metrics are calculated</strong><br><br>
+Datadog analyzes experiments at the <strong>subject</strong> level—the unit you configured when you set up the experiment, typically a user. Datadog computes a metric value for each enrolled subject (for example, revenue per user or whether the user completed a signup). These per-subject values form a distribution for each variant. Datadog's statistical engine then compares these distributions between control and treatment.<br><br>
+<strong>Relative lift</strong> measures how much the treatment shifted the average per-subject metric value compared to the control:<br><br>
+<pre><code>Relative lift = (Treatment − Control) / Control</code></pre>
+A relative lift of 10% means the treatment group's average per-subject value is 10% higher than the control group's average. Negative lift means the treatment performed worse on average.
+</div>
```
brett0000FF: This reads a bit too big to fit into a Note callout. I'd recommend pulling it back out into a section, or if you want to deemphasize it, you could put it inside a collapse-content shortcode.

brett0000FF (approved):
Thanks! The only blocking feedback from Docs is that we need to add back the deleted image.

PR #37816, opened 2026-06-29
Remove migration banner from findings schema page

hugoqnc:

Removes the learning center callout banner ("Learn about migrating to this new schema so you can avoid any interruptions to your workflows.") from the [Security Findings Schema Reference](https://docs.datadoghq.com/security/guide/findings-schema/) page.

image


estherk15 (approved):
Thanks for the update!

PR #37859, opened 2026-06-30
[SVLS-9408] Clarify cost-saving techniques for Step Functions, Logic Apps, and App insights Traces

ryanstrat:

Fixes SVLS-9408

Adds a clarification, shared across all logs to traces backed integration products, that log indexing costs can be reduced without breaking traces or enhanced metrics, as long as exclusion happens at the index level (after ingestion) rather than before logs reach Datadog.

Changes:
- New shared partial `layouts/partials/log-to-trace-indexing-note.html` so the wording stays consistent across products.
- Surfaced on the three integrations that rely on backend L2T conversion, at each page's indexing/exclusion decision point:
  - AWS Step Functions: `content/en/serverless/step_functions/troubleshooting.md`
  - Azure Logic Apps: `content/en/serverless/logic_apps/troubleshooting.md`
  - Azure App Insights integration: `content/en/tracing/guide/serverless_enable_azure_app_insights.md`

### Merge instructions

Merge readiness:
- [x] Ready for merge

### Additional notes

Vale passes with no new warnings on the changed files.


nine5two7 (approved):
LGTM

layouts/partials/log-to-trace-indexing-note.html
```diff
@@ -0,0 +1,3 @@
+<div class="alert alert-info">
+<strong>Note</strong>: To lower log indexing costs without losing traces or enhanced metrics, use <a href="/logs/log_configuration/indexes/#exclusion-filters">exclusion filters</a> at the index level instead of dropping logs before they reach Datadog. Index-level exclusion filters apply after ingestion, so they reduce what Datadog indexes and stores while leaving trace generation intact. Filtering or dropping these logs before they are ingested (for example, at the log source or forwarder) prevents Datadog from generating traces and enhanced metrics.
```
buraizu:
```suggestion
To lower log indexing costs without affecting traces or enhanced metrics, use <a href="/logs/log_configuration/indexes/#exclusion-filters">exclusion filters</a> at the index level instead of dropping logs before they reach Datadog. Index-level exclusion filters apply after ingestion, so they reduce what Datadog indexes and stores while leaving trace generation intact. Filtering or dropping these logs before they are ingested (for example, at the log source or forwarder) prevents Datadog from generating traces and enhanced metrics.
```

buraizu: The `alert-info` box automatically displays the text "Note", so it's unnecessary to add it here (screenshot):
alert-info-note

buraizu (changes requested):
Thanks for the PR, just requesting a couple of minor updates

layouts/partials/log-to-trace-indexing-note.html
ryanstrat: Applied your suggestion — changed "losing" to "affecting". Thanks!

buraizu (approved):

ryanstrat:
I see this is pending on a review from Corpweb. Should I not have used a partial for this?

buraizu (commented):
@ryanstrat thanks for following up. Taking a step back, I just wanted to confirm a couple of points about this change:

1. It looks like the new information hasn't been added to a couple of the pages mentioned in your description and _has_ been added to a page _not_ mentioned in your description. Specifically:

- `content/en/serverless/logic_apps/troubleshooting.md` and `content/en/serverless/step_functions/troubleshooting.md` don't receive the new shortcode, but only some unrelated changes. 
- `content/en/serverless/logic_apps/installation.md` does receive the new shortcode, but isn't mentioned in the description.

Could you please confirm if this is expected?

2. Additionally, the new note's placement on `/tracing/guide/serverless_enable_azure_app_insights/` (between "How it works" and "Prerequisites") seems potentially confusing. A user at this point isn't even sure if they meet the prerequisites, so a billing-forward note about log indexes seems out of place.

   A couple of potential solutions come to mind:
      1. Move the note underneath the numbered list in the "Prerequisites" section, which seems like the natural place for when a user would be making decisions about log inclusion/exclusion.
      2. Drop the note from this page entirely. This seems safe for this page, since it never otherwise mentions log indexing.

What do you think?

ryanstrat:
I forgot to update the PR description. Will fix that.

I pulled the note from the Step Functions page because it was a little redundant to the content on that page. I still kept it as a partial because I anticipate needing this note on additional future products that share the same backend infra as Step Functions / Logic Apps / App Insights. 

I'll move the note to the bottom of the prerequisites setting. It is intended to address a customer concern that might exist while configuring the logs forwarding automation. It might be a little premature while this integration is in preview.

PR #37888, opened 2026-07-01
Document how monitor renotification impacts On-Call Pages

SimoneT-DD:
### Merge readiness

- [x] Ready for merge

### DIFF
--- a/content/en/incident_response/on-call/pages/_index.md
+++ b/content/en/incident_response/on-call/pages/_index.md
@@ -45,6 +45,14 @@ If you send a Page through a monitor alert and the Team's routing rule uses dyna
 - If the WARN threshold is crossed, the Page urgency is set to `low`.
 - If the ALERT threshold is crossed, the Page urgency is set to `high`.
 
+#### Monitor renotification
+
+When a monitor is configured to [send renotifications][8] to an On-Call Team, the behavior depends on the current state of the Page:
+
+- **Page is resolved**: The monitor re-triggers and creates a new Page, which is routed through the Team's escalation policies.
+- **Page is acknowledged**: The monitor re-triggers and moves the existing Page back to `Triggered` state, restarting the escalation policies from the beginning.
+- **Page is triggered and escalation is exhausted**: If all steps in the escalation policy have been executed but no one has acknowledged the Page, a renotification from the monitor re-triggers the escalation policy from the beginning.
+
 ### Trigger Pages through email
 
 Generate a unique email address to trigger a Page directly to a Team's on-call responders. Emails sent to this address follow the Team's configured routing and escalation policies.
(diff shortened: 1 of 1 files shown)


content/en/incident_response/on-call/pages/_index.md
```diff
@@ -45,6 +45,14 @@ If you send a Page through a monitor alert and the Team's routing rule uses dyna
 - If the WARN threshold is crossed, the Page urgency is set to `low`.
 - If the ALERT threshold is crossed, the Page urgency is set to `high`.
 
+#### Monitor renotification
+
+When a monitor is configured to [send renotifications][8] to an On-Call Team, the behavior depends on the current state of the Page:
+
+- **Page is resolved**: The monitor re-triggers and creates a new Page, which is routed through the Team's escalation policies.
+- **Page is acknowledged**: The monitor re-triggers and moves the existing Page back to `Triggered` state, restarting the escalation policies from the beginning.
```
OctaviPascual: `the monitor re-triggers` -> I don't think this is the right wording, a monitor re-notifies, it doesn't re-trigger. It's the Page that is re-triggered.

`restarting the escalation policies from the beginning` -> this is false, the behavior is this one: `Restart the same step that was being executed when the Page was acknowledged` as stated in [the RFC](https://datadoghq.atlassian.net/wiki/spaces/Bugle/pages/6228443307/RFC+Re-trigger+page+on+monitor+re-notification#Problem)

content/en/incident_response/on-call/pages/_index.md
```diff
@@ -45,6 +45,14 @@ If you send a Page through a monitor alert and the Team's routing rule uses dyna
...
+- **Page is triggered**: If all steps in the escalation policy have been executed already but no one has acknowledged the Page, a renotification from the monitor re-triggers the escalation policy from the beginning.
```
OctaviPascual: `If all steps ...` -> I would add in the end: `Otherwise, the renotification has no effect`. An alternative can be: `The renotification has no effect unless all steps in the escalation policy have been executed. If they have, the Page re-triggers the escalation policy from the beginning.`

`re-triggers the escalation policy from the beginning` -> this is true today, but the desired behaviour is `If the EP completed all its steps, the re-notification will re-trigger the page on the last step (even if the page is in Triggered state).`
We identified the bug with @ClairePhi, the problem is that it's the whole re-trigger mechanism that has this bug, so it also impacts ack timeout and that's why we haven't fixed it yet. But soon this `from the beginning` won't be true anymore, so let's not forget to update it.

estherk15: I changed to `escalation policy` for consistency, but if plural is correct, I recommend changing all three bullets.

Also, I updated wording to address @OctaviPascual's comments.

```suggestion
- **Page is resolved**: The monitor renotifies and creates a new Page, which is routed through the Team's escalation policy.
- **Page is acknowledged**: The monitor renotifies, and the Page is re-triggered, resuming the escalation policy at the step that was in progress when the Page was acknowledged. 
- **Page is triggered**: If all steps in the escalation policy have already run but no one has acknowledged the Page, the monitor renotifies and the Page is re-triggered, restarting the escalation policy from the beginning.
```

estherk15 (approved):
Left a suggestion to address feedback, but this look good, thanks for the update!

PR #37890, opened 2026-07-01
[SDTEST-3880] Document Android test support in Java Test Optimization

ypopovych:

Fixes SDTEST-3880

Adds an Android section to the Java Test Optimization setup page:
- Android unit tests are fully supported.
- UI and integration tests are supported only through Robolectric, with a link to the Robolectric documentation.
- Tests running on an Android emulator or a physical device are not supported.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes

### DIFF
--- a/content/en/tests/setup/java.md
+++ b/content/en/tests/setup/java.md
@@ -57,6 +57,14 @@ Other build systems, such as Ant, Bazel, or SBT are supported with the following
 - Automatic coverage configuration and reporting is not supported.
 - When building a multi-module project, every module is reported in a separate trace.
 
+### Android
+
+Android unit tests are fully supported. UI and integration tests are supported only through [Robolectric][11], which lets these tests run on the JVM. If you haven't used Robolectric before, see the [Robolectric documentation][11] to learn how to set it up.
+
+Tests that run on an Android emulator or a physical device **are not supported**.
+
+Configure Android tests the same way as other Gradle-based Java tests, applying `GRADLE_OPTS` to the Gradle task that runs your Robolectric-based tests (for example, `testDebugUnitTest`). See the [Gradle tab](#running-your-tests) for configuration details.
+
 ## Setup
 
 You may follow interactive setup steps on the [Datadog site][2] or the instructions below.
@@ -571,3 +579,4 @@ To disable all integrations, augment the list of `-javaagent` arguments with `dd
 [8]: /tests/#parameterized-test-configurations
 [9]: https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-display-names
 [10]: /tracing/trace_collection/compatibility/java#integrations
+[11]: https://robolectric.org/getting-started/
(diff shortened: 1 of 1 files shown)


content/en/tests/setup/java.md
```diff
@@ -57,6 +57,14 @@ Other build systems, such as Ant, Bazel, or SBT are supported with the following
 - Automatic coverage configuration and reporting is not supported.
 - When building a multi-module project, every module is reported in a separate trace.
 
+### Android
+
+Android unit tests are fully supported. UI and integration tests are supported only through [Robolectric][11], which lets these tests run on the JVM. If you haven't used Robolectric before, see the [Robolectric documentation][11] to learn how to set it up.
```
daniel-mohedano: The Robolectric documentation itself describes the framework as "Robolectric is a framework that brings fast and reliable unit tests to Android", so I'm wary of specifying unit tests vs UI and integration tests.

I'd rephrase the sentence to something along the lines of "Android testing is supported when using the Robolectric framework. Tests that run on an Android emulator or physical device **are not supported**." 

I'm also not sure about linking to their documentation, as we don't do that for other frameworks.

content/en/tests/setup/java.md
```diff
@@ -57,6 +57,14 @@ Other build systems, such as Ant, Bazel, or SBT are supported with the following
...
+
+Tests that run on an Android emulator or a physical device **are not supported**.
+
+Configure Android tests the same way as other Gradle-based Java tests, applying `GRADLE_OPTS` to the Gradle task that runs your Robolectric-based tests (for example, `testDebugUnitTest`). See the [Gradle tab](#running-your-tests) for configuration details.
```
daniel-mohedano: If we rephrase the first two paragraphs, we could move them between the compatibility table and the "If your test framework is not supported..." sentence, which might flow better to the supported build systems section after and let us remove this paragraph. 

If you think it might be better to keep the specific `Android` section, then I'd simplify the wording to something like "Configure Android tests the same way as other Gradle-based JVM tests. See the [Gradle section](#running-your-tests) for configuration details." Or even pointing them just to the `Setup` section overall, given that auto instrumentation should also work for these users

daniel-mohedano (changes requested):

content/en/tests/setup/java.md
ypopovych: Main problem that we also support JUnit4/5 JVM tests on Android and a lot of them don't need Robolectric at all. So Robolectric is needed only if you tests depend on some Android API. And Robolectric by itself isn't a test framework, it is more like a mock framework with mocks collection for the Android APIs. We need to tell somehow that we don't support emulator/device but you can use Robolectric to mock them in your tests

content/en/tests/setup/java.md
```diff
@@ -57,6 +57,12 @@ Other build systems, such as Ant, Bazel, or SBT are supported with the following
 - Automatic coverage configuration and reporting is not supported.
 - When building a multi-module project, every module is reported in a separate trace.
 
+### Android
+
+Android tests that run on the JVM are supported. Tests that depend on the Android API, such as Espresso tests, Compose UI tests, and some unit tests, are supported only with the [Robolectric][11] framework.
+
+Tests that require an emulator or a physical device **are not supported**.
```
domalessi:
```suggestion
Tests that require an emulator or a physical device are not supported.
```

domalessi (approved):
One minor tweak, but otherwise looks good!

daniel-mohedano (approved):

PR #37909, opened 2026-07-01
(maint) Add claude code default permissions

hestonhoffman:

Adds some sane (read-only) defaults for Claude Code.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -13,6 +13,8 @@
   },
   "permissions": {
     "allow": [
+      "Grep",
+      "Ls",
       "Bash(gh pr view *)",
       "Bash(gh pr list *)",
       "Bash(gh pr diff *)",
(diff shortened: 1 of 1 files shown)


domalessi (approved):
Ty!

.claude/settings.json
```diff
@@ -10,5 +10,31 @@
   "enabledPlugins": {
     "marketplace-auto-update@datadog-claude-plugins": true,
     "documentation@datadog-claude-plugins": true
+  },
+  "permissions": {
+    "allow": [
+      "Grep",
+      "Ls",
```
clreaume, line 17:
```suggestion
      "Bash(grep *)",
      "Bash(ls *)"
```
when I asked Claude to add this to my local file, it did so in this fashion, which matches the gh pattern below. I'm not sure your version wouldn't work, though!

hestonhoffman: I thought of doing this here, but the grep command in particular is often combined with/piped into write commands, which makes me a bit more nervous about allowing the wildcard in this case 😬

hestonhoffman: I think this is fine for a local solution though, if you're comfortable with it

clreaume: ah, that makes a lot of sense! thank youu

PR #37941, opened 2026-07-02
Add CUPED variance reduction documentation page

tbuffington7:
## Summary

Adds a new experiments methodology page documenting CUPED (Controlled-experiment Using Pre-Experiment Data), our default variance reduction technique.

The page covers:
- Overview: why variance reduction matters and what CUPED does
- How CUPED works: zero-in-expectation adjustments added to the raw lift
- Default covariates (30-day pre-exposure aggregations + subject properties) and missing-value handling
- How coefficients are selected (variance minimization, not regression adjustment)
- Variance reduction across mean, ratio, and percentile metrics
- FAQ addressing common misconceptions

Also wires the page into the experiments nav and cross-links it from Analysis Methods.


buraizu:
Thanks for the PR! Created DOCS-15065 for documentation team review

lukasgoetzweiss (approved):

content/en/experiments/statistics/analysis_methods.md
```diff
@@ -73,7 +76,7 @@ The prior matters most when sample sizes are small. With enough data, the poster
 
 Analysis methods are only one part of the statistical analysis plan. Datadog Experiments also supports modifying the following [settings][1]:
 
-CUPED
+[CUPED](/experiments/statistics/cuped)
```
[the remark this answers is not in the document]
drichards-87: Just adding one more bit of context: another reason we use the reference-style link format is for localization and translation purposes.

content/en/experiments/statistics/cuped.md
```diff
@@ -0,0 +1,139 @@
...
+
+### If CUPED and non-CUPED results differ, which one should I trust?
+
+Use CUPED results when CUPED's weak assumptions are met: they have the same expectation as non-CUPED results, but with lower variance. Proper data management practices reduce the risk of assumption violations, and diagnostic checks can detect pre-experiment imbalance and dimensional sample ratio mismatch.
+
+<div class="alert alert-info">Decide whether the decision is based on the CUPED or non-CUPED result before the experiment starts. This avoids selecting the result that best supports a preferred decision.</div>
+
+### Does CUPED correct for bias in an improperly randomized experiment?
+
+No. CUPED assumes that the experiment is properly randomized, so it does not recover statistically valid estimates from systematically biased experiments. However, random covariate imbalances still arise in properly randomized experiments. CUPED corrects for these imbalances in the sense that the CUPED-adjusted estimates are *conditionally* unbiased for a specific observed covariate imbalance.
```
drichards-87:
```suggestion
## Assumptions and interpretation

### Modeling assumptions

CUPED does not assume linear relationships, homoscedasticity, or the absence of interactions. Its main assumption is that the covariate means across variants are equal in expectation. The other assumptions are common to digital experimentation in general, such as no confounding through proper randomization, and samples large enough for the estimators to be approximately normal by the central limit theorem.

### Why CUPED lift estimates differ from non-CUPED estimates

CUPED does not produce the same lift estimates with narrower confidence intervals. If CUPED and non-CUPED estimates were always identical, their variance, and therefore their confidence intervals, would also be identical. The difference between CUPED and non-CUPED results is the exact mechanism that drives CUPED's reduced variance.

### Choosing between CUPED and non-CUPED results

Use CUPED results when CUPED's assumptions are met. They have the same expectation as non-CUPED results, but with lower variance. Proper data management practices reduce the risk of assumption violations, and diagnostic checks can detect pre-experiment imbalance and dimensional sample ratio mismatch.

<div class="alert alert-info">Before the experiment starts, decide whether to base your decision on the CUPED or non-CUPED result. This avoids selecting the result that best supports a preferred outcome.</div>

### CUPED and improperly randomized experiments

CUPED does not correct for bias caused by improper randomization. It assumes the experiment is properly randomized, so it cannot recover statistically valid estimates from systematically biased experiments. Random covariate imbalances still arise in properly randomized experiments, and CUPED corrects for these: the CUPED-adjusted estimates are *conditionally* unbiased for a specific observed covariate imbalance.
```

drichards-87: One thing to flag here: we generally steer away from Q&A-style FAQ sections in Datadog docs. The Content Types guidance in our internal Confluence has the reasoning behind this, but the short version:

- We tend not to use question-and-answer content, since headings phrased as questions are harder to scan and not great for SEO.
- "FAQ" also means something specific in our information architecture: a private, un-indexed page in an /faq folder for edge-case workarounds shared by direct link, which is different from what's here.

So my suggestion is mostly about moving the existing content into a regular section with descriptive headings instead of questions. I've left some suggested wording that keeps everything intact.

content/en/experiments/statistics/cuped.md
```diff
@@ -0,0 +1,139 @@
...
+
+Standard experiment analysis compares metric data from subjects exposed to a treatment with metric data from subjects in a control group. All data used in that comparison is collected during the experiment. Teams often have information about subjects from before the experiment started, especially historical metric data. CUPED uses this pre-experiment data to control for some of the metric variance that comes from randomly assigning variants to subjects. For example, one variant might have more active subjects than another variant by random chance. CUPED reduces the effect of this random variation by controlling for different activity levels across variants.
+
+CUPED is enabled by default. With CUPED enabled, the displayed lift and metric values may differ from the naive estimates calculated from the raw data. The raw data is unchanged; CUPED changes only the estimator used for analysis. This difference is expected and is the mechanism that drives the reduced variance. You can toggle CUPED in the experiment's [statistical analysis plan][2].
```
drichards-87:
```suggestion
CUPED is enabled by default. With CUPED enabled, the displayed lift and metric values may differ from the unadjusted estimates calculated from the raw data. The raw data is unchanged; CUPED changes only the estimator used for analysis. This difference is expected and is the mechanism that drives the reduced variance. You can toggle CUPED in the experiment's [statistical analysis plan][2].
```

drichards-87 (changes requested):
Left some feedback from Docs.

drichards-87:
@tbuffington7 I just pushed a change to fix the merge conflict. I'll approve the PR as soon as the build job checks finish running.

drichards-87 (approved):

PR #37785, opened 2026-06-26
📝 Add schema view and drift comparison to API endpoints docs

mauricesvay:
### What does this PR do? What is the motivation?

Adds a new "View and compare API schemas" section to the API endpoints page, covering:
- How to read an endpoint's inferred OpenAPI schema (Preview and YAML views)
- How to compare declared vs inferred schemas to detect drift
- Setup requirements (AAP enabled, Catalog registration)

### Merge readiness

- [ ] Ready for merge

### AI assistance

Used Claude Code for initial draft.

### Additional notes

PR #37946, opened 2026-07-03
📝 Crop API endpoint schema screenshot

mauricesvay:

Crops the screenshot added in #37785 to remove the left-hand navigation, per review feedback.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes


estherk15 (approved):
Thanks for the update! A heads up, deleting the image may cause cacheing issues. Quick fix is to add the new image with a new filename without deleting the old image.

mauricesvay:
> Thanks for the update! A heads up, deleting the image may cause cacheing issues. Quick fix is to add the new image with a new filename without deleting the old image.

@estherk15 Thanks for the review, I've updated the screenshot path to avoid caching issues.

PR #37952, opened 2026-07-03
djm k8s ssi doc fix: remove enabledNamespaces, just keep targets

aboitreaud:

With the version requirements we make, targets namespaces and enabledNamespaces are not supported when both set at the same time so injection is not working. 

targets match names namespaceSelector is the correct way of specifying which namespace should be injected

### Merge readiness

- [ ] Ready for merge

## For Datadog employees:


estherk15 (approved):
Thanks for the update!

PR #37995, opened 2026-07-06
Add SQL template variables to warehouse metric and exposure docs

htrajan:
## What does this PR do?

Documents the `{{variable}}` SQL template variables that Datadog Experiments customers can use in **Metric SQL Models** and **Exposure SQL Models** for warehouse-native experiments. Template variables let customers push Datadog's date filters into their own SQL so each pipeline run scans only the data it needs.

Adds a **"Filter with template variables"** section to two existing articles:

- `content/en/experiments/defining_metrics.md` (Warehouse tab, Metric SQL Model)
- `content/en/experiments/concepts/exposure_sql.md` (Exposure SQL Model)

Each section includes a reference table of the supported variables, the ISO 8601 / quoting rules, and a worked SQL example:

- `{{analysis_start_timestamp}}` / `{{analysis_end_timestamp}}` — the window the current update materializes (full window on a full refresh; the current tranche on an incremental refresh)
- `{{experiment_events_start_timestamp}}` / `{{experiment_events_end_timestamp}}` — the full experiment event window, even during incremental refreshes
- `{{assignments_start_timestamp}}` / `{{assignments_end_timestamp}}` — the assignment window
- `{{experiment_key}}`, `{{experiment_keys}}`, `{{allocation_key}}` — flag-allocation keys, rendered as quoted string literals

## Motivation

Customers with large warehouse tables need a documented way to limit what the experiment pipeline scans on each run. These variables already exist in the product but were undocumented.

## Review notes

- Content-only change to two existing articles (+55 lines); no menu or structural changes.
- Uses plain `{{var}}` syntax, consistent with how existing docs (for example, `monitors/types/slo.md`) render literal template variables.

Made with [Cursor](https://cursor.com)


brett0000FF (approved):

lukasgoetzweiss:
@htrajan I worry that this current placement takes a lot of real estate for people first learning about warehouse native architecture. This is a bit of an advanced feature, so I wonder if we can break it out into a separate page under "concepts" and link there from the exposure and metric SQL models page. That also reduces the need to duplicate the list of variables on multiple pages

Thoughts?

htrajan:
> I wonder if we can break it out into a separate page under \"concepts\" and link there from the exposure and metric SQL models page. That also reduces the need to duplicate the list of variables on multiple pages

@lukasgoetzweiss Agreed—I've implemented this. The full SQL template variable reference now lives on a new Concepts page, both the Metric and Exposure SQL Model pages link to it, and the approved variable descriptions and examples are preserved on the consolidated page.

lukasgoetzweiss (approved):

PR #38009, opened 2026-07-07
Document IP allowlist support for the MCP Server

rgwood-dd:

Admins looking to restrict who can connect to the Datadog MCP Server had no way to find that the IP allowlist covers it. This cross-links the two docs: the MCP Server setup page now points to the IP allowlist as a way to block connections from unapproved origins, and the IP allowlist page lists the MCP Server among the resources it protects.

This is motivated by some users asking for this functionality in the MCP server and not realizing we already have it.

### Merge instructions

Merge readiness:
- [x] Ready for merge


cswatt (approved):

aaronherrmann (commented):
LGTM! 👍

PR #38019, opened 2026-07-07
[OTAGENT-1152] Document custom DDOT Collector config on ECS Fargate

songy23:

Fixes OTAGENT-1152

Adds a "Use a custom Collector configuration" section to the DDOT Collector ECS Fargate install guide, explaining how to supply a custom `otel-config.yaml` by building a custom Agent image, since ECS Fargate doesn't support host volumes and the Collector always reads its config from a fixed in-container path.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes

### DIFF
--- a/content/en/opentelemetry/setup/ddot_collector/install/ecs_fargate.md
+++ b/content/en/opentelemetry/setup/ddot_collector/install/ecs_fargate.md
@@ -164,6 +164,37 @@ Refer to your language's OpenTelemetry SDK documentation to add the ECS resource
 - **Go**: [go.opentelemetry.io/contrib/detectors/aws/ecs][5]
 - **All languages**: See the [ECS resource detector documentation][6] in the OpenTelemetry Collector contrib repository.
 
+### Use a custom Collector configuration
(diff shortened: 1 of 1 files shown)


songy23:
### Validation

Tested this end to end on a real ECS Fargate deployment to confirm the documented steps work as written:

- Built a custom image (`FROM public.ecr.aws/datadog/agent:latest-full` + `COPY otel-config.yaml /etc/datadog-agent/otel-config.yaml`)
- Pushed it to ECR and ran it as an ECS Fargate task
- Confirmed the custom `otel-config.yaml` is present at the fixed path inside the running container, and that the DDOT Collector starts successfully and builds its pipelines from the custom configuration

This confirms the custom-image approach is correct, and that there's no other supported way to override the Collector config path on ECS Fargate (no host volumes, and the entrypoint always passes a hardcoded `--config` flag). All test resources were torn down after validation.

brett0000FF (approved):
Looks great! 🚀

PR #38026, opened 2026-07-08
Add attribute precedence table for log pipelines/processors

10kaoru12:

Adds an **Attribute precedence** section to the Log Pipelines page that explains, in one place, what happens when the same attribute is set by more than one processor across matching pipelines.

Today this behavior is only documented piecemeal on individual processor pages (for example, the "only the first one is considered" note on the log status remapper, and the "last one is taken into account" note on the log date remapper), and several processors document nothing at all. Customers configuring multiple pipelines/processors have no single reference for which write wins.

The new section groups all processors into three behaviors:
- **Last write wins** (Grok parser, Category, Arithmetic, String builder, Lookup, URL/User-Agent/GeoIP parsers, Decoder)
- **Depends on `override_on_conflict`** (Remapper, Array processor; default `false` = do not override)
- **First write wins** (Log status/Service/Log message/Trace/Span remappers; the Log date remapper is the exception where the last one wins)

Wording is reused verbatim from the existing processor pages to stay consistent.

### Merge readiness

- [ ] Ready for merge

## For Datadog employees:

- Branch follows the `<name>/<description>` convention.

### AI assistance

Drafted with Claude Code. Wording was reused verbatim from existing processor pages (log_status_remapper, service_remapper, log_message_remapper, log_date_remapper, remapper). The precedence behavior was independently verified against the logs-backend source before writing.

### Additional notes

English (`content/en`) only, per the contributing guidelines (translated files are managed by the translation service). Opened as a draft to review the preview build and Vale results before marking ready.


estherk15:
Thanks for the update @10kaoru12, I moved the section down the page so that it follows the processor information(https://github.com/DataDog/documentation/pull/38042). Let me know if you have any questions

10kaoru12:
@estherk15 
Thanks for the update! I've approved #38042. Let me know if there's anything else needed.

content/en/logs/log_configuration/pipelines.md
```diff
@@ -286,6 +286,18 @@ Move a pipeline into another pipeline to make it into a nested pipeline:
...
 1. Click {{< ui >}}Move{{< /ui >}}.
 
+## Attribute precedence when multiple processors match {#attribute-precedence}
+
+When multiple processors within matching pipelines set the same attribute, the outcome depends on the processor type. There are three behaviors:
+
+| Behavior | Description | Processors |
+| --- | --- | --- |
+| Last write wins | The value set by the later processor (further down the order) overrides the earlier value. | Grok parser, Category processor, Arithmetic processor, String builder processor, Lookup processor, URL parser, User-Agent parser, GeoIP parser, Decoder processor |
+| Depends on `override_on_conflict` | Follows the `override_on_conflict` parameter. By default (`false`), the target element is not overridden if it is already set. | Remapper, Array processor |
```
estherk15: @10kaoru12 This row was flagged. The backend points to array may process not array processor. Can you confirm before merging?

estherk15 (approved):

content/en/logs/log_configuration/pipelines.md
10kaoru12: @estherk15 Thanks for the note! 
Could you share a bit more detail on what you're seeing? 
My understanding from looking at the source is that ArrayProcessor (array-processor type) does use override_on_conflict, specifically in its ExtractKeyValueOperation — which controls whether an already-set target attribute gets overwritten when the extracted key-value map is written out.

estherk15: Maybe I'm looking at the wrong file, but I only see this on the Array Map Processor, not the Array processor: https://github.com/DataDog/datadog-api-spec/blob/24f67ad08ba98c351d3f2ddc2be6c6859466eff4/spec/v1/logs_pipelines.yaml#L1384

10kaoru12: Fixed the table entry from "Array processor" to "Array Map processor" in de703f82. Thanks for catching this!

PR #38130, opened 2026-07-13
fix: remove incorrect escaping from $basearch in the Datadog Security CLI repository URL 

nikhilvarpe:
Remove the unnecessary escape character before the $basearch variable in the Datadog Security CLI YUM repository URL for Red Hat/CentOS.

The escaped value (\$basearch) caused YUM/DNF to generate an invalid repository path, resulting in a 404 error while fetching repository metadata. Using $basearch directly allows YUM/DNF to correctly resolve the system architecture (for example, x86_64) and access the appropriate package repository.

This change ensures that the Datadog Security CLI can be installed successfully using the package repository instructions for Red Hat/CentOS and other supported RPM-based systems.

**The attached file consist of code snippet of the tested code :** [security-cli-code-snippet.txt](https://github.com/user-attachments/files/29977240/security-cli-code-snippet.txt)


maycmlee:
Hi @nikhilvarpe, thanks for opening this PR! We'll review and be back in touch.

Bit-Doctor (approved):

janine-c (approved):
Got approval from an engineer too for due diligence's sake 🙂 Thank you for the catch, @nikhilvarpe!

PR #38182, opened 2026-07-15
Add US1-FED support warning for Case Management AI features

roxanne-moslehi:

Adds inline warnings noting that AI features for Case Management (MCP Server integration and custom agents) are not supported on the US1-FED site.

- `incident_response/case_management/ai/_index.md`
- `incident_response/case_management/ai/custom_agents.md`

### Merge instructions

Merge readiness:
- [x] Ready for merge

### Additional notes


domalessi (commented):
Thanks for filing, @roxanne-moslehi! Since the banner applies to the full page, the preferred approach is [this](https://datadoghq.atlassian.net/wiki/spaces/docs4docs/pages/5201857960/Documenting+site+support#Use-a-site-support-ID-in-the-front-matter-%5BinlineExtension%5D). Let me know if you need any help with that!

roxanne-moslehi:
@domalessi that pages mentions to use the manual entry like I have if we need specific language. I don't want it to say the entire product isn't supported since it's just specific features in case management that aren't supported in gov. Lmk what we should do

domalessi:
@roxanne-moslehi Got it! In that case, I'm cool with this PR. Only other Q - do we need this call-out for FED-2 as well?

roxanne-moslehi:
fed2 is not supported for the full product for now so it already has a warning

domalessi (approved):

PR #38350, opened 2026-07-21
Remove unstable tags from Case Management APIs

sysy6r:
## What changed

Removes `x-unstable` annotations from all 43 Case Management API v2 operations and regenerates the API navigation metadata.

## Why

These operations are documented under `/api/v2`, but their stale annotations caused the documentation to display them as unstable.

## Impact

Case Management v2 API endpoints will no longer display the preview/unstable label or be hidden as unstable in the API navigation 
image

The generated clients will also don't marking those apis as unstable

### DIFF
--- a/config/_default/menus/api.en.yaml
+++ b/config/_default/menus/api.en.yaml
@@ -8926,8 +8926,7 @@ menu:
           - v2
         operationids:
           - UpdateMaintenanceWindow
-        unstable:
-          - v2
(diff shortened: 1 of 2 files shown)


PezCoder (approved):

sysy6r:
this is wrong, I will instead do a PR in datadog-api-spec repo, that will automatically update this one repo

PR #38363, opened 2026-07-21
[DOCS-14949] Add troubleshooting item about source sends canceled

maycmlee:
### Merge readiness

- [ ] Ready for merge

## For Datadog employees:


jhgilbert:
[Automated message] This PR has merge conflicts created by the [docs repo reorg project](https://github.com/DataDog/documentation/blob/master/REPO_REORG.md). Because this PR is marked as a work in progress, no attempt was made to auto-resolve the conflicts. When your PR is finished, you can queue your PR for an auto-fix by removing the `WORK IN PROGRESS` label and the `astro-reorg-skip` label. Your PR will be processed in the next batch of attempted auto-fixes.

hugo/content/en/observability_pipelines/monitoring_and_troubleshooting/troubleshooting.md
```diff
@@ -144,6 +144,18 @@ The curl command you use is based on the port you are using, as well as the path
 
 If you see the error `Too many files` and the Worker processes repeatedly restart, it could be due to a low file descriptor limit on the host. To resolve this issue for Linux environments, set `LimitNOFILE` in the systemd service configuration to `65,536` to increase the file descriptor limit.
 
+### Source send canceled error
+
+If you see error logs with the reason `Source send canceled`, events at the source level are getting dropped. This could be due to several reasons, including backpressure or Worker shut down or restarts. In the case of shut down or restarts, the error occurs when a source shuts down or restarts while events are in transit.
```
maycmlee:
```suggestion
If you see error logs with the reason `Source send canceled`, events at the source level are getting dropped. This could be due to several reasons, including backpressure or Worker shut down or restarts. An error is logged when a source shuts down or restarts while events are in transit.
```

p-parekh (approved):
lgtm

hugo/content/en/observability_pipelines/monitoring_and_troubleshooting/troubleshooting.md
```diff
@@ -151,6 +151,18 @@ The curl command you use is based on the port you are using, as well as the path
 
 If you see the error `Too many files` and the Worker processes repeatedly restart, it could be due to a low file descriptor limit on the host. To resolve this issue for Linux environments, set `LimitNOFILE` in the systemd service configuration to `65,536` to increase the file descriptor limit.
 
+### Source send canceled error
```
jackie8c: This error message actually updated [here](https://github.com/vectordotdev/vector/commit/c96234d1d6c9ecc1b78765fbb0e2c9bb322d567c) in Vector to `"Source send interrupted mid-flight; pipeline may be overloaded or shutting down."`

hugo/content/en/observability_pipelines/monitoring_and_troubleshooting/troubleshooting.md
```diff
@@ -151,6 +151,18 @@ The curl command you use is based on the port you are using, as well as the path
...
+
+If you see error logs with the reason `Source send canceled`, events at the source level are getting dropped. This could be due to several reasons, including backpressure or Worker shut down or restarts. An error is logged when a source shuts down or restarts while events are in transit.
+
+To determine if the error is due to backpressure, use the [Observability Pipelines Overview][29] dashboard to troubleshoot. You can filter by pipelines ID, host, Worker ID, and components. Check the following:
```
jackie8c: Could we soften the “determine if” to something like “help determine whether”? Identifying backpressure requires correlating buffer utilization and CPU metrics with shutdown, restart, or configuration-reload logs around the same timestamp- the available telemetry may not necessarily distinguish the cause conclusively

hugo/content/en/observability_pipelines/monitoring_and_troubleshooting/troubleshooting.md
```diff
@@ -151,6 +151,18 @@ The curl command you use is based on the port you are using, as well as the path
 
 If you see the error `Too many files` and the Worker processes repeatedly restart, it could be due to a low file descriptor limit on the host. To resolve this issue for Linux environments, set `LimitNOFILE` in the systemd service configuration to `65,536` to increase the file descriptor limit.
 
+### Source send canceled error
+
+If you see error logs with the reason `Source send canceled`, events at the source level are getting dropped. This could be due to several reasons, including backpressure or Worker shut down or restarts. An error is logged when a source shuts down or restarts while events are in transit.
```
jackie8c: 1. Before saying that “events at the source level are getting dropped,” could we explain a little more why they are considered dropped? Something like: 

> “... the send operation was interrupted before all events were handed off downstream. The remaining events are counted as dropped for that send attempt.”

2. Could we remove the last sentence of this paragraph? I think it may get confusing if we say that this error is logged when a source shuts down/restarts, when technically, the error is triggered when the source send action is interrupted with events still unsent. 

Also, depending on the source's behavior, the events may be retried or redelivered, so we could add that?

hugo/content/en/observability_pipelines/monitoring_and_troubleshooting/troubleshooting.md
maycmlee:
```suggestion
### Source send interrupted mid-flight
```

hugo/content/en/observability_pipelines/monitoring_and_troubleshooting/troubleshooting.md
maycmlee: > The remaining events are counted as dropped for that send attempt. 

When you say counted as dropped, does that mean the events aren't actually dropped?

> depending on the source's behavior, the events may be retried or redelivered 

Is there a specific example of when there would be a retry or redelivery that we could provide? What's the difference between retry and redeliver? Or am I getting too much into the weeds? 😅 

Also, is there a way for the user to know that the error is due to a Worker shut down or restart, rather than backpressure?

```suggestion
If you see `Source send interrupted mid-flight; pipeline may be overloaded or shutting down` error logs, the send operation was interrupted before all events were handed off downstream. Any remaining events are counted as dropped for that send attempt. Possible causes include backpressure, Worker shutdown, or Worker restarts. Depending on the source, the events may be retried or redelivered.
```

hugo/content/en/observability_pipelines/monitoring_and_troubleshooting/troubleshooting.md
maycmlee: What do you think about using `investigate`? I think that might leave it more open and less conclusive?

```suggestion
To investigate whether the error is due to backpressure, use the [Observability Pipelines Overview][29] dashboard to troubleshoot. You can filter by pipelines ID, host, Worker ID, and components. Check the following:
```

jackie8c: yep that sounds good!

hugo/content/en/observability_pipelines/monitoring_and_troubleshooting/troubleshooting.md
jackie8c: > When you say counted as dropped, does that mean the events aren't actually dropped?

the events are indeed dropped by the Worker and not handed downstream. OPW also increments the `component_discarded_events_total` metric.

> Is there a specific example...? 

In this case ([jira](https://datadoghq.atlassian.net/browse/DOCS-14949?focusedCommentId=3379308)), the pipeline used an OpenTelemetry source. If the source’s downstream send is canceled, any events from that send operation that have not yet entered the rest of the pipeline are dropped by the Worker. The client sending data to the source may retry the request, but the Worker itself does not retain or retry the dropped events

> What's the difference between retry and redeliver? Or am I getting too much into the weeds? 

I agree that this is probably too much detail for this troubleshooting section, so I’d omit the retry/redelivery sentence actually 😅

> Also, is there a way for the user to know that the error is due to a Worker shut down or restart, rather than backpressure?

The error itself doesn’t distinguish shutdown from backpressure, unfortunately. Users can correlate its timestamp with Worker lifecycle logs, such as `Vector has stopped`, `Shutting down...`, or for pod/container restart events around the same time. High buffer or CPU utilization can indicate backpressure, but aren’t definitive proof

maycmlee: Thanks @jackie8c! I made some updates. Let me know your thoughts and if anything is incorrect. 

```suggestion
If you see `Source send interrupted mid-flight; pipeline may be overloaded or shutting down` error logs, an issue interrupted the send operation before the Worker sent all events in the batch downstream. The Worker drops any remaining events in that batch and increments the `component_discarded_events_total` metric. Possible causes for the interruption can include backpressure, Worker shutdown, or Worker restarts.

To investigate whether the interruption was due to a Worker restart or shutdown, try correlating the timestamp of the error with Worker lifecycle logs, such as `Vector has stopped`, `Shutting down...`, or with pod or container restart events around the same time.
```

jackie8c: @maycmlee lgtm thanks!

jackie8c (approved):

hestonhoffman (approved):

PR #38381, opened 2026-07-22
Document container-based primary tags as enabled by default (Agent 7.65.0)

ichinaski:
Container-based primary tags are enabled by default since Datadog Agent 7.65.0, so this rewrites the opt-in `enable_cid_stats` section of the primary tags guide as an opt-out one using `disable_cid_stats`, and notes the automatic-in-Fargate behavior for 7.35.0 through 7.64.x. Also fixes a pre-existing invalid Docker Compose `environment` mapping in the example.

### DIFF
--- a/content/en/tracing/guide/setting_primary_tags_to_scope.md
+++ b/content/en/tracing/guide/setting_primary_tags_to_scope.md
@@ -89,14 +89,24 @@ If you change a previously set primary tag, be aware of the following:
 
 ## Container-based additional primary tags
 
-You can index your trace metrics based on the tags derived from Docker containers and Kubernetes pod metadata on Linux-based platforms. Container-based primary tags are available in Datadog Agent versions 7.35.0 and later.
+You can index your trace metrics based on the tags derived from Docker containers and Kubernetes pod metadata on Linux-based platforms.
 
-To enable container-based primary tags, install Agent version 7.35.0 or later, update the CID stats setting as described below, and restart the Agent. The procedure for enabling depends on how you installed the Agent:
+Container-based primary tags are enabled by default in Datadog Agent versions 7.65.0 and later. In versions 7.35.0 through 7.64.x, the feature is opt-in outside of Fargate and enabled automatically in Fargate environments.
+
+With Agent 7.65.0 or later, no feature flag is required. Go to the [APM Settings][6] page and select the additional primary tag you want to use. It can take up to two hours for changes to this setting to take effect.
+
+Now you can filter your services in the [Catalog][7] by the tag being sent by your containerized services. Trace metrics used by Dashboards and Monitors can also be aggregated by the container primary tag.
+
+**Note**: Primary tag values should not contain capital letters or special characters (aside from underscores, minuses, colons, periods, and slashes). If they do, some features may not work properly.
+
+### Disable container-based primary tags
+
+To turn off container-based primary tags, set the `disable_cid_stats` APM feature and restart the Agent. If `DD_APM_FEATURES` is already set, append `disable_cid_stats` to its comma-separated list. The procedure depends on how you installed the Agent:
 
 {{< tabs >}}
 {{% tab "Helm" %}}
 
-Using the Datadog Helm chart version 2.26.2 or later, add the following to your values file:
+Add the following to your values file:
 
 ```yaml
 #...
@@ -104,7 +114,7 @@ datadog:
   #...
   env:
     - name: DD_APM_FEATURES
-      value: 'enable_cid_stats'
+      value: 'disable_cid_stats'
 ```
 
 {{% /tab %}}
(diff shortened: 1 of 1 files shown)


ajgajg1134 (approved):

content/en/tracing/guide/setting_primary_tags_to_scope.md
```diff
@@ -89,22 +89,32 @@ If you change a previously set primary tag, be aware of the following:
 
 ## Container-based additional primary tags
 
-You can index your trace metrics based on the tags derived from Docker containers and Kubernetes pod metadata on Linux-based platforms. Container-based primary tags are available in Datadog Agent versions 7.35.0 and later.
+You can index your trace metrics based on the tags derived from Docker containers and Kubernetes pod metadata on Linux-based platforms.
 
-To enable container-based primary tags, install Agent version 7.35.0 or later, update the CID stats setting as described below, and restart the Agent. The procedure for enabling depends on how you installed the Agent:
+Container-based primary tags are enabled by default in Datadog Agent versions 7.65.0 and later. In versions 7.35.0 through 7.64.x, the feature is opt-in outside of Fargate and enabled automatically in Fargate environments.
```
OliviaShoup: Optional clarity tweak — the Fargate exception is stated in both directions, which reads a little dense.

```suggestion
Container-based primary tags are enabled by default in Datadog Agent versions 7.65.0 and later. In Agent versions 7.35.0 through 7.64.x, container-based primary tags are enabled automatically in Fargate and are opt-in everywhere else.
```

content/en/tracing/guide/setting_primary_tags_to_scope.md
```diff
@@ -89,22 +89,32 @@ If you change a previously set primary tag, be aware of the following:
...
+
+With Agent 7.65.0 or later, no feature flag is required. Go to the [APM Settings][6] page and select the additional primary tag you want to use. It can take up to two hours for changes to this setting to take effect.
+
+Now you can filter your services in the [Catalog][7] by the tag being sent by your containerized services. Trace metrics used by Dashboards and Monitors can also be aggregated by the container primary tag.
```
OliviaShoup: Minor — Vale flags "Now" as a temporal word (`Datadog.tense`). Pre-existing text just relocated by this PR, so not blocking, but since this hunk is already touched:

```suggestion
You can filter your services in the [Catalog][7] by the tag being sent by your containerized services. Trace metrics used by Dashboards and Monitors can also be aggregated by the container primary tag.
```

content/en/tracing/guide/setting_primary_tags_to_scope.md
```diff
@@ -89,22 +89,32 @@ If you change a previously set primary tag, be aware of the following:
...
+
+### Disable container-based primary tags
+
+To turn off container-based primary tags, set the `disable_cid_stats` APM feature and restart the Agent. If `DD_APM_FEATURES` is already set, append `disable_cid_stats` to its comma-separated list. The procedure depends on how you installed the Agent:
 
 {{< tabs >}}
 {{% tab "Helm" %}}
 
-Using the Datadog Helm chart version 2.26.2 or later, add the following to your values file:
+Add the following to your values file:
```
OliviaShoup, line 107: The removed sentence dropped the minimum Helm chart version needed to set `DD_APM_FEATURES` (previously: "Using the Datadog Helm chart version 2.26.2 or later, add the following to your values file:"). Can you confirm this constraint no longer applies (e.g., that chart version is old enough it's a non-issue), or should it be kept for the disable flow?

OliviaShoup (changes requested):
Thanks for cleaning this up! The default-on behavior for 7.65.0+ and the Docker Compose `environment` syntax fix are both correct and clear.

One thing before merging: the rewrite drops instructions for the pre-7.65.0/non-Fargate opt-in case. The doc states (line 94) that Agent versions 7.35.0–7.64.x still require opt-in outside of Fargate, but the tabs below now only document `disable_cid_stats`. There's no `enable_cid_stats` guidance left for that cohort anywhere on the page. Could you either restore opt-in instructions for that case, or point to where they now live?

A couple of minor inline suggestions as well, nothing blocking.

content/en/tracing/guide/setting_primary_tags_to_scope.md
ichinaski: That version is [4.5 years old](https://github.com/DataDog/helm-charts/releases/tag/datadog-2.26.2), we can safely drop it from the sentence IMO.

ichinaski:
> Thanks for cleaning this up! The default-on behavior for 7.65.0+ and the Docker Compose `environment` syntax fix are both correct and clear.
> 
> One thing before merging: the rewrite drops instructions for the pre-7.65.0/non-Fargate opt-in case. The doc states (line 94) that Agent versions 7.35.0–7.64.x still require opt-in outside of Fargate, but the tabs below now only document `disable_cid_stats`. There's no `enable_cid_stats` guidance left for that cohort anywhere on the page. Could you either restore opt-in instructions for that case, or point to where they now live?
> 
> A couple of minor inline suggestions as well, nothing blocking.

I simplified the version note to a "since 7.65.0" one-liner.

OliviaShoup (approved):

PR #38411, opened 2026-07-22
[DOCS-15168] Add units note and worked example for the time() function

OliviaShoup:

Fixes DOCS-15168

Expands the `time()` function section on the Dashboards Querying page to prevent a common unit-mismatch error. `time()` returns Unix **seconds**, but log and event timestamps (such as `@timestamp`) are in Unix **milliseconds**, so `time() - max:@timestamp` returns a large negative number. This adds a units note plus worked examples showing how to convert between seconds and milliseconds, and a note that `time()` must live inside the formula.

### Merge instructions

Two open items before this is ready to merge (both marked with `TODO(DOCS-15168)` HTML comments in the diff):

1. **SME confirmation** of the "`time()` resolves to the end of the selected time window / use a live or rolling range" behavior. I could not verify this against a primary source, so it needs a dashboards/query SME sign-off. Remove the paragraph if it turns out to be inaccurate.
2. **Screenshot** of the scalar table widget example still needs to be captured and added (placeholder comment in the diff).

Merge readiness:
- [ ] Ready for merge

### Additional notes

The unit facts are verified: the existing section already documents that `time()` returns seconds, and `@timestamp` being Unix milliseconds is confirmed in `dashboards/guide/custom_time_frames.md`. Kept "Unix" (not "UNIX") to match the existing intro sentence in the same section and the "Unix milliseconds" phrasing used elsewhere in the timestamp docs.


jhgilbert:
[Automated message] This PR has merge conflicts created by the [docs repo reorg project](https://github.com/DataDog/documentation/blob/master/REPO_REORG.md). Because this PR is marked as a work in progress, no attempt was made to auto-resolve the conflicts. When your PR is finished, you can queue your PR for an auto-fix by removing the `WORK IN PROGRESS` label and the `astro-reorg-skip` label. Your PR will be processed in the next batch of attempted auto-fixes.

PR #38470, opened 2026-07-24
Add Claude Code hook to block git commit/push to master

hestonhoffman:

Adds a Claude Code `PreToolUse` hook that blocks Bash-invoked `git commit`/`git push` commands that would write to `master`, including via detached HEAD or an explicit refspec (e.g. `git push origin some-branch:master`). It also blocks `--no-verify`, `HUSKY=0`, and force-push flags on any branch.

This runs before git starts, so it catches bypasses that `.husky/pre-commit` and `.husky/pre-push` can't defend against on their own. The `.husky` hooks are unchanged and remain the guard for commits made outside Claude Code — this is a supplementary layer, not a replacement.

`.claude/settings.json` is updated to register the hook, and `CLAUDE.md`'s Critical Rules section notes that these rules are now enforced mechanically so Claude treats a block as intentional rather than something to route around.

Reviewers should note: after pulling this branch, Claude Code will prompt to review the new hook configuration on first use.

### Merge instructions

Merge readiness:
- [ ] Ready for merge

### Additional notes


brett0000FF (approved):
🚀

PR #38478, opened 2026-07-27
Add Cross-App Access (XAA) setup documentation for Okta

anthony-dgx:

Adds a new page documenting how to set up Cross-App Access (XAA) with Okta: **Organization Settings > Cross-App Access** (`/account_management/org_settings/cross_app_access/`).

Cross-App Access lets AI agents call the Datadog API on behalf of users already authorized in Okta, without a per-user consent screen. The page covers:

- Configuring the feature in Datadog, and which values to copy in each direction between Datadog and Okta
- Scoping an agent through Application Scope Management, including how Datadog resolves scopes sent by Okta
- The Okta-side setup (resource server, AI agent, delegations, resource connection)
- Enabling the connector in the agent, verification, and troubleshooting

Also adds:
- `mcp_xaa_resource_url` and `mcp_xaa_issuer_url` region params, so the page shows the correct URLs for the reader's selected Datadog site
- A menu entry in `main.en.yaml` and one screenshot

### Merge instructions

Draft while the feature is in Preview. Holding for technical review from the feature team, and the screenshot needs retaking once the page shows production values.

Merge readiness:
- [ ] Ready for merge


assets/scripts/config/regions.config.js
```diff
@@ -845,6 +845,28 @@ export default {
...
+  mcp_xaa_resource_url: {
+    us: 'https://mcp.datadoghq.com',
+    us3: 'https://mcp.us3.datadoghq.com',
+    us5: 'https://mcp.us5.datadoghq.com',
+    eu: 'https://mcp.datadoghq.eu',
+    ap1: 'https://mcp.ap1.datadoghq.com',
+    ap2: 'https://mcp.ap2.datadoghq.com',
+    uk1: 'https://mcp.uk1.datadoghq.com',
+    gov2: 'Cross-App Access is not supported for US2-FED.',
+    gov: 'Cross-App Access is not supported for US1-FED.'
```
hay-sam, line 857: https://mcp.datadoghq.com/.well-known/oauth-protected-resource/v1/mcp

these should also be /v1/mcp

content/en/account_management/org_settings/cross_app_access.md
```diff
@@ -0,0 +1,149 @@
...
+      text: 'Datadog MCP Server'
+---
+
+<div class="alert alert-info">Cross-App Access is in Preview.</div>
+
+## Overview
+
+Cross-App Access (XAA) lets AI agents call the Datadog API on behalf of users your organization already authorized in Okta. Without it, every user authorizes each agent individually through a browser consent screen. With it, your Okta administrator grants that access once, centrally, and users skip the per-user consent step.
+
+Okta issues each agent a short-lived token called an ID-JAG (Identity Assertion JWT Authorization Grant). The agent presents this token to Datadog, and Datadog exchanges it for an access token scoped to the user Okta named. Because Okta mints the token, your administrators control which agents reach Datadog and revoke that access from Okta.
```
hay-sam:
```suggestion
Okta issues each agent a short-lived token called an ID-JAG (Identity Assertion JWT Authorization Grant). The agent presents this token to Datadog, and Datadog exchanges it for an access token scoped to the user Okta named. Because Okta mints the token, your administrators grant and revoke Datadog access for AI agents from Okta.
```

/nit something about this doesn't read quite right 

> Because Okta mints the token, your administrators control which agents reach Datadog and revoke that access from Okta.

content/en/account_management/org_settings/cross_app_access.md
```diff
@@ -0,0 +1,149 @@
...
+Datadog derives the token signing keys location from the Okta issuer URL, so the value must be exact:
+
+- Use `https`.
+- Enter the bare host with no path, port, query string, fragment, username, or password.
+- Use a subdomain of `.okta.com`, `.oktapreview.com`, or `.okta-emea.com`. Datadog rejects the apex domain, so `example.okta.com` works and `okta.com` does not.
+
+Clearing the field and saving removes the issuer and stops Datadog from accepting tokens.
+
+## Control what an agent can do
+
```
hay-sam, line 98: I think we can be more explicit in saying that although you can do the scopes restrictions in either place (ddog or okta) - you only _need_ to pick 1 if you want the restrictions to apply & we suggest ddog

static/images/account_management/cross_app_access/cross-app-access-settings.png
hay-sam, line 1: Image

I do think we should update the screenshot to have a uuid as the Public Id field

We are specifically expecting the uuid/id & not the public_id (could be uuid or random string like the example shown for older orgs) - this is different from the existing "company id" field in okta which is the public id

As an aside I'm wondering if we should switch the component where the org uuid is listed to look less editable  (it looks like the very editable issuer field above)
+ in the client table, should we move the copy button next to the client id so it's just manage app on the right?

assets/scripts/config/regions.config.js
anthony-dgx: Fixed - the resource values now include `/v1/mcp`.

Confirmed against the live metadata on all seven commercial sites: `/.well-known/oauth-protected-resource/v1/mcp` returns `resource` and `authorization_servers` both equal to `<host>/v1/mcp`, so the resource URL and the issuer URL are the same value per site. Kept them as two separate site parameters since Okta exposes two fields and they could diverge later.

content/en/account_management/org_settings/cross_app_access.md
anthony-dgx: Applied your suggestion verbatim. Reads better, thanks.

content/en/account_management/org_settings/cross_app_access.md
anthony-dgx: Rewritten. The section now leads with: you can restrict in either place, both take effect, so you only need to set it in one - and then gives the reasons we recommend Datadog. The Okta **Scope Condition** row is now conditional too: "**Allow all**, if you control scopes in Datadog ... Otherwise, list the scopes this agent is allowed."

static/images/account_management/cross_app_access/cross-app-access-settings.png
anthony-dgx: Agreed the screenshot needs re-shooting - holding off until one open question is settled, because I think there is a mismatch to fix first.

`PublicOrgIdField` reads `getOrgField('public_id')`, but the token exchange resolves the tenant with `GetOrgByUuid(aud_tenant)`. For an org whose `public_id` is not a UUID, the value the UI hands the admin gets rejected as an unrecognized `aud_tenant`, with no hint that the wrong identifier was copied.

Which side should change - the field reading the org UUID, or the lookup accepting `public_id`? Once that is decided I will reshoot the screenshot with a correct value and make the copy instruction name it precisely. For now the doc says "copy your organization ID from the **Public Org ID** field" and notes it is not the same as the company ID Okta asks for elsewhere.

On the two UI asides (making the org ID field look read-only rather than editable, and moving the copy button next to the client ID so only **Manage app** sits on the right) - both make sense to me, but they are web-ui changes rather than doc changes, so I am routing them to the team rather than tracking them here.

content/en/account_management/org_settings/cross_app_access.md
```diff
@@ -0,0 +1,225 @@
...
+Datadog derives the token signing keys location from this value, so it must be exact.
+
+```shell
+curl -X PUT "{{< region-param key="dd_api" >}}/api/v2/login/org_configs/mcp_cross_app_access_issuer_url" \
+  -H "Content-Type: application/json" \
+  -H "Authorization: Bearer ${DD_TOKEN}" \
+  -d '{
+    "data": {
+      "type": "org_config",
```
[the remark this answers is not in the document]
anthony-dgx: It's a good callout, but this is what we have today in our API documentation. I'lll ignore for now.

assets/scripts/config/regions.config.js
```diff
@@ -845,6 +845,28 @@ export default {
...
+  mcp_xaa_issuer_url: {
+    us: 'https://mcp.datadoghq.com/v1/mcp',
+    us3: 'https://mcp.us3.datadoghq.com/v1/mcp',
+    us5: 'https://mcp.us5.datadoghq.com/v1/mcp',
+    eu: 'https://mcp.datadoghq.eu/v1/mcp',
+    ap1: 'https://mcp.ap1.datadoghq.com/v1/mcp',
+    ap2: 'https://mcp.ap2.datadoghq.com/v1/mcp',
+    uk1: 'https://mcp.uk1.datadoghq.com/v1/mcp',
+    gov2: 'Cross-App Access is not supported for US2-FED.',
+    gov: 'Cross-App Access is not supported for US1-FED.'
```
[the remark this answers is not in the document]
anthony-dgx: Today they are the same, but they might be different in the future

assets/scripts/config/regions.config.js
```diff
@@ -845,6 +845,28 @@ export default {
...
+  mcp_xaa_resource_url: {
+    us: 'https://mcp.datadoghq.com/v1/mcp',
+    us3: 'https://mcp.us3.datadoghq.com/v1/mcp',
+    us5: 'https://mcp.us5.datadoghq.com/v1/mcp',
+    eu: 'https://mcp.datadoghq.eu/v1/mcp',
+    ap1: 'https://mcp.ap1.datadoghq.com/v1/mcp',
+    ap2: 'https://mcp.ap2.datadoghq.com/v1/mcp',
+    uk1: 'https://mcp.uk1.datadoghq.com/v1/mcp',
+    gov2: 'Cross-App Access is not supported for US2-FED.',
+    gov: 'Cross-App Access is not supported for US1-FED.'
```
brett0000FF, line 857: Since this is not supported for gov sites, we should add an entry to `config/_default/params.yaml` under `unsupported_sites`.

For example:
```
cross_app_access: [gov,gov2]
```

This will put an unsupported banner at the top of the page if a use has selected one of those sites.

content/en/account_management/org_settings/cross_app_access.md
```diff
@@ -0,0 +1,218 @@
...
+3. Set the following fields.
+
+  | Okta field                | Value                                                                                                |
+  | ------------------------- | ---------------------------------------------------------------------------------------------------- |
+  | **Client ID at resource** | `391e6845-8153-4de1-bbf0-c1b6ef7fdc14`                                                               |
+  | **Scope Condition**       | **Allow all**, the only supported value. See [Control scopes in Datadog](#control-scopes-in-datadog) |
+
+4. Activate the agent from the **Actions** menu.
+
+### Control scopes in Datadog
```
brett0000FF: This section is currently nested under "Finish the setup in Okta," but it is Datadog config. Consider promoting this to h2 or moving it under "Configure cross-app access in Datadog."

content/en/account_management/org_settings/cross_app_access.md
```diff
@@ -0,0 +1,218 @@
...
+**Note**: Changing the issuer URL later requires deleting and recreating the resource connection described in [Connect Claude to the Datadog application](#connect-claude-to-the-datadog-application).
+
+### Register Claude as an AI Agent
+
+Create an AI Agent entry for Claude in Okta, then exchange keys with Anthropic. Anthropic signs the requests Okta receives, so Okta needs Anthropic's public key before it issues any token.
+
+1. Create the AI Agent entry for Claude.
+2. Assign owners to the agent. Okta requires an owner before you can activate it.
+3. Send the AI Agent ID that Okta generates to Anthropic.
+4. Add the public key that Anthropic returns to the AI Agent entry, on the **Credentials** tab.
```
brett0000FF: Please wrap UI elements like buttons and tabs in this `ui` tag. This prevents our translation service from trying to translate UI strings.
```suggestion
4. Add the public key that Anthropic returns to the AI Agent entry, on the {{< ui >}}Credentials{{< /ui >}} tab.
```

content/en/account_management/org_settings/cross_app_access.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Overview
+
+Cross-App Access (XAA) lets AI agents call the Datadog API on behalf of users your organization already authorized in Okta. Without it, every user authorizes the agent individually through a browser consent screen. With it, your Okta administrator grants that access once, centrally, and users skip the per-user consent step.
+
+Okta issues the agent a short-lived token called an ID-JAG (Identity Assertion JWT Authorization Grant). The agent presents this token to Datadog, and Datadog exchanges it for an access token owned by the user who initiated the call. Because Okta mints the token, your administrators grant and revoke Datadog access for AI agents from Okta.
+
+In Preview, Cross-App Access supports Okta as the only identity provider and Claude as the only agent.
+
+## How it works
```
brett0000FF: I'd recommend cutting this section. It mostly restates some of the text in Overview, and it pushes the important prereq and setup content further down on the page, which is what a user is actually looking for.

content/en/account_management/org_settings/cross_app_access.md
```diff
@@ -0,0 +1,218 @@
...
+To set the scopes Claude is allowed:
+
+1. Navigate to [**Organization Settings > Mobile and Third-Party Access**](https://app.datadoghq.com/organization-settings/mobile-third-party-access).
+2. Select the Claude application, then select the **Scopes** tab.
+3. Use the **Allowed** checkbox for each scope to control what Claude reaches.
+4. Click **Enable** to save.
+
+Adding or removing a scope affects every user in your organization, and removing a scope revokes existing authorizations that rely on it. See [Application Scope Management](/account_management/org_settings/mobile_third_party_access/#application-scope-management).
+
+A scope that is not allowed in Datadog is never granted, whatever the token requests.
```
brett0000FF:
```suggestion
A scope that is not allowed in Datadog is never granted, regardless of what the token requests.
```

brett0000FF (approved):
Looks good! Thanks for the content. I just left a few suggestions.

jhgilbert:
[Automated message] This PR has merge conflicts caused by the [recent docs repo reorg](https://github.com/DataDog/documentation/blob/master/REPO_REORG.md) (files moved from the repo root into `hugo/`). A new PR with your commits translated to the correct paths has been opened: #38908

Please follow the instructions in the PR description.

PR #38514, opened 2026-07-28
[DSEM-999] Add deprecated warning banner to ddsql_schema dataset pages

irenegaunab:
## Summary

Dataset pages with `status: deprecated` now show an alert warning banner explaining the table no longer returns data.

This supports the REDAPL→UDM migration where ~30 retired cloud resource tables were deprecated (DSEM-999). [Here](https://datadoghq.atlassian.net/wiki/spaces/ADVANCED/pages/6537905128/2025-07-08+Supporting+Retired+Cloud+Resource+Tables#Full-list-of-retired-tables) you can see a list of deprecated tables. Without this change, the public docs such as [this one](https://docs.datadoghq.com/ddsql_reference/data_directory/aws/aws.resource_tags.dataset/) would show no indication that a table is deprecated.

## What does this PR do? What is the motivation?

Adds a conditional block to `layouts/ddsql_schema/single.html` that renders an `alert-warning` banner when `.Params.status == "deprecated"`:

> **Deprecated**: This table is deprecated and no longer returns data. Querying it returns empty results.

## Testing

Verified the deprecated warning banner renders correctly on the branch preview:
https://docs-staging.datadoghq.com/irene.gauna/add-deprecation-status/ddsql_reference/data_directory/aws/aws.resource_tags/

Screenshot 2026-07-29 at 11 48 05


layouts/ddsql_schema/single.html
```diff
@@ -14,6 +14,12 @@ <h1 id="pagetitle">{{ .Title }}</h1>
     {{ partial "site_support_banner/site_support_banner.html" . }}
 
     <div class="ddsql-schema-single">
+        {{ if eq .Params.status "deprecated" }}
+        <div class="alert alert-warning">
+            <strong>Deprecated</strong>: This table is deprecated and may return no data.
```
petitcl, line 19: super nit: the text was changed and the PR screenshot does not match the text anymore.

layouts/ddsql_schema/single.html
```diff
@@ -14,6 +14,12 @@ <h1 id="pagetitle">{{ .Title }}</h1>
     {{ partial "site_support_banner/site_support_banner.html" . }}
 
     <div class="ddsql-schema-single">
+        {{ if eq .Params.status "deprecated" }}
```
petitcl, line 17: I found the place where we filter out non stable datasets and it's [here](https://github.com/ddoghq/websites-sources/blob/fef89c4b9542d2daa58ff48b3ffe3a792f426d71/Makefile#L367-L369). This logic should probably be updated to be "status == stable | deprecated" (ie allowlist of known supported statuses in case we want to add more statuses in the future).

layouts/ddsql_schema/single.html
irenegaunab: haha done :)!

Jackie-Robson (approved):

PR #38538, opened 2026-07-28
[DO NOT MERGE] Add Cdocs to Astro

jhgilbert:
## Env setup (requires Node 24)

```shell
cd astro
nvm use           # or whatever you use to set your Node version
yarn install
yarn dev
```

Then visit `localhost:4321/dd_e2e` in your browser. If you just started `yarn dev`, it can take a minute to load. This page is drafty, but it gives you a good idea of what already exists.

The API docs are also available on this branch, if you would like to look at them: `localhost:4321/api/latest`.

## Key folders (all in `astro/src`):

- `lib`: Utils, helpers, and anything else that warrants a plain TypeScript function. `lib/api` is what processes the YAML spec, for example.
- `components`: All components, whether they're vanilla `.astro` or Preact, and whether they're used in Cdocs, the API docs, or both. Each component folder contains text, plaintext rendering logic (where it exists so far), HTML markup, etc.
- `customization_config`: A temporary place to put Cdocs traits/options/option groups for testing purposes, until Hugo and Astro have a true `shared` folder.
- `partials`: The location of partial `.mdoc` files. The `.mdoc` files actually use an alias, `@partials`, so this folder can move and be re-aliased if we prefer to put it elsewhere.
- `content/en/dd_e2e/cdocs`: Test pages for Cdocs. There's not much here yet, just the basic proof of concept.
- `pages`: The `.astro` components for various pages. Some are HTML pages, some are scripts that look up entries in `content/en` and render them. The folder structure in `pages` provides URL matching, which is why some folders have strange names like `[...lang]`.

## Key files
- `pages/[...slug].astro`: The catch-all Cdocs HTML route.
- `pages/[...slug].md.ts`: The catch-all Cdocs plaintext rendering route.

## For Datadog employees:


astro/astro.config.mjs
hestonhoffman, line 22: @jhgilbert this won't necessarily match the path to websites-modules on every machine. Are we expecting this be a user-cloned repo on the same level as the documentation repo (`../../websites-modules`)? Or am I missing some tooling that should install websites-modules to this path? 

Changing it to my local path (here and in `astro/tsconfig.json`) allowed me to run the tests etc

jhgilbert: @hestonhoffman Right now, yeah, you have to have the repo cloned there. David is going to look at our options for handling these kinds of dependencies, but right now, there are a few brittle issues like that, where Astro needs this repo, or Astro is reaching into Hugo for something.

PR #38606, opened 2026-07-30
MySQL DBM: recommend REFERENCES instead of SELECT for schema collection

azhou-datadog:
### What does this PR do?

Changes the "Collecting schemas" section of the MySQL Database Monitoring setup pages to recommend `GRANT REFERENCES` instead of `GRANT SELECT`, drops the per-column scoping option, and adds a troubleshooting entry for the "no tables were found" warning.

Affects all five MySQL setup pages (self-hosted, RDS, Aurora, Azure, Cloud SQL), since they share the `dbm-mysql-agent-config-examples` shortcode.

### Motivation

MySQL exposes a table in `INFORMATION_SCHEMA` only to users that hold a privilege on that table. That is why this page currently asks customers to grant the Agent `SELECT` on their tables, and why it has to reassure them in prose that "the Agent does not use SELECT to access or read your table data."

The `REFERENCES` privilege satisfies the same visibility requirement and grants no ability to read data at all, so we can stop asking for read access on customer data and the reassurance becomes a property of the grant rather than a promise. Testing across MySQL 5.7 / 8.0 / 8.4 / 9.x and MariaDB 10.11 / 11.4 confirmed the Agent collects a byte-identical schema payload under `REFERENCES`, while `SELECT` on a user table returns `ERROR 1142`. The privilege is grantable on RDS/Aurora, Cloud SQL, and Azure Database for MySQL.

The per-column option is removed because a column-level grant exposes only the granted columns, so it yields an incomplete schema.

### DIFF
--- a/content/en/database_monitoring/setup_mysql/troubleshooting.md
+++ b/content/en/database_monitoring/setup_mysql/troubleshooting.md
@@ -220,6 +220,20 @@ GRANT EXECUTE ON PROCEDURE datadog.enable_events_statements_consumers TO datadog
 
 <!-- TODO: add a custom query recipe for getting the max sql text length -->
 
+### Tables are missing from collected schemas
+
+If the Agent logs this warning:
+```
+No tables were found across any of the N databases. This may indicate insufficient privileges to view table metadata. The datadog user needs REFERENCES (or SELECT) privileges on the tables.
+```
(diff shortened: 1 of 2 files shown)


brett0000FF (approved):

jhgilbert:
[Automated message] This PR has merge conflicts caused by the [recent docs repo reorg](https://github.com/DataDog/documentation/blob/master/REPO_REORG.md) (files moved from the repo root into `hugo/`). A new PR with your commits translated to the correct paths has been opened: #38894

Please follow the instructions in the PR description.

PR #38624, opened 2026-07-30
Fix misleading Docker volume path guidance for private locations

julinvictus:
## Summary
- Fixes DOCS-14963: the Docker tab instructions for Synthetic private locations said the worker config file must be placed in `/etc/docker`, but the example command mounts from `$PWD` — these two statements contradicted each other.
- Clarifies that the host-side path (before the colon in the `-v` flag) can be any directory; only the container-side path (`/etc/datadog/synthetics-check-runner.json`) needs to match exactly, since that's what the worker looks for inside the container.

## Test plan
- [ ] Docs preview renders the "Docker" tab correctly under **Synthetic Monitoring > Settings > Private Locations**
- [ ] Verify the corrected wording matches the actual `-v` bind-mount semantics used by the `docker run` command

Jira: https://datadoghq.atlassian.net/browse/DOCS-14963

### DIFF
--- a/content/en/synthetics/platform/private_locations/_index.md
+++ b/content/en/synthetics/platform/private_locations/_index.md
@@ -178,7 +178,7 @@ Launch your private location on:
 {{< tabs >}}
 {{% tab "Docker" %}}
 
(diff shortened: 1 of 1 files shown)


julinvictus:
Closing in favor of a new PR using a branch name that follows the <name>/<description> convention.

PR #38692, opened 2026-08-03
Sync PR description template in CLAUDE.md with .github/PULL_REQUEST_TEMPLATE.md

evazorro:

The PR description template embedded in `CLAUDE.md` had drifted from the real one in `.github/PULL_REQUEST_TEMPLATE.md`, so Claude Code drafts PR descriptions against a stale copy.

### Merge readiness

- [x] Ready for merge

### AI assistance

Used Claude Code to make this PR.

### Additional notes

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -92,19 +92,29 @@ CRITICAL: Always use format `<name>/<description>` with forward slash. Without t
 Use the Jira ticket key in square brackets: `[DOCS-XXXXX] Brief description`
 
 ### PR Description Template
+The canonical template is `.github/PULL_REQUEST_TEMPLATE.md`. Read that file when
+drafting a PR description rather than relying on the copy below, which can drift.
+
 ```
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -92,20 +92,30 @@ CRITICAL: Always use format `<name>/<description>` with forward slash. Without t
 Use the Jira ticket key in square brackets: `[DOCS-XXXXX] Brief description`
 
 ### PR Description Template
+The canonical template is `.github/PULL_REQUEST_TEMPLATE.md`. Read that file when
```
evazorro: @brett0000FF Claude added this but also updated the template below. I feel like we probably want one or the other, not both. Do you have a preference?

brett0000FF: I think we should just remove this pointer. The drift it's guarding against is pretty low-stakes and not worth it imo.

brett0000FF (approved):

PR #38969, opened 2026-08-06
[SVLS-9609] Explicitly instruct setting 'Allow access to all app settings' in AAS Linux sidecar docs

kathiehuang:
Our public docs currently present "set required vars explicitly on both containers" and "enable Allow access to all app settings" as interchangeable alternatives. However, some reserved/platform-injected env vars like `WEBSITE_SITE_NAME` do not propagate to the sidecar if "Allow access to all app settings" is unchecked. This explicitly instructs customers to set this in the Azure portal.

https://datadoghq.atlassian.net/browse/SVLS-9609

### Merge readiness

- [x] Ready for merge

## For Datadog employees:

### DIFF
--- a/hugo/content/en/serverless/azure_app_service/linux_code.md
+++ b/hugo/content/en/serverless/azure_app_service/linux_code.md
@@ -442,10 +442,11 @@ Path to the instrumentation library loaded by the .NET runtime.<br>
       - {{< ui >}}Registry server URL{{< /ui >}}: `index.docker.io`
       - {{< ui >}}Image and tag{{< /ui >}}: `datadog/serverless-init:latest`
       - {{< ui >}}Port{{< /ui >}}: 8126
-   1. Select {{< ui >}}Apply{{< /ui >}}.
-   1. On the sidecar container's settings, enable the {{< ui >}}Allow access to all app settings{{< /ui >}} option.
+      - Under {{< ui >}}Environment variables{{< /ui >}}, enable the {{< ui >}}Allow access to all app settings{{< /ui >}} option.
 
-      {{< img src="serverless/azure_app_service/app_settings.png" alt="In Azure, an Environment Variables section. An 'Allow access to all app settings' option is enabled with a checkbox." >}}
+        {{< img src="serverless/azure_app_service/app_settings.png" alt="In Azure, an Environment Variables section. An 'Allow access to all app settings' option is enabled with a checkbox." >}}
+
+   1. Select {{< ui >}}Apply{{< /ui >}}.
 
 3. **Restart your application**.
 
--- a/hugo/content/en/serverless/azure_app_service/linux_code.md
+++ b/hugo/content/en/serverless/azure_app_service/linux_code.md
@@ -442,8 +442,10 @@ Path to the instrumentation library loaded by the .NET runtime.<br>
       - {{< ui >}}Registry server URL{{< /ui >}}: `index.docker.io`
       - {{< ui >}}Image and tag{{< /ui >}}: `datadog/serverless-init:latest`
       - {{< ui >}}Port{{< /ui >}}: 8126
-      - {{< ui >}}Environment Variables{{< /ui >}}: Include all previously configured Datadog environment variables.
    1. Select {{< ui >}}Apply{{< /ui >}}.
+   1. On the sidecar container's settings, enable the {{< ui >}}Allow access to all app settings{{< /ui >}} option.
+
+      {{< img src="serverless/azure_app_service/app_settings.png" alt="In Azure, an Environment Variables section. An 'Allow access to all app settings' option is enabled with a checkbox." >}}
 
 3. **Restart your application**.
 
--- a/hugo/content/en/serverless/azure_app_service/linux_container.md
+++ b/hugo/content/en/serverless/azure_app_service/linux_container.md
@@ -487,9 +487,7 @@ See the [{{< ui >}}Manual{{< /ui >}} tab](?tab=manual#instrumentation) for descr
 
 #### Application settings
 
-In your {{< ui >}}App settings{{< /ui >}} in Azure, set the following environment variables on your main container and enable the {{< ui >}}Allow access to all app settings{{< /ui >}} option.
-
-{{< img src="serverless/azure_app_service/app_settings.png" alt="In Azure, an Environment Variables section. An 'Allow access to all app settings' option is enabled with a checkbox." >}}
+In your main container's {{< ui >}}App settings{{< /ui >}} in Azure, set the following environment variables.
 
 - `DD_API_KEY`: Your [Datadog API key][701]
 - `DD_SERVICE`: How you want to tag your service. For example, `sidecar-azure`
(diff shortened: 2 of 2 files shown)


hugo/content/en/serverless/azure_app_service/linux_code.md
```diff
@@ -442,8 +442,10 @@ Path to the instrumentation library loaded by the .NET runtime.<br>
       - {{< ui >}}Registry server URL{{< /ui >}}: `index.docker.io`
       - {{< ui >}}Image and tag{{< /ui >}}: `datadog/serverless-init:latest`
       - {{< ui >}}Port{{< /ui >}}: 8126
-      - {{< ui >}}Environment Variables{{< /ui >}}: Include all previously configured Datadog environment variables.
    1. Select {{< ui >}}Apply{{< /ui >}}.
+   1. On the sidecar container's settings, enable the {{< ui >}}Allow access to all app settings{{< /ui >}} option.
```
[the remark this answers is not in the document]
kathiehuang: The templates leave `inheritAppSettingsAndConnectionStrings` unset, which defaults to `null`, [which has the same behavior as](https://azure.github.io/AppService/2025/03/06/Sidecars-Deep-Dive-Part1.html#:~:text=All%20application%20settings%20are%20passed%20to%20all%20containers%20as%20environment%20variables%20unless%20overridden%20using%20inheritAppSettingsAndConnectionStrings%3Dfalse%20in%20the%20sidecar%20spec.) `true`

hugo/content/en/serverless/azure_app_service/linux_container.md
```diff
@@ -487,7 +487,7 @@ See the [{{< ui >}}Manual{{< /ui >}} tab](?tab=manual#instrumentation) for descr
 
 #### Application settings
 
-In your {{< ui >}}App settings{{< /ui >}} in Azure, set the following environment variables on both your main container and the sidecar container. Alternatively, set these variables on your main container and enable the {{< ui >}}Allow access to all app settings{{< /ui >}} option.
+In your {{< ui >}}App settings{{< /ui >}} in Azure, set the following environment variables on your main container and enable the {{< ui >}}Allow access to all app settings{{< /ui >}} option.
```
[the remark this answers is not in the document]
kathiehuang: https://github.com/DataDog/documentation/pull/38969/commits/f1791b7d146e341a08c18ed4728e3a24520479be

hugo/content/en/serverless/azure_app_service/linux_code.md
OliviaShoup: The previous step ends on "Select Apply," so a reader may not realize they need to reopen/select the container afterward.

```suggestion
   1. Select the sidecar container to open its settings, then in the {{< ui >}}Environment variables{{< /ui >}} section, enable the {{< ui >}}Allow access to all app settings{{< /ui >}} option.
```

hugo/content/en/serverless/azure_app_service/linux_container.md
```diff
@@ -499,6 +497,10 @@ In your {{< ui >}}App settings{{< /ui >}} in Azure, set the following environmen
 - `DD_AAS_INSTANCE_LOGGING_ENABLED`: When `true`, log collection is automatically configured for an additional file path: `/home/LogFiles/*$COMPUTERNAME*.log`
 - `DD_AAS_INSTANCE_LOG_FILE_DESCRIPTOR`: An optional file descriptor used for more precise file tailing. Recommended for scenarios with frequent log rotation. For example, setting `_default_docker` configures the log tailer to ignore rotated files and focus only on Azure's active log file.
 
+On the Datadog sidecar container's settings, enable the {{< ui >}}Allow access to all app settings{{< /ui >}} option.
```
OliviaShoup: Same navigation gap as the linux_code.md version of this step — doesn't say how to get to the sidecar container's settings or that the option lives in the "Environment variables" section.

```suggestion
Select the sidecar container to open its settings, then in the {{< ui >}}Environment variables{{< /ui >}} section, enable the {{< ui >}}Allow access to all app settings{{< /ui >}} option.
```

OliviaShoup (commented):
Hey, thanks for the PR! Looks great. I left two comments that I think should be addressed before we merge. Please reach out if you have any questions!

hugo/content/en/serverless/azure_app_service/linux_code.md
kathiehuang: Good point! The customer can actually just enable "Allow access to all app settings" before selecting Apply so I decided to just move that instruction before "Apply", WDYT?

image

https://github.com/DataDog/documentation/pull/38969/commits/4c5b785ff4073101ee3b2879db3cf131cc41b200

hugo/content/en/serverless/azure_app_service/linux_container.md
kathiehuang: I made the same change as above - but now I'm wondering why for AAS Linux container we give instructions for configuring the sidecar container before app settings, and vice versa for linux code 🤔

hugo/content/en/serverless/azure_app_service/linux_code.md
OliviaShoup: this looks great! thanks :)

hugo/content/en/serverless/azure_app_service/linux_container.md
OliviaShoup: I don't think there's a functional reason the two need to differ. now that the checkbox is enabled at container-creation time, the sidecar picks up app settings whenever they're saved, so either order works

one thing worth flagging though: `linux_container.md` has no restart step at all, while `linux_code.md` has an explicit one noting that Azure auto-restarts when Application Settings are saved. so saving app settings last in linux_container may be what implicitly triggers the restart. if we flip the order there, we'd want to add a restart step to match

there's also some broader drift between the two Manual tabs (linux_code uses a numbered 1-3 list, linux_container uses H4 sections). you don't have to expand the scope of this PR for that though. want me to open a follow-up ticket to align them, and confirm the restart behavior with your team at the same time?

kathiehuang: That makes sense! A follow-up ticket sounds good! I agree that if we flip the order we should have a restart step to match!

OliviaShoup: Filed it: DOCS-15348 (https://datadoghq.atlassian.net/browse/DOCS-15348)

It covers all three of the differences between the two pages (the ordering, the numbered-list vs H4-sections structure, and the missing restart step), and calls out the open question about whether the ordering in this file is load-bearing because of the missing restart step. I flagged that part as inference from reading the docs rather than something I verified against Azure, so worth a sanity check from your team before anyone edits.

Approving this one now, thanks for the quick turnaround!

OliviaShoup (approved):
Approving. Both fixes look good, and putting the checkbox inside the Edit container form before Apply is a better solution than what I originally suggested.

The Manual-tab alignment discussion is tracked separately in DOCS-15348 and doesn't block this.

Lewis-E (approved):

PR #39006, opened 2026-08-07
Remove Agent config template build dependency

brett0000FF:
### What changed

- Removes the generated Agent configuration block on the English logs getting-started page (doesn't seem to have much value?)
- Keeps a temporary shortcode fallback for localized pages until the translation service propagates the English change.
- Removes `config_template.yaml` from both the preview and production pull configurations.

### Why

[DataDog/datadog-agent#54160](https://github.com/DataDog/datadog-agent/pull/54160) removes `pkg/config/config_template.yaml`. The Documentation build currently pulls and parses that file, so the Agent PR cannot merge safely until this dependency is removed.

### Related

[DOCS-15330](https://datadoghq.atlassian.net/browse/DOCS-15330)

[DOCS-15330]: https://datadoghq.atlassian.net/browse/DOCS-15330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #39062, opened 2026-08-10
[DOCS-15330] Remove links to deleted Agent config_template.yaml

brett0000FF:

Fixes DOCS-15330

[DataDog/datadog-agent#54160](https://github.com/DataDog/datadog-agent/pull/54160) deleted `pkg/config/config_template.yaml`. [#39006](https://github.com/DataDog/documentation/pull/39006) removed the build-time dependency on that file; this PR is the follow-up that clears the remaining prose links, which now 404.

Only `blob/main/…` and `blob/master/…` links are affected. Commit-pinned (`blob/<sha>/…`) and tag-pinned (`blob/7.53.0/…`) links still resolve, so those are left as-is.

Also corrects a pre-existing mismatch on the Agent 5 configuration files page, where the prose said "sample `config_template.yaml` file" but the link pointed at `dd-agent/datadog.conf.example`.

### Merge readiness

- [x] Ready for merge

### AI assistance

Claude Code was used to audit the repository for references.

### Additional notes

Agent-side follow-up: Maxime proposed committing a generated `pkg/config/schema/example/datadog.yaml` at each Agent release. That file does not exist yet. If it lands, the pages in this PR can be repointed at it.

### DIFF
--- a/hugo/content/en/account_management/billing/aws.md
+++ b/hugo/content/en/account_management/billing/aws.md
@@ -77,7 +77,7 @@ For technical questions, contact [Datadog support][6].
 For billing questions, contact your [Customer Success][7] Manager.
 
 [1]: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html
-[2]: https://github.com/DataDog/datadog-agent/blob/main/pkg/config/config_template.yaml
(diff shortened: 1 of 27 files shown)


maycmlee (approved):
LGTM!

PR #39074, opened 2026-08-10
Update tag_enrichment.md

marianafrangos:
update tag enrichment documentation to include new telemetry types offered to add tags to

### Merge readiness

- [ ] Ready for merge

## For Datadog employees:

### DIFF
--- a/hugo/content/en/tracing/services/tag_enrichment.md
+++ b/hugo/content/en/tracing/services/tag_enrichment.md
@@ -16,7 +16,7 @@ Tag enrichment is in Preview. To request access, fill out this form.
 
 ## Overview
 
-Use tag enrichment rules to add tags to your logs, spans, trace metrics, and data jobs (more telemetry coming soon) without code changes or redeployment. You can use values from service metadata you've already defined in Catalog, the value of another tag, or a fixed value.
+Use tag enrichment rules to add tags to your logs, APM spans, trace metrics, and Data Jobs Monitoring (DJM) telemetry without code changes or redeployment. You can use values from service metadata you've already defined in Catalog, the value of another tag, or a fixed value.
 
 ## Prerequisites
 
--- a/hugo/content/en/tracing/services/tag_enrichment.md
+++ b/hugo/content/en/tracing/services/tag_enrichment.md
@@ -16,7 +16,7 @@ Tag enrichment is in Preview. To request access, fill out this form.
 
 ## Overview
 
-Use tag enrichment rules to add tags to your Logs and APM spans without code changes or redeployment. You can use values from service metadata you've already defined in Catalog, the value of another tag, or a fixed value.
+Use tag enrichment rules to add tags to your logs, spans, trace metrics, and data jobs (more telemetry coming soon) without code changes or redeployment. You can use values from service metadata you've already defined in Catalog, the value of another tag, or a fixed value.
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


hugo/content/en/tracing/services/tag_enrichment.md
```diff
@@ -16,7 +16,7 @@ Tag enrichment is in Preview. To request access, fill out this form.
 
 ## Overview
 
-Use tag enrichment rules to add tags to your Logs and APM spans without code changes or redeployment. You can use values from service metadata you've already defined in Catalog, the value of another tag, or a fixed value.
+Use tag enrichment rules to add tags to your logs, spans, trace metrics, and data jobs (more telemetry coming soon) without code changes or redeployment. You can use values from service metadata you've already defined in Catalog, the value of another tag, or a fixed value.
```
OliviaShoup: Two things here.

**1. "(more telemetry coming soon)"** — this is an undated roadmap promise on a page with no expiry, so it goes stale the moment Metrics ships and nobody comes back to delete it. It's also not a pattern we use: "coming soon" appears in only 4 of the 5,761 files in `content/en`, and none of those use it in body prose (one is inside an HTML comment, one is a table cell, two are `***Status:*** *coming soon*` in a single guide).

The Preview callout at the top of this page already tells readers this is early and evolving, so the parenthetical isn't buying much.

**2. This list doesn't match the one in "Tag enrichment behavior"** (line 64), which says "Logs, APM, and DJM **only**" — three items to this line's four. A reader can't tell whether trace metrics are supported. That's the part I'd most like your answer on, since only you know which is right.

Assuming trace metrics *are* supported and "spans + trace metrics" is just APM broken out:

```suggestion
Use tag enrichment rules to add tags to your logs, APM spans, trace metrics, and Data Jobs Monitoring (DJM) telemetry without code changes or redeployment. You can use values from service metadata you've already defined in Catalog, the value of another tag, or a fixed value.
```

hugo/content/en/tracing/services/tag_enrichment.md
```diff
@@ -61,7 +61,7 @@ Click {{< ui >}}Add Tags{{< /ui >}} to open the tag enrichment rule modal pre-po
 
 ## Tag enrichment behavior
 
-- **Impacted telemetry**: Tag enrichment applies to Logs and APM spans only.
+- **Impacted telemetry**: Tag enrichment applies to Logs, APM, and DJM only. More telemetry, including Metrics, DBM, Profiling, Kubernetes, USM, and Events, is coming soon.
```
OliviaShoup: Same "coming soon" point as line 19, plus a few smaller things:

- **"only" and "more is coming soon" cancel each other out** in the same bullet. One says the list is closed, the next says it isn't.
- **Abbreviations land undefined.** DJM, DBM, and USM aren't expanded. DJM is worth special care: it appears just twice in all of `content/en`, and the one place it's properly introduced writes it as "Data Observability: Jobs Monitoring (DJM)" — so readers genuinely won't recognize it here.
- **Casing differs from line 19** ("Logs, APM" vs "logs, spans") for the same concepts on the same page.

```suggestion
- **Impacted telemetry**: Tag enrichment applies to Logs, APM, and Data Jobs Monitoring (DJM). Support for additional telemetry types, including Metrics, Database Monitoring, Profiling, Kubernetes, Universal Service Monitoring, and Events, is planned.
```

If you'd rather not signal direction at all, dropping the second sentence is cleaner still — the Preview callout covers it.

OliviaShoup (commented):
Thanks for updating this! Left two comments. Let me know if you have questions!

OliviaShoup (approved):

PR #39093, opened 2026-08-11
Document Source to Target monitors for Data Observability

kevinzenghu:

Documents Source to Target monitors on the Data Observability monitor page. This monitor type is in Preview and was not documented.

Adds:
- A `Source to Target monitors` section with a Preview disclaimer, an explanation of what the monitor compares, creation steps that follow the monitor creation flow, guidance on choosing a format, how Custom SQL works across the two sides, and evaluation behavior.
- An example monitor tab for detecting rows lost between a source and its destination.
- A pointer from the Overview to the new section.

Labels match the monitor creation UI: the monitor type is **Source to Target**, the source and the target are selected in separate steps, and the comparison control is **Format**.

No new pages, so no sidebar navigation change is needed.

### Merge readiness

- [ ] Ready for merge

## For Datadog employees:

### AI assistance

Drafted with Claude Code from the existing page structure, the monitor's implemented behavior, and screenshots of the creation flow.

One label still to confirm: the second **Format** option is written as `Percent Difference`. Only `Difference` was visible in the screenshots used here.

### Additional notes

A screenshot of the source and target selection, or of the Format control, would fit the creation steps and match the images used elsewhere on this page.

Two observations outside the scope of this PR, for a possible follow-up:
- The `Detection method` section uses `Anomaly` and `Threshold`, while the creation flow shows `Anomalies` and `Thresholds`.
- The `Monitor schedule` section documents `Hourly`, `Daily`, and `Manual`, while the creation flow shows `Scheduled` and `Manual`.


hugo/content/en/monitors/types/data_observability.md
```diff
@@ -245,6 +247,51 @@ Observed value {{observed}} is within the expected range.
...
+Source to target monitors are available in all regions except GovCloud.
+
+### Create a source to target monitor
+
+1. Navigate to [{{< ui >}}Monitors{{< /ui >}} > {{< ui >}}New Monitor{{< /ui >}}][6] and select {{< ui >}}Source to Target{{< /ui >}}.
+2. Under {{< ui >}}Choose source{{< /ui >}}, select the warehouse that holds the source data, then select the data to compare.
+3. Under {{< ui >}}Choose target{{< /ui >}}, do the same for the destination. The source and the target can be in different data warehouses or in the same one.
+4. Under {{< ui >}}Select your metric type{{< /ui >}}, choose the metric to compare. Source to target monitors support the same metric types as other Data Observability monitors, including row count, freshness, nullness, uniqueness, cardinality, and {{< ui >}}Custom SQL{{< /ui >}}.
+5. Set {{< ui >}}Format{{< /ui >}} to control how the comparison is expressed:
+    - {{< ui >}}Difference{{< /ui >}}: the absolute difference between the source and target values.
```
jenhuang: the diff isn't absolute. it's target - source.

hugo/content/en/monitors/types/data_observability.md
```diff
@@ -245,6 +247,51 @@ Observed value {{observed}} is within the expected range.
...
+    - {{< ui >}}% Difference{{< /ui >}}: the difference expressed as a percentage.
```
jenhuang: the diff % is relative to the source. not sure if we need to call that out.

hugo/content/en/monitors/types/data_observability.md
```diff
@@ -245,6 +247,51 @@ Observed value {{observed}} is within the expected range.
...
+When the metric type is {{< ui >}}Custom SQL{{< /ui >}}, supply one query for the source and one query for the target. A {{< ui >}}WHERE{{< /ui >}} clause is not accepted for this metric type. Include any filtering in each query.
+
+### Evaluation
+
+The difference between the source and the target is recorded as its own metric, so a source to target monitor is evaluated by the same detection methods as any other Data Observability monitor, including anomaly detection. Both sides are measured on a synchronized schedule, so the two values are captured at the same time rather than following each warehouse's default collection cadence.
+
+The following conditions apply:
+
+- The comparison runs when both the source and the target measurements succeed. If either measurement fails, the monitor skips that evaluation.
+- The source and target measurements must be captured within 30 minutes of each other. If the gap between them is larger, the monitor skips that evaluation.
```
jenhuang: I'm not sure we need this extra info. I think 99% of the time, the evaluation will work exactly how they expect. until we see customers complain that we're missing evaluations, we probably don't need to document the implementation details.

jenhuang (approved):

drichards-87 (approved):

hugo/content/en/monitors/types/data_observability.md
kevinzenghu: Good catch, fixed in 968063783f. Changed to "the target value minus the source value," and added that a negative value means the target has less than the source, since the sign is what distinguishes loss from duplication.

hugo/content/en/monitors/types/data_observability.md
kevinzenghu: Made it explicit in 968063783f: "the same difference as a percentage of the source value." It was one clause and it keeps the two bullets parallel, but happy to drop it back to just "as a percentage" if you would rather not spell it out.

hugo/content/en/monitors/types/data_observability.md
kevinzenghu: Removed in 968063783f. You own the behavior, so deferring to you on this one.

Only note for the record: the reason I had included it was that when the comparison is skipped there is no user-visible signal explaining why. If that ever does generate tickets, this is the section to add it back to.

PR #39098, opened 2026-08-11
[WEB-9790] Scaffold hand-written API pages as content-collection plac…

bgdeutsch:
- adds 4 hand-written `mdoc` pages with a todo comment so the content can be pasted in before launch.
- replaces the four per-page routes with one dynamic route and one
  plaintext route.
- derives `pages.json` / `llms.txt` root pages from the content collection instead
  of a hardcoded list, so adding a `.mdoc` page lists it in both without a second edit.

### Merge readiness

- [ ] Ready for merge

## For Datadog employees:

### Additional notes
Steps to verify:

- `yarn dev`
- hardcoded pages appear: http://localhost:4321/api/latest/scopes/, http://localhost:4321/api/latest/rate-limits/, http://localhost:4321/api/latest/using-the-api/
- `/` route still redirects to `/api/latest`
- side nav works as expected
-`yarn build` is successful
- check output of `pages.json` and `llms.txt`, they contain the hardcoded api pages
- passing tests: `yarn test:headless-ai src/components/ApiSideNav src/lib/pagesListing \
  tests/headless/pagesJson.test.ts tests/headless/llmsTxt.test.ts`


astro/src/pages/[...slug].astro
```diff
@@ -5,7 +5,8 @@ import type { GetStaticPaths } from 'astro';
 import BaseLayout from '@layouts/BaseLayout.astro';
 
 export const getStaticPaths: GetStaticPaths = async () => {
-  const docs = await getCollection('en');
+  // api/latest/* has its own route, which renders it with ApiLayout.
+  const docs = await getCollection('en', (entry) => !entry.id.startsWith('api/latest'));
```
Comment (jhgilbert) [inline astro/src/pages/[...slug].astro]: I suggest something like `isApiContent(id)` as a helper (you could export it from overviewPages.ts), so you don't have a repeat of a hard-coded URL like "api/latest" (which also shows up in `[...page].astro`). In practice this probably won't be a problem, but might as well sidestep any need to update multiple hard-coded strings.

I would also add a slash on the end of your matcher, just to avoid matching a string like "api/latest-somethingelse". Again in reality, probably not very likely, so this is bonus safety stuff.

jhgilbert (approved):

PR #39121, opened 2026-08-11
Align tag enrichment telemetry scope and correct Jobs Monitoring product name

OliviaShoup:

Follow-up to #39074, which updated the telemetry types tag enrichment applies to. That PR left the Overview and the **Impacted telemetry** bullet listing different sets of telemetry at different levels of detail, so the two sections gave different answers to "which telemetry gets enriched?"

This PR:

- Makes the Overview list and the **Impacted telemetry** bullet match: logs, APM spans, and trace metrics.
- Scopes the Data Observability: Jobs Monitoring claim. Confirmed with the feature owner that Jobs Monitoring telemetry is enriched because its job telemetry is sent as APM spans, rather than Jobs Monitoring being supported as its own telemetry type. The page now says so, and notes that Jobs Monitoring metrics are not enriched. Previously it listed Jobs Monitoring as a fourth telemetry type alongside APM spans, which implied broader coverage than exists.
- Corrects the product name to **Data Observability: Jobs Monitoring** and links it on first mention. The former name appears once in `content/en`, and the page title is the current name.
- Removes the `(DJM)` abbreviation, which was defined twice and never used afterward. No page in `content/en` introduces it and then reuses it.
- Restores the restrictive "only" that was dropped from the behavior bullet, so the page states a support boundary again, with corrected modifier placement ("applies only to").
- Replaces "Support for additional telemetry types ... is planned" with explicit negation. `is not supported` is the established pattern in the docs, and the previous phrasing goes stale silently as types ship.
- Distinguishes custom and infrastructure metrics from trace metrics, since trace metrics are listed as supported.

Note: `tracing/services/service_remapping_rules.md` lists a broader set of telemetry types for remapping rules, including several this page now lists as unsupported. That page is left unchanged here. It uses six abbreviations that are never expanded, and its scope list belongs to a different feature, so it is better handled separately.

### Merge readiness

- [x] Ready for merge

## For Datadog employees:

### AI assistance

Used Claude Code to research existing naming and phrasing conventions across `content/en` and to draft the wording. All claims were verified against the repo and reviewed manually.

### Additional notes

Vale passes on the changed lines. The one remaining Vale suggestion on this file is pre-existing and outside this diff.

### DIFF
--- a/hugo/content/en/tracing/services/tag_enrichment.md
+++ b/hugo/content/en/tracing/services/tag_enrichment.md
@@ -45,7 +45,7 @@ Custom rules let you target a specific set of services and configure exactly how
    - Select the `team` tag, `system` tag, `custom` tag, or multiple.
    - For each tag, select whether the tag value comes from Entity Metadata, the value of a different tag, or a fixed value.
    - Choose whether the value is applied only when it doesn't already exist, or is appended to the current list of values for that tag.
-1. By default, tags are added to all telemetry.
+1. By default, tags are added only when the value is missing for a given telemetry item.
 1. Optionally, enter a descriptive name for the rule.
 1. Review and save your rule. After you save the rule, it can take up to an hour for enrichment to be fully applied to incoming telemetry.
 
--- a/hugo/content/en/tracing/services/tag_enrichment.md
+++ b/hugo/content/en/tracing/services/tag_enrichment.md
@@ -16,7 +16,7 @@ Tag enrichment is in Preview. To request access, fill out this form.
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


marianafrangos (approved):

hugo/content/en/tracing/services/tag_enrichment.md
rtrieu:
```suggestion
1. By default, tags are added only when the value is missing for a given telemetry item.
```

rtrieu (approved):
minor suggestion but otherwise looks good

PR #39146, opened 2026-08-12
Add Cloud SIEM Content Pack filter types

janine-c:
A designer and PM flagged that the Cloud SIEM Content Pack documentation was missing the types that users can filter Content Packs by in the UI. This adapts the tooltips we have in the UI to make this information surfaceable in the docs and in Ask AI.

### Merge readiness
Will check this box when I've gotten confirmation that this is what they were looking for!
- [x] Ready for merge

## For Datadog employees:


hugo/content/en/security/cloud_siem/ingest_and_enrich/content_packs.md
```diff
@@ -46,6 +46,12 @@ further_reading:
 - [Workflow Automation][4], to automate actions and accelerate investigation and remediation of issues
 - Configuration guides
 
+You can filter Content Packs by the following types:
+- **Content Packs**: Integrations bundled with security-relevant content such as detection rules, SOAR workflows, and custom tooling
```
buraizu:
```suggestion
- **Content Packs**: Integrations bundled with security-relevant content such as detection rules, SOAR workflows, and custom tooling
```

SOAR doesn't seem to be spelled out anywhere, and isn't on the list of well-known terms from the editorial reference. Would it potentially be helpful to spell that out here?

buraizu (commented):
Looks good overall, just posted one question

hugo/content/en/security/cloud_siem/ingest_and_enrich/content_packs.md
janine-c: Great catch, thanks as always!

hugo/content/en/security/cloud_siem/ingest_and_enrich/content_packs.md
```diff
@@ -46,6 +46,12 @@ further_reading:
 - [Workflow Automation][4], to automate actions and accelerate investigation and remediation of issues
 - Configuration guides
 
+You can filter Content Packs by the following types:
+- **Content Packs**: Integrations bundled with security-relevant content such as detection rules, Security Orchestration, Automation, and Response (SOAR) workflows, and custom tooling
+- **Enrichment Packs**: Content to add valuable context to SIEM analysis, such as vulnerabilities or third-party insights, to improve investigations
+- **Integration Packs**: Content curated from Datadog's catalog to be relevant for use with Cloud SIEM
+- **Entity Packs**: Integrations and bundled content that power User and Entity Behavior Analytics by modeling normal activity for users and entities and surfacing risky anomalies in Cloud SIEM
```
buraizu: It looks like an acronym for **User and Entity Behavior Analytics** (UEBA) is introduced here:
https://docs.datadoghq.com/integrations/appomni/#overview

Just flagging in case you want to update this for consistency

hugo/content/en/security/cloud_siem/ingest_and_enrich/content_packs.md
buraizu: Thanks for making that edit. Taking another look, the full form nested inside the series is a bit hard to parse as it nests a comma-separated 3-item series inside another comma-separated 3-item series. Did a bit of digging into this, and came up with a couple of potential alternatives:

1. SOAR is already spelled out on [Respond (SOAR) and Report](https://docs.datadoghq.com/security/cloud_siem/respond_and_report/), so linking there avoids a second copy to maintain and keeps the bullet scannable:

```
- **Content Packs**: Integrations bundled with security-relevant content such as detection rules, [SOAR][5] workflows, and custom tooling
```

That needs a matching reference definition with the others at the bottom of the page:

`[5]: /security/cloud_siem/respond_and_report/`

2. If you'd rather keep the expansion inline, leading with the acronym lets the parentheses do the grouping the commas can't:

```
- **Content Packs**: Integrations bundled with security-relevant content such as detection rules, SOAR (Security Orchestration, Automation, and Response) workflows, and custom tooling
```

hugo/content/en/security/cloud_siem/ingest_and_enrich/content_packs.md
janine-c: Thanks for these suggestions! How do you think it looks now? I was thinking about how to make the acronym definitions a little less intrusive, so I tried putting them in tooltips instead. I'd love to hear your thoughts on whether it solves the problem!

buraizu: @janine-c thanks for trying this — it probably does keep the bullets scannable, although I wasn't able to confirm as the `build_preview` job failed and the tooltips aren't rendering in the preview.

However, digging into this a bit, it looks like the way the shortcode renders might create some other issues for screen readers and keyboard-only users:

- a screen reader reading the page linearly hears "detection rules, SOAR button, workflows, and custom tooling" — the expansion only surfaces if they tab to it.
- a keyboard user who tabs to the button sees nothing appear.

Existing usage in the repo suggests tooltips work best as a supplement rather than the sole source. For example, [this doc](https://docs.datadoghq.com/network_monitoring/devices/snmp_traps/?tab=yaml#resolution) uses `text="object identifier (OID)"` and lets the tooltip carry the definition, so the expansion stays in the visible text either way.

As such, my suggestion is still to keep both expansions in the text — the `[SOAR][5]` link for the first one, since it's already spelled out on `Respond (SOAR) and Report`, and an inline parenthetical for UEBA to match the AppOmni page.

Happy to reevaluate if you have objections or a different perspective. Let me know what you think!

janine-c: Yeah, I'm not sure what's going on with the build; it worked fine locally! I'll keep investigating.

Your feedback about the tooltips is definitely valid, but because the issue isn't unique to this PR and they exist elsewhere in the docs, I don't think it's necessarily worth deciding not to use them here. I'll make a ticket with Websites so we can address it separately, though, because I definitely agree it's something we should change.

For this situation, I think I'd rather have acronyms spelled out on this page. I don't think a user needs to fully understand the concept to get unblocked in understanding what this category is, so I'd rather provide the spelled-out acronym on this page rather than have them click to another one and then find it themselves on that page. That might fall into a bit of a grey area with regards to how much cognitive load is too much, but I prefer to keep things light if I can.

I understand that this is something that we might have to agree to disagree on! Because the designer I was talking to has already approved it over Slack, would it be okay to get an approval on this one so we can put it to bed, and we can always come back and edit it in the future if we decide that it's not doing the job?

buraizu: Thanks for confirming, and sharing your perspective! In that case, would you be open to option 2 that I mentioned earlier?

> - **Content Packs**: Integrations bundled with security-relevant content such as detection rules, SOAR (Security Orchestration, Automation, and Response) workflows, and custom tooling

janine-c: Yes, applied! Thanks for the discussion as always!

buraizu (approved):

PR #39167, opened 2026-08-13
Update Intelligent Correlation enablement steps

fbasciettog-dd:
## Summary
- Preview Cases is no longer part of the Intelligent Correlation onboarding flow.
- Rewrites the **Enable Intelligent Correlation** section on `/events/correlation/intelligent/` to document the current flow: click Edit on the Intelligent Correlator settings card, optionally scope by tag under Define Intelligent Correlator, then Save & Enable.
- Adds two new screenshots (`intelligent_correlator_card.png`, `intelligent_correlator_define.png`) reflecting the updated UI.
- Both existing reference links (`[1]` settings page, `[2]` intelligent rule edit URL) are unchanged destinations — only how they're used in the copy changed.
- Only the "Enable Intelligent Correlation" section changed; Overview, "Receiving your first case", and Further Reading are untouched.

## Test plan
- [ ] Build/preview locally and confirm both screenshots render at full width under their respective steps
- [ ] Confirm reference links `[1]` and `[2]` still resolve
- [ ] Spot-check field labels against the live app (Settings → Intelligent Correlation → Edit)


hugo/content/en/events/correlation/intelligent.md
```diff
@@ -14,11 +14,18 @@ Intelligent Correlation uses a Machine Learning modeling approach. It automatica
 ## Enable Intelligent Correlation
 
 To get started:
-1. Navigate to the [{{< ui >}}Correlation Settings{{< /ui >}}][1] page, and click [{{< ui >}}Preview Cases{{< /ui >}}][2]. 
-1. From there you can preview the intelligent correlations that are created from your organization.
 
+1. On the [{{< ui >}}Intelligent Correlation{{< /ui >}} settings][1] page, click [{{< ui >}}Edit{{< /ui >}}][2] on the {{< ui >}}Intelligent Correlator{{< /ui >}} card.
```
clreaume: reordering the sentence to orient the reader before describing the action
```suggestion
1. On the [{{< ui >}}Intelligent Correlation{{< /ui >}} settings][1] page, on the {{< ui >}}Intelligent Correlator{{< /ui >}} card, click [{{< ui >}}Edit{{< /ui >}}][2].
```

hugo/content/en/events/correlation/intelligent.md
```diff
@@ -14,11 +14,18 @@ Intelligent Correlation uses a Machine Learning modeling approach. It automatica
...
 
-{{< img src="events/correlation/intelligent/intelligent_config_updated.png" alt="Configure intelligent correlation" style="width:100%;" >}}
+   {{< img src="events/correlation/intelligent/intelligent_correlator_card.png" alt="Intelligent Correlation settings page showing the Intelligent Correlator card, currently off, with an Edit button" style="width:100%;" >}}
```
clreaume, line 24: a couple comments on this image:

- noting that link [1] doesn't actually bring me to this Settings > Intelligent Correlation page in the few DD instances I tested in. is this feature/UI enabled for customers?
- I recommend cropping it; it would be easier to see the text if some of the empty white space at the bottom was removed (and optionally, if the window width was reduced)

hugo/content/en/events/correlation/intelligent.md
```diff
@@ -14,11 +14,18 @@ Intelligent Correlation uses a Machine Learning modeling approach. It automatica
...
+
+1. Under {{< ui >}}Define Intelligent Correlator{{< /ui >}}, optionally narrow the events the correlator evaluates, then click {{< ui >}}Save & Enable{{< /ui >}}.
```
clreaume:
```suggestion
1. Under {{< ui >}}Define Intelligent Correlator{{< /ui >}}, optionally narrow the events the correlator evaluates:
```

hugo/content/en/events/correlation/intelligent.md
```diff
@@ -14,11 +14,18 @@ Intelligent Correlation uses a Machine Learning modeling approach. It automatica
...
+   - {{< ui >}}Consider events with any of these tags{{< /ui >}}: only events matching one of these tags are correlated. Leave this empty to consider all Monitor alerts.
+   - {{< ui >}}Exclude events with any of these tags{{< /ui >}}: events matching one of these tags are never correlated, which is useful for filtering out noisy or non-production alerts.
+
+   {{< img src="events/correlation/intelligent/intelligent_correlator_define.png" alt="Define Intelligent Correlator form with fields to include and exclude events by tag, and a Save and Enable button" style="width:100%;" >}}
+
```
clreaume, line 31:
```suggestion

2. Click {{< ui >}}Save{{< /ui >}}.
```
moving this step below the bullets above describing the include/exclude fields. also, it looks like the save button text is just "Save"

hugo/content/en/events/correlation/intelligent.md
```diff
@@ -14,11 +14,18 @@ Intelligent Correlation uses a Machine Learning modeling approach. It automatica
...
-1. Navigate to the [{{< ui >}}Correlation Settings{{< /ui >}}][1] page, and click [{{< ui >}}Preview Cases{{< /ui >}}][2]. 
-1. From there you can preview the intelligent correlations that are created from your organization.
 
+1. On the [{{< ui >}}Intelligent Correlation{{< /ui >}} settings][1] page, click [{{< ui >}}Edit{{< /ui >}}][2] on the {{< ui >}}Intelligent Correlator{{< /ui >}} card.
 
-{{< img src="events/correlation/intelligent/intelligent_config_updated.png" alt="Configure intelligent correlation" style="width:100%;" >}}
+   {{< img src="events/correlation/intelligent/intelligent_correlator_card.png" alt="Intelligent Correlation settings page showing the Intelligent Correlator card, currently off, with an Edit button" style="width:100%;" >}}
+
+1. Under {{< ui >}}Define Intelligent Correlator{{< /ui >}}, optionally narrow the events the correlator evaluates, then click {{< ui >}}Save & Enable{{< /ui >}}.
+   - {{< ui >}}Consider events with any of these tags{{< /ui >}}: only events matching one of these tags are correlated. Leave this empty to consider all Monitor alerts.
```
clreaume:
```suggestion
   - {{< ui >}}Consider events with any of these tags{{< /ui >}}: Only events matching one of these tags are correlated. To consider all monitor alerts, leave this field empty.
```

hugo/content/en/events/correlation/intelligent.md
```diff
@@ -14,11 +14,18 @@ Intelligent Correlation uses a Machine Learning modeling approach. It automatica
...
+   - {{< ui >}}Exclude events with any of these tags{{< /ui >}}: events matching one of these tags are never correlated, which is useful for filtering out noisy or non-production alerts.
```
clreaume:
```suggestion
   - {{< ui >}}Exclude events with any of these tags{{< /ui >}}: Events matching one of these tags are not correlated. This may be useful for filtering out noisy or non-production alerts.
```

hugo/content/en/events/correlation/intelligent.md
```diff
@@ -14,11 +14,18 @@ Intelligent Correlation uses a Machine Learning modeling approach. It automatica
...
+
+   {{< img src="events/correlation/intelligent/intelligent_correlator_define.png" alt="Define Intelligent Correlator form with fields to include and exclude events by tag, and a Save and Enable button" style="width:100%;" >}}
+
+Enabling the Intelligent Correlator applies to your entire organization. After it's enabled, it will evaluate incoming Monitor alerts and group related ones into cases automatically. You can return to the same page at any time to adjust the tag filters or disable the correlator.
```
clreaume:
```suggestion
```
suggest moving this above, to be an introductory paragraph in this section. will add another suggestion above inserting it

hugo/content/en/events/correlation/intelligent.md
wjiang720: This description isn't accurate anymore since we removed the ML model 
"Intelligent correlation automatically aggregates Monitor alerts into work items based on related infrastructure dependencies, underlying telemetry, and other heuristics." - this is what i took from the current setup UI

hugo/content/en/events/correlation/intelligent.md
```diff
@@ -14,11 +14,18 @@ Intelligent Correlation uses a Machine Learning modeling approach. It automatica
 ## Enable Intelligent Correlation
```
clreaume, line 15: moving the paragraph below up here with slight wording tweaks
```suggestion
## Enable Intelligent Correlation

Enabling the Intelligent Correlator applies to your entire Datadog organization. After it's enabled, it evaluates incoming monitor alerts, automatically grouping related ones into cases. You can return to the same page at any time to adjust the tag filters or disable the Correlator.
```

hugo/content/en/events/correlation/intelligent.md
wjiang720: This is the part worth clarifying we're already by default limiting to only paging monitor/whatever the filter we have?

clreaume (approved):
hey there, thanks for drafting this change! I left some questions and suggestions, but nothing is blocking

hugo/content/en/events/correlation/intelligent.md
```diff
@@ -14,11 +14,18 @@ Intelligent Correlation uses a Machine Learning modeling approach. It automatica
...
+
+1. Under {{< ui >}}Define Intelligent Correlator{{< /ui >}}, optionally narrow the events the correlator evaluates, then click {{< ui >}}Save & Enable{{< /ui >}}.
+   - {{< ui >}}Consider events with any of these tags{{< /ui >}}: only events matching one of these tags are correlated. Leave this empty to consider all Monitor alerts.
+   - {{< ui >}}Exclude events with any of these tags{{< /ui >}}: events matching one of these tags are never correlated, which is useful for filtering out noisy or non-production alerts.
+
+   {{< img src="events/correlation/intelligent/intelligent_correlator_define.png" alt="Define Intelligent Correlator form with fields to include and exclude events by tag, and a Save and Enable button" style="width:100%;" >}}
+
+Enabling the Intelligent Correlator applies to your entire organization. After it's enabled, it will evaluate incoming Monitor alerts and group related ones into cases automatically. You can return to the same page at any time to adjust the tag filters or disable the correlator.
```
wjiang720: Maybe add a warning section here explaining that intelligent correlation relies on the relationships we identify from the underlying telemetry. Customers may not see any correlated results if there isn’t enough telemetry to establish those relationships. We also only process paging monitors, so customers without any paging monitors will not see results.

hugo/content/en/events/correlation/intelligent.md
fbasciettog-dd: Applied, though restructured a bit further to also address the settings-link comment below — step 1 now reads: "On the Settings page, find Intelligent Correlation under Projects, then click Edit on the Intelligent Correlator card."

hugo/content/en/events/correlation/intelligent.md
fbasciettog-dd: Good catch — you're right that there's no dedicated deep link; Intelligent Correlation is just one item under Projects in general Settings, and which project ID/position it has varies by org. Updated step 1 to reflect that rather than implying a direct link. Link [1] itself is unchanged (`/event/settings/correlation`) since it already lands on the right general page.

hugo/content/en/events/correlation/intelligent.md
fbasciettog-dd: Applied — now ends in a colon before the bullets, and the Save & Enable click moved to its own numbered step (see below).

hugo/content/en/events/correlation/intelligent.md
fbasciettog-dd: Moved to its own step. On the label: it's conditional in code — `Save & Enable` shows on first-time setup (no org config yet), `Save` shows once config already exists. Since this section is about first-time enablement, we're keeping `Save & Enable`. Let us know if the org you tested in already had config present — that'd explain why you saw `Save`.

hugo/content/en/events/correlation/intelligent.md
fbasciettog-dd: Applied your wording almost verbatim — capitalized "Only," reordered the "leave empty" clause.

hugo/content/en/events/correlation/intelligent.md
fbasciettog-dd: Applied — "not correlated" instead of "never correlated," capitalized "Events."

hugo/content/en/events/correlation/intelligent.md
fbasciettog-dd: Applied both together — the paragraph now opens the "Enable Intelligent Correlation" section instead of closing it, with your wording (org-wide scope, "evaluates... automatically grouping," return-to-adjust-or-disable).

hugo/content/en/events/correlation/intelligent.md
fbasciettog-dd: Applied both together — the paragraph now opens the "Enable Intelligent Correlation" section instead of closing it, with your wording (org-wide scope, "evaluates... automatically grouping," return-to-adjust-or-disable).

hugo/content/en/events/correlation/intelligent.md
fbasciettog-dd: Agreed and fixed — Intelligent Correlation no longer runs on the data science/ML engine. Overview now reads: "Intelligent Correlation automatically aggregates Monitor alerts into cases based on related infrastructure dependencies, underlying telemetry, and other heuristics." Kept "cases" rather than "work items" — that rebrand is still behind an experiment flag and isn't the current default term.

hugo/content/en/events/correlation/intelligent.md
fbasciettog-dd: Confirmed — by default IC only considers Monitor alerts that notify a paging integration (PagerDuty, Opsgenie, VictorOps, or on-call). Added a note below the enable steps to call this out; see the reply on the warning-section comment below for the full text.

hugo/content/en/events/correlation/intelligent.md
fbasciettog-dd: Added a note right after the enable steps:

"Intelligent Correlation relies on relationships identified from your infrastructure and underlying telemetry. If there isn't enough telemetry to establish those relationships, you might not see correlated cases. By default, Intelligent Correlation only considers Monitor alerts that notify a paging integration. If your organization has no monitors configured to page, Intelligent Correlation automatically considers high-priority (P1 or P2) alerts instead."

One nuance we're leaving out for now to keep this readable: orgs get auto-promoted to the P1/P2 fallback only after sustained zero paging volume (and can be demoted back), so it's not an instant switch. Happy to add that detail too if you think customers need it, but wanted to keep the note focused on the two points you raised.

hugo/content/en/events/correlation/intelligent.md
fbasciettog-dd: Swapped in tighter crops for both screenshots — pushed.

hugo/content/en/events/correlation/intelligent.md
wjiang720: How about: Intelligent Correlation relies on relationships identified within your infrastructure based on the underlying telemetry. It considers only alerts from monitors that are important to your organization, such as monitors that notify a paging integration or are tagged as high priority. If there is not enough telemetry to establish these relationships, or if your monitors do not meet these criteria, you may not see any correlated cases.

lotannauo (approved):
Thanks Fran!

hugo/content/en/events/correlation/intelligent.md
```diff
@@ -10,16 +10,27 @@ further_reading:
...
+
 ## Enable Intelligent Correlation
 
+Enabling the Intelligent Correlator applies to your entire Datadog organization. After it's enabled, it evaluates incoming monitor alerts, automatically grouping related ones into cases. You can return to the same page at any time to adjust the tag filters or disable the Correlator.
+
 To get started:
-1. Navigate to the [{{< ui >}}Correlation Settings{{< /ui >}}][1] page, and click [{{< ui >}}Preview Cases{{< /ui >}}][2]. 
-1. From there you can preview the intelligent correlations that are created from your organization.
 
+1. On the [{{< ui >}}Settings{{< /ui >}}][1] page, find {{< ui >}}Intelligent Correlation{{< /ui >}} under {{< ui >}}Projects{{< /ui >}}, then click [{{< ui >}}Edit{{< /ui >}}][2] on the {{< ui >}}Intelligent Correlator{{< /ui >}} card.
```
clreaume:
```suggestion
1. On the [{{< ui >}}Settings{{< /ui >}}][1] page, under {{< ui >}}Projects{{< /ui >}}, click {{< ui >}}Intelligent Correlation{{< /ui >}}.
2. On the {{< ui >}}Intelligent Correlator{{< /ui >}} card, click [{{< ui >}}Edit{{< /ui >}}][2].
```

clreaume: added this suggestion to improve the new wording you pushed, but I think we need to add specificity to "Settings page"—which settings page, and is there a link to it? (I understand from your reply below that there's no deep link to the Intelligent Correlation settings page, but I would expect there to be one to Settings)

clreaume: the existing [1] link redirects for me. should it just be https://app.datadoghq.com/event/settings/ ? (though that seems to redirect to https://app.datadoghq.com/event/settings/work-attributes , which isn't what we want)

hugo/content/en/events/correlation/intelligent.md
clreaume: I think this is great context, but it belongs in the Overview instead of in the enablement instructions section

fbasciettog-dd: Applied — moved a tightened version of this into the Overview section, done.

hugo/content/en/events/correlation/intelligent.md
fbasciettog-dd: Good call, done — split into two separate steps.

fbasciettog-dd: Confirmed with the team — this is expected, not a docs issue. Intelligent Correlation is a per-org dynamic "project" without a stable/knowable ID we could link to directly, so there's no static URL that goes straight to it. Linking to the general Settings page is the correct approach here; if that link lands on a different default tab (e.g. work items) for some orgs, that's the app falling back when no project ID is present, and it's independent of which link we choose. The in-app steps (Settings → Projects → Intelligent Correlation → Edit) handle the navigation from there regardless of the landing tab.

fbasciettog-dd: Tthis is expected for now, not a docs issue. Intelligent Correlation is a per-org dynamic "project" without a stable/knowable ID we could link to directly, so there's no static URL that goes straight to it. Linking to the general Settings page is the correct approach here; if that link lands on a different default tab (e.g. work items) for some orgs, that's the app falling back when no project ID is present, and it's independent of which link we choose. The in-app steps (Settings → Projects → Intelligent Correlation → Edit) handle the navigation from there regardless of the landing tab.

As a follow up, I'll review if we can make a tweak to the settings page to allow direct linking to it.

PR #39105, opened 2026-08-11
[CRED-2815] Document PAT/SAT scope, TTL, and app-key restrictions

emubello:
## What changed

Adds a Restrictions section to the Personal Access Tokens and Service Access Tokens documentation.

The new sections document containment rules for API calls authenticated with a PAT or SAT:

- Cannot create or update application keys. Revocation is still allowed.
- Cannot create or update another token with scopes broader than its own.
- Cannot create a token with a TTL that outlives itself.

Calls that violate these restrictions return a `403 Forbidden` response.

## Why

This redoes the changes from #38945 on a fresh branch after the docs repo reorg. The content now targets the current `hugo/content/...` paths.

## Validation

- `git diff --check`
- `vale hugo/content/en/account_management/personal-access-tokens.md hugo/content/en/account_management/service-access-tokens.md`

Vale reports no errors. Remaining warnings/suggestions are pre-existing lines in these files.

PR #39173, opened 2026-08-13
Drop confusing 'still' from PAT/SAT application key restriction

OliviaShoup:

Follow-up to #39105, addressing [this review comment](https://github.com/DataDog/documentation/pull/39105#discussion_r3768491601) that arrived after the other suggestions had already been committed.

"Revoking application keys is **still** allowed" implies the other two restrictions in the list were previously permitted and have now changed. The real contrast is with the sibling bullets, not with past behavior, so "still" reads as a change note that isn't one. This drops the word from both pages.

The comment was left on the PAT page, but the same sentence appears on the SAT page, so both are updated:

- `personal-access-tokens.md:73`
- `service-access-tokens.md:98`

No other changes.

### Merge readiness

- [x] Ready for merge

### AI assistance

Claude Code made the edit and confirmed Vale reports nothing on the changed lines.


clreaume (approved):
woop

PR #39211, opened 2026-08-14
De-duplicate Hex integration docs

evazorro:

Right now there are two copies of the Hex integration docs: 

* under [integrations/](https://docs.datadoghq.com/integrations/hex/), and
* under [data_observability/](https://docs.datadoghq.com/data_observability/quality_monitoring/business_intelligence/hex/)

I want to consolidate so that there's only a single doc for this integration. In my opinion, the best way to do this is to use the `integrations/` version. But we can leave info about Hex in the Data Observability docs for discoverability, and just link from there to the integrations docs.

This PR removes the content from the duplicate version, and sets up a few links and redirects to the integration docs.

#### Some technical details for the Docs team reviewer

* ~~Normally we add the redirect to the _receiving_ page, but that's not possible here because integration docs don't have the same frontmatter. That's why I'm putting this one on the old page and using `external_redirect`. There's precedent for doing this: see [this old setup guide](https://github.com/DataDog/documentation/blob/master/hugo/content/en/integrations/guide/oracle-fusion-integration-setup.md?plain=1) that now redirects to the main Oracle Fusion integration doc.~~
* I'm also adding Hex to the left navigation, because that was missed in the original PR where it was added.

### Merge readiness

- [ ] Ready for merge

### AI assistance

Used Claude Code to explore different options for restructuring and redirecting, and to make the nav changes. Information architecture, testing, and PR descriptions by me.

### Additional notes
There's a [corresponding PR in integrations-internal-core](https://github.com/DataDog/integrations-internal-core/pull/3907) to make a couple updates to the Hex README. I may move some of the content I deleted here into that version, if the DO team wants (_update_ -> I moved deleted content in [this PR](https://github.com/DataDog/integrations-internal-core/pull/3918)).

#### Also

I am planning to repeat this process with the other duplicate integration docs under [data_observability/quality_monitoring/](https://docs.datadoghq.com/data_observability/quality_monitoring/), if everyone is on board after seeing this proof of concept.


kevinzenghu:
Hi Eva! Apologies for not following up on the Slack thread, and thanks for picking this up.

Agreed on deduplicating when the content is identical. My one ask is on direction: could we keep the DO doc as canonical and link out from `/integrations/hex/` instead?

The reason is that Hex, like most of our Business Intelligence integrations for Data Observability, doesn't do anything on its own. It has no standalone metrics or tile install. It depends on a warehouse like Snowflake being connected first to establish lineage. Someone landing on `/integrations/hex/` expecting a normal integration setup would hit a dead end, so I'd rather reinforce that this is a DO integration.

Longer term, if we can extract generally useful Datadog metrics from Hex, that would justify a broader integration page in addition to the DO content!

estherk15 (approved):

evazorro:
(For posterity: Kevin and I met to discuss, and we're going ahead with the approach originally proposed here, plus a few tweaks to add framing to the landing pages in the DO docs. I'll push an update to this PR.)

PR #39214, opened 2026-08-14
(feat) Img component

hestonhoffman:
Adds Img component with support for images and videos.

- ImgVideo is a child of Img
- Images use CDN links instead of loading from repo. Local dev + staging use the staging URL. Prod uses prod.
- Added a lightbox to replicate behavior on Hugo.
- Added a plaintext twin that renders an img tag with the prod src, caption, and alt text. For inline images, it drops the tag. I decided to just grab the prod source for the image here instead of rendering the path for each environment. I don't think we gain much from rendering the local path in local etc.
- I removed image properties that appear to be dead in Hugo (`href`, `target`, `wide`, `img_param`, `pop_param`, `figure_class`, and `figure_style`).
- I replaced the `style` prop with `widthPercent` and added a validation error that prevents writers from using widthPercent with height/width.

🟢  All tests pass
🟢  You can preview this using `preview` or `dev` now, since they use the same image URLs.


estherk15 (approved):

astro/src/components/Img/plaintext/Img.ts
```diff
@@ -0,0 +1,35 @@
...
+  caption,
+  video,
+  inline,
+}: ImgNodeInput): MarkdocNode {
+  const attributes: Record<string, unknown> = { src };
+  if (alt) attributes.alt = alt;
+  if (video) attributes.video = true;
+  if (caption) attributes.caption = caption;
+  if (inline) attributes.inline = true;
+  return tag("img", attributes);
```
jhgilbert, line 25:
```suggestion
  return tag("image", attributes);
```

astro/src/components/Img/plaintext/tests/unit.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+describe("imgNode", () => {
+  it("renders an img tag with src and alt", () => {
+    const node = imgNode({
+      src: "https://prod.img.url",
+      alt: "Browse the integration pipeline library",
+    });
+    const result = format(node);
+
+    expect(result).toContain("{% img");
```
jhgilbert, line 13:
```suggestion
    expect(result).toContain("{% image");
```

Super minor but for bots, we usually use the most obvious name possible, so it would be `image`. Easy change though, you can just change the rendered Markdoc tag name.

astro/src/components/Img/plaintext/Img.ts
```diff
@@ -0,0 +1,35 @@
...
+ * image, since `src` here is the resolved full-size CDN URL, not the
+ * content-relative path authors write. Layout-only attributes (`width`,
+ * `height`, `widthPercent`, `popup`) are dropped: none of them affect a
+ * plaintext consumer.
+ */
+
+import type { Node as MarkdocNode } from "@markdoc/markdoc";
+import { tag } from "@lib/plaintext/helpers";
+
+export interface ImgNodeInput {
```
jhgilbert: I'm curious why the Preact Img props wouldn't become a shared type instead? You could ignore any prop keys you don't need for the plaintext.

astro/src/components/Img/ImgController.module.css
```diff
@@ -0,0 +1,81 @@
+.img__figure {
+  margin: 0 0 1rem;
+}
+
+.img__image {
+  max-width: 100%;
+  height: auto;
+  /* matches Hugo's $ddgray (#d6d6d6); no equivalent design token exists yet */
+  border: 1px solid #d6d6d6;
```
jhgilbert: You would still use a `hugo` token, declared in the same way the other `hugo` legacy tokens are declared, rather than hardcoding the value here. That way, we'll have it next time, and we also have an obvious hugo token to integrate or deprecate.

astro/src/components/Img/ImgController.module.css
```diff
@@ -0,0 +1,81 @@
...
+}
+
+.img-lightbox__image {
+  max-width: 90vw;
+  max-height: 90vh;
+  object-fit: contain;
+}
+
+.img-lightbox__caption {
+  color: #ffffff;
```
jhgilbert: Should probably use a token to support dark mode later

astro/src/components/Img/ImgController.module.css
```diff
@@ -0,0 +1,81 @@
...
+  border-radius: 50%;
+  width: 5em;
+  height: 5em;
+}
+
+.img-lightbox__spinner {
+  font-size: 10px;
+  position: relative;
+  text-indent: -9999em;
+  border-top: 0.55em solid rgba(234, 234, 234, 0.5);
```
jhgilbert: I won't keep commenting on it, but we're generally trying to avoid hard-coded colors in the CSS, even if it means we have to create a new token to avoid it (but often a reasonable alternative already exists in the tokens). This element is not required to match Hugo exactly, so don't worry about that

astro/src/components/Img/Img.astro
```diff
@@ -0,0 +1,60 @@
+---
+import ImgController from "./ImgController";
+import ImgVideo from "./ImgVideo";
+import { IMAGES_URL } from "@config/images";
+
+interface Props {
```
jhgilbert: I've seen a few flavors of these props, they can probably be rolled into one shared type?

astro/src/components/Img/ImgController.tsx
```diff
@@ -0,0 +1,319 @@
...
+        dialogElement,
+        imageElement.naturalWidth,
+        imageElement.naturalHeight,
+      );
+    }
+    setLoading(false);
+  }
+
+  let media: JSX.Element;
+  const showsLightbox = !inline && popup;
```
jhgilbert: It seems like this file includes quite a bit of lightbox logic that you might not need (if you wind up using `PictureImage` instead of `FigureWithLightbox`), but would still get loaded into the page. It's probably more idiomatic to pull that logic into a `Lightbox` component that can wrap whatever you need it to. (I don't think you need to call it `FigureWithLightbox`, the figure is just the HTML implementation of the lightbox, if I'm understanding what's happening correctly.) That `Lightbox` component can have its own file, allowing you to set aside (both mentally and in the "client bundle" sense) complex logic when it doesn't apply. Preact components can act as wrappers, so anything wrapped inside `<Lightbox></Lightbox>` can still render, similar to how slots work in Astro.

You can do things like this (made-up example) to avoid duplicating `PictureImage` markup the way you have needed to in this file:

```
const picture = <Picture src={src} alt={alt} />;

return useLightbox ? <Lightbox>{picture}</Lightbox> : picture;
```

astro/src/components/Img/Img.astro
```diff
@@ -0,0 +1,55 @@
...
+  popup,
+  srcset,
+  popupHref,
+  imageUrl,
+};
+
+const needsLightbox = !inline && popup;
+---
+
+{
```
hestonhoffman, line 42: We're back to a ternary here, but I felt it was necessary because it allowed me to conditionally hydrate the lightbox.

astro/src/components/Img/ImgController.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+type ImgControllerProps = PictureImageProps & {
+  imageUrl: string;
+  popupHref: string;
+  caption?: string;
+  inline?: boolean;
+  popup?: boolean;
+};
+
+function PictureImage({
```
jhgilbert: Does `Picture` clash with something in Astro?

astro/src/components/Img/ImgVideo.tsx
```diff
@@ -0,0 +1,36 @@
...
+const cl = classListFactory(styles);
+
+interface ImgVideoProps {
+  imageUrl: string;
+  width?: string;
+  height?: string;
+  widthPercent?: number;
+}
+
+export default function ImgVideo({
```
jhgilbert: Does `Video` clash with something in Astro?

astro/src/components/Img/Img.astro
jhgilbert: I think some ternary behavior is hard to avoid with `.astro`, I at least find this short and easy to understand!

astro/src/components/Img/Lightbox.module.css
```diff
@@ -0,0 +1,72 @@
+/* Ported from hugo/layouts/partials/global-modals/global-modals.html +
+   hugo/assets/scripts/components/global-modals.js (Bootstrap 5 Modal). */
+
+/* The trigger link that opens the lightbox. Lives here rather than in
+   ImgController.module.css because Lightbox renders the link itself. */
+.img__link--popup {
+  cursor: zoom-in;
+}
+
+.img-lightbox__overlay {
```
jhgilbert, line 10: I don't object to separate CSS modules for various components if you can get away with it, but just wanted to note as an FYI that I often start with this approach and wind up combining them because I find I need to refer to one of the parent component classes/states for whatever reason. If you run into that, it's OK to combine modules as needed. Separate modules are nicer for codesplitting, so it's a good place to start!

astro/src/components/Img/Img.astro
```diff
@@ -0,0 +1,55 @@
...
+  video ? (
+    <ImgVideo
+      imageUrl={imageUrl}
+      width={width}
+      height={height}
+      widthPercent={widthPercent}
+    />
+  ) : needsLightbox ? (
+    <ImgController client:idle {...imgControllerProps} />
```
jhgilbert, line 51: Love the conditional hydration, that's a clever approach!

astro/markdoc.schema.mjs
```diff
@@ -89,5 +89,33 @@ export default {
...
+        if (widthPercent != null && (width != null || height != null)) {
+          return [
+            {
+              id: "img-widthPercent-conflict",
+              level: "error",
+              message:
+                "The `widthPercent` attribute can't be combined with `width` or `height`.",
+            },
+          ];
+        }
```
jhgilbert, line 115: Claude suggested this validation since we don't use the `caption` prop when it's a `video`:

```
  if (video && caption != null) {
    errors.push({
      id: "img-video-caption-conflict",
      level: "error",
      message: "The `caption` attribute is not supported on video images.",
    });
  }
```

I thought that seemed reasonable, but leave it to you

jhgilbert (approved):

PR #39217, opened 2026-08-15
Fix the annotation key names listed in the AD Annotations v1 tab

jumpe1:

In [Getting Started with Containers → Autodiscovery](https://docs.datadoghq.com/getting_started/containers/autodiscovery/?tab=adannotationsv1), the **AD Annotations v1** tab says the check configuration key is `ad.datadoghq.com/<container-name>.checks`. That is the v2 key — the sentence is duplicated from the v2 tab, where it is correct. The YAML example in the same tab correctly uses `check_names`, `init_configs`, and `instances`, so the tab contradicts itself.

This is worth more than a typo fix. Autodiscovery applies only one annotation format per container, preferring v2 over v1, so a reader who follows both the v1 example and its prose ends up setting v1 and v2 keys together — and the v1 configuration is the one that gets silently dropped.

### Merge readiness

- [x] Ready for merge


estherk15 (approved):
Thanks for the update!

PR #39246, opened 2026-08-17
Clarify Okta SCIM role profile selection

dhdaviddai:
## What changed

- Clarified that Okta administrators must select the user profile for the application configured for Datadog SCIM before adding the `roles` attribute.
- Distinguished that profile from the default Okta user profile.

## Why

The default Okta user profile does not expose the app-specific external name and namespace fields used by the SCIM role mapping.

## Impact

Okta administrators can find the documented fields without changing product behavior.

## Validation

- Vale 3.9.1: 0 errors and no alerts on the changed line
- `git diff --check`


estherk15 (approved):
Thanks for the update!

PR #39314, opened 2026-08-19
Align IaC Security docs with Rule Type terminology

whitemerch:
## What does this PR do? What is the motivation?

Aligns IaC Security public documentation with the product UI, which uses **Rule Type** (not "category") for IaC rule classification.

## Changes

- **`iac_security/configuration.md`**: Update user-facing prose and the section heading from "Categories" to "Rule types". YAML keys (`ignore-categories`, `only-categories`) are unchanged, as this is how they are programmatically defined to be handled by the scanner, it would be a breaking change.
- **`layouts/iac_security/list.html`**: Rename the public IaC Rules browser filter label from "Categories" to "Rule Types".

## Merge readiness

- [x] Ready for merge

### DIFF
--- a/hugo/content/en/security/code_security/iac_security/configuration.md
+++ b/hugo/content/en/security/code_security/iac_security/configuration.md
@@ -18,7 +18,7 @@ further_reading:
     text: "IaC Security Rules"
 ---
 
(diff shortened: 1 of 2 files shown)


maycmlee (approved):

StefonSimmons (approved):
reasonably contained for a UI update

PR #39327, opened 2026-08-19
Update github login button image on team page

jcharry:

Updates the Log in to GitHub image on the team detail page. The team detail page was recently redesigned and the old image is no longer accurate.

### Merge readiness

- [ ] Ready for merge

## For Datadog employees:


buraizu:
@jcharry thanks for the PR! The new image looks fine — but the old image needs to stay in the repo, and the new image should be uploaded with a new name (and `img` elements likewise updated in any docs that use it). This is because of how translations work in the docs. Let me know if you'd like more details, and I'll be happy to point you to some internal resources.

jcharry:
Thank you for the heads up @buraizu. Now I know :)

I've updated the pull request with a new image and updated the places where it's used.

buraizu (approved):

buraizu:
@jcharry thanks for following up, approved. For future reference, there's no need to update the translated docs — this is managed through an automated process, and your change to the `es` directory may be overwritten at some point before the change makes its way through the translation process. Not a breaking issue here — but just in case you see the image revert to its previous version in the Spanish docs, you'll know the cause.

PR #39386, opened 2026-08-21
RUM: document exclusion filters and restructure the retention filters page

qsellem:

Documents **exclusion filters**, a new type of RUM filter evaluated between the permanent and the custom retention filters, and reorganizes the retention filters page around the three filter types that now exist.

An event matching an exclusion filter is not evaluated against the custom retention filters, so it can never be the reason its session is retained. Today customers achieve this by repeating negative queries inside every filter — one org maintains the same bot exclusion across 179 custom retention filters — which is the pain this feature removes.

**New page structure** (previous order in parentheses):

| Section | Change |
|---|---|
| Overview | edited — names the three filter types |
| How it works | edited and shortened |
| **Types of filters** | **new** — the path an event takes, plus a table of what a match does |
| Permanent retention filters | unchanged content, new screenshot |
| **Exclusion filters** | **new**, marked Preview, with the Error Tracking filter as a subsection |
| **Custom retention filters** | **new wrapper** around *Excluding sessions using retention filters* (was `##`) |
| **Setup** | **new** — how to reach the page, with *Creating a filter*, *Modifying filters* and *Capping retention with quotas* as subsections |
| How retention filters work with replays | moved down |
| Cross-product retention filters, Best practices, API, Next steps | unchanged |

Sections follow the order in which an event is evaluated: permanent, exclusion, custom.

**The distinction the page now makes explicit:** retention filters are session-scoped — one matching event retains the whole session. Exclusion filters are event-scoped — a matching event is skipped, but its session can still be retained by another event. This is the single most confusable part of the feature, so it gets a callout in *Exclusion filters*.

**Other changes**

- Rewrote *Excluding sessions using retention filters* rather than deleting it: in-query exclusions still work but only apply to the filter that contains them, with a pointer to exclusion filters for configuration-wide exclusion.
- Updated `guide/retention_filter_best_practices.md` to mention exclusion filters, and fixed its link to `retention_filters/#logic` — an anchor that does not exist on the page (now `#how-it-works`).
- Heading **text** for `Modifying filters` and `Excluding sessions using retention filters` is unchanged even though both are now nested, so existing deep links from the best practices guides still resolve.

### Merge readiness

- [ ] Ready for merge

Remaining before merge:

1. **Two screenshots still show the previous UI**: `modifying-filters.png`, and the `rum-without-limits-how-retention-filters-work-2.png` flow diagram at the top of *How it works*, which still shows the two-stage model. The three filter-group screenshots and the permanent filters screenshot are updated to the new grouped UI.
2. **Preview launch timing** — the Exclusion filters section links to https://www.datadoghq.com/product-preview/rum-exclusion-filters/, which 404s until the form is published.

### AI assistance

Used Claude Code to draft the restructure from the product brief and to audit inbound anchor links across the docs. Reviewed manually.

### Additional notes

Open questions for reviewers:

- **Error Tracking exclusion filter default state** — the page says you can enable or disable it but not delete it. It does not say whether it is on or off by default for existing orgs, because the brief doesn't specify. Needs a sentence once decided.
- **API** — the API section is unchanged. The API reference does not document exclusion-filter endpoints yet (`/api/latest/` has `rum-retention-filters` and `rum-retention-quota` only), so linking there now would send readers to a page that can't help them. Follow-up once the endpoints publish, along with Terraform coverage if applicable.
- **UI steps for creating an exclusion filter** were written against the design prototype in the brief and should be checked against the shipped UI.


hugo/content/en/real_user_monitoring/rum_without_limits/retention_filters.md
```diff
@@ -70,72 +84,122 @@ There are three permanent retention filters:
...
+To delete a filter:
 
 1. Hover over the filter and click the {{< ui >}}Delete{{< /ui >}} icon.
 1. Click {{< ui >}}Confirm{{< /ui >}}.
 
-### Disable a filter
+#### Disable a filter
 
-Disabled filters simply ignore events and do not make any sampling decisions. Events flowing in the list will skip disabled filters.
+Disabled filters simply ignore events. A disabled custom retention filter makes no sampling decision, and a disabled exclusion filter excludes no event.
```
OliviaShoup: "simply" is on our remove-filler-word list

```suggestion
Disabled filters ignore events. A disabled custom retention filter makes no sampling decision, and a disabled exclusion filter excludes no event.
```

hugo/content/en/real_user_monitoring/rum_without_limits/retention_filters.md
```diff
@@ -70,72 +84,122 @@ There are three permanent retention filters:
...
+Custom retention filters are the filters you create and control. Each one pairs an event type with a query and a retention rate, and each one can retain sessions.
 
-To create a retention filter:
+### Excluding sessions using retention filters
+
+Custom retention filters specify which sessions to keep, rather than which to exclude. You cannot set a retention percentage to 0% (the minimum is 0.1%). Additionally, setting low retention percentages is not an effective exclusion strategy because sessions may still be retained by other filters in your configuration.
+
+To keep a single filter from matching a subset of events, add exclusions **inside that filter's query**. For example:
+
+- Adding `-version:(1* OR 2*)` to a retention filter ensures that filter never keeps events from older versions 1 and 2 of your application.
```
OliviaShoup: this is a style guide thing

```suggestion
- Adding `-version:(1* OR 2*)` to a retention filter helps ensure that filter never keeps events from older versions 1 and 2 of your application.
```

hugo/content/en/real_user_monitoring/rum_without_limits/retention_filters.md
```diff
@@ -40,27 +42,39 @@ A session is stored as soon as a retention filter matches one of its constitutin
 
 The logical flow of retention filters is the following:
 
-- All RUM events are evaluated against each filter in sequence, starting with the first one received.
-- When an event `A` matches a filter, a decision is made based on the retention rate to either sample the entire session in, or wait for future events to be evaluated. In both cases, event `A` is not evaluated further against subsequent retention filters. This is why the **order of retention filters matters**.
-- Retained sessions are saved and accessible in the Session Explorer and other RUM pages. New events coming from this session do not go through the list of retention filters, but are automatically kept to ensure complete visibility.
+- Each event is evaluated against your filters in sequence.
+- When an event matches a filter, the retention rate decides retention for the whole session. Either way, the event is not evaluated against the following filters, which is why the **order of retention filters matters**.
+- Once a session is retained, it is available in the Session Explorer and other RUM pages, and its later events are kept automatically instead of going through the filters.
```
OliviaShoup: totally optional 
```suggestion
- After a session is retained, it is available in the Session Explorer and other RUM pages, and its later events are kept automatically instead of going through the filters.
```

OliviaShoup (approved):
this looks great! 

one non-blocking structural note. `## Exclusion filters` and `### Excluding sessions using retention filters` (a subsection two headings later, under Custom retention filters) are very similarly named for two different concepts. I see the PR description explains the heading text is intentionally unchanged to preserve existing deep links but flagging since a reader or AI agent skimming headings alone could conflate the two

qsellem:
Good catch — renamed `### Excluding sessions using retention filters` to `### Excluding events with a filter query` (b63a258), and updated the link to it in `retention_filter_best_practices.md` in the same commit.

The reasoning: "query" is the word that separates the workaround from the feature, and the section is now mostly a pointer to exclusion filters anyway. The anchor argument turned out to be weaker than I thought — the only in-repo link was one this PR already touches, so what breaks is external deep links (Slack threads, support macros, bookmarks), which Hugo can't alias for anchors. Seemed worth a one-time cost against a permanent conflation risk.

Does the new heading read clearly enough to you, or would something like `Excluding events without exclusion filters` land better? Happy to change it again.

Also applied your three suggestions.

PR #39370, opened 2026-08-20
Add missing Observability Pipelines packs to left nav

daniellepark-datadog:
Several packs merged in prior PRs were never added to the left nav menu, so they weren't discoverable (e.g. MITRE ATT&CK Windows Enrichment). This adds all 101 existing packs to the nav in alphabetical order.

### Merge readiness

- [ ] Ready for merge

## For Datadog employees:

- ⚠️ Your branch name **MUST** follow the `<name>/<description>` convention and include the forward slash (`/`). If you've already created your PR with an incorrect branch name, please rename your branch and open a fresh PR.
- 🤖 **New**: Comment with `/review` to run an automated check that catches common issues before a Documentation team member reviews your PR.

PR #39395, opened 2026-08-21
Use a filterable table for Observability Pipelines packs

domalessi:
## Summary

- Replaces the 101-item packs bullet list (and the 25 packs already in the left nav) with a `multifilter-search` table on the Packs hub, filterable by source, destination, and type.
- Keeps **Packs** as a single left-nav entry. Individual pack pages stay published and linked from the table, but they are no longer listed in the sidebar.
- Drops `private: true` on the packs section so pack pages can show up in search.

This is an alternative to #39370, which adds all 101 packs to the left nav. No other left-nav group on the site is larger than ~29 items. Large parallel catalogs elsewhere use a hub page instead: SDS library rules and APM recommendations (`multifilter-search`), dashboard widget types (`whatsnext` cards), integrations, and OOTB security rules.

## Test plan

- [ ] Preview the [Packs](https://docs.datadoghq.com/observability_pipelines/packs/) hub: table renders, search works, and the Source / Destination / Type filters narrow the list.
- [ ] Confirm the left nav under Observability Pipelines shows **Packs** as one item (no pack children).
- [ ] Open several pack rows (a processing pack, a destination-mapping pack, and a MITRE enrichment pack) and confirm the linked pages still load.
- [ ] Spot-check a few filter values (for example Source = AWS, Destination = Microsoft Sentinel, Type = Enrichment).

Made with [Cursor](https://cursor.com)


OliviaShoup (approved):
chef's kiss

PR #39430, opened 2026-08-24
Add skill-callout shortcode for agent skill install commands

brett0000FF:

Replaces the generic tip admonition used for the `dd-apm` skill install command with a bordered card that has a heading and supporting text, so the skill install step reads as a distinct call to action rather than an aside.

The markup lives in a shared partial with a new `skill-callout` shortcode in front of it, so other skill install commands can use the same treatment. `dd-apm-skill` becomes a thin wrapper over that partial, so the six pages that already use it need no changes.

The command inside the card renders as an ordinary shell code block. It keeps the site's standard code block styling and the hover copy button that `copy-code.js` appends to every fenced shell block, rather than introducing a separate copy control.

Card typography and shadow values are taken from existing site styles: body copy matches the site's paragraph scale, the inline code pill matches the site's inline code sizing, and the shadow reuses the value already used for tooltips.

Usage:

```
{{< skill-callout
    title="Set up APM with an agent"
    text="Install the `dd-apm` skill in your AI coding agent for guided APM setup." >}}
npx skills add https://github.com/datadog-labs/agent-skills --skill dd-apm --full-depth -y
{{< /skill-callout >}}
```

### Merge readiness

- [ ] Ready for merge

## For Datadog employees:

- :warning: Your branch name **MUST** follow the `<name>/<description>` convention and include the forward slash (`/`). If you've already created your PR with an incorrect branch name, please rename your branch and open a fresh PR.
- :robot: **New**: Comment with `/review` to run an automated check that catches common issues before a Documentation team member reviews your PR.

### AI assistance

Claude Code was used to build the shortcode and styles, and to verify the rendered output against a local build.


jeff-morgan-dd (approved):

hugo/layouts/shortcodes/skill-callout.html
```diff
@@ -0,0 +1,42 @@
...
+@param {string} text Supporting sentence, rendered as markdown so backticks
+    become inline code. Optional.
+@param {string} action_name Action name reported to RUM, so this command's copy
+    button is distinguishable from other copy buttons on the page. Optional.
+*/ -}}
+{{- $command := trim .Inner "\n\r\t " -}}
+{{- if not $command -}}
+    {{- errorf "skill-callout: needs the install command as inner content (called from %s)" .Page.File.Path -}}
+{{- end -}}
+<div class="skill-callout">
```
jhgilbert: If you add `plaintext-section` as a class as well, this will be wrapped in a box in the plaintext build. Optional, just depends on if you want it to act like a standalone section in the build / have the bottom boundary marked out clearly.

jhgilbert (approved):
I left an optional comment, but also wanted to remind you to add a VS code snippet for the new component. I think `$ddpurple` is also already available in SCSS (I see it used elsewhere, like in the tooltip for example) but my SCSS knowledge is not amazing, so you may have a reason for doing it the way you did.

Beyond that, LGTM, thanks!

PR #39485, opened 2026-08-26
astro Docs: `oneOf` schema support 

StefonSimmons:

Ports the `oneOf` option labeling from [#38288](https://github.com/DataDog/documentation/pull/38288) (Hugo) into the `astro/` API docs, so both sites label the same way.

Union options in the API schema tables were labeled positionally. They now take a label derived from a discriminant property (`type` or `data_source`, falling back to `Object N` when no discriminant uniquely identifies a branch:

**Example:**
| Before | After |
| --- | --- |
| `Option 1` | `<type=timeseries>` |
| `Option 2` | `<data_source=metrics>` |
| `Option 3` | `Object 3` |

Discriminant detection matches [#38288](https://github.com/DataDog/documentation/pull/38288): a candidate property must be a string enum in every branch that defines it, a branch keeps its name only when its single-value enum is unique across branches

**motivation**: <https://datadoghq.atlassian.net/browse/WEB-9792>

### What changed

- `astro/src/lib/api/refResolver.ts` — adds `getBestDiscriminant` and `unionOptionLabels`, with `groupBy` and `resolveProperty` helpers. Both union call sites (`schemaToFields` and `propertyToField`) now use them.
- `astro/src/lib/api/refResolver.test.ts` — 12 new tests covering discriminant selection, the acceptance thresholds, `$ref` resolution, and label fallback. Two existing tests updated for the new fallback label.

### Pages to check

Schema table rows render collapsed on both sites, so each path below needs expanding before the labels appear. Each entry links to the equivalent page on the live Hugo site — the labels should match.

1. **Logs Pipelines — Get all pipelines** — `/api/latest/logs-pipelines/get-all-pipelines/` ([Live Hugo version](https://docs.datadoghq.com/api/latest/logs-pipelines/#get-all-pipelines))

   Response → `processors`. 20 of 22 options named: `<type=grok-parser>`, `<type=date-remapper>`, `<type=attribute-remapper>`. One level of expansion, so this is the quickest check.

2. **Action Connection — Create a new Action Connection** — `/api/latest/action-connection/create-a-new-action-connection/` ([Live Hugo version](https://docs.datadoghq.com/api/latest/action-connection/#create-a-new-action-connection))

   Request body → `data` → `attributes` → `integration`. 24 of 24 named: `<type=AWS>`, `<type=Anthropic>`, `<type=Azure>`. The largest union in either spec that discriminates perfectly.

3. **Synthetics — Create a test** — `/api/latest/synthetics/create-a-test/` ([Live Hugo version](https://docs.datadoghq.com/api/latest/synthetics/#create-a-test))

   Request body → `config` → `request` → `basicAuth`. 7 of 7 named: `<type=web>`, `<type=sigv4>`, `<type=ntlm>`, `<type=digest>`. Deeply nested, so it exercises whether labels survive several levels down.

4. **Service Definition — Get all service definitions** — `/api/latest/service-definition/get-all-service-definitions/` ([Live Hugo version](https://docs.datadoghq.com/api/latest/service-definition/#get-all-service-definitions))

   Response → `data` → `attributes` → `schema`. The discriminant here is `schema-version`, not `type`, giving `<schema-version=v1>` through `<schema-version=v2.2>`. Confirms detection is not hardcoded to `type`.

5. **Metrics — Query scalar data across multiple products** — `/api/latest/metrics/query-scalar-data-across-multiple-products/` ([Live Hugo version](https://docs.datadoghq.com/api/latest/metrics/#query-scalar-data-across-multiple-products))

   Request body → `data` → `attributes` → `queries`. The mixed case: 6 of 8 named, so `Object 1` and `Object 2` sit directly beside `<data_source=apm_resource_stats>` and `<data_source=slo>`. Best page for judging whether the `Object N` fallback reads well next to real labels.

### Differences from the Hugo implementation

Two deliberate divergences, each driven by how the two sites read the spec:

- Hugo operates on a fully dereferenced spec, so `item.properties.type` is already the enum schema. Astro reads the raw spec, where the discriminant is almost always a `$ref` (for example `#/components/schemas/AlertValueWidgetDefinitionType`). `resolveProperty` follows that hop; without it no branch gets a name.
- Labels carry raw `<` and `>` instead of `&lt;`/`&gt;`. Astro escapes at render time, and the plaintext renderer needs the unescaped form.

Checked against Hugo's exported `getBestDiscriminant` across the live v1 and v2 specs: 167 unions compared, 0 label mismatches.

The `widgets.json` half of #38288 has no counterpart here, since the Astro site does not generate that file. (See [Additional notes](#3-additional-notes) for more)

### Merge readiness

- [x] Ready for merge

## For Datadog employees:

- ⚠️ Your branch name **MUST** follow the `<name>/<description>` convention and include the forward slash (`/`). If you've already created your PR with an incorporate branch name, please rename your branch and open a fresh PR.

### AI assistance

Claude Desktop drafted the implementation and tests; I reviewed the result and verified its labeling against the live Hugo site. I also ran tests and update snapshots

### Additional notes
Why the widget half of PR#38288 isn't represented here

The `addWidget` recursion walks `WidgetDefinition.oneOf` and writes `hugo/content/en/api/v1/dashboards/widgets.json`, a lookup keyed by widget type. The `{{< dashboards-widgets-api >}}` shortcode reads that file to render the `## API` section on each page under `/dashboards/widgets/`.

It lives in `build-api-pages.js` because it uses that file's `schemaTable()` and `filterExampleJson()` helpers along with the `full_spec.yaml` input. The dependency is on those two functions rather than on the generation of API Docs/pages. 

So, none of this applies to `astro/` which is only serving API Docs/pages `full_spec.yaml` right now


janine-c (commented):
Looks like we don't need to approve this to merge, but commenting to get it out of our review queue :D

jhgilbert:
I'm seeing some variation between Hugo and Astro, like this one (the Get All Pipelines page linked above, in the processors row of the table).

Astro:
Screenshot 2026-08-28 at 1 53 52 PM

Hugo: 
Screenshot 2026-08-28 at 1 53 31 PM

Note also that the boundary of one table cell is pushed out oddly in Astro, that might be a pre-existing issue with longer cells though.

StefonSimmons:
@jhgilbert  Follow up `astro` `api/` items. Reviewing this PR against the Hugo equivalent surfaced some more differences. None of them relate to `oneOf`, so are excluded here and belong in a separate PR. In our `astro` `api/` implementation:

1. Enumerated fields report their `type` instead of _`enum`_
   - Seen at http://localhost:4321/api/latest/synthetics/create-a-test `Request Body → config → assertions → Object 1 → operator`. Hugo renders that row as `operator [required] | enum |...`. Astro should match that instead of  `operator [required] | string |...`
   - The fix is to return "enum" when `schema.enum` is present. 
1. Fields with a `format` prop report both the `type` and the `format`
   - Seen at http://localhost:4321/api/latest/dashboard-lists/get-a-dashboard-list  `Response`, where Hugo renders `id` as `int64` and `created` and `modified` as `date-time`. Astro should match that instead of adding the `type` there
   - The fix is to return `format` only when present, matching Hugo's `format || type` evaluation.
1. **Request** bodies render `readOnly: true` fields instead of omitting
   - Seen at http://localhost:4321/api/latest/synthetics/create-a-test `Request Body`, where Astro lists `creator`, `monitor_id` and `public_id`. Also at http://localhost:4321/api/latest/action-connection/create-a-new-action-connection/ `Request Body → data`, where Astro lists `id` but Hugo doesnt. 
   - Hugo omits `readOnly: true` fields from **Request** tables. Astro renders them, dimmed and italicised via `.schema-table__row--readonly`.

StefonSimmons:
> I'm seeing some variation between Hugo and Astro, like this one (the Get All Pipelines page linked above, in the processors row of the table).
> 
> Astro: Screenshot 2026-08-28 at 1 53 52 PM
> 
> Hugo: Screenshot 2026-08-28 at 1 53 31 PM
> 
> Note also that the boundary of one table cell is pushed out oddly in Astro, that might be a pre-existing issue with longer cells though.

@jhgilbert thanks for catching this. I have a fix ready to review

jhgilbert (approved):

PR #39558, opened 2026-08-28
Hide write-only fields from response examples

GabeWM:

API reference response examples are synthesized from their schemas. The renderer filters `readOnly` properties from requests but currently includes `writeOnly` properties in responses, which can produce payloads that the API never returns.

Exclude `writeOnly` properties from synthesized response examples for all API reference pages.

### Merge readiness

- [x] Ready for merge

### AI assistance

Used pi to trace the renderer behavior, implement the filter, and add the regression test.

### Additional notes

Found while reviewing the Workflow Automation preview for DataDog/datadog-api-spec#6506.


jhgilbert (approved):

jhgilbert:
Verified locally: 
Screenshot 2026-09-01 at 1 18 53 PM

PR #39567, opened 2026-08-28
[DOCS-15410] Correct log-based metric stream filter query syntax

maxvp:

Addresses DOCS-15410.

- Replace the "syntax is the same as Log Explorer Search" claim on the log-based metric filter step
- Document that full-text search (`*:search_term`) is unsupported in log-based metric filters
- Add log-based metric filters to the full-text search limitation banner on the Log Explorer search syntax page

### Merge readiness

- [x] Ready for merge

## For Datadog employees:

### AI assistance

Used Claude Code to survey existing filter-query guidance across the docs, draft the changes, and design the in-app test matrix. Product behavior was verified manually in the Datadog app.

### Additional notes

When creating a full-text query in `Generate Metric`, the application surfaces a warning that full-text- is not supported for custom metric filters.


hugo/content/en/logs/log_configuration/logs_to_metrics.md
```diff
@@ -46,13 +46,15 @@ You can also create metrics from an Analytics search by selecting the {{< ui >}}
...
 
-1. {{< ui >}}Input a query to filter the log stream{{< /ui >}}: The query syntax is the same as for the [Log Explorer Search][6]. Only logs ingested with a timestamp within the past 20 minutes are considered for aggregation. The index must be excluded from the query.
+1. {{< ui >}}Input a query to filter the log stream{{< /ui >}}: Write the query using the [Log Explorer search syntax][6]. Datadog evaluates the stream filter query against the log stream at ingest time, not against indexed logs. The filter does not support every Log Explorer search feature, including [full-text search][12] (`*:search_term`). A query that returns results in the Log Explorer can still match no logs here. Aggregation only considers logs ingested with a timestamp within the past 20 minutes. Exclude the index from the query.
+
+   After you save the metric, confirm that it produces datapoints in the [Metrics Explorer][13] before you rely on it. If the query matches logs in the Log Explorer but the metric stays empty, rewrite the term as a quoted phrase, for example `message:"Database operation failed."`.
```
janine-c:
```suggestion
   After you save the metric, confirm that it produces datapoints in the [Metrics Explorer][13] before you rely on it. If the query matches logs in the Log Explorer but the metric stays empty, rewrite the term as a quoted phrase; for example, `message:"Database operation failed."`.
```

hugo/content/en/logs/log_configuration/logs_to_metrics.md
```diff
@@ -46,13 +46,15 @@ You can also create metrics from an Analytics search by selecting the {{< ui >}}
...
 2. {{< ui >}}Select the field you would like to track{{< /ui >}}: Select `*` to generate a count of all logs matching your query or enter a log attribute (for example, `@network.bytes_written`) to aggregate a numeric value and create its corresponding `count`, `min`, `max`, `sum`, and `avg` aggregated metrics. If the log attribute facet is a [measure][7], the value of the metric is the value of the log attribute.
 3. {{< ui >}}Add dimensions to `group by`{{< /ui >}}: By default, metrics generated from logs do not have any tags unless explicitly added. Any attribute or tag dimension that exists in your logs (for example, `@network.bytes_written`, `env`) can be used to create metric [tags][8]. Metric tags names are equal to the originating attribute or tag name, without the @.
```
janine-c:
```suggestion
3. {{< ui >}}Add dimensions to `group by`{{< /ui >}}: By default, metrics generated from logs do not have any tags unless explicitly added. Any attribute or tag dimension that exists in your logs (for example, `@network.bytes_written`, `env`) can be used to create metric [tags][8]. Metric tags names are equal to the originating attribute or tag name, without the `@`.
```
Can we sneak these backticks in?

hugo/content/en/logs/log_configuration/logs_to_metrics.md
```diff
@@ -46,13 +46,15 @@ You can also create metrics from an Analytics search by selecting the {{< ui >}}
 
 {{< img src="logs/processing/logs_to_metrics/create_custom_metrics2.png" alt="Create a Logs to metric" style="width:80%;">}}
 
-1. {{< ui >}}Input a query to filter the log stream{{< /ui >}}: The query syntax is the same as for the [Log Explorer Search][6]. Only logs ingested with a timestamp within the past 20 minutes are considered for aggregation. The index must be excluded from the query.
+1. {{< ui >}}Input a query to filter the log stream{{< /ui >}}: Write the query using the [Log Explorer search syntax][6]. Datadog evaluates the stream filter query against the log stream at ingest time, not against indexed logs. The filter does not support every Log Explorer search feature, including [full-text search][12] (`*:search_term`). A query that returns results in the Log Explorer can still match no logs here. Aggregation only considers logs ingested with a timestamp within the past 20 minutes. Exclude the index from the query.
```
janine-c, line 49: This is a lot of information! I don't think it's wrong to have all in one place like this, per se, but I do get concerned about big piles of conceptual information potentially muddying the user's ability to follow the procedural information in this list.

I wonder if it would make sense to add a "how it works" section or similar before this procedure, where we can have stuff like what Datadog does separate from what the user needs to do. I think that reduces some cognitive load for users; they can have their expectations set before they start their task, and then just focus on executing it while they're reading this, rather than trying to think through what comes next. If you decide to go that route, it definitely doesn't have to be in this PR, though; all this information is correct, and you can totally move it around in the future if you want to.

janine-c (approved):
Looks great! I made an IA suggestion that I think might make it easier for users to cleanly separate out the conceptual vs. procedural information they have to know to get what they want out of their queries, but like I said in my comment, it doesn't have to be something we think about now. If you want to chat through options or anything like that, don't hesitate to let me know!

hugo/content/en/logs/log_configuration/logs_to_metrics.md
maxvp: Thanks for the review! Agree on adding a new section; filed DOCS-15574 to avoid scope creeping this PR.

PR #39583, opened 2026-08-31
Document Published Analysis MCP tools

vitalii-dd:
Summary:
- Document Published Analysis MCP tools in the DDSQL tool docs.
- Add example prompts for create, update, and delete workflows.
- Mention ddsql_schema_search_tables as a way to find a Published Analysis before deletion.

Ticket: https://datadoghq.atlassian.net/browse/NBK-4732
Related implementation PR: https://github.com/DataDog/dd-source/pull/456699

Recreated from closed PR #37548, rebased onto current master (repo reorg moved the file to hugo/content/en/mcp_server/tools.md).

PR #39609, opened 2026-09-01
Document the notebooks MCP toolset

andyjmorgan:
### What does this PR do?

Adds a `## Notebooks` section to the MCP Server tools page documenting the `notebooks` toolset and the two tools that will serve it: `create_notebook_comments` and `add_notebooks_to_collection`.

Placed between Networks and Onboarding to keep the toolset sections alphabetical.

### Motivation

Neither tool is documented today. Both currently sit on the `experimental` toolset, which is not disclosed anywhere on this page, so this is their first public documentation rather than a relabelling.

Agent Experiences is retiring `experimental` org-wide at the end of Q3. The `notebooks` toolset is where the notebook tools move, so it needs a public entry before that move is customer-visible.

### Additional Notes

The section uses the existing non-default-toolset alert, matching `cases`, `experiments`, and `product-analytics`, since `core` is the only default toolset.

Depends on two dd-source changes, and should not merge before the first has deployed:

- DataDog/dd-source#76032 marks the `notebooks` toolset GA and client-visible. Until that ships the toolset is absent from the public endpoint, so a customer following these docs could not enable it.
- DataDog/dd-source#76035 moves the two tools onto the toolset.

Open question for reviewers, and the reason this is a draft: both tools are additionally gated per org by feature flag (`notebook-collections` and `notebooks-comments-mcp`). The alert box covers the toolset not being enabled by default, but nothing on this page documents per-tool flags, so a customer who enables `?toolsets=notebooks` may still see no tools. If that is not acceptable, the alternative is the stronger Preview alert used by `apm`, `code-exec`, and `remote-actions`, which would need a product-preview signup URL.

Internal review from the notebooks team first, then the Documentation team, following the same path as #39583.

### DIFF
--- a/hugo/content/en/mcp_server/tools.md
+++ b/hugo/content/en/mcp_server/tools.md
@@ -1260,6 +1260,30 @@ Retrieves all network interfaces for a specific device.
 - Show me all interfaces on device `device:abc123`.
 - List the interface statuses for my core router.
 
+## Notebooks
+
+<div class="alert alert-info">The <code>notebooks</code> toolset is not enabled by default. See <a href="/mcp_server/setup">Set Up the Datadog MCP Server</a> for instructions on enabling toolsets.</div>
+
+Extended tools for [notebooks][57], beyond the notebook tools included in the `core` toolset.
+
+### `create_notebook_comments`
+*Toolset: **notebooks***\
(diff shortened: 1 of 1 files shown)


andyjmorgan:
Superseded: wrong scope. Replaced by a PR that documents the notebooks toolset in the Available toolsets list rather than documenting individual tools.

PR #39610, opened 2026-09-01
Document the notebooks MCP toolset

andyjmorgan:
### What does this PR do?

Adds `notebooks` to the **Available toolsets** list on the MCP Server setup page. One bullet, placed alphabetically between `networks` and `onboarding`.

### Motivation

The `notebooks` toolset is not currently listed, so there is no public reference for it — a customer has no way to know the toolset name to pass in `?toolsets=`.

The notebook tools that will serve it are moving off the `experimental` toolset, which Agent Experiences is retiring org-wide at the end of Q3.

### Additional Notes

Listed under **Available toolsets** rather than **Preview toolsets** because DataDog/dd-source#76032 marks the toolset `isGA`, which puts it in the `all` alias. That PR needs to merge and deploy before this one — the list is introduced as "These toolsets are generally available", so listing it earlier would describe a toolset customers cannot yet enable.

Deliberately scoped to the toolset. This PR does not document individual tools; the tool entries on the tools page can follow once the tools have moved and their per-org gating is settled.


andyjmorgan:
Superseded by #39615, which pushes the same signed commit to a branch in this repo rather than a fork, so the preview-link and check-site-region jobs can post their comments.

PR #39617, opened 2026-09-01
(WEB-9885) Migrate the card-grid component to Astro

hestonhoffman:
Migrates the Hugo `card-grid` / `image-card` shortcode pair to the Astro site as Markdoc tags, with a plaintext twin and a hover tooltip. The tooltip shows on hover and keyboard focus, and dismisses on Escape.

In the end, there wasn't very much overlap with the img component. Just one shared function for building an image URL. I moved that into lib and exported it.

Included:

- `card-grid` and `image-card` Markdoc tags, with schema validation for invalid children, empty grids, and cards with no content.
- `image_width` inheritance from the grid down to each card, matching Hugo's `.Parent.Get` behavior
- A Preact tooltip island, hydrated on idle.
- A plaintext twin, kept in sync with the component.
- Design tokens for the new spacing, shadow, and tooltip colors.
- Unit and browser test coverage for the component, schema, transform, and twin.
- Claude noticed several issues with tooltip accessibility on Hugo and fixed them for Astro. I've kept Claude's code comments related to accessibility.

### AI assistance

Built with Claude Code from an approved spec, task by task, with a code review between tasks and a whole-branch review at the end. Every review finding was verified independently before it was acted on; one was narrowed and closed differently than proposed, and two earlier AI decisions were overturned by review.


astro/src/components/CardGrid/ImageCard.module.css
```diff
@@ -0,0 +1,41 @@
+.image-card {
+  display: flex;
+  border: var(--border-width) solid var(--color-border-card);
```
jhgilbert: Design tokens usually shouldn't be coupled to a particular component; we would just have a border color (or maybe multiple border colors, like border-dark etc.) and use it on all components. The idea of design tokens is that you can change one border token, and all your components then have the new border color. In practice sometimes "one token" is not really achievable, but that's the overall idea.

Same for a few other `card` scoped tokens (I think there's at least one other one).

astro/src/components/CardGrid/ImageCard.module.css
```diff
@@ -0,0 +1,41 @@
...
+.image-card__body {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  text-align: center;
+  gap: var(--space-xs);
+  padding: var(--space-sm) var(--space-xs);
+  width: 100%;
+  min-height: var(--card-grid-card-min-height);
```
jhgilbert: This token might be an exception to "tokens should be component agnostic", since I can't think of a more generic token you would use. There's an argument against using a token at all since the scope of the value is just in this stylesheet and nowhere else. It's not a theme variable, it's a value specific to how _this_ component operates and it's unlikely that anything else would reuse it. I would probably hardcode it, though I'm learning as I go here!

astro/src/components/CardGrid/ImageCard.module.css
```diff
@@ -0,0 +1,41 @@
...
+}
+
+.image-card--has-title .image-card__body {
+  min-height: var(--card-grid-card-min-height-titled);
```
jhgilbert: Probably also a hardcoded value, depending on what you decide to do above

astro/src/lib/images/buildImageUrl.ts
```diff
@@ -0,0 +1,23 @@
+import { IMAGES_URL } from "@config/images";
+
+export interface BuiltImageUrls {
```
jhgilbert: We wouldn't want a plural since it's not a list, and `Built` is just how we got there. `ImageUrl`? `ImageUrlAttrs`? I think I would expect `ImageUrl` to be a string, but that's fine, that's what types are for if I need to hover

astro/src/lib/plaintext/twinTransform.ts
```diff
@@ -38,6 +42,23 @@ type TwinAdapter = (node: MarkdocNode) => MarkdocNode | MarkdocNode[];
 const attr = (node: MarkdocNode, name: string): unknown =>
   node.attributes?.[name];
 
+/**
+ * Yield a node's children, replacing paragraph and inline wrappers with the
+ * nodes they contain. Markdoc only makes a tag a direct child when it sits
+ * alone on its line; inline siblings end up nested inside a paragraph.
+ */
+function* childrenUnwrappingParagraphs(
```
jhgilbert: `unwrapChildParagraphs` to keep it verby? Though this function name does evoke pleasant "Christmas morning" vibes ...

astro/markdoc.config.mjs
```diff
@@ -7,6 +7,32 @@ import {
...
+ * Two self-closing tags on one line are inline siblings, so Markdoc wraps
+ * them in a paragraph. A transform reading `node.children` directly would
+ * not see them and would silently render nothing, so container tags that
+ * only accept tag children walk through paragraphs instead.
+ *
+ * This works on AST nodes, before `transformChildren`, so callers can still
+ * read raw attributes off each child (`card-grid` needs that for its
+ * `image_width` inheritance).
+ */
+function* childrenUnwrappingParagraphs(node) {
```
jhgilbert: Are we even paying these children? Is this legal? (I'd go with `unwrapChildParagraphs` again)

astro/src/content/en/dd_e2e/components/card-grid.mdoc
```diff
@@ -0,0 +1,73 @@
+---
+title: "Card grid"
+description: "Places a clickable card grid on the page"
+type: static
+---
+
```
jhgilbert, line 6: I'd like to see a multi-row grid, especially one where the bottom row falls short of the top rows, just to verify how that's handled

astro/src/components/CardGrid/plaintext/CardGrid.ts
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * Only `href`, `title`, and `subtitle` survive; a grid carries no semantics
+ * once its images are stripped.
+ */
+
+import type { Node as MarkdocNode } from "@markdoc/markdoc";
+import { inline, link, list, listItem } from "@lib/plaintext/helpers";
+
+export interface PlaintextCard {
+  href: string;
```
jhgilbert, line 20: For the plaintext URLs, I noticed they're relative (`/some/path`). AI doesn't do well with that sometimes, could it be the full URL instead?

astro/src/components/CardGrid/ImageCard.module.css
hestonhoffman: It turns out, `color-border-card` was almost identical to the existing `--color-border` 😅

astro/src/components/CardGrid/ImageCard.module.css
hestonhoffman: I'm went with hard-coding

jhgilbert (approved):

PR #39667, opened 2026-09-02
[DOCS-15586] Escape markdown special characters inside ui shortcode

hestonhoffman:

Fixes DOCS-15586

The `ui` shortcode was recently updated to run its content through `markdownify` (#39571) so that Markdown inside `{{< ui >}}...{{< /ui >}}` renders correctly. A side effect is that a bare `+`, `-`, or `>` inside the shortcode is now interpreted as Markdown list or blockquote syntax instead of rendering as the literal character.

This PR escapes those characters (`\+`, `\-`, `\>`) wherever they appear alone inside `{{< ui >}}` tags across the English content, so they render as literal text again.

### Merge readiness

- [ ] Ready for merge


estherk15 (approved):
Thanks for the fix Heston!

PR #39671, opened 2026-09-02
Update workflow AI build instructions

GabeWM:
The workflow AI instructions no longer match the current workflow editor and Bits Chat experience.

Updates the primary creation flow to start from **New workflow**, and documents **Ask Bits** from anywhere in Datadog as another entry point. Clarifies that **Edit with AI** is an unlabeled AI toolbar icon, and covers follow-up questions and confirmation. Removes the obsolete warning that workflow AI cannot answer product questions.

Related: https://dd.slack.com/archives/D0B6JFXSFKK/p1788363428287229

### DIFF
--- a/hugo/content/en/actions/workflows/build.md
+++ b/hugo/content/en/actions/workflows/build.md
@@ -51,7 +51,7 @@ To create a workflow:
 1. Click the up arrow ({{< ui >}}↑{{< /ui >}}) to submit the prompt.
 1. Answer any follow-up questions, then confirm the workflow when prompted. Bits creates the workflow and opens it in the workflow editor.
 
-You can also create a workflow from anywhere in Datadog with Bits Chat. Open Bits Chat by clicking {{< ui >}}Ask Bits{{< /ui >}} in the top-right navigation bar or pressing <kbd>Cmd</kbd>/<kbd>Ctrl</kbd> + <kbd>I</kbd>, then ask Bits to create a workflow.
+You can also create a workflow from anywhere in Datadog with Bits Chat. Open Bits Chat by clicking {{< ui >}}Ask Bits{{< /ui >}} in the top-right navigation bar or pressing <kbd>Cmd</kbd> or <kbd>Ctrl</kbd> + <kbd>I</kbd>, then ask Bits to create a workflow.
 
 To update an existing workflow:
 1. Open the workflow and click the AI icon <i class="icon-bits-ai"></i> in the left toolbar. Bits Chat opens with the workflow attached as context.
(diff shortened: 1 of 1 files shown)


hugo/content/en/actions/workflows/build.md
```diff
@@ -43,20 +43,21 @@ You can create workflows or edit existing workflows from the [Workflow Automatio
...
+1. From the [Workflow Automation][1] page, click {{< ui >}}New workflow{{< /ui >}}.
+1. Under {{< ui >}}Build a workflow with AI{{< /ui >}}, enter a detailed prompt or select a suggested prompt.
+1. Click the up arrow ({{< ui >}}↑{{< /ui >}}) to submit the prompt.
+1. Answer any follow-up questions, then confirm the workflow when prompted. Bits creates the workflow and opens it in the workflow editor.
 
-To iterate on an existing workflow:
-1. From an existing workflow, click {{< ui >}}Edit with AI{{< /ui >}}.
-1. Enter a detailed prompt for the behavior you'd like to add to your workflow. Include the integrations and actions you'd like to use.
-1. Click the up arrow ({{< ui >}}↑{{< /ui >}}) to add the functionality to your workflow.
+You can also create a workflow from anywhere in Datadog with Bits Chat. Open Bits Chat by clicking {{< ui >}}Ask Bits{{< /ui >}} in the top-right navigation bar or pressing <kbd>Cmd</kbd>/<kbd>Ctrl</kbd> + <kbd>I</kbd>, then ask Bits to create a workflow.
```
buraizu: Suggesting to replace the slash for accessibility.

```suggestion
You can also create a workflow from anywhere in Datadog with Bits Chat. Open Bits Chat by clicking {{< ui >}}Ask Bits{{< /ui >}} in the top-right navigation bar or pressing <kbd>Cmd</kbd> or <kbd>Ctrl</kbd> + <kbd>I</kbd>, then ask Bits to create a workflow.
```

buraizu (commented):
Thanks for the PR! This looks good overall, with one minor suggestion for accessibility.

Separately, I'm wondering if this new content might be better served with its own subsection, perhaps with another new subsection for the existing content (something like "Create a workflow from Workflow Automation" for the existing content and "Create a workflow from anywhere" for the new content). That might make it easier for humans/AI agents to realize that there are two separate entry points for this functionality. Let me know what you think!

GabeWM:
> Thanks for the PR! This looks good overall, with one minor suggestion for accessibility.
> 
> Separately, I'm wondering if this new content might be better served with its own subsection, perhaps with another new subsection for the existing content (something like "Create a workflow from Workflow Automation" for the existing content and "Create a workflow from anywhere" for the new content). That might make it easier for humans/AI agents to realize that there are two separate entry points for this functionality. Let me know what you think!

@buraizu thanks for reviewing and I like the idea. We need to make some larger changes to how we approach AI in our docs. Is it alright if I leave this change for the future? I do agree it would be a good one, but making the minimal amount of changes to maintain accuracy as of right now.

buraizu (approved):

PR #39770, opened 2026-09-08
Update Security Findings schema reference page

filipelteixeira:
## Summary

This PR was generated by the Security Finding Schema Tool (CLI) to update the Security Finding Schema artifacts.

Schema source: `ddoghq/security-finding-schema@main`.

---

### Generated by

- **Tool**: Security Finding Schema Tool (CLI)

---

### CI Failures

If CI is failing, it may be because updating the schema broke some code integrations. In that case, a human should manually fix the code to make it work before merging. This is the responsibility of the PR creator.

---

cc @ddoghq/k9-finding-schema-governance

### DIFF
--- a/content/en/security/guide/findings-schema/generated/examples.md
+++ b/content/en/security/guide/findings-schema/generated/examples.md
@@ -1,0 +1,1736 @@
+{{< tabs >}}
+{{% tab "API Security" %}}
+
+```json
+{
+  "api_endpoint": {
+    "method": "GET",
+    "operation_name": "http.request",
+    "path": "/api/v2/users/{userID}/profile",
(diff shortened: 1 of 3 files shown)


filipelteixeira:
Closing: the generated artifacts target `content/en/security/guide/findings-schema/generated/*.md`, but the live page is the single file `content/en/security/guide/findings-schema.md`. Publishing this would add 7.4k lines at a new path that shadows the existing page's URL rather than updating it. Needs a fix in the publish tool's documentation output paths first. The batch's actual schema change (`custom._dd.sds_info`) is internal-only and does not surface in the customer-facing reference, so documentation has nothing to ship this batch.

PR #39800, opened 2026-09-09
Update canceled maintenance screenshot for Status Pages

jbenais:
## Summary
- Updates the cancel maintenance screenshot for Status Pages with a current one to stay consistent with other examples in the page.


reaganrazon (approved):

buraizu (changes requested):
@jbenais thanks for the PR! However, for the docs we shouldn't replace images with new versions that have the same filename. Instead, we need to upload new images with **new filenames**, and then update the `img` references in the doc to point to the new image.

This prevents certain caching issues with translated doc pages — let me know if you want more information, and I can direct you to some internal resources.

jbenais:
@buraizu thanks for the context! Fixed in [#a084820](https://github.com/DataDog/documentation/pull/39800/commits/a084820df3ab0fb72cb518d582038e15f44c01df): reverted `cancel-maintenance.png` back to the original, added the new screenshot as `cancel-maintenance-window.png`, and updated the doc's `img` reference to point to the new filename.

buraizu (approved):

PR #39810, opened 2026-09-09
Update AI Credit limits permission

iFallUpHill:
## Summary

- Document that only billing_edit is required to view and set AI Credit limits.
- Remove the outdated user_access_manage and org_management requirements.
- Confirm that the usage-quota API references already require only billing_edit.

## Testing

- git diff --check


buraizu (commented):
Thanks for the PR! Since there's now only a single permission, it might be helpful for customers if this link went directly to the relevant section. Unfortunately, I'm not able to make a suggestion where this link is defined. That's on line `69`, and it would need to be updated to the value below:
`/account_management/rbac/permissions/#billing-and-usage`

buraizu (approved):

PR #39889, opened 2026-09-11
Esther/on call integrations list proposal

estherk15:
- Reverts css style changes
- Converts the integration tiles to a list of available integrations
- Converts html to md table


SimoneT-DD:
Looks amazing now! Thanks Esther
When do you think we will able to merge this?

PR #39982, opened 2026-09-15
Add site-support banner

hestonhoffman:

Brings the site-support banner to the Astro API docs. A page for a product that isn't supported on the reader's selected Datadog site now shows the same caution banner it does in Hugo. The plaintext (`.md`) version of the page carries a similar note.

The support data moves to `shared/site_support.yaml` so both sites can read one copy. Hugo keeps its own copy for now, and an equivalence test fails if the two drift apart.

Two differences from Hugo, both deliberate:
- Hugo uses path-segment walking or front-matter matching to add the banner to pages. To add the banner to API pages in Hugo, we used a `cascade` feature that isn't available in Astro. I replaced both path-segment walking and the cascade feature with a `url_paths` feature in the shared config, which allows us to match paths explicitly. I've retained frontmatter `id` matching for backwards compatibility with Hugo.
- The support banner names the reader's site at build time rather than substituting it with client-side JS, so there is one rendered variant per unsupported site.

### AI assistance

Designed and implemented with Claude Code, test-first throughout.

### Testing
 
1. Install dependencies (this PR adds picomatch)
   ```
   yarn install
   ```
2. `yarn build:en`
3. `yarn preview`
4. Check the following pages for banner behavior:
    - [Test page](http://localhost:4322/dd_e2e/components/site-support-banner/?site=us): Banner appears on US1-FED and US2-FED.
    - [App Builder](http://localhost:4322/api/latest/app-builder/?site=gov2): Banner appears only on US2-FED.
    - [Agentless scanning](http://localhost:4322/api/latest/agentless-scanning/?site=gov2): Banner appears on US2-FED and US1-FED. 
    - [Agent observability](http://localhost:4322/api/latest/agent-observability/?site=us): No banners on any site.
    
### Additional notes

`region-param` (the client-side site-name substitution) and the preview/banner `callout` component are out of scope and still to come.


brett0000FF (approved):
Tested locally and looks good!

bgdeutsch (approved):
Non-blocking: `ApiLayout` never forwards `site_support_id`, so the frontmatter
tier is dead on API HTML pages.

[...page].md.ts:44   siteSupportNoteNodes(url.pathname, lang, entry.data.site_support_id)
ApiLayout.astro:72   <SiteSupportBanner pathname={Astro.url.pathname} />



Set `site_support_id` on an API page and `.md` gets a banner, the real page
doesn't. Silently. Nothing hits it today since everything uses `url_paths`.

Fix is threading the prop through — `entry` is already in scope at
`[...page].astro:39`.

PR #39999, opened 2026-09-16
[DOCS-15627] Document below-threshold detection for Cloud SIEM Threshold rules

janine-c:

Fixes DOCS-15627

Adds below-threshold detection coverage to the Cloud SIEM Custom Detection Rules docs. The Threshold method previously only described above-threshold (`a > N`) cases; this PR documents below-threshold cases (`a < N` or `a <= 0`) for detecting silent sources, along with the sliding evaluation window behavior that determines when a below-threshold case becomes true.

### Merge readiness

I made a couple wording changes from the original request, so I've pinged Jackson on Slack for approval. Will check this box when I have his rubber stamp 🙂 

- [ ] Ready for merge

## For Datadog employees:

### AI assistance

Used Claude Code to draft the documentation changes based on the Jira ticket's recommended text and to open this PR.

### Additional notes


hugo/content/en/security/cloud_siem/detect_and_monitor/custom_detection_rules/_index.md
```diff
@@ -64,7 +64,7 @@ You can create the following types of custom detection rules:
 
 The following detection methods are available when you create a custom detection rule or historical job:
 
-- Threshold: Detects when events exceed a user-defined threshold.
+- Threshold: Detects when the query count meets a user-defined case, including above-threshold (`a > N`) and below-threshold (`a < N` or `a <= 0`) cases.
```
melamos: It's not clear to me whether you can specify a threshold that defines multiple cases, or only one? You say "a case" singular but then use "and" in the includes example, which implies multiples.

fwiw, I also prefer "specified" over "user-defined". It's assumed the user does the specifying, and you avoid the unfriendlish "user".

hugo/layouts/shortcodes/mdoc/en/security/cloud_siem/set_conditions_threshold.mdoc.md
```diff
@@ -10,9 +10,10 @@
...
     - A rule condition contains logical operations (`>`, `>=`, `<`, `&&`, `||`) to determine if a signal should be generated based on the event counts in the previously defined queries.
-    - The ASCII lowercase query labels are referenced in this section. An example rule condition for query `a` is `a > 3`.
+    - The ASCII lowercase query labels are referenced in this section. An example rule condition for query `a` is `a > 3`. To detect a silent source, use a below-threshold condition, such as `a < 1` or `a <= 0`.
       {% alert level="info" %}
-      The query label must precede the operator. For example, `a > 3` is allowed; `3 < a` is not allowed.
+      - The query label must precede the operator. For example, `a > 3` is allowed; `3 < a` is not allowed.
+      - For below-threshold cases such as `a < 1` or `a <= 0`, the sliding evaluation window must be empty of matching events before the case is true. After you enable the rule, it must match at least one event before creating cases that indicate that there were no detections. If you group by an attribute, Cloud SIEM only evaluates groups that have matching events.
```
melamos: I read this several times and don't quite get it. This seems to say the rule must match an event before it can create a case that says no events matched, which seems odd. Can you clarify whether/why that's true, and reword so it doesn't read as contradictory?

melamos (commented):
I have questions!

hugo/content/en/security/cloud_siem/detect_and_monitor/custom_detection_rules/_index.md
janine-c: Thanks for this! Replaced `user-defined` with `specified`, and `and` with `or`.

hugo/layouts/shortcodes/mdoc/en/security/cloud_siem/set_conditions_threshold.mdoc.md
janine-c: Yeah, I think this behaviour is what originally confused the user. Jackson provided some alternative wording; does that clear it up at all?

melamos: The revision clarifies some unclear wording, but it doesn't really address the inconsistency. Digging into this a bit more:

1. Sequencing of the two below-threshold requirements is easy to misread as contradictory (window must be empty vs. must match an event). Should we make the sequence explicit? Like:

▎ "For below-threshold cases such as `a < 1` or `a <= 0`, after you enable the rule, it must first match at least one event to establish a baseline. Afterwards, the rule may evaluate as true when the sliding evaluation window is empty of matching events."

2. Should the grouping caveat cover an "all groups silent" scenario? (TBH, this is 100% Claude's overzealous question, but I raise it here in case it's important.)

▎ "If you group by an attribute, Cloud SIEM only evaluates groups that have matching events."

This reads as "a fully silent source with grouping enabled will never trigger a below-threshold signal". The Log Monitor "No data and below alerts" sibling pattern handles this by firing when there are no logs for all groups, not just when one group's count drops below threshold. Does Threshold have an equivalent case? Like:
  
  ▎ "If you group by an attribute, the rule evaluates only groups that have matching events, unless all groups are silent, in which case the rule evaluates as true."

melamos (approved):
I'm still a bit fuzzy on how evaluation works, but I'm approving so as not to stand in your way. Feel free to entertain or ignore my second round of Qs.

hugo/layouts/shortcodes/mdoc/en/security/cloud_siem/set_conditions_threshold.mdoc.md
jztegodd: Thanks for digging into this, @melamos and @janine-c!

Regarding **1 (sequencing):**
Making the order explicit is a huge improvement. The only tweak I'd suggest is dropping the word **"baseline"** -- in Cloud SIEM, "baseline" is a term of art specific to Anomaly detection (historical statistical baselines). Threshold rules don't calculate baselines; they just need an initial matching log to arm/initialize evaluation for that query. 

Greater aligned like this:
> "For below-threshold cases such as `a < 1` or `a <= 0`, after you enable the rule, it must match at least one event before it can evaluate a below-threshold condition. After this initial match, the rule triggers a security signal when the sliding evaluation window contains no matching events."

Regarding **2 (grouping / all groups silent):**
Your suspicion that Claude was overzealous is spot on, haha; Cloud SIEM Threshold rules do **not** have an "all groups silent" evaluation. That fallback exists in Log Monitors ("No data and below alerts"), but was never part of Cloud SIEM real-time rule evaluation. 

In Cloud SIEM Threshold, if you group by an attribute, it strictly tracks and evaluates groups that receive matching events. If a source goes completely silent, it does not trigger an "all groups silent" signal. Keeping the original statement as-is accurately reflects the engine's behavior:
> "If you group by an attribute, Cloud SIEM only evaluates groups that have matching events."

domalessi (approved):

hugo/layouts/shortcodes/mdoc/en/security/cloud_siem/set_conditions_threshold.mdoc.md
janine-c: Perfect, thank you both! Applied the edit and will publish once the checks pass 🙌🏻

PR #39707, opened 2026-09-03
[DOCS-15652] Add Experiments project access control docs

imran-at-datadog:
### What does this PR do? What is the motivation?

Fixes DOCS-15652

Documents Data Access Control for Agent Observability **Experiments projects**, which lets an admin restrict a single project so only chosen teams or roles can see it.

Adds `llm_observability/improve/access_control` covering what a restriction hides (list views, direct links, spans, evaluation metrics, dataset records, and writes), how to author the policy, the API equivalent, and the limitations.

Also corrects the **Agent Observability** entry on the Data Access Control page. Before this PR it read:

> Only the event data is restricted—experiment list views and metadata are not. Datasets, annotation queues, and managed prompts are not supported.

Both statements describe an earlier state of the feature. List views, metadata, and Experiments datasets are restricted as of this rollout; annotation queues and managed prompts remain unsupported.

| File | Change |
| --- | --- |
| `llm_observability/improve/access_control.md` | New page |
| `llm_observability/improve/_index.md` | Section link to the new page |
| `account_management/rbac/data_access.md` | Corrected the Agent Observability entry, added a link |
| `llm_observability/data_governance.md` | One-line pointer to the new page |
| `config/_default/menus/main.en.yaml` | Nav entry under Improve |

### Merge readiness

- [ ] Ready for merge

**The `Do Not Merge` label has been removed, but please read this before merging.** The feature is behind a rollout gate that is not yet enabled in production, so the correction to `data_access.md` describes behavior that is not live. Merging now makes that page accurate for the rollout and inaccurate until it happens. Reviews are welcome in the meantime; I will comment here when the rollout completes.

## For Datadog employees:

Review history and one open question:

**Resolved in review.** The first draft opened the UI procedure by explaining that access is keyed on the `ml_app` tag and that the same key covers both applications and projects. That framing is gone — `ml_app` is an implementation detail that the Experiments UI does not surface, so the procedure now just says to select the project. The tag survives only where it is load-bearing: the API payload, a note that the filter key may be locked to a value the org already uses, and the older access control page where the value is typed by hand. The `data_access.md` entry was reworded for the same reason; its previous phrasing ("Set the `ml_app` tag to match the project ID") read as though you tag your own telemetry, which is not what happens.

The page also moved from `Improve > Experiments` to `Improve`, so it can cover Datasets as well — which it already does, since datasets are restricted by their project — and was retitled *Data Access Control in Agent Observability*.

[...]

PR #40017, opened 2026-09-17
[DOCS-15744] Document annotation queue coverage in Agent Observability Data Access Control

imran-at-datadog:

Fixes [DOCS-15744](https://datadoghq.atlassian.net/browse/DOCS-15744)

Documents Data Access Control coverage for Agent Observability **annotation queues**. Restricting a project now also hides its annotation queues, the traces under review in them, the labels and notes reviewers applied, and the queue's label schema.

Two pages currently tell customers the opposite. `account_management/rbac/data_access.md` reads:

> **Not supported**: Annotation queues and managed prompts.

That was accurate when #39707 merged. Annotation queue enforcement has since been built and verified end-to-end in production, so the statement is about to become wrong for every customer.

Note that annotation queues are **not** required to belong to a project historically, so the page is careful to say a queue *that belongs to* a project inherits its restriction, and to call a project-less queue out as unrestrictable.

This PR extends the existing `llm_observability/improve/access_control` page rather than adding a new one. That page was deliberately moved up from *Improve > Experiments* to *Improve* in #39707 so it could cover more than experiments, and queues are restricted by exactly the same object (a Restricted Dataset with an `ml_app` filter set to a project ID). No new page and no nav change.

| File | Change |
| --- | --- |
| `llm_observability/improve/access_control.md` | Broadened the overview from Experiments projects to Agent Observability projects; added what a restriction hides on a queue; new **Annotation queues** section distinguishing Data Access Control from a queue's own reviewer/assignee settings; two limitations added, one corrected |
| `account_management/rbac/data_access.md` | Annotation Queues added to supported telemetry; **Not supported** narrowed to managed prompts |
| `llm_observability/investigate/annotation_queues.md` | Pointer from **Managing queue access** to the access control page, plus a further-reading link |
| `llm_observability/data_governance.md` | One-line pointer broadened to mention queues |

### Merge readiness

- [ ] Ready for merge

**Please read this before merging.** The annotation-queue kill switch is enabled in one verification org, not for all production customers. The flip is imminent and is what this PR is preparing for, but until it happens the corrected `data_access.md` entry describes behavior that is not live everywhere. Reviews are very welcome in the meantime; I will comment here when the rollout completes.

## For Datadog employees:

Verification this is based on: **Annotation Queues DAC — Prod verification results (MLOB-8128)**, 31 PASS / 0 LEAK / 0 FAIL / 0 5xx in Greg's demo org on us5, plus a pre-restriction baseline and a post-restriction integrity pass confirming that every refused write changed nothing. The wording of what a restriction hides is taken from surfaces that were actually exercised (list views, annotated interactions, label schema, every write route, and both export routes), not from the design doc.

Three things reviewers should know. The first and third want a decision; the second is settled and recorded here for the record.

**1. Two behaviors customers will hit are stated explicitly.**

- *A queue with no project cannot be restricted* (a Limitations bullet). Enforcement deliberately lets queues that have no project through — hiding them would have blacked out pre-existing queues in every org at flip time. Any queue created before an org configured Data Access Control may be in this state and is unrestrictable, which is a real hole worth stating plainly.
- *Every annotation queue you create must belong to a project* (in the **Annotation queues** section, since it is a requirement rather than a limitation). This is a user-visible API behavior change and it is worth reviewers noting how broad it is: `project_id` is optional on the create DTO and is enforced in the service by `strings.TrimSpace(params.ProjectID) == "" && s.dataAccessEnabled(ctx, orgID)`. That predicate is the **kill-switch flag**, not "this org has configured Data Access Control", so at full rollout the requirement lands on every targeted org, including orgs with no Restricted Datasets that never use the feature. The annotation queues page already documents `project_id` as required on create, which becomes accurate for everyone at that point.

**2. The remedy for an orphan queue is UI-only, by design.** The limitation says to *move the queue into a project, or recreate it in one*. Checked against `origin/main`, that is accurate but route-dependent:

- The service supports reassignment and checks `requireProjectWritable` on **both** the source and destination project, so a caller cannot read a project they cannot see by moving a queue out of it.
- Moving an **orphan** queue into a project is permitted: an empty source project short-circuits that check (`queue_access.go:268`, and the comment above it says so explicitly), so the caller only needs write access on the destination.
- Clearing a queue's project is always refused (`project_id cannot be cleared`).
- But `project_id` is **not** on the public `AnnotationQueueUpdateRequestV2` DTO, so this is reachable only through the UI route, which embeds `UpdateAnnotationQueueParams` (that one does carry `ProjectID`).

So the remedy works in the app and not through the documented API. The annotation queues page already lists Update as `name`, `description`, or `annotation_schema` only, so the two pages don't contradict each other — but if Docs would rather be explicit, the limitation can say to move the queue *in the UI*.

**3. One verified gap is deliberately not documented.** Creating a queue into a project the caller cannot see answers 404, while creating into a project UUID that does not exist at all answers 201 — so on the create path a 404 means "this project exists and you cannot see it". It is filed as MLOB-8151 and is not a data leak: no queue or project content is exposed and the caller must already hold the UUID. I judged it too narrow and too internal for customer docs. Say the word if you would rather it appear.

Inherited from #39707 and unchanged here: this page says **policy** where `data_access.md` says **Restricted Dataset** throughout, bridged only by the info callout in step 1. Still a find-and-replace if Docs would rather the two pages agree today.

Also omitted deliberately: comment content on traces is not covered by project-keyed restrictions, but "comments" are not a documented Agent Observability feature anywhere in these docs, so a Limitations bullet would be introducing the feature in order to disclaim it. The near-term plan is to state it in the product UI where comments are written instead. Flagging it here rather than in the page.

### AI assistance

Drafted with Claude Code from the MLOB-8128 production verification results and the enforcement PRs, then reviewed and corrected by hand. Two claims were pulled back out of the pages during that review (the comments limitation, and an over-confident remedy for orphan queues) and moved into this description; the orphan-queue one was then settled by reading `origin/main`, and two further errors were caught the same way — a flat assertion that queues belong to a project, and a requirement wrongly conditioned on having configured Data Access Control.

### Additional notes

Preview links should show the updated page under **Agent Observability > Use Agent Observability to Improve Applications > Data Access Control**, with the new **Annotation queues** section between *Restrict a project through the API* and *Find a project's ID*.

[DOCS-15744]: https://datadoghq.atlassian.net/browse/DOCS-15744?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gsvigruha (approved):

hugo/content/en/llm_observability/improve/access_control.md
```diff
@@ -80,6 +81,14 @@ You can also create a restriction with the Data Access Control [Datasets API][5]
 }
 ```
 
+## Annotation queues
+
+An annotation queue that belongs to a project inherits that project's restriction. Restricting a project hides its queues, the traces they hold for review, the labels reviewers applied, and each queue's label schema. Users outside the granted teams or roles cannot annotate, edit, or delete a queue, and cannot export its annotated interactions to a dataset or to CSV.
```
estherk15:
```suggestion
An annotation queue that belongs to a project inherits that project's restriction. Restricting a project hides its queues, the traces they hold for review, the labels that reviewers applied, and each queue's label schema. Users outside the granted teams or roles cannot annotate, edit, or delete a queue, and cannot export its annotated interactions to a dataset or to CSV.
```

estherk15 (approved):

PR #40025, opened 2026-09-17
[DOCS-14483] Document datadog.yaml config options supported in Fleet Automation

evazorro:

You can deploy configuration changes in Fleet Automation by providing a `datadog.yaml` file. However, Fleet Automation doesn't support all possible fields right now. Customers previously had to guess which fields they could provide in YAML, and would hit validation errors when they used invalid fields. 

This PR adds a reference for which fields are supported. @StefonSimmons did work for us on the `website-sources` side to make data from the source file available in the `documentation` repo, and I created a partial that parses that data and turns it into table form.

### Merge readiness

- [ ] Ready for merge

### AI assistance

Used Claude Code to create the partials and shortcodes. Intro text and PR description by me.


hugo/layouts/shortcodes/fa-config-fields.html
```diff
@@ -0,0 +1,78 @@
+{{- /*
+  Renders the supported Fleet Automation datadog.yaml configuration fields as
+  collapsible sections, each containing a table of fields. Sourced from the
+  config-validator schema vendored at _vendor/data/agent_config_schema.json
```
dustmop: What is the path `_vendor/data/agent_config_schema.json` referring to?

dustmop (approved):
LGTM, but I have one question that I'm mainly asking out of curiosity.

hugo/layouts/shortcodes/fa-config-fields.html
evazorro: When the validator schema gets pulled in from the `dd-source` repo, that's where the file with that info ends up. PR here if you want to check it out: https://github.com/ddoghq/websites-sources/pull/694

dustmop: I see now, thanks! LGTM

hugo/layouts/shortcodes/fa-config-fields.html
```diff
@@ -0,0 +1,78 @@
+{{- /*
+  Renders the supported Fleet Automation datadog.yaml configuration fields as
+  collapsible sections, each containing a table of fields. Sourced from the
```
brett0000FF: Not blocking, but it would be nice if the headings showed up in the right-hand ON THIS PAGE section.

hugo/content/en/agent/fleet_automation/configure_agents.md
```diff
@@ -63,7 +63,7 @@ The example below shows the `logs_enabled` field changed from `false` to `true`,
 
 Fleet Automation provides an API to apply configuration updates programmatically. Deploy changes to any group of hosts with filter queries, supplying either full configuration files or targeted patches. Push configuration on demand or integrate it into your existing automation workflows. For full details, see the [Fleet Automation API][4].
 
-**Note**: The API does not support all Agent configuration fields. Settings related to Agent connection or secrets (`site`, `api_key`, and other authentication parameters) cannot be managed through the API.
+**Note**: The API does not support all Agent configuration fields. Settings related to Agent connection or secrets (`site`, `api_key`, and other authentication parameters) cannot be managed through the API. For the fields you can manage, see [Supported datadog.yaml Configuration Fields][11].
```
brett0000FF, line 66: This page seems useful not just for the API but also for the section above? Might be useful to link from there too.

brett0000FF (approved):
Very cool! Just a few bits of feedback:
- Headings don't show in the right-hand TOC. Not blocking but would be nice.
- Valid values uses exponent notation. This seems unusual, so wanted to flag in case that's not intentional. For example, `max_events_per_second` has `0–1e+09`.

evazorro:
Thanks Brett! Re this point:

> * Headings don't show in the right-hand TOC. Not blocking but would be nice.

Agreed it's not the best. The TOC gets built from Markdown headings and I wanted to _just_ use the collapse blocks on the page (which get built as raw HTML by the shortcode, so there are no Markdown headers for it to pick up).

This is another option I considered (real headings with a collapse block under each), but I didn't like how it looked:

image

Another option, suggested by Claude, is that we could add a new special case to this [table-of-contents](https://github.com/DataDog/documentation/blob/master/hugo/layouts/partials/table-of-contents/table-of-contents.html) file. That looks right, but I don't think we want to add any new custom stuff ahead of the Astro migration:

image

Idk. Maybe even if it's visually less appealing (to me), the best option is to do the headers version so we get the TOC. What do you think?

hugo/content/en/agent/fleet_automation/configure_agents.md
evazorro: Yup, good call! Added to the page intro.

Also fixed the bug with the exponent notation -- good catch!!

bgdeutsch (approved):

PR #40081, opened 2026-09-21
Align MCP cases tool docs with Work Management naming and fix create_datadog_case permissions

gondyb:

Two changes to the Cases (Work Management) section of the MCP Server tools page.

**Corrects the permissions listed for `create_datadog_case`.** It requires `Cases Write` **and** `Cases Read`, not `Cases Write` alone. The tool can resolve a project from a project key or a project name, which reads the project list, so a caller with only write access gets a permission error. The page currently understates this, so anyone granting permissions from these docs ends up with a tool that fails. This part is a straight correction and is accurate today.

**Aligns the wording with Work Management.** The section heading already says "Cases (Work Management)", but the body still said "case" throughout. Descriptions now lead with "work item" while keeping "case" alongside, and a short note up front explains that the two terms mean the same thing and that the tool names, the `case_id` argument, and the returned keys still use *case*. The example prompts deliberately use a mix of both so it is clear either phrasing works.

No behavior is documented as changing beyond the permission correction.

### Merge readiness

- [x] Ready for merge

The permission correction is accurate now and could merge immediately. The terminology wording is matched by an in-flight update to the tool descriptions themselves; if you would prefer to land only the permission fix first, I am happy to split it.

### AI assistance

Drafted with Claude Code. The permission correction and the terminology were checked against the tool implementations and the live tool schemas rather than written from memory.

### Additional notes

Only `content/en` is edited; translations are managed separately.


brett0000FF (approved):
Looks good! It seems like we should eventually just mention that cases = work items once at the top, rather than throughout each section. But since the toolset and tool names still use "case," the repetition seems valuable to avoid confusion for users.

gondyb:
Yes, the next step is to GA and update the MCP tools name

PR #40110, opened 2026-09-21
Switch Kafka DSM demo consumers from Python to Java

shaina-patel-23:
**Motivation**
Previously we weren't able to see the consumer group members and configuration details for python consumers. This PR fixes that.

**Changes**
- Rewrites the Python consumers to Java

**Testing**
| Before | After |
|--------|--------|
| BeforePR1 | AfterPR1 |
| BeforePR2 | AfterPR2 |

### Merge readiness

- [ ] Ready for merge

### For Datadog employees:


hugo/static/resources/yaml/data_streams/docker-compose-kafka-demo.yaml
chnn: (nit) It seems a bit hacky to have a second unique cluster that shares the same cluster ID. All of the metrics reported from this cluster will be indistinguishable from metrics reported from the first cluster, which means the metrics queried in the UI will be some aggregate of the two.

I don't know if this will create problems, maybe it's fine? Alternatively, we could give the second cluster a unique ID and `KAFKA_ADVERTISED_LISTENERS`. Then (I think) the UI would show the one function cluster, and the offline partition warnings from the second cluster separately.

chnn (approved):
One nit, looks good though!

hugo/static/resources/yaml/data_streams/docker-compose-kafka-demo.yaml
shaina-patel-23: Applied suggestion, ss is in the PR for the current state. We are now displaying two distinct cluster IDs so they're no longer indistinguishable. 
- Original cluster: cluster:kafka-demo + kafka_cluster_id:Mka3OEVBNTcwNTJENDM2Qg                          
- Chaos cluster: cluster:kafka-demo-chaos + kafka_cluster_id:Y2hhb3NDbHVzdGVyMDAwMA

OliviaShoup (approved):
thanks for the PR!

